PR #12, opened 2023-03-22
Move some utils from client to types

thibault-martinez:


Thoralf-M (dismissed):

Thoralf-M (dismissed):

Thoralf-M (approved):

types/src/block/output/alias_id.rs
```diff
@@ -19,6 +22,11 @@ impl AliasId {
     pub fn or_from_output_id(self, output_id: &OutputId) -> Self {
         if self.is_null() { Self::from(output_id) } else { self }
     }
+
+    /// Returns the bech32 encoding of the alias ID.
+    pub fn to_bech32(&self, bech32_hrp: &str) -> String {
```
kwek20, line 27: maybe we should make this a trait?

thibault-martinez: Yeah I've been thinking about that as well, I don't like that Bech32 is not strongly typed. I'll open an issue. A topic for another day.

thibault-martinez: https://github.com/iotaledger/iota-sdk/issues/18

kwek20 (approved):
1 comment but fine to merge

PR #21, opened 2023-03-22
Validate bech32 hrp of provided addresses

### LINKED ISSUE #1954: Validate bech32 addresses against the network
## Description

When wallet.rs expects bech32 addresses as arguments, the corresponding functions should validate whether that address belongs to the network of the account manager

## Motivation

To mitigate users sending funds from one network to the other.

## Requirements

It shouldn't be possible for  

1. a prepare/send an output with a profile manager from the SMR network
2. with a recipient address of another network (e.g. `iota...`) to succeed

## Open questions (optional)

Use this section to ask any questions that are related to the feature.

No.

Thoralf-M:
# Description of change

Validate bech32 hrp of provided addresses

Fixes https://github.com/iotaledger/wallet.rs/issues/1954

## Type of change

Added tests

## Change checklist

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -22,6 +22,7 @@ use crate::{
     builder::{ClientBuilder, NetworkInfo},
     constants::DEFAULT_TIPS_INTERVAL,
     error::Result,
+    Error,
 };
 
 /// An instance of the client using HORNET or Bee URI
@@ -199,5 +200,17 @@ impl Client {
             .map_or(NetworkInfo::default().fallback_to_local_pow, |info| {
                 info.fallback_to_local_pow
             })
+    }
+
+    /// Validates if a bech32 HRP matches the one from the connected network.
+    pub async fn validate_bech32_hrp(&self, bech32_hrp: &str) -> Result<()> {
+        let expected_bech32_hrp = self.get_bech32_hrp().await?;
+        if bech32_hrp != expected_bech32_hrp {
+            return Err(Error::InvalidBech32Hrp {
+                provided_bech32_hrp: bech32_hrp.to_string(),
+                expected_bech32_hrp,
+            });
+        };
+        Ok(())
     }
 }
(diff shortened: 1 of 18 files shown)


thibault-martinez (changes requested):
Conflicts

client/src/client.rs
```diff
@@ -200,4 +201,16 @@ impl Client {
                 info.fallback_to_local_pow
             })
     }
+
+    /// Validates if a bech32 HRP matches the one from the connected network.
+    pub async fn validate_bech32_hrp(&self, bech32_hrp: &str) -> Result<()> {
```
kwek20: this method name sounds like it could be used to validate if an hrp is valid, not if its for the current network.
i know the comment says that, but maybe we rename to `bech32_hrp_matches` and return Result<bool> instead?
A whole error thrown seems excessive to me

kwek20: hmm it does make the code below less readable having to do a `if matches {}`
Lets leave the result as is

Thoralf-M: Renamed

kwek20 (dismissed):

client/src/error.rs
```diff
@@ -53,6 +53,14 @@ pub enum Error {
     /// Invalid BIP32 chain data
     #[error("invalid BIP32 chain data")]
     InvalidBIP32ChainData,
+    /// Invalid bech32 HRP, should match the one from the used network
+    #[error("invalid bech32 hrp for the connected network: {provided_bech32_hrp}, expected: {expected_bech32_hrp}")]
+    InvalidBech32Hrp {
+        /// The bech32 human readable part from the provided address.
+        provided_bech32_hrp: String,
```
thibault-martinez: I'd personally not repeat the bech32_hrp part here and expected but not a big deal, we just usually don't

thibault-martinez (dismissed):

thibault-martinez (approved):

kwek20 (approved):

PR #33, opened 2023-03-24
SDK crate

thibault-martinez:


sdk/Cargo.toml
```diff
@@ -68,6 +70,7 @@ stronghold = [ "iota_stronghold" ]
 events = [ ]
 ledger_nano = [ "iota-ledger-nano" ]
 participation = [ "storage" ]
+message_interface = [ "backtrace", "fern-logger" ]
```
Thoralf-M, line 78: Why `"fern-logger"` here?

thibault-martinez: It's only used for the `init_logger` function of the message interface

Thoralf-M: Ah Brord added the function also to main crate, I thought it was only in the bindings

.github/workflows/bindings-nodejs.yml
```diff
@@ -11,7 +11,7 @@ on:
       - "!**/examples/**" # Exclude all examples
       - "!**/tests/**" # Exclude all tests
       - "!client/bindings/**" # Exclude all bindings
-      - "client/bindings/nodejs/**" # Re-include nodejs bindings
+      - "sdk/src/client/bindings/nodejs/**" # Re-include nodejs bindings
```
Thoralf-M, line 14: What about the wallet bindings?
Exclude should be like this 
```
      - "!/bindings/**" # Exclude all bindings
```
and include not sure, maybe we wait until we merged it

thibault-martinez: Yeah tbh I think I'd merge without taking care too much of the CI for now, it needs a lot of changes because of the merge.

thibault-martinez: I've done the bare minimum so we could merge and rebase other PRs ASAP, I'll properly fix them all later

Thoralf-M (approved):

kwek20 (approved):

PR #39, opened 2023-03-27
Bindings/wasm cache

### LINKED ISSUE #60: Cache `get_network_info` on wasm builds
To avoid an HTTP call every time.

kwek20:
# Description of change

Adds a cache over the NetworkInfo getter in wasm
Cache only holds for 1 second instead of the 1 minute in the issue. was probably a typo

Also updated the init_logger for wasm, and moved all unix_timestamp to a single method

fixes [https://github.com/iotaledger/iota-sdk/issues/60](https://github.com/iotaledger/iota-sdk/issues/60)

## Type of change

called account methods with a test to see if we hit the cache or the real getter


sdk/src/client/constants.rs
```diff
@@ -22,6 +22,9 @@ pub(crate) const DEFAULT_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/",
 pub(crate) const MAX_PARALLEL_API_REQUESTS: usize = 100;
 /// Max allowed difference between the local time and latest milestone time, 5 minutes in seconds
 pub(crate) const FIVE_MINUTES_IN_SECONDS: u32 = 300;
+/// Delay for caching a node info response in WASM runtime
+#[cfg(target_family = "wasm")]
+pub(crate) const CACHE_NETWORK_INFO_TIMEOUT_IN_SECONDS: u32 = 1;
```
Thoralf-M: Bech32, network id... basically never changes, only the milestone timestamp which we use in `get_time_checked()` and there we just need to be somehow correct (currently it's +-5 minutes), so should be fine with 60 seconds?
Only 1 will still send many more requests than needed

kwek20: Hmm fair point. but maybe we need to set it to something like FIVE_MINUTES_IN_SECONDS / 5, so if we ever change that, its still ok?

kwek20: idk its hard to test the cache in an automated way

Thoralf-M: Don't think we need to make it depending on the other value, it's rather depends on the milestone interval
For example if we at some point get protocol updates via milestones it should probably be lower, but I think there is no timeline for this

Thoralf-M (dismissed):

thibault-martinez (changes requested):
Doesn't compile on WASM

Thoralf-M (approved):

thibault-martinez (approved):

PR #145, opened 2023-03-29
CLI: init with default values when called with no parameters

### LINKED ISSUE #85: CLI: Calling ´wallet´ without parameters should also initialise the wallet if no account exists
## Description

I often just type `wallet` as the first thing. No we can discuss if I'm just stupid :trollface:, but wouldn't it make sense to run `wallet init` internally than if nothing was created yet?

## Motivation

Feels more intuitive to me and I saw some people which did the same thing as I did 😅 

## Requirements

1. Run `wallet` 
2. If no db or stronghold file was created yet, execute the init routine.
3. Enjoy initialised wallet

## Open questions (optional)

Should we just check if the folders/files weren't created yet, or what would be the best way to do it?

No

thibault-martinez:
Fixes #85

### DIFF
--- a/cli/src/account_history.rs
+++ b/cli/src/account_history.rs
@@ -21,6 +21,7 @@ impl Default for AccountHistory {
 
 impl<T: ToString> History<T> for AccountHistory {
     fn read(&self, pos: usize) -> Option<String> {
+        println!("test");
         self.history.get(pos).cloned()
     }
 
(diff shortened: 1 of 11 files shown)


cli/src/account_history.rs
```diff
@@ -21,6 +21,7 @@ impl Default for AccountHistory {
 
 impl<T: ToString> History<T> for AccountHistory {
     fn read(&self, pos: usize) -> Option<String> {
+        println!("tet");
```
Thoralf-M:
```suggestion
```
:trollface:

thibault-martinez: :trollface:

Thoralf-M (approved):

kwek20 (approved):

PR #149, opened 2023-03-29
CLI: improve help and add autocompletion

### LINKED ISSUE #74: CLI: The in-command help function doesn't show all possible options for mint-native-token like reading metadata from a json file (which is available)
Actually check for all commands.

### LINKED ISSUE #140: CLI: investigate dialoguer completion feature

thibault-martinez:
Fixes https://github.com/iotaledger/iota-sdk/issues/74
Fixes https://github.com/iotaledger/iota-sdk/issues/140

### DIFF
--- a/cli/src/account_completion.rs
+++ b/cli/src/account_completion.rs
@@ -1,0 +1,62 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use dialoguer::Completion;
+
+pub(crate) struct AccountCompletion<'a> {
+    options: [&'a str; 33],
+}
+
+pub(crate) const ACCOUNT_COMPLETION: AccountCompletion = AccountCompletion {
+    options: [
+        "addresses",
+        "balance",
+        "burn-native-token",
+        "burn-nft",
+        "claim",
+        "consolidate",
+        "create-alias-output",
+        "decrease-native-token-supply",
+        "destroy-alias",
+        "destroy-foundry",
+        "exit",
+        "faucet",
+        "increase-native-token-supply",
+        "mint-native-token",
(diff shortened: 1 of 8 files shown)


cli/src/account_completion.rs
```diff
@@ -0,0 +1,62 @@
...
+        "help",
+    ],
+};
+
+impl<'a> Completion for AccountCompletion<'a> {
+    fn get(&self, input: &str) -> Option<String> {
+        let matches = self
+            .options
+            .iter()
+            .filter(|option| option.starts_with(input))
```
kwek20, line 53: maybe we want to switch this to contains? just typing power, would be nice that all options show up

thibault-martinez: Yeah but you have to return only one String, so what do you return when you provide power ?

kwek20: ah i see, thats silly. never mind then

cli/src/command/account.rs
```diff
@@ -41,110 +41,182 @@ pub enum AccountCommand {
...
+    },
+    /// Burn an NFT.
+    BurnNft {
+        /// NFT ID to be burnt, e.g. 0xecadf10e6545aa82da4df2dfd2a496b457c8850d2cab49b7464cb273d3dffb07.
+        nft_id: String,
+    },
     /// Claim outputs with storage deposit return, expiration or timelock unlock conditions.
-    Claim { output_id: Option<String> },
+    Claim {
+        /// Output IDs to be claimed.
```
Thoralf-M:
```suggestion
        /// Output ID to be claimed.
```

Thoralf-M (approved):

kwek20 (approved):

PR #179, opened 2023-04-03
Generic builder methods

thibault-martinez:


sdk/src/types/block/output/nft.rs
```diff
@@ -94,22 +94,27 @@ impl NftOutputBuilder {
...
     #[inline(always)]
-    pub fn add_unlock_condition(mut self, unlock_condition: UnlockCondition) -> Self {
-        self.unlock_conditions.push(unlock_condition);
+    pub fn add_unlock_condition(mut self, unlock_condition: impl Into<UnlockCondition>) -> Self {
+        self.unlock_conditions.push(unlock_condition.into());
         self
     }
 
-    ///
+    /// Sets a collection of [`UnlockCondition`] to the builder.
```
DaughterOfMars: It would probably be good to specify in the comment that this is a replacement, rather than an append. Same for the other variants of course.

thibault-martinez: I was under the impression `Sets` would mean that, but I can clarify for sure

DaughterOfMars (dismissed):

sdk/src/types/block/output/nft.rs
thibault-martinez: Done

Thoralf-M (approved):

kwek20 (approved):

PR #188, opened 2023-04-03
Only run coverage tests on `iota-sdk`

### LINKED ISSUE #80: Fix coverage CI

DaughterOfMars:
# Description of change

Changed the `--all` flag to specifically select `--package iota-sdk`.

Closes #80

## Type of change

## Change checklist


.github/workflows/build-and-test.yml
```diff
@@ -96,7 +99,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --all-features --release
+          args: --all-features
```
thibault-martinez, line 102: We sure about this one? Running our tests on debug bug is notoriously slow, I'd say the compilation time difference is way smaller than the execution time difference so not sure if it's worth it?

DaughterOfMars: That's surprising. I can revert it, but I think it's unusual that debug vs release would have a meaningful impact on an application like this.

thibault-martinez: I think it's down to hashes, they are very slow on debug mode for some reason

Thoralf-M: For PoW there is a big difference, but since the minimum pow score is 0 in our private tangle it should still be fine?

thibault-martinez: Idk, maybe I just have some PTSDs from bee but I remember debug was just way too slow

.github/workflows/coverage.yml
```diff
@@ -67,12 +70,20 @@ jobs:
...
         run: cargo install cargo-binutils rustfilt
 
+      - name: Build for tests
+        uses: actions-rs/cargo@v1
+        env:
+          RUSTFLAGS: "-C instrument-coverage"
+          LLVM_PROFILE_FILE: "iota-%m.profraw"
+        with:
+          command: test
+          args: --tests --package iota-sdk --all-features --no-run -- --include-ignored
```
thibault-martinez, line 80: Our of curiosity, can you elaborate a bit on this new step?

DaughterOfMars: I originally added it simply to try to separate building and running the tests, to see if that would help identify the problem. Then I thought it might be smart to hold off on starting the private tangle until the build is done, otherwise it would run the entire 20+ minutes.

.github/workflows/coverage.yml
```diff
@@ -67,12 +70,20 @@ jobs:
       - name: Install binutils and rustfilt
         run: cargo install cargo-binutils rustfilt
 
+      - name: Build for tests
+        uses: actions-rs/cargo@v1
+        env:
+          RUSTFLAGS: "-C instrument-coverage"
+          LLVM_PROFILE_FILE: "iota-%m.profraw"
```
Thoralf-M, line 77: Are they needed here? Isn't it enough if they get set in coverage.sh?

DaughterOfMars: They are. Without them, it would rebuild again in the script.

.github/workflows/coverage.yml
thibault-martinez: Ok, I doubt the private tangle running in the meantime would be an issue though, it doesn't do anything if there is no activity

thibault-martinez (approved):

Thoralf-M (approved):

PR #195, opened 2023-04-04
fix: support macOS 10.13

Tuditi:
# Description of change

Use latest XCode version and patch `librocksdb-sys` to be able to build nodeJS bindings compatible with macOS 10.13. They should be run on the latest macOS version, since that's the runner that supports the newest Xcode.

The `build.rs` for `librocksdb-sys` is patched so that it is compiled with `-faligned-allocation`. This patch can be removed once an Xcode version that uses Clang 15 is released.

https://github.com/iotaledger/firefly/issues/4766

## Type of change

- Documentation Fix

## Change checklist


thibault-martinez (dismissed):

.github/workflows/build-and-test.yml
```diff
@@ -126,6 +126,12 @@ jobs:
         run: echo "LIBCLANG_PATH=$((gcm clang).source -replace "clang.exe")" >> $env:GITHUB_ENV
         if: matrix.os == 'windows-latest'
```
rajivshah3: We need to add this in order to verify it's working correctly:

```
- name: Set deployment target (macOS)
  run: echo "MACOSX_DEPLOYMENT_TARGET=10.13" >> $GITHUB_ENV
  if: matrix.os == 'macos-latest'
```

rajivshah3 (changes requested):

.github/workflows/build-and-test.yml
```diff
@@ -126,6 +126,12 @@ jobs:
...
 
+      - name: Select Xcode (macOS)
+        uses: maxim-lobanov/setup-xcode@
```
rajivshah3:
```suggestion
        uses: maxim-lobanov/setup-xcode@v1
```

Tuditi:
This PR is in draft, since XCode 14.3 requires the runners to be compatible with `macos-13`. Which hasn't been released yet.
https://github.com/actions/runner-images/issues/7375

thibault-martinez:
`macos-13` has been released, right?

Tuditi:
> `macos-13` has been released, right?

Yes, I updated it now, so let's see if the tests go through.

DaughterOfMars (dismissed):

.github/workflows/bindings-wallet-nodejs.yml
```diff
@@ -45,14 +45,23 @@ jobs:
...
+        os: [windows-latest, macos-13, ubuntu-latest]
         node: ["18"]
 
     steps:
       - name: Checkout the Source Code
         uses: actions/checkout@v3
 
+      - name: Select Xcode (macOS)
+        uses: maxim-lobanov/setup-xcode@v1
+        if: matrix.os == 'macos-latest'
```
thibault-martinez: Just checking, is this supposed to be `macos-latest` ?

Tuditi: It's not, good catch!

thibault-martinez:
@Tuditi or @rajivshah3 Can one of you have a look at the failing CI please 🙏🏻

rajivshah3:
Looks like the same errors are still happening 😢 . It seems like even though Apple's fork [says it's using LLVM 15](https://github.com/apple/llvm-project/blob/f0fb631dd1a3a2988b23ba5057cd9106713cd0b4/llvm/CMakeLists.txt#L14), `clang --version` is reporting `Apple clang version 14.0.3` and the [line](https://github.com/apple/llvm-project/blob/f0fb631dd1a3a2988b23ba5057cd9106713cd0b4/clang/include/clang/Basic/AlignedAllocation.h#L30) we need fixed still says 10.14. We could potentially get away with using an Xcode beta once it comes out, though. Xcode 14.3 uses the `swift-5.8-RELEASE` branch, but the latest `swift-5.9` development branch includes [the fix](https://github.com/apple/llvm-project/commit/4722b5d1330a1104a0aa256804b9537c4d2c39d4). I assume the next Xcode beta will be released at WWDC 2023 which is scheduled for June 5

patches/rocksdb_faligned_allocation.patch
thibault-martinez, line 1: Can you maybe put them somewhere else than the root? Or at least `.patches`

rajivshah3: Yep, `.patches` might be easiest to do

rajivshah3 (approved):

DaughterOfMars (approved):
Is this necessary?

rajivshah3:
@Alexandcoats it would be ideal as we've received multiple requests for macOS 10.13 support in Firefly. We were previously able to support it before RocksDB introduced features from C++17. There are more details in the second bullet point of [this issue](https://github.com/iotaledger/wallet.rs/issues/1635). This workaround won't be necessary for too long though, as I think the next major version of Xcode will fix the root issue

thibault-martinez (approved):

thibault-martinez:
Thanks <3

PR #203, opened 2023-04-04
chore(examples): improving syncing in rust examples

### LINKED ISSUE #119: Add `sync` calls to Rust examples.
## Description

Just like in the examples of the other languages, we should have `sync` calls in the Rust examples. For example running the transaction example multiple times currently fails with `Node(NoInputs)`, because the account is never synced in that script.

- [ ] Add `account.sync(None).await?` where appropriate, with a small comment why you'd want to call `sync`.

No.

DaughterOfMars:
# Description of change

Improved the logic in some of the rust examples to avoid needing to manually copy and paste IDs. Added syncing in places where it may be prudent.

- Closes #119 

## Type of change

I ran the examples.

## Change checklist


sdk/examples/wallet/06_send_micro_transaction.rs
```diff
@@ -20,27 +20,31 @@ async fn main() -> Result<()> {
...
-    println!(
-        "Transaction: {} Block sent: {}/api/core/v2/blocks/{}",
-        transaction.transaction_id,
-        &env::var("NODE_URL").unwrap(),
-        transaction.block_id.expect("no block created yet")
-    );
+    // May want to ensure the account is synced before sending a transaction.
+    let balance = account.sync(None).await?;
+
+    if balance.base_coin.available >= 1 {
```
Thoralf-M: Not sure about this check, we check it in `send_micro_transaction()` and it adds noise to the example
Also >= 1 could still not be enough, because the output needs to have 1 + the minimum require storage deposit for a basic output

sdk/examples/wallet/09_mint_native_token.rs
```diff
@@ -52,6 +50,10 @@ async fn main() -> Result<()> {
     };
 
     let transaction = account.mint_native_token(native_token_options, None).await?;
+
+    // Ensure the account is synced after minting.
```
Thoralf-M, line 57: `mint_native_token()` only creates the transaction, but it doesn't wait for it to be confirmed, so if the tx isn't confirmed yet, this is useless
One could add `account.retry_transaction_until_included(transaction.transaction_id, None, None).await?;` to wait for the confirmation (also tries to get it confirmed)

sdk/examples/wallet/06_send_micro_transaction.rs
DaughterOfMars: Okay, makes sense. Is there a way a user would be able to check if they can send micro without trying and failing?

sdk/examples/wallet/09_mint_native_token.rs
DaughterOfMars: Is this also true for the mint_nft example?

Thoralf-M: Yes, actually for all functions that create a transaction

sdk/examples/wallet/06_send_micro_transaction.rs
Thoralf-M: One could build an output like here https://github.com/iotaledger/iota-sdk/blob/aa68f6b6879ae17d661b53469e81add37d707592/sdk/examples/client/output/micro_transaction.rs#L55-L66, just with `BasicOutputBuilder::new_with_minimum_storage_deposit(rent_structure)?` (rent_structure can be retrieved with `client.get_rent_structure().await?`) and then check if the available amount is >= (even then you can't be sure, because if it's not ==, then you need a remainder output which also needs at least the minimum required storage deposit of an output)

So for example minimum required storage deposit is 1_000_000, your available balance is 2_000_000, then you want to send an output with 1_500_000, this would leave only 500_000 behind which would be not enough to cover the storage deposit

DaughterOfMars: Probably better to just try and fail then :/

sdk/examples/wallet/03_get_funds.rs
```diff
@@ -29,6 +29,9 @@ async fn main() -> Result<()> {
     let faucet_response =
         request_funds_from_faucet(&env::var("FAUCET_URL").unwrap(), &address[0].address().to_bech32()).await?;
 
+    // Ensure the account is synced after receiving funds.
+    account.sync(None).await?;
+
     println!("{faucet_response}");
```
Thoralf-M: The faucet also needs some time to send a transaction, so maybe something like this would be better
```suggestion
    println!("{faucet_response}");
    
    // Wait a few seconds for the faucet transaction to get processed
    tokio::time::sleep(std::time::Duration::from_secs(15)).await;

    // Ensure the account is synced after receiving funds.
    account.sync(None).await?;
```

DaughterOfMars: Oof, is there really not a better way than that?

Thoralf-M: Not really, in the tests we do this https://github.com/iotaledger/iota-sdk/blob/f5d4d22ee792300b3664c6a17d42a8916c09973b/sdk/tests/wallet/common/mod.rs#L60-L69
The faucet API doesn't return any information about when the transaction will be sent and even then we don't know when it will be confirmed

sdk/examples/wallet/17_check_unlock_conditions.rs
```diff
@@ -16,6 +16,8 @@ async fn main() -> Result<()> {
 
     // Get the account we generated with `01_create_wallet`
     let account = manager.get_account("Alice").await?;
+    // May want to ensure the account is synced before using output information.
+    account.sync(None).await?;
```
Thoralf-M: For this example it's useless since we build a new output and don't get one from the account

Thoralf-M (dismissed):

kwek20 (dismissed):

thibault-martinez (changes requested):
Needs rebasing

sdk/examples/wallet/16_destroy_alias.rs
```diff
@@ -23,26 +18,28 @@ async fn main() -> Result<()> {
 
     // Get the account we generated with `01_create_wallet`
     let account = wallet.get_account("Alice").await?;
+    // May want to ensure the account is synced before sending a transaction.
+    let balance = account.sync(None).await?;
 
-    let balance = account.balance().await?;
-    println!("Balance before destroying:\n{balance:?}",);
+    // Get the first foundry
```
Thoralf-M:
```suggestion
    // Get the first alias
```

sdk/examples/wallet/12_increase_native_token_supply.rs
```diff
@@ -23,6 +23,8 @@ async fn main() -> Result<()> {
 
     // Get the account we generated with `01_create_wallet`
     let account = wallet.get_account("Alice").await?;
+    // May want to ensure the account is synced before sending a transaction.
+    account.sync(None).await?;
 
     let balance = account.balance().await?;
```
Thoralf-M: Syncing also returns the balance, so would be enough here
```suggestion
    let balance = account.sync(None).await?;
```

Thoralf-M (dismissed):

thibault-martinez (changes requested):
Conflicts

Thoralf-M (dismissed):

sdk/examples/wallet/03_get_funds.rs
```diff
@@ -21,13 +21,33 @@ async fn main() -> Result<()> {
...
+    let start = std::time::Instant::now();
+    let balance = loop {
+        if start.elapsed().as_secs() > 60 {
+            panic!("took too long waiting for funds")
+        };
+        let balance = account.sync(None).await?;
+        if balance.base_coin.available > funds_before {
+            break balance;
+        }
+    };
```
thibault-martinez, line 50: Don't we want a small sleep or something? Looks very expensive and spamming the node

sdk/examples/wallet/05_transaction.rs
```diff
@@ -18,25 +18,34 @@ async fn main() -> Result<()> {
...
+        )];
+        let transaction = account.send_amount(outputs, None).await?;
+
+        // Wait for transaction to get included
+        account
+            .retry_transaction_until_included(&transaction.transaction_id, None, None)
+            .await?;
+
+        println!(
+            "Transaction: {} Block sent: {}/api/core/v2/blocks/{}",
```
thibault-martinez: Do you mind putting a new line after the transaction?

Thoralf-M (dismissed):

sdk/examples/wallet/05_transaction.rs
thibault-martinez: Sorry I didn't want to do one comment for each but there is more

sdk/examples/wallet/08_send_nft.rs
```diff
@@ -23,26 +18,35 @@ async fn main() -> Result<()> {
...
+
+        let transaction = account.send_nft(outputs, None).await?;
+
+        // Wait for transaction to get included
+        account
+            .retry_transaction_until_included(&transaction.transaction_id, None, None)
+            .await?;
+
+        println!(
+            "Transaction: {} Block sent: {}/api/core/v2/blocks/{}",
```
thibault-martinez: :trollface:

sdk/examples/wallet/wallet.rs
```diff
@@ -68,11 +68,11 @@ async fn main() -> Result<()> {
         1_000_000,
     )];
     let tx = account.send_amount(outputs, None).await?;
```
Thoralf-M:
```suggestion
    let transaction = account.send_amount(outputs, None).await?;
```

Thoralf-M (approved):

thibault-martinez (approved):

PR #207, opened 2023-04-05
Unlock unused inputs

### LINKED ISSUE #170: [Bug]: Input locked forever when transaction with ledger fails
### Issue description

```
After I wanted to withdraw my governance vote after the termination, this did not work at first (I used a ledger). 
I've been waiting forever for confirmation! 
At some point I closed the Shimmer wallet and then suddenly the transaction confirmation came on the ledger. 
I then rejected this because shimmer was already closed.
and now shimmer is telling me I can't withdraw the vote because they are being used in another transaction. 
```

This can probably also happen if a program is terminated while being in the transaction function, we could maybe check the locked inputs during syncing and unlock them if they aren't in a pending transaction, just needs to be done carefully so it can't happen for inputs that are selected for a transaction at the same time (not sure if the locks would currently prevent that)

### Version

-

### Expected behaviour

Input should get unlocked again if something in the transaction function fails

### Actual behaviour

Input locked forever

No

_No response_

### Errors

_No response_

### Duplicate declaration

Thoralf-M:
# Description of change

Unlock unused inputs during wallet building
Improve a few debug logs

#170 

## Type of change

Manually locked inputs and checked if they get unlocked again

## Change checklist

### DIFF
--- a/sdk/src/wallet/wallet/builder.rs
+++ b/sdk/src/wallet/wallet/builder.rs
@@ -1,9 +1,12 @@
 // Copyright 2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::sync::{
-    atomic::{AtomicU32, AtomicUsize},
-    Arc,
+use std::{
+    collections::HashSet,
+    sync::{
+        atomic::{AtomicU32, AtomicUsize},
+        Arc,
+    },
 };
 #[cfg(feature = "storage")]
 use std::{path::PathBuf, sync::atomic::Ordering};
@@ -21,7 +24,7 @@ use crate::wallet::storage::adapter::memory::Memory;
 use crate::wallet::storage::{constants::default_storage_path, manager::ManagerStorage};
 use crate::{
     client::secret::SecretManager,
-    wallet::{AccountHandle, ClientOptions, Wallet},
+    wallet::{account::Account, AccountHandle, ClientOptions, Wallet},
 };
 
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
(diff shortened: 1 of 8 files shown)


sdk/src/wallet/wallet/builder.rs
```diff
@@ -191,7 +194,11 @@ impl WalletBuilder {
         let event_emitter = Arc::new(Mutex::new(EventEmitter::new()));
 
         #[cfg(feature = "storage")]
-        let accounts = storage_manager.lock().await.get_accounts().await.unwrap_or_default();
+        let mut accounts = storage_manager.lock().await.get_accounts().await.unwrap_or_default();
+        #[cfg(feature = "storage")]
+        // It happened that inputs got locked, the transaction failed, but they weren't unlocked again, so we do this
```
thibault-martinez, line 198: Comment before attribute

sdk/src/wallet/wallet/builder.rs
```diff
@@ -191,7 +194,11 @@ impl WalletBuilder {
         let event_emitter = Arc::new(Mutex::new(EventEmitter::new()));
 
         #[cfg(feature = "storage")]
-        let accounts = storage_manager.lock().await.get_accounts().await.unwrap_or_default();
+        let mut accounts = storage_manager.lock().await.get_accounts().await.unwrap_or_default();
+        // It happened that inputs got locked, the transaction failed, but they weren't unlocked again, so we do this
+        // here
+        #[cfg(feature = "storage")]
+        unlock_unused_inputs(&mut accounts)?;
```
thibault-martinez, line 201: Is it enough to do that only when building a wallet ?

Thoralf-M: I think so, normally this shouldn't even happen, that's why I also added more matches on the errors to unlock the inputs if something in the transaction process fails
Also tried to clean it up during syncing, but that doesn't work in a safe way with the way we create the transactions. To make that safe we would need to add a new lock or change it so the account gets locked the whole time

sdk/src/wallet/wallet/builder.rs
```diff
@@ -251,3 +258,26 @@ impl WalletBuilder {
...
+    for account in accounts.iter_mut() {
+        let mut used_inputs = HashSet::new();
+        for transaction_id in account.pending_transactions() {
+            if let Some(tx) = account.transactions().get(transaction_id) {
+                for input in &tx.inputs {
+                    used_inputs.insert(input.metadata.output_id()?);
+                }
+            }
+        }
+        for input in account.locked_outputs().clone() {
```
thibault-martinez: Can you maybe use retain to avoid cloning ?

Thoralf-M: Done

thibault-martinez (dismissed):

thibault-martinez (approved):

Alex6323 (approved):

PR #229, opened 2023-04-09
add class IotaBlock to python binding

mp-v2:
# Description of change

This PR adds to the python binding to make encoding data to byte format and then to hex format easier.
It adds the class IotaBlock in the same structure as the IotaClient class, and uses similar terminology to existing python binding example scripts.

## Type of change

Used the class, ran some different tests manually.
python3.5 or higher needed for the hex conversion builtin functions of python

## Change checklist

### DIFF
--- a/sdk/src/client/bindings/python/iota_client/block.py
+++ b/sdk/src/client/bindings/python/iota_client/block.py
@@ -1,0 +1,113 @@
+from iota_client import IotaClient
+import os
+
+class IotaBlock:
+    """
+    A class for handling the block and payload.
+
+    Load data into the block with .load_string() or .load_file()
+    Review the block payload with .summary()
+    Yield the payload with .payload()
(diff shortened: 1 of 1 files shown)


Thoralf-M:
Hello, thanks for this PR, but this class looks very specific with all the prints and load_file option, very different to the other methods in the library
Also, we want to align our bindings as much as possible, so we try to avoid adding classes only for a single language
If you want to contribute, maybe you would like to add more types as visible in https://github.com/iotaledger/iota-sdk/pull/256? The Rust definitions are here https://github.com/iotaledger/iota-sdk/tree/develop/sdk/src/types/block and the types the bindings get are the DTOs, for example https://github.com/iotaledger/iota-sdk/blob/33158aa3cbcf58b17145e3330d94098d5d5bce68/sdk/src/types/block/block.rs#L265

PR #236, opened 2023-04-11
feat(cli): prompt on first run

### LINKED ISSUE #146: CLI: prompt to create a new account on first run

### LINKED ISSUE #251: CLI: Make prompt distinguishable from command output
## Description

I'ld enjoy a different style of prompt in the CLI. Can be a different color or a different font style.

## Motivation

It makes working with the CLI wallet a bit more pleasant as it helps with better distinguishing between what the users enters (the command) and the command output like balances, addresses, etc.

## Requirements

1. add `console = { version = "0.15.5", default-features = false }` to Cargo.toml of the CLI
2. add this line: `console::style(format!("Account \"{}\"", alias)).green().to_string())`
3. enjoy a green prompt (for example)

Yes.

Alex6323:
# Description of change

This PR prompts for a decision on how to proceed on first run. Currently the wallet app would just create a wallet with default parameters. The new behavior asks the user whether it should do that. Additionally if the answer was "yes", the wallet will ask about creating a named initial account.

List of changes:
* guides the user towards creating an initial wallet and inital account
* reports an error if either the wallet database or the stronghold snapshot file is missing (either none of them can exist (initial state), or both of them exist (any valid future state))
* makes the wallet command handling more organized and easier to read
* adds some more dialoguer helper fns
* shows a hint how to see all available commands when the initial account has been created
* prints a help dialog when the answer to create an initial wallet was "no"
* colorizes the account prompt to make it easier to distinguish from command outputs

Closes #146 
Closes #251 

## Type of change

## Change checklist


cli/src/wallet.rs
```diff
@@ -28,25 +28,25 @@ pub async fn new_wallet(cli: WalletCli) -> Result<(Option<Wallet>, Option<String
         || "./stardust-cli-wallet-db".to_string(),
         |os_str| os_str.into_string().expect("invalid WALLET_DATABASE_PATH"),
     );
-    let snapshot_path = std::path::Path::new("./stardust-cli-wallet.stronghold");
-    let snapshot_exists = snapshot_path.exists();
+    let stronghold_path = std::path::Path::new("./stardust-cli-wallet.stronghold");
```
thibault-martinez: Revert please, snapshot is correct and consistently used internally.

Alex6323: okay

Thoralf-M (dismissed):

cli/src/wallet.rs
```diff
@@ -70,24 +74,47 @@ pub async fn new_wallet(cli: WalletCli) -> Result<(Option<Wallet>, Option<String
...
+            if get_decision("Initialize a default account?")? {
+                println_log_info!("Initializing default account.");
+                let account = new_command(&wallet, None).await?;
+                (Some(wallet), Some(account))
+            } else {
+                (Some(wallet), None)
+            }
+        } else if let Some(account_index) = pick_account(&wallet).await? {
+            let accounts = wallet.get_accounts().await?;
+            let account = accounts[account_index as usize].alias().await;
```
Thoralf-M: This should also work
```suggestion
            let account = wallet.get_account(account_index).await?;
```

Thoralf-M: Actually `pick_account()` can return the account handle directly I think

cli/src/wallet.rs
```diff
@@ -3,117 +3,105 @@
...
-            if get_decision("Initialize a default account?")? {
-                println_log_info!("Initializing default account.");
-                let account = new_command(&wallet, None).await?;
-                (Some(wallet), Some(account))
-            } else {
-                (Some(wallet), None)
+                    // ask the new user whether a default account should be created
+                    if get_decision("Initialize a default account?")? {
+                        println_log_info!("Initializing default account.");
+                        let account= add_account(&wallet, None).await?;
```
Thoralf-M: fmt

Thoralf-M (dismissed):

thibault-martinez:
Conflicts

thibault-martinez (changes requested):
Conflicts

Thoralf-M (dismissed):

Thoralf-M (dismissed):

cli/src/helper.rs
```diff
@@ -22,26 +22,42 @@ pub fn get_password(prompt: &str, confirmation: bool) -> Result<String, Error> {
...
 
-pub async fn pick_account(wallet: &Wallet) -> Result<Option<u32>, Error> {
-    let accounts = wallet.get_accounts().await?;
+pub fn get_decision(prompt: &str) -> Result<bool, Error> {
+    let input = Input::<String>::new()
+        .with_prompt(prompt)
+        .default("Yes".into())
+        .interact_text()?;
+
+    match input.as_str() {
```
DaughterOfMars: Probably should do `input.to_lowercase()` and only match on lowercase values.

cli/src/wallet.rs
```diff
@@ -1,88 +1,105 @@
...
-            };
+            }
+            (false, false) => {
+                if get_decision("Initialize a new wallet with default values?")? {
+                    let wallet = init_command(storage_path, snapshot_path, InitParameters::default()).await?;
+                    println_log_info!("Initialized wallet with default values.");
 
-            (wallet, account)
+                    // ask the user whether a default account should be created
+                    if get_decision("Initialize a default account?")? {
```
DaughterOfMars: Can this allow naming the account?

cli/src/wallet.rs
```diff
@@ -1,88 +1,105 @@
...
-                WalletCommand::ChangePassword => change_password_command(&wallet, &password).await?,
-                WalletCommand::New { alias } => account = Some(new_command(&wallet, alias).await?),
-                WalletCommand::SetNode { url } => set_node_command(&wallet, url).await?,
-                WalletCommand::Sync => sync_command(&wallet).await?,
-                // PANIC: this will never happen because these variants have already been checked.
-                WalletCommand::Init(_) | WalletCommand::Mnemonic | WalletCommand::Restore { .. } => unreachable!(),
-            };
+            }
+            (false, false) => {
+                if get_decision("Initialize a new wallet with default values?")? {
```
DaughterOfMars: What if they say no? Should we print help about initializing the wallet then exit?

cli/src/wallet.rs
```diff
@@ -1,88 +1,105 @@
...
+                    let wallet = init_command(storage_path, snapshot_path, InitParameters::default()).await?;
+                    println_log_info!("Initialized wallet with default values.");
 
-            (wallet, account)
+                    // ask the user whether a default account should be created
+                    if get_decision("Initialize a default account?")? {
+                        let account = add_account(&wallet, None).await?;
+                        println_log_info!("Initialized default account.");
+                        (Some(wallet), Some(account))
```
DaughterOfMars: Once the account is created, you are forwarded to the account prompt with no instruction. I think it should automatically print the help page here.

cli/src/wallet.rs
```diff
@@ -1,88 +1,105 @@
...
-                WalletCommand::New { alias } => account = Some(new_command(&wallet, alias).await?),
-                WalletCommand::SetNode { url } => set_node_command(&wallet, url).await?,
-                WalletCommand::Sync => sync_command(&wallet).await?,
-                // PANIC: this will never happen because these variants have already been checked.
-                WalletCommand::Init(_) | WalletCommand::Mnemonic | WalletCommand::Restore { .. } => unreachable!(),
-            };
+            }
+            (false, false) => {
+                if get_decision("Initialize a new wallet with default values?")? {
+                    let wallet = init_command(storage_path, snapshot_path, InitParameters::default()).await?;
```
DaughterOfMars, line 72: This command prompts the user to store the mnemonic without telling them what is happening. It is a bit confusing:

```
Initialize a new wallet with default values?: y
Stronghold password: [hidden]
Mnemonic has been generated.
❯ Write to console only
  Write to file only
  Write to console and file
```

It should perhaps say: `Where do you want to store the generated mnemonic?`

cli/src/helper.rs
Alex6323: yeah, although if someone writes YES or yEs s/he belongs to a prison!

cli/src/wallet.rs
Alex6323: maybe ... I wanted to be as less intrusive and change inducing as possible, because people will have very different and possibly strong opinions on that. I would - for example - enforce "named accounts" only, hence getting rid of numbered accounts. But that's something that should be discussed in the team, and probably be a separate PR then.

cli/src/wallet.rs
Alex6323: if they say "no" they have to go through the subcommands and provide specifc, i.e. non-default, params.

cli/src/wallet.rs
Alex6323: I tend to agree but there may be different opinions here which is why I didn't include that.

cli/src/wallet.rs
Alex6323: I have no problem with changing it, but the term "store" would be confusing here IMO because in the text that comes afterwards we are requesting the user to store it. So it's really just written either to the console or to a file, which then should be stored by the user, either by copying it to a secure location from the console, or by moving the created "mnemonic.txt" file to a secure/encrypted device.

cli/src/wallet.rs
DaughterOfMars: Exactly, so then we should print the help page for that so it's clear what needs to happen.

cli/src/wallet.rs
thibault-martinez: I think I would tend to agree here, I would ask for an account name

cli/src/wallet.rs
thibault-martinez: I don't disagree, but maybe for consistency there are other places where this should be done?

cli/src/wallet.rs
thibault-martinez: I think it would already be a lot clearer if we say "Write it to..." then it's obvious we're speaking about the mnemonic

cli/src/helper.rs
Alex6323: Done

cli/src/wallet.rs
Alex6323: Done

cli/src/wallet.rs
Alex6323: right ... one small word can make all the difference

cli/src/wallet.rs
Alex6323: can you elaborate Thibault? Why not just print the help on first run?

cli/src/helper.rs
```diff
@@ -25,14 +25,14 @@ pub fn get_password(prompt: &str, confirmation: bool) -> Result<String, Error> {
...
     let input = Input::<String>::new()
         .with_prompt(prompt)
-        .default("Yes".into())
+        .default("yes".into())
         .interact_text()?;
 
-    match input.as_str() {
-        "Yes" | "yes" | "y" => Ok(true),
-        "No" | "no" | "n" => Ok(false),
+    match input.to_lowercase().as_str() {
```
DaughterOfMars: Might be worth a loop here too, in case someone typos

cli/src/helper.rs
```diff
@@ -27,18 +27,23 @@ pub fn get_password(prompt: &str, confirmation: bool) -> Result<String, Error> {
...
-            expected: "yes|y or no|n".to_string(),
-            found: input,
-        }),
+    loop {
+        match input.to_lowercase().as_str() {
+            "yes" | "y" => return Ok(true),
+            "no" | "n" => return Ok(false),
+            _ => {
+                println_log_info!("Accepted input values are: yes|y|no|n");
+                input = Input::<String>::new()
```
DaughterOfMars: Can reduce code by moving this to the top of the loop.

cli/Cargo.toml
```diff
@@ -15,6 +15,7 @@ path = "src/main.rs"
 iota-sdk = { path = "../sdk", default-features = false, features = [ "wallet", "tls", "storage", "rocksdb", "stronghold", "participation" ] }
 
 clap = { version = "4.2.0", default-features = false, features = [ "std", "color", "help", "usage", "error-context", "suggestions", "derive", "env" ] }
+console = { version = "0.15.5", default-features = false }
```
DaughterOfMars: This crate seems iffy. Maybe [colored](https://docs.rs/colored/latest/colored/) is better?

thibault-martinez: `colored` is also used by fern so it should already be in the dependency tree

cli/src/helper.rs
```diff
@@ -61,9 +106,9 @@ pub async fn generate_mnemonic() -> Result<String, Error> {
     println_log_info!("Mnemonic has been generated.");
```
kwek20, line 129: For me it wasnt clear that the write Select was related to the mnemonic. Lets add
`Mnemonic has been generated. Do you want to: `

cli/src/wallet.rs
```diff
@@ -1,88 +1,110 @@
...
-            (wallet, account)
+                    // ask the user whether a default account should be created
+                    if get_decision("Create initial account?")? {
+                        let alias = get_account_name("New account name", &wallet).await?;
+                        let account = add_account(&wallet, Some(alias)).await?;
+                        println_log_info!("Created initial account.");
+                        print_account_help();
+                        (Some(wallet), Some(account))
+                    } else {
+                        (Some(wallet), /*  */ None)
```
kwek20: leftover comment section here

cli/src/wallet.rs
kwek20: I thought it was suddenly a lot of text, maybe just say "type help for a list of commands" ?

sdk/src/wallet/account/update.rs
```diff
@@ -24,7 +24,7 @@ use crate::{
 
 impl AccountHandle {
     /// Set the alias for the account
-    pub async fn set_alias(&self, alias: &str) -> crate::wallet::Result<()> {
+    pub async fn set_alias(&mut self, alias: &str) -> crate::wallet::Result<()> {
```
kwek20: this isnt needed right?

Alex6323: right, good catch

cli/src/wallet.rs
Alex6323: thanks

cli/src/helper.rs
Alex6323: okay, I understand. but we should probably put it in a separate line with just a `println!` so it doesn't get logged.

cli/src/wallet.rs
Alex6323: yeah, I thought about that as well ... also a very good option I think.

kwek20 (dismissed):

DaughterOfMars:
inb4: `final final touches`

Alex6323:
very finalst touches incoming

DaughterOfMars:
that's cheating

kwek20:
Conflicts :trollface:

Alex6323:
I feel like someone overtook me on the right :racing_car:

Thoralf-M (changes requested):
Conflicting again :sadcat:

Thoralf-M (dismissed):

cli/src/wallet.rs
```diff
@@ -1,88 +1,108 @@
...
+                (Some(wallet), None)
+            }
+            WalletCommand::Mnemonic => {
+                mnemonic_command().await?;
+                return Ok((None, None));
+            }
+        }
+    } else {
+        // no command provided, i.e. `> ./wallet`
+        match (storage_path.exists(), snapshot_path.exists()) {
```
DaughterOfMars, line 57: Seems like there is a lot of duplicated code here. Can you not have separate match statements to dedup?

Alex6323: hmmm. There's duplicated code indeed. You mean this here?
```
if get_decision("Create initial account?")? {
 ...    
}
```
This could be factored out into a method, yes.

DaughterOfMars: I meant more that there's only one match that handles various "states" when it should be more like a waterfall of states.

DaughterOfMars: Could use a loop around the match, for instance.

Alex6323: hmmm, thinking about it

cli/src/account.rs
```diff
@@ -41,21 +41,18 @@ pub async fn account_prompt(account: Account) -> Result<(), Error> {
 
 // loop on the account prompt
 pub async fn account_prompt_internal(account: Account, history: &mut AccountHistory) -> Result<bool, Error> {
+    use colored::Colorize;
```
thibault-martinez: Put it with the others

cli/src/command/wallet.rs
```diff
@@ -168,16 +204,35 @@ pub async fn restore_command(
...
     wallet.set_client_options(ClientOptions::new().with_node(&url)?).await?;
 
-    Ok(())
+    Ok(wallet)
 }
 
-pub async fn sync_command(wallet: &Wallet) -> Result<(), Error> {
+pub async fn sync_command(storage_path: &Path, snapshot_path: &Path) -> Result<Wallet, Error> {
+    let password = get_password("Stronghold password", !snapshot_path.exists())?;
+    let wallet = unlock_wallet(storage_path, snapshot_path, &password).await?;
```
thibault-martinez, line 204: What's up with all this duplicated code? Looks like it was better before.

cli/src/helper.rs
```diff
@@ -23,30 +27,72 @@ pub fn get_password(prompt: &str, confirmation: bool) -> Result<String, Error> {
...
                 .interact_on(&Term::stderr())?;
 
-            Ok(Some(index as u32))
+            Ok(Some(accounts.swap_remove(index)))
         }
     }
 }
 
+pub fn print_wallet_help() {
+    use clap::Parser;
```
thibault-martinez: Put with the others

cli/src/command/wallet.rs
Alex6323: well then let's not merge it

thibault-martinez: I'm just speaking about this specific change...

Alex6323: yeah, talking about the same (not merging this change) ... if we undo that we get back to how `new_wallet` fn was implemented before.

thibault-martinez: Ah alright, your answer seemed very passive aggressive :trollface:

Alex6323: yeah, I realized it could be picked up this way ... but I tried to reduce repetition as much as possible. I was aware of that. I can look into deduping a bit more as I did with `unlock_wallet`, but honestly ... many commands require similar preconditions, but not identical, otherwise I would have passed them in.

DaughterOfMars: I think, as I said in the other comment, that we can dedup by changing the structure of that code rather than separating into another method.

Alex6323: But you were talking about a different module, Alex (src/wallet.rs) while this is about (src/command/wallet.rs)

DaughterOfMars: Ah, ignore me!

Alex6323: all good
@thibault-martinez what exactly is duplicated code? Those are just function calls, not repeated logic code.

Alex6323: I found the change worth it since it makes the `match` in the `new_wallet` fn really flat and deduped(!!) compared to what we had before.

Alex6323: like this in particular was really disturbing to me (which is repeated code):
```
// PANIC: this will never happen because these variants have already been checked.
WalletCommand::Init(_) | WalletCommand::Mnemonic | WalletCommand::Restore { .. } => unreachable!(),
```

cli/src/wallet.rs
Alex6323: it's hard for me to say what to make from your proposal ... maybe @thibault-martinez wants to chime in on it. 
The way I see it (which may be absolutely wrong), we have 4 possible situations/states we can run into when starting up the app.
1. `.stronghold` file and `wallet-db` exist -> all good -> log in -> send me all your tokens
2. neither of them exist -> all good ... we can fix that -> guide the user through the setup -> enter account -> send me all tokens
3. `wallet-db` exists, but `.stronghold` file does not -> now that is weird (was .stronghold moved?) something fishy seems to be going -> for security reasons abort the app with error message
4. `.stronghold` exists, but `wallet-db` does not -> seems fixable but it's still strange, because on init we usually create both, so something fishy may be going on ... I decided FOR NOW that we should abort as well.

Alex6323: And my problem now is that I cannot bring my logic together with your `loop`ing idea :sweat_smile:

DaughterOfMars: The loop would just match to check what state we're in, then perform only the thing to go to the next state.

Alex6323: Feel free to push your idea to the branch if you want, and we can discuss it.

DaughterOfMars: I took a closer look and changed my mind.

cli/src/command/wallet.rs
thibault-martinez: I don't think this is repeated code. It was to be explicit but we can easily just replace the whole thing by `_`

sdk/src/wallet/wallet/mod.rs
```diff
@@ -74,6 +74,16 @@ impl Wallet {
         Ok(self.accounts.read().await.clone())
     }
 
+    /// Get all account aliases
+    pub async fn get_account_aliases(&self) -> crate::wallet::Result<Vec<String>> {
```
thibault-martinez, line 78: Changelog

cli/src/command/wallet.rs
```diff
@@ -93,31 +98,40 @@ impl Default for InitParameters {
...
-
-    wallet.change_stronghold_password(current, &new).await?;
+pub async fn change_password_command(storage_path: &Path, snapshot_path: &Path) -> Result<Wallet, Error> {
+    let password = get_password("Stronghold password", !std::path::Path::new(snapshot_path).exists())?;
+    let wallet = unlock_wallet(storage_path, snapshot_path, &password).await?;
+    let new_password = get_password("Stronghold new password", true)?;
+    wallet.change_stronghold_password(&password, &new_password).await?;
 
-    Ok(())
+    Ok(wallet)
```
kwek20, line 117: why do we return wallet here?

cli/src/helper.rs
```diff
@@ -25,30 +33,70 @@ pub fn get_password(prompt: &str, confirmation: bool) -> Result<String, Error> {
...
+            "yes" | "y" => return Ok(true),
+            "no" | "n" => return Ok(false),
+            _ => {
+                println_log_error!("Accepted input values are: yes|y|no|n");
+            }
+        }
+    }
+}
+
+pub async fn get_account_name(prompt: &str, wallet: &Wallet) -> Result<String, Error> {
```
thibault-martinez: Rename to `get_account_alias`

cli/src/wallet.rs
```diff
@@ -1,88 +1,103 @@
...
+        }
     };
+    Ok((wallet, account))
+}
 
-    Ok((Some(wallet), account))
+async fn create_initial_account(wallet: Wallet) -> Result<(Option<Wallet>, Option<String>), Error> {
+    // Ask the user whether an initial account should be created.
+    if get_decision("Create initial account?")? {
+        let alias = get_account_name("New account name", &wallet).await?;
```
thibault-martinez: alias

cli/src/helper.rs
```diff
@@ -25,30 +33,70 @@ pub fn get_password(prompt: &str, confirmation: bool) -> Result<String, Error> {
...
+    let accounts = wallet.get_account_aliases().await?;
+    loop {
+        let input = Input::<String>::new().with_prompt(prompt).interact_text()?;
+        if input.is_empty() || !input.is_ascii() {
+            println_log_error!("Invalid input, please choose a non-empty name consisting of ASCII characters.");
```
thibault-martinez: Is it even possible to create an empty one? Don't think Input lets you do it

cli/src/helper.rs
Alex6323: Done

cli/src/wallet.rs
Alex6323: Done

cli/src/helper.rs
Alex6323: you're right, but I didn't want to rely on that if they decide to change that behavior. Easy to forget adding this check then.

sdk/src/wallet/wallet/mod.rs
thibault-martinez: It conflicts

cli/src/helper.rs
thibault-martinez: Hmm ok

cli/src/command/wallet.rs
Alex6323: I still think that the new impl of `new_wallet` is a lot cleaner and less repetitive because it's doing only one pattern matching on the `cli.command` while the current impl does it 3 times.

thibault-martinez: The difference in processing of these variants exists and we can't do anything about it. You may be removing some matches but you are just moving the difference down the line in the individual functions, adding some more code in each of them that was not previously needed. I guess it's a matter or preference but I personally don't really find it cleaner.

Alex6323: yeah, I am moving as much as possible into those functions because that is where I think the logic should be. `new_wallet` for me is a typical control-flow like function that just delegates/points to some logic, but doesn't contain a lot of logic itself. It's a preference ... I agree and if people don't like it I revert.

cli/src/command/wallet.rs
```diff
@@ -93,31 +98,40 @@ impl Default for InitParameters {
...
 
     Ok(())
 }
 
-pub async fn change_password_command(wallet: &Wallet, current: &str) -> Result<(), Error> {
-    let new = get_password("Stronghold new password", true)?;
-
-    wallet.change_stronghold_password(current, &new).await?;
+pub async fn change_password_command(storage_path: &Path, snapshot_path: &Path) -> Result<Wallet, Error> {
+    let password = get_password("Stronghold password", !std::path::Path::new(snapshot_path).exists())?;
```
thibault-martinez: It's already a Path?

cli/src/command/wallet.rs
```diff
@@ -142,9 +156,21 @@ pub async fn mnemonic_command() -> Result<(), Error> {
...
+) -> Result<(Wallet, String), Error> {
+    let password = get_password("Stronghold password", !snapshot_path.exists())?;
+    let wallet = unlock_wallet(storage_path, snapshot_path, &password).await?;
 
+    let alias = add_account(&wallet, alias).await?;
+
+    Ok((wallet, alias))
+}
+
+pub async fn add_account(wallet: &Wallet, alias: Option<String>) -> Result<String, Error> {
```
thibault-martinez: Can you put this at the bottom pls

cli/src/command/wallet.rs
```diff
@@ -142,9 +156,21 @@ pub async fn mnemonic_command() -> Result<(), Error> {
...
+    let mut builder = wallet.create_account();
     if let Some(alias) = alias {
         builder = builder.with_alias(alias);
```
thibault-martinez: FYI, this now takes an `impl Into<Option<String>>` so we don't need to if let anymore

cli/src/command/wallet.rs
Alex6323: lol, yeah (Rust is really good for refactoring, but nothing is perfect and I am ... unfortunately just a human). Good catch!!

thibault-martinez (approved):
I don't like all of this and maybe we can get back to it later, but I'll approve for now, It's overall a good PR.

DaughterOfMars (approved):

PR #267, opened 2023-04-16
SecretManage::Error and StorageProvider::Error

### LINKED ISSUE #45: Investigate if SecretManage::Error is possible
No way to get dedicated errors for ledger and stronghold without it.
- [x] Ledger error
- [x] Stronghold error

thibault-martinez:
Fixes https://github.com/iotaledger/iota-sdk/issues/45


sdk/src/client/secret/ledger_nano.rs
```diff
@@ -179,11 +242,6 @@ impl SecretManageExt for LedgerSecretManager {
             });
         }
 
-        if coin_type.is_none() || account_index.is_none() {
-            return Err(InputSelectionError::NoAvailableInputsProvided)?;
-        }
```
thibault-martinez, line 184: This should never happen anyway?

Thoralf-M: Well yes it shouldn't happen, but since the values are optional it can happen and then it would panic if we don't return an error before

thibault-martinez: But I mean, we already return an error if it's None on L227 so this is never going to happen ?

thibault-martinez: Wait is it legal to have a `PreparedTransactionData` with no input at all?

Thoralf-M: That can't be correct, but yeah currently it's possible to construct it since it's public

thibault-martinez: Meh, I'll revert this then

sdk/src/client/stronghold/error.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+/// Error type of the iota client crate.
```
Thoralf-M:
```suggestion
/// Error type of the stronghold module.
```
or whatever

sdk/src/client/secret/ledger_nano.rs
thibault-martinez: Done

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -42,30 +38,99 @@ use crate::{
 /// See also: <https://wiki.trezor.io/Hardened_and_non-hardened_derivation>.
 pub const HARDENED: u32 = 0x8000_0000;
 
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
Thoralf-M, line 43: Missing documentation?

thibault-martinez: Done

Thoralf-M (approved):

sdk/src/client/secret/mod.rs
```diff
@@ -235,17 +240,13 @@ impl SecretManage for SecretManager {
     ) -> crate::client::Result<Vec<Address>> {
         match self {
             #[cfg(feature = "stronghold")]
-            Self::Stronghold(secret_manager) => {
-                secret_manager
-                    .generate_addresses(coin_type, account_index, address_indexes, options)
```
kwek20, line 240: Why are we wrapping in Ok here and the other places?

thibault-martinez: because I'm using `?` to convert the error (which is not a ClientError anymore) so we need to rewrap it

cli/src/error.rs
```diff
@@ -32,3 +32,9 @@ impl From<ClientError> for Error {
         Error::Client(Box::new(error))
     }
 }
+
+impl From<iota_sdk::client::stronghold::Error> for Error {
+    fn from(error: iota_sdk::client::stronghold::Error) -> Self {
+        Self::Client(Box::new(iota_sdk::client::Error::Stronghold(error)))
```
kwek20, line 38: Why box here but not on the spot above? Box is only needed for recursive errors and large enum things right?

thibault-martinez: Not sure what you are asking, what is the "spot above" ?

kwek20: bindings/core/src/error.rs

thibault-martinez: Ah, I see there is a `#[allow(clippy::large_enum_variant)]` there. We have an issue to change that already 👍🏻

sdk/src/client/secret/mod.rs
kwek20: Oh i see, ? acts as .into() but unwraps. didnt expect that it worked like that :)

kwek20 (approved):

PR #304, opened 2023-04-21
Builders use `impl Into<Option<V>>` instead of `V`

### LINKED ISSUE #223: Use `set_v(v: impl Into<Option<V>>)` instead of `set_v(v: V)`
Helps with overall usability.
We may have some `set_*`, `with_*`, ...

thibault-martinez:
Fixes https://github.com/iotaledger/iota-sdk/issues/223


DaughterOfMars (dismissed):
Looks good. Only consideration I would add is that we could look into adding [Non-Generic Inner Functions](https://www.possiblerust.com/pattern/non-generic-inner-functions) for reduced binary sizes.

DaughterOfMars (dismissed):

DaughterOfMars (approved):

kwek20 (approved):

PR #335, opened 2023-04-25
Add amount claim

### LINKED ISSUE #212: Add OutputsToClaim::Amount variant
## Description

Add Amount variant to OutputsToClaim

## Motivation

Claim only outputs where we get more amount, for example basic output with expiration unlock condition

## Requirements

1. Add `Amount` variant to https://github.com/iotaledger/iota-sdk/blob/b5a72f8aea9be4dc4bc0b59087e910f6d24bcc35/sdk/src/wallet/account/operations/output_claiming.rs#L25
2. Handle this case here https://github.com/iotaledger/iota-sdk/blob/b5a72f8aea9be4dc4bc0b59087e910f6d24bcc35/sdk/src/wallet/account/operations/output_claiming.rs#L73-L87 and only add outputs where we get more amount

No.

DaughterOfMars:
# Description of change

Adds the Amount claiming variant to claim outputs that net amount to the account.

- Fixes #212 

## Type of change

Added a test case.


sdk/src/wallet/account/operations/output_claiming.rs
```diff
@@ -27,7 +27,8 @@ pub enum OutputsToClaim {
     MicroTransactions = 1,
     NativeTokens = 2,
     Nfts = 3,
-    All = 4,
+    Amount = 4,
+    All = 5,
```
thibault-martinez: Do they actually have to have discriminants ?

sdk/src/wallet/account/operations/output_claiming.rs
```diff
@@ -50,88 +51,55 @@ impl Account {
...
+                            &[],
+                            output_data,
+                            local_time,
+                            // Not relevant without alias addresses
+                            None,
+                        )?
+                    {
+                        match outputs_to_claim {
+                            OutputsToClaim::MicroTransactions => {
+                                if let Some(sdr) = unlock_conditions.storage_deposit_return() {
```
Thoralf-M, line 77: I know, not the part you added, but what do you think about adding this here?
```suggestion

                                if let Some(sdr) = unlock_conditions.storage_deposit_return() {
                                    // If expired, it's not a micro transaction anymore
                                    if unlock_conditions.is_expired(local_time) {
                                        continue;
                                    }
```

sdk/src/wallet/account/operations/output_claiming.rs
DaughterOfMars: I don't think so. I was wondering the same thing tbh

Thoralf-M (dismissed):

sdk/src/wallet/account/operations/output_claiming.rs
```diff
@@ -50,88 +50,63 @@ impl Account {
         let mut output_ids_to_claim: HashSet<OutputId> = HashSet::new();
         for (output_id, output_data) in &account_details.unspent_outputs {
             // Don't use outputs that are locked for other transactions
-            if !account_details.locked_outputs.contains(output_id) {
-                if let Some(output) = account_details.outputs.get(output_id) {
-                    match &output.output {
-                        Output::Basic(basic_output) => {
```
thibault-martinez, line 56: Is it correct to remove Basic/Nft match? Isn't it going to pick aliases and foundries?

Thoralf-M (approved):

sdk/src/wallet/account/operations/output_claiming.rs
Thoralf-M: Good catch, still needs to be handled

Thoralf-M (changes requested):
Unresolved comment

sdk/src/wallet/account/operations/output_claiming.rs
thibault-martinez: Maybe you can simply filter the iterator by `is_basic || is_nft` @Alexandcoats

Thoralf-M (approved):

thibault-martinez (approved):

PR #339, opened 2023-04-26
Feat/node error enum

### LINKED ISSUE #190: Split Error::Node(String) into multiple variants
## Description

Split Error::Node(String) into multiple error variants on which one can match (without matching on the String :trollface: )

## Motivation

Improve error handling, less mistakes possibles

## Requirements

1. Split errors we can have there into multiple variants
2. Also check other error variants that contain a String if it makes sense to split them

No.

kwek20:
# Description of change

Adds a node Error type with Node specific errors

Fixes #190 

## Type of change

Ran on a private tangle


sdk/src/client/node_api/mod.rs
```diff
@@ -4,6 +4,8 @@
 //! node API modules
 
 pub mod core;
+
```
thibault-martinez:
```suggestion
```

sdk/src/client/node_manager/http_client.rs
```diff
@@ -52,11 +52,23 @@ impl HttpClient {
         if status.is_success() {
             Ok(Response(response))
         } else {
-            Err(Error::ResponseError {
-                code: status.as_u16(),
-                text: response.text().await?,
-                url: url.to_string(),
-            })
+            let text = response.text().await?;
+            // hornet and bee return different error blocks
```
thibault-martinez: Don't bother about bee

sdk/src/client/node_api/indexer/routes.rs
```diff
@@ -59,7 +59,7 @@ impl Client {
             .get_output_ids(&route, QueryParameters::empty(), true, false)
             .await?
             .first()
-            .ok_or_else(|| crate::client::Error::Node("no output id for alias".to_string()))?))
+            .ok_or_else(|| Error::NoOutput("alias"))?))
```
thibault-martinez: Shouldn't this be a node error?

kwek20: Ive tried to divide it between results from 1 node, and node_manager/quorum stuff, so a single node will return 404 not found(node_api:Error::NotFound), then at node_manager will do error.replace() or return NotFound. Since theres 3 cases for this, i made it specific for NoOutput of all nodes (so not a specific node error).

But yea the NoResult, NotFound and NoOutput are difficult to separate tbh

sdk/src/wallet/account/operations/syncing/foundries.rs
```diff
@@ -29,7 +29,7 @@ impl Account {
                 task::spawn(async move {
                     match client.foundry_output_id(foundry_id).await {
                         Ok(output_id) => Ok(Some(client.get_output(&output_id).await?)),
-                        Err(crate::client::Error::NotFound(_)) => Ok(None),
```
kwek20, line 32: so here we used to return notFound when ResponseError { 404 }, which now turns per node into NotFound, which turns into a NoOutput here https://github.com/iotaledger/iota-sdk/pull/339/files#diff-66ee4221c238c720dbef1e5f6706bbeeeb1d3655a41bef1ee411cfeed99c2eb2R87

Its confusing :D

sdk/src/client/error.rs
```diff
@@ -217,6 +201,13 @@ pub enum Error {
     Stronghold(#[from] crate::client::stronghold::Error),
 }
 
+// Used in faucet funds request util
+impl From<reqwest::Error> for Error {
```
DaughterOfMars: If this Error can be created from a reqwest error, then the variant should probably just stay.

sdk/src/client/node_api/core/routes.rs
```diff
@@ -236,55 +227,45 @@ impl Client {
...
-                    } else {
-                        return Err(Error::Node(e));
-                    }
-                } else {
-                    return Err(e);
+            Err(Error::Node(crate::client::node_api::error::Error::UnavailablePow)) => {
+                if !self.get_fallback_to_local_pow() {
+                    return Err(Error::Node(crate::client::node_api::error::Error::UnavailablePow));
+                }
+                // Without this we get:within `impl Future<Output = [async output]>`, the trait `Send` is not
```
DaughterOfMars: Is this still true?

sdk/src/client/node_api/indexer/routes.rs
DaughterOfMars: I think it's fine to have multiple copies of the same error. We should be changing the error messages for wrapped errors to specify what component they are from anyway.

sdk/src/client/node_manager/http_client.rs
```diff
@@ -52,11 +53,23 @@ impl HttpClient {
         if status.is_success() {
             Ok(Response(response))
         } else {
-            Err(Error::ResponseError {
-                code: status.as_u16(),
-                text: response.text().await?,
-                url: url.to_string(),
-            })
+            let text = response.text().await?;
+            // Different urls, nodes and versions give different replies
```
DaughterOfMars, line 57: ewwww

sdk/src/client/node_manager/mod.rs
```diff
@@ -214,11 +216,8 @@ impl NodeManager {
                                 result_counter += 1;
                             },
                         ),
-                        Err(Error::ResponseError { code: 404, url, .. }) => {
-                            error.replace(crate::client::Error::NotFound(url));
-                        }
                         Err(err) => {
-                            error.replace(err);
+                            error.replace(err.into());
```
DaughterOfMars, line 219: Maybe a list of errors would be better here?

DaughterOfMars: Or a map of node id to error

sdk/src/client/error.rs
kwek20: I felt like since its only the faucet, which is sort of a "node" its better to not make the From explicit for entire crate. Would it be better if i made the error from the funds request directly?

sdk/src/client/node_manager/mod.rs
kwek20: i agree, but thats quite out of scope of the issue i think. Wed still only return 1 error though so which one would you pick?

sdk/src/client/node_api/core/routes.rs
kwek20: yes 👍

DaughterOfMars: blast

sdk/src/client/node_manager/mod.rs
DaughterOfMars: Well, probably it would make sense to have an error type that has multiple inner errors.

sdk/src/client/error.rs
DaughterOfMars: You mean create a `Faucet` variant in the error?

sdk/src/client/node_api/indexer/routes.rs
Thoralf-M: I think this should rather be something like a `InvalidNodeResponse` error, because if the node responds with 200 this Vec should contain 1 output id, if not the node is doing something wrong

kwek20: Thats what i mean, 200 will be fine, and otherwise its 404, which will lead to NoOutput("alias") because it doesnt exist. We can easily just request a nonexistant one right?

sdk/src/client/error.rs
kwek20: Hmm didnt think of that, maybe thats a good idea!

sdk/src/client/node_api/error.rs
```diff
@@ -0,0 +1,29 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+/// Type alias of `Result` in Node errors
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// Stronghold errors.
```
Thoralf-M:
```suggestion
/// Node errors.
```

sdk/src/client/node_api/error.rs
```diff
@@ -0,0 +1,29 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    /// The requested data was not found. (404)
+    #[error("The requested data {0} was not found.")]
+    NotFound(String),
```
Thoralf-M, line 12: Why have this variant and not also use `ResponseError` for this case?

kwek20: Because then we can conveniently match on 404, i thought it was a bit more explicit then match on `ResponseError { code: 404, .., .. }`

kwek20: I mena, `ResponseError ` is just a way to display generic other errors, we dont actually match on it anywhere

Thoralf-M: Where do we match on NotFound?

Thoralf-M: Nvm, found it

sdk/src/wallet/account/operations/syncing/foundries.rs
```diff
@@ -29,7 +29,7 @@ impl Account {
...
+                        Err(crate::client::Error::NoOutput("foundry")) => Ok(None),
```
Thoralf-M: Maybe debug print the foundry_id here? Just foundry isn't very helpful

sdk/CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Lots of builder setters are now taking an `impl Into<Option<T>>` instead of a `T` parameter;
 - All `ledger_nano` related errors have been moved from the client error to a newly created `client::secret::ledger_nano::Error`;
 - All `stronghold` related errors have been moved from the client error to a newly created `client::stronghold::Error`;
+- All `Node` related errors have been moved from the client error to a newly created `client::node_api::Error`;
```
thibault-martinez: Rebase and move to 0.4

sdk/examples/client/block/transaction.rs
```diff
@@ -19,7 +19,7 @@ async fn main() -> Result<()> {
     let client = Client::builder().with_node(&node_url)?.finish()?;
 
     let secret_manager =
-        SecretManager::try_from_mnemonic(&std::env::var("NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC_1").unwrap())?;
+        SecretManager::try_from_mnemonic(&std::env::var("NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC").unwrap())?;
```
thibault-martinez: This doesn't exist?

sdk/src/wallet/account/operations/syncing/foundries.rs
```diff
@@ -29,7 +29,10 @@ impl Account {
                 task::spawn(async move {
                     match client.foundry_output_id(foundry_id).await {
                         Ok(output_id) => Ok(Some(client.get_output(&output_id).await?)),
-                        Err(crate::client::Error::NotFound(_)) => Ok(None),
+                        Err(crate::client::Error::NoOutput("foundry", id)) => {
+                            log::debug!("{}", crate::client::Error::NoOutput("foundry", id));
```
Thoralf-M: Sorry, I actually meant debug "print" it like this :sweat_smile: because then you also know the type of the id in the error and don't have to add it manually as a second str
```suggestion
                        Err(crate::client::Error::NoOutput(&format!("{id:?}")) => {
```

kwek20: But this is a match statement :) I'm not sure I follow

Thoralf-M: Ah sorry it should be like this when the error is created and not in the match

sdk/src/client/node_api/indexer/routes.rs
```diff
@@ -84,7 +84,7 @@ impl Client {
             .get_output_ids(&route, QueryParameters::empty(), true, false)
             .await?
             .first()
-            .ok_or_else(|| crate::client::Error::Node("no output id for foundry".to_string()))?))
+            .ok_or_else(|| Error::NoOutput("foundry", foundry_id.to_string()))?))
```
Thoralf-M: @kwek20 I meant why not like this?
```suggestion
            .ok_or_else(|| Error::NoOutput(format!("{foundry_id:?}")))?))
```

sdk/src/wallet/account/operations/syncing/foundries.rs
Thoralf-M: Commented at the correct place

sdk/src/client/node_api/indexer/routes.rs
kwek20: Wouldnt you still want to know what type of output? this would just give the id.

Thoralf-M: Because of :? it prints FoundryId(0x..) no?

kwek20: oh true! Smart

kwek20: Hm that would leave us unable to do filter on just foundry though. Or i need to split up the error but that seems excessive
```
match err  {
     Ok() => {}
     Err(crate::client::Error::NoOutput("foundry")) => {}
}
```

Thoralf-M: Where do we need that? `Err(crate::client::Error::NoOutput(_))` is not fine?

Thoralf-M (approved):

Thoralf-M (changes requested):
conflicting

sdk/src/client/node_manager/mod.rs
```diff
@@ -228,54 +227,39 @@ impl NodeManager {
             for node in nodes {
                 match self.http_client.get(node.clone(), timeout).await {
                     Ok(res) => {
-                        match res.status() {
-                            200 => {
```
Thoralf-M, line 231: Will this not return a serialization error now instead of a better node one?

kwek20: naw those are already checked at the http.get().parse:response().
Only thing here could be into_json() failing, turning it into Error::Node(rewest), i guess you mean that. What would you want to do with that?

otherwise, `error.replace(err.into());` the err will be Error::node()

Thoralf-M: Ah right the parse_response() was the part I was missing, should be all good :+1:

Thoralf-M (dismissed):

sdk/src/client/node_api/core/routes.rs
```diff
@@ -167,61 +167,52 @@ impl Client {
...
+                // implemented for `std::sync::RwLockWriteGuard<'_, NetworkInfo>`
+                {
+                    let mut client_network_info = self
+                        .inner
+                        .network_info
+                        .write()
+                        .map_err(|_| crate::client::Error::PoisonError)?;
+                    // switch to local PoW
+                    client_network_info.local_pow = true;
+                }
```
Alex6323: you can just do this tho:
```rust
self
    .inner
    .network_info
    .write()
    .map_err(|_| crate::client::Error::PoisonError)?
    .local_pow = true;
```
to have the read/write lock guard only live as long as necessary.

sdk/src/client/node_api/core/routes.rs
Alex6323: same as above ... I would just not create the local var `client_network_info`.

sdk/src/client/node_api/core/routes.rs
```diff
@@ -246,59 +237,49 @@ impl Client {
...
+                let block_res = self.finish_block_builder(None, block.payload().cloned()).await;
+                let block_with_local_pow = match block_res {
+                    Ok(block) => {
+                        // reset local PoW state
+                        let mut client_network_info = self
+                            .inner
+                            .network_info
+                            .write()
+                            .map_err(|_| crate::client::Error::PoisonError)?;
+                        client_network_info.local_pow = false;
```
Alex6323: same here

sdk/src/client/node_manager/mod.rs
```diff
@@ -227,54 +226,39 @@ impl NodeManager {
...
-                                    let node_info: InfoResponse = res.into_json().await?;
-                                    let wrapper = crate::client::node_api::core::routes::NodeInfoWrapper {
-                                        node_info,
-                                        url: format!("{}://{}", node.url.scheme(), node.url.host_str().unwrap_or("")),
-                                    };
-                                    let serde_res = serde_json::to_string(&wrapper)?;
-                                    return Ok(serde_json::from_str(&serde_res)?);
-                                }
+                        // Handle node_info extra because we also want to return the url
+                        if path == "api/core/v2/info" {
```
Alex6323: Should this path be better stored in a constant at the top of the module?

kwek20: good idea!

sdk/src/client/node_api/core/routes.rs
kwek20: Done

Thoralf-M: You only did two other places?

sdk/src/client/error.rs
```diff
@@ -86,25 +86,22 @@ pub enum Error {
...
     /// Error on API request
     #[error("node error: {0}")]
-    Node(String),
+    Node(#[from] crate::client::node_api::error::Error),
     /// The block doesn't need to be promoted or reattached
     #[error("block ID `{0}` doesn't need to be promoted or reattached")]
     NoNeedPromoteOrReattach(String),
     /// The requested data was not found.
     #[error("the requested data {0} was not found.")]
     NotFound(String),
```
thibault-martinez: Why is this `NotFound` still used if we have one in the new Error?

kwek20: Yes it's still used. Not found is when a quorum question has no result. The other one, noresult is per node basis

thibault-martinez: I'm not speaking of any `noresult`. We have both `ClientError::NotFound` and `NodeError::NotFound` but in `node_api` module we use the client one where I'd have expected a node error to be used? And why wouldn't node quorum error be in node error? Maybe they need to be renamed for a bit of clarity, it's a bit confusing.

kwek20: IM open to suggestions:
The node `NotFound` gets thrown when the page returns 404
The client `NotFound` gets returned when on a call to multiple nodes, no result is found ever. However i do think we can unwrap there because that means the `self.http_client.get_bytes` doesnt return anything ever, no ok or err. Lets remove it and ill see if i can get rid of NotFound in client::Error

thibault-martinez: Just curious: why do we prefer unwrapping than just using the other NotFound instead?

kwek20: So its not confusing which one means what, rather just have it gone completely in this case i think

thibault-martinez: Hmm ok, as long as you are 200% sure it can't panic

kwek20: Solar flares ;)

Thoralf-M (approved):

thibault-martinez (approved):

PR #343, opened 2023-04-26
Move boxes from `Payload` to higher types

DaughterOfMars:
# Description of change

This moves the `Box`ed variants from the `Payload` type to the higher types `OptionalPayload` and `RegularTransactionEssence`. The result is a slightly cleaner implementation.

## Type of change


DaughterOfMars:
In hindsight, the boxes are important to ensure that the variants are equally sized.

PR #353, opened 2023-04-27
Use Vec<u32> instead of Chain for InputSigningDataDto

### LINKED ISSUE #242: Replace `Chain` in DTOs to Vec<u32>
## Description

Replace `Chain` in DTOs to Vec<u32>, for example here https://github.com/iotaledger/iota-sdk/blob/87cd53903a6bb49147e28caf27d2adeea2fa288a/sdk/src/client/secret/types.rs#L163

## Motivation

Much easier to provide from bindings, currently it's like this
https://github.com/iotaledger/iota-sdk/blob/dd66e1ce0f94d3755b6eecc0ccb2afefba0a2ff8/sdk/src/client/bindings/nodejs/test/client/utilityMethods.spec.ts#L111-L118
instead of `[44, 4218, 0, 0, 0]`

## Requirements

1. Replace `Chain` in DTOs
2. Update bindings
3. Make sure that everything still works

No.

Thoralf-M:
# Description of change

Use Vec<u32> instead of Chain for InputSigningDataDto
Use Segment::HARDEN_MASK instead of own HARDENED const

Fixes #242 

## Type of change

Tests

## Change checklist

### DIFF
--- a/bindings/nodejs/lib/client/Client.ts
+++ b/bindings/nodejs/lib/client/Client.ts
@@ -35,6 +35,7 @@ import type {
     ITreasury,
     INodeInfoProtocol,
     UnlockTypes,
+    HexEncodedString,
 } from '@iota/types';
 import type { INodeInfoWrapper } from '../../types/client/nodeInfo';
 import { SecretManagerType } from '../../types/secretManager/secretManager';
@@ -273,9 +274,7 @@ export class Client {
      */
     async signatureUnlock(
         secretManager: SecretManagerType,
-        // Uses `Array<number>` instead of `Uint8Array` because the latter serializes
-        // as an object rather than an array, which results in errors with serde.
-        transactionEssenceHash: Array<number>,
+        transactionEssenceHash: HexEncodedString,
         chain: IBip32Chain,
     ): Promise<UnlockTypes> {
         const response = await this.methodHandler.callMethod({
--- a/bindings/nodejs/tests/client/utilityMethods.spec.ts
+++ b/bindings/nodejs/tests/client/utilityMethods.spec.ts
@@ -116,7 +116,7 @@ describe('Client utility methods', () => {
         const message = '0x494f5441';
         const signature = await new SecretManager(secretManager).signEd25519(
             message,
-            // HD-Wallet type, IOTA coin type, first account, first public address
+            // HD-Wallet type, IOTA coin type, first account, public, first address
             [44, 4218, 0, 0, 0],
         );
 
--- a/bindings/nodejs/tests/client/utilityMethods.spec.ts
+++ b/bindings/nodejs/tests/client/utilityMethods.spec.ts
@@ -116,14 +116,8 @@ describe('Client utility methods', () => {
         const message = '0x494f5441';
         const signature = await new SecretManager(secretManager).signEd25519(
             message,
-            // [44, 4218, 0, 0, 0] IOTA coin type, first account, first public address
-            [
-                { hardened: true, bs: [128, 0, 0, 44] },
-                { hardened: true, bs: [128, 0, 16, 123] },
-                { hardened: true, bs: [128, 0, 0, 0] },
-                { hardened: true, bs: [128, 0, 0, 0] },
-                { hardened: true, bs: [128, 0, 0, 0] },
-            ],
+            // HD-Wallet type, IOTA coin type, first account, first public address
+            [44, 4218, 0, 0, 0],
         );
 
         const bech32Address = Utils.hexPublicKeyToBech32Address(
--- a/bindings/python/tests/test_offline.py
+++ b/bindings/python/tests/test_offline.py
@@ -33,28 +33,26 @@ def test_mnemonic_address_generation():
 
 
 def test_sign_verify_ed25519():
-    secret_manager = MnemonicSecretManager(Utils.generate_mnemonic())
+    secret_manager = MnemonicSecretManager(
+        "acoustic trophy damage hint search taste love bicycle foster cradle brown govern endless depend situate athlete pudding blame question genius transfer van random vast")
     message = utf8_to_hex('IOTA')
 
     secret_manager = SecretManager(secret_manager)
     signature = secret_manager.sign_ed25519(
         message,
-        # [44, 4218, 0, 0, 0] IOTA coin type, first account, first public address
-        [
-            {'hardened': True, 'bs': [128, 0, 0, 44]},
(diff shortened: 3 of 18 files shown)


sdk/src/client/secret/types.rs
```diff
@@ -169,23 +169,24 @@ pub struct InputSigningDataDto {
...
 
+#[allow(missing_docs)]
 impl InputSigningData {
-    pub(crate) fn try_from_dto(input: &InputSigningDataDto, token_supply: u64) -> Result<Self> {
+    pub fn try_from_dto(input: &InputSigningDataDto, token_supply: u64) -> Result<Self> {
         Ok(Self {
             output: Output::try_from_dto(&input.output, token_supply)?,
             output_metadata: OutputMetadata::try_from(&input.output_metadata)?,
-            chain: input.chain.clone(),
+            chain: input.chain.clone().map(Chain::from_u32_hardened),
```
thibault-martinez, line 181: Is this correct? We consider everything to be hardened now? Which wasn't the case before?

Thoralf-M: Before it was a Chain already
`from_u32_hardened` does the hardening

kwek20 (dismissed):

Thoralf-M:
Waiting for https://github.com/iotaledger/crypto.rs/pull/187 since this might change something

thibault-martinez:
> Waiting for [iotaledger/crypto.rs#187](https://github.com/iotaledger/crypto.rs/pull/187) since this might change something

It's been merged

Thoralf-M:
Tried with the new crypto.rs version and still works fine, but didn't update yet because stronghold still requires the older version

bindings/nodejs/lib/client/Client.ts
```diff
@@ -273,9 +274,7 @@ export class Client {
      */
     async signatureUnlock(
         secretManager: SecretManagerType,
-        // Uses `Array<number>` instead of `Uint8Array` because the latter serializes
-        // as an object rather than an array, which results in errors with serde.
-        transactionEssenceHash: Array<number>,
+        transactionEssenceHash: HexEncodedString,
```
thibault-martinez, line 277: Should we also do this to the current nodejs bindings?

bindings/nodejs/tests/client/utilityMethods.spec.ts
```diff
@@ -116,14 +116,8 @@ describe('Client utility methods', () => {
...
             message,
-            // [44, 4218, 0, 0, 0] IOTA coin type, first account, first public address
-            [
-                { hardened: true, bs: [128, 0, 0, 44] },
-                { hardened: true, bs: [128, 0, 16, 123] },
-                { hardened: true, bs: [128, 0, 0, 0] },
-                { hardened: true, bs: [128, 0, 0, 0] },
-                { hardened: true, bs: [128, 0, 0, 0] },
-            ],
+            // HD-Wallet type, IOTA coin type, first account, first public address
```
thibault-martinez: Can you maybe slightly change the last item of the comment? Even if correct, the way it's written it feels like an item is missing

bindings/python/tests/test_offline.py
```diff
@@ -33,28 +33,26 @@ def test_mnemonic_address_generation():
...
         message,
-        # [44, 4218, 0, 0, 0] IOTA coin type, first account, first public address
-        [
-            {'hardened': True, 'bs': [128, 0, 0, 44]},
-            {'hardened': True, 'bs': [128, 0, 16, 123]},
-            {'hardened': True, 'bs': [128, 0, 0, 0]},
-            {'hardened': True, 'bs': [128, 0, 0, 0]},
-            {'hardened': True, 'bs': [128, 0, 0, 0]},
-        ],
+        # HD-Wallet type, IOTA coin type, first account, first public address
```
thibault-martinez: same

sdk/CHANGELOG.md
```diff
@@ -50,11 +50,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Wallet::emit_test_event` no longer returns a `Result`;
 - `AccountBuilder::new` now takes a wallet;
 
+### Changed
```
thibault-martinez: You are adding a new Changed section

sdk/src/client/secret/types.rs
```diff
@@ -195,7 +196,13 @@ impl From<&InputSigningData> for InputSigningDataDto {
         Self {
             output: OutputDto::from(&input.output),
             output_metadata: OutputMetadataDto::from(&input.output_metadata),
-            chain: input.chain.clone(),
+            chain: input.chain.as_ref().map(|chain| {
+                chain
+                    .segments()
+                    .iter()
+                    .map(|seg| u32::from_be_bytes(seg.bs()) ^ Segment::HARDEN_MASK)
```
thibault-martinez: Is this `u32::from_be_bytes(seg.bs())` still necessary?

bindings/nodejs/lib/client/Client.ts
Thoralf-M: Client nodejs binding got removed and in the wallet we don't have it

thibault-martinez: Ah right, my bad

sdk/src/client/secret/types.rs
Thoralf-M: Currently yes, when crypto.rs is updated maybe not anymore

thibault-martinez: Right

thibault-martinez: Can you put a small TODO ?

Thoralf-M: Actually it's private, so without exposing it in crypto.rs it's not possible

semenov-vladyslav: What is the purpose of ` ^ Segment::HARDEN_MASK`? It'll convert non-hardened segment value to a hardened one and vice versa. Is that the intent? If you need to ensure that the segment is hardened then use `|  Segment::HARDEN_MASK`. If you need to ensure that the segment is non-hardened then use `& !Segment::HARDEN_MASK`.

Thoralf-M: Thanks, used `& !Segment::HARDEN_MASK`
I think here we always want the non hardened version because that's the values the users usually provide (since it needs to be hardened when using it with ed25519, it doesn't really matter)

thibault-martinez (approved):

Alex6323 (approved):

PR #358, opened 2023-04-28
Use a tip as block id instead of creating a new block

### LINKED ISSUE #64: Unify "Get a Block" example
## Description

The [Get a Block](https://wiki.iota.org/shimmer/iota.rs/how_tos/get_block/) examples explain different features in different languages.
This issue is part of the efforts in iotaledger/iota-sdk#59 

## Motivation

We should show/explain the same features in all languages (if they exist).

## Requirements

Let's show in every example how to:

- [ ] Don't show again on how to create a block. We already know that. Instead do it the Rust way and use a random tip. (That way we even include new info on how to fetch tips :trollface: ) We could probably even get rid of the get id from env part in the rust example
    - [ ] Rust (Remove env part and just use a random tip)
    - [ ] Nodejs
    - [ ] Python
    - [ ] Java

- [ ] Explain how to get block metadata
    - [ ] Rust
    - [x] Nodejs
    - [x] Python
    - [ ] Java 

No.

Thoralf-M:
# Description of change

Use a tip as block id instead of creating a new block

Fixes #64

## Type of change

- Documentation Fix

Running the examples

## Change checklist

### DIFF
--- a/bindings/nodejs/examples/client/07_get_block_data.ts
+++ b/bindings/nodejs/examples/client/07_get_block_data.ts
@@ -20,16 +20,16 @@ async function run() {
     });
 
     try {
-        // Create block with no payload.
-        const blockIdAndBlock = await client.buildAndPostBlock();
-        console.log('Block:', blockIdAndBlock, '\n');
+        // Fetch a block ID from the node.
+        const blockIds = await client.getTips();
+        console.log('Block IDs:', blockIds, '\n');
 
         // Get the metadata for the block.
-        const blockMetadata = await client.getBlockMetadata(blockIdAndBlock[0]);
+        const blockMetadata = await client.getBlockMetadata(blockIds[0]);
         console.log('Block metadata: ', blockMetadata, '\n');
 
         // Request the block by its id.
-        const blockData = await client.getBlock(blockIdAndBlock[0]);
+        const blockData = await client.getBlock(blockIds[0]);
         console.log('Block data: ', blockData, '\n');
     } catch (error) {
         console.error('Error: ', error);
--- a/sdk/examples/client/get_block.rs
+++ b/sdk/examples/client/get_block.rs
@@ -1,0 +1,32 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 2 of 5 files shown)


Dr-Electron:
🏃‍♂️

Dr-Electron:
Didn't test the code, but looks good to me

DaughterOfMars:
For those of us simpletons who are less familiar with the systems, could you provide an explanation for why this is okay to do?

Thoralf-M:
@Alexandcoats there must always be at least one tip (block that isn't referenced by other blocks yet) at the end of the Tangle
Therefore we can always retrieve a new block, which we can then also request from the node without fearing that it has already been pruned

DaughterOfMars (approved):

sdk/examples/client/get_block.rs
thibault-martinez, line 1: I don't have anything against that example but we already had get_block and get_block_metadata with tip, right?

thibault-martinez: Ah alright you wanted one that does both for consistency, ignore me

thibault-martinez (approved):

bindings/nodejs/examples/client/07_get_block_data.ts
```diff
@@ -20,16 +20,16 @@ async function run() {
...
-        const blockIdAndBlock = await client.buildAndPostBlock();
-        console.log('Block:', blockIdAndBlock, '\n');
+        // Fetch a block ID from the node.
+        const blockIds = await client.getTips();
+        console.log('Block IDs:', blockIds, '\n');
 
         // Get the metadata for the block.
-        const blockMetadata = await client.getBlockMetadata(blockIdAndBlock[0]);
+        const blockMetadata = await client.getBlockMetadata(blockIds[0]);
         console.log('Block metadata: ', blockMetadata, '\n');
```
Alex6323, line 29: so ... we are only interested in ONE block id, right? Isn't it enough then to log only this one block id?

Thoralf-M: Yes, might change it in another PR :+1:

PR #365, opened 2023-05-01
Fix no_std support for types

### LINKED ISSUE #279: Can `types` be no_std ?

DaughterOfMars:
# Description of change

This PR fixes `no_std` support for the `types` module, and reorganizes the dependencies and features so that there is no unnecessary importing of the `std` feature.

- Closes #279 

## Type of change

Built the sdk with the following command.

```
cargo build --no-default-features --features serde,events,ledger_nano,message_interface,participation,rocksdb,serde,storage,stronghold,tls,rand -p iota-sdk
```

All of the above features support `no_std` now.

## Open Questions

The `regex` crate is undergoing a major [rewrite](https://github.com/rust-lang/regex/pull/978) right now, and it is very close to being merged. Once it is, I believe it will also support `no_std`, thereby allowing us to remove the `std` feature from `mqtt`.


BurntSushi:
> The `regex` crate is undergoing a major [rewrite](https://github.com/rust-lang/regex/pull/978) right now, and it is very close to being merged. Once it is, I believe it will also support `no_std`, thereby allowing us to remove the `std` feature from `mqtt`.

This is correct, but note that it will still require `alloc`.

thibault-martinez (changes requested):
CI fails

thibault-martinez:
Do you mind adding a no_std target workflow so we don't accidentally remove no_std support in the future pls? Something like we had in bee https://github.com/iotaledger/bee/blob/shimmer-develop/.github/workflows/ci.yml#L84

sdk/CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `types::block::DtoError`, `client::Error::BlockDto` and `wallet::Error::BlockDto`;
 - `BasicOutput`, `AliasOutput`, `FoundryOutput`, `NftOutput` - `new_with_amount` and `new_with_minimum_storage_deposit` functions;
 - `OutputsToClaim::None` variant;
+- `client_without_tls` feature in favor of separate `client` and `tls` features;
```
thibault-martinez: Sorry I've released 0.3 this morning, can you move it into 0.4 please?

thibault-martinez (changes requested):

sdk/CHANGELOG.md
DaughterOfMars: yarp

.github/workflows/lint.yml
```diff
@@ -153,3 +153,20 @@ jobs:
...
+        uses: actions/checkout@v3
+
+      - name: Set up Rust
+        uses: ./.github/actions/setup-rust
+
+      - name: Check
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
+          args: --no-default-features -F serde,events,ledger_nano,message_interface,participation,rocksdb,serde,storage,stronghold,tls,rand
```
thibault-martinez: I think the features are a bit confusing tbh, it works because client and wallet are not here so most of these feature aren't actually used but they are definitely not no_std

DaughterOfMars: Prefer to just check types then?

thibault-martinez: Yeah I think so, I don't even think we'll ever be able to make anything else no_std anyway

thibault-martinez: Maybe pow, not even sure

.github/workflows/lint.yml
```diff
@@ -153,3 +153,20 @@ jobs:
...
+        uses: actions/checkout@v3
+
+      - name: Set up Rust
+        uses: ./.github/actions/setup-rust
+
+      - name: Check
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
+          args: --no-default-features -F serde -p iota-sdk
```
thibault-martinez: How is this checking for no_std compatibility? Missing a target?

DaughterOfMars: It's just the default features minus `std`.

thibault-martinez: Yeah but we need to compile it for a no_std target otherwise we may have std dependencies and we wouldn't know about it

DaughterOfMars: Not really true, since `#[no_std]` is set by omitting the `std` feature. However, this IS slightly more error prone as changing the `serde` feature to include `std` would break it.

thibault-martinez: In an ideal world yes, but in the end `#[no_std]` is not enough. We had the same issue with bee and we opened this issue https://github.com/rust-lang/cargo/issues/8798. You can actually compile something with no_std that wouldn't compile on an actual no_std platform.

DaughterOfMars: How annoying. I'll add the target then.

sdk/Cargo.toml
```diff
@@ -74,24 +74,23 @@ dotenvy = { version = "0.15.7", default-features = false }
 fern-logger = { version = "0.5.0", default-features = false }
 
 [features]
-default = [ "std", "serde" ]
+default = [ "std", "serde", "tls", "client", "wallet" ]
```
thibault-martinez: Do we keep std and serde here? It's already in the other features now

DaughterOfMars: Good point

thibault-martinez (approved):

Thoralf-M (approved):

PR #373, opened 2023-05-02
chore(examples): use consts in wallet examples

### LINKED ISSUE #121: Improve clarity of values to be updated in examples
## Description

In the examples, sometimes addresses or IDs need to be changed to run successfully. Currently those are "magic values" provided directly to the function calls. We probably should change that to be clearly named variables near the top of the example for clarity.

No.

### LINKED ISSUE #125: Align Create a Wallet Account examples
## Description

- [ ] Either use `.env` or clearly set a required constants near the top of the example.
- [ ] Only create one account in Node.js.
- [ ] Use same paths for Stronghold and database locations across languages.
- [ ] Print similar expected output across languages.

No.

Alex6323:
# Description of change

Makes wallet examples easier to modify by placing changeable things in consts at the top of each example or makes them env vars.

Also fixes some of the examples that weren't working properly, like:
* the background syncing example
* the ping/pong/threads examples (yanked in favor of the `multithread` example)
* the update alias output example
* and others

I took the opportunity to make them as consistent as possible to give users/developers a good experience. The examples will print a lot more useful information and in a way that's not overloading the terminal.

Closes #125 
Closes #121 

## Type of change


sdk/examples/wallet/01_create_wallet.rs
```diff
@@ -2,9 +2,13 @@
...
 
 //! In this example we will create a new wallet.
-//! Rename `.env.example` to `.env` first.
 //!
-//! `cargo run --example create_wallet --release`
+//! Make sure there's no `example.stronghold` file and no `example.walletdb` folder yet!
+//!
+//! Rename `.env.example` to `.env` first, then run the command:
+//! ```sh
+//! cargo run --release --all-features --example create_wallet
```
thibault-martinez: Don't know if we should suggest all features for every example, maybe be way longer than necessary...

Alex6323: yeah, I was thinking about that for a bit ... I get that compiling all features takes a bit longer, on the other hand it simplifies running the examples. All you have to do to run another example is get the last shell command and change the name. No C&P necessary. Also ... the shell history is not invalidated if a particular example changes its feature requirements. This makes it also easier for us because we don't need to change those docs anymore. So I preferred convenience over efficiency which I think - considering that this is not production code - would be the right choice here.

thibault-martinez (changes requested):
Conflicts

sdk/Cargo.toml
```diff
@@ -608,11 +603,6 @@ name = "storage"
 path = "examples/wallet/storage.rs"
 required-features = [ "wallet", "storage" ]
 
-[[example]]
-name = "threads"
-path = "examples/wallet/threads.rs"
```
Thoralf-M, line 673: This was my good spammer example :sob:

sdk/examples/wallet/07_send_native_tokens.rs
```diff
@@ -15,40 +20,51 @@ use iota_sdk::{
 };
 use primitive_types::U256;
 
+// The account alias used in this example
+const ACCOUNT_ALIAS: &str = "Alice";
```
Thoralf-M: Isn't this really annoying if one wants to change it, because then it needs to be changed in all examples?
Maybe most examples should just get the first account by index 0 and not use the alias

sdk/examples/wallet/07_send_native_tokens.rs
```diff
@@ -15,40 +20,51 @@ use iota_sdk::{
...
+// The native token amount to send
+const SEND_NATIVE_TOKEN_AMOUNT: u64 = 10;
+// The address to send the tokens to
+const RECV_ADDRESS: &str = "rms1qpszqzadsym6wpppd6z037dvlejmjuke7s24hm95s9fg9vpua7vluaw60xu";
+// The wallet database folder
+const WALLET_DB_PATH: &str = "./example.walletdb";
```
Thoralf-M: Same here, looks annoying to change, maybe make it a env var?

sdk/Cargo.toml
thibault-martinez: Why is it removed anyway?

Alex6323: there are no good spammers :trollface: 
seriously tho ... can you check with the `multithread` example @Thoralf-M ? I thought they were basically the same, and `threads` wouldn't add much else anymore.

sdk/examples/wallet/07_send_native_tokens.rs
Alex6323: Maybe ... I basically started with what Jeroen suggested because I thought that's what people wanted. I can give it more thought about whether it makes actually sense.

sdk/examples/wallet/07_send_native_tokens.rs
Alex6323: I agree :+1:

sdk/Cargo.toml
Thoralf-M: It's only doing 4 rounds and missing the syncing when required, the threads you could just run to spam, this helped me to find some race conditions which is also harder if you have multiple accounts doing txs at the same time

sdk/examples/wallet/07_send_native_tokens.rs
Alex6323: Done

sdk/examples/wallet/07_send_native_tokens.rs
Alex6323: Changed.

sdk/.env.example
```diff
@@ -1,8 +1,22 @@
...
 NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC_1="endorse answer radar about source reunion marriage tag sausage weekend frost daring base attack because joke dream slender leisure group reason prepare broken river"
 NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC_2="width scatter jaguar sponsor erosion enable cave since ancient first garden royal luggage exchange ritual exotic play wall clinic ride autumn divert spin exchange"
+# Seeds used in the examples (don't ever use those for real funds)
 NON_SECURE_USE_OF_DEVELOPMENT_SEED_1=0x256a818b2aac458941f7274985a410e57fb750f3a3a67969ece5bd9ae7eef5b2
 NON_SECURE_USE_OF_DEVELOPMENT_SEED_2=0x256a818b2aac458941f7274985a410e57fb750f3a3a67969ece5bd9ae7eef5b3
+# The account aliases used in the examples
+ACCOUNT_ALIAS_1="Alice"
+ACCOUNT_ALIAS_2="Bob"
+# The wallet database folder used in the examples
+WALLET_DB_PATH="./example.walletdb"
```
Thoralf-M: maybe
```suggestion
WALLET_DB_PATH="./example-walletdb"
```
because it's a folder and not a file

Alex6323: yeah, sure

sdk/Cargo.toml
Alex6323: I am adding it back as indeed a value spammer is a good and fun thing to have.

sdk/.env.example
```diff
@@ -1,8 +1,22 @@
...
+# The Stronghold snapshot file used in the examples
+STRONGHOLD_SNAPSHOT_PATH="./example.stronghold"
+# The password to the Stronghold snapshot file
 STRONGHOLD_PASSWORD="some_hopefully_secure_password"
+# The node to issue transactions with in the examples
```
Thoralf-M: with in the examples sounds weird

Alex6323: it has the same structure as "the guys I went to school with in my teenage years"   which doesnt sound weird to me tbh

Thoralf-M: Ok maybe it's just me
Another thing, all env values are for the examples, so why mention examples above most of them? If you want to mention that they are for the examples, I would just put a single comment on top for all

Alex6323: fair enough :+1:

sdk/.env.example
```diff
@@ -1,8 +1,26 @@
...
+# any interference by other users running those examples.
+
+# Mnemonics (Don't ever use them to manage real funds!)
 NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC_1="endorse answer radar about source reunion marriage tag sausage weekend frost daring base attack because joke dream slender leisure group reason prepare broken river"
 NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC_2="width scatter jaguar sponsor erosion enable cave since ancient first garden royal luggage exchange ritual exotic play wall clinic ride autumn divert spin exchange"
-NON_SECURE_USE_OF_DEVELOPMENT_SEED_1=0x256a818b2aac458941f7274985a410e57fb750f3a3a67969ece5bd9ae7eef5b2
-NON_SECURE_USE_OF_DEVELOPMENT_SEED_2=0x256a818b2aac458941f7274985a410e57fb750f3a3a67969ece5bd9ae7eef5b3
-STRONGHOLD_PASSWORD="some_hopefully_secure_password"
+# Seeds (Don't ever use them to manage real funds!)
+NON_SECURE_USE_OF_DEVELOPMENT_SEED_1="0x256a818b2aac458941f7274985a410e57fb750f3a3a67969ece5bd9ae7eef5b2"
```
kwek20, line 10: lets get rid of seed completely?

sdk/examples/wallet/11_decrease_native_token_supply.rs
```diff
@@ -2,45 +2,74 @@
...
+// also needs to be available. You can check this by running the `get_balance` example. You can mint a new native token
+// by running the `mint_native_token` example.
+const TOKEN_ID: &str = "0x08847bd287c912fadedb6bf38900bda9f2d377b75b2a0bece8738699f56ebca4130100000000";
+// The amount of native tokens to melt
+const MELT_AMOUNT: u64 = 10;
+
 #[tokio::main]
 async fn main() -> Result<()> {
+    if TOKEN_ID == "0x08847bd287c912fadedb6bf38900bda9f2d377b75b2a0bece8738699f56ebca4130100000000" {
+        println!("You need to change the TOKEN_ID constant before you can run this example successfully!");
```
kwek20, line 32: nice addition

sdk/examples/wallet/17_check_unlock_conditions.rs
```diff
@@ -13,22 +20,31 @@ use iota_sdk::{
...
     let account_addresses: Vec<Bech32Address> = account
         .addresses()
         .await?
         .into_iter()
-        .map(|a| a.address().clone())
+        .map(|address| address.address().clone())
         .collect();
 
-    let output = BasicOutputBuilder::new_with_amount(1_000_000)
+    println!("Account addresses: {:#?}", account_addresses);
```
kwek20: dont think we need this log

sdk/examples/wallet/accounts.rs
```diff
@@ -1,92 +1,107 @@
...
-    println!("Syncing took: {:.2?}", now.elapsed());
-    println!("Balance: {balance:?}");
+    let addresses = account2.addresses().await?;
+    println!("Number of addresses in {alias2}'s account: {}", addresses.len());
+    println!("{alias2}'s base coin balance:\n{:#?}", balance.base_coin());
 
     Ok(())
 }
+
+async fn get_or_create_account(wallet: &Wallet, alias: &str) -> Result<Account> {
```
kwek20, line 100: we could use this in more places i think. Maybe even an actual wallet method?

sdk/examples/wallet/ledger_nano.rs
```diff
@@ -17,64 +24,78 @@ use iota_sdk::{
...
 
-    let client_options = ClientOptions::new().with_node(&std::env::var("NODE_URL").unwrap())?;
-
+    let client_options = ClientOptions::new().with_node(&var("NODE_URL").unwrap())?;
     let secret_manager = LedgerSecretManager::new(true);
-
     let wallet = Wallet::builder()
         .with_secret_manager(SecretManager::LedgerNano(secret_manager))
-        .with_storage_path("ledger_nano_walletdb")
+        .with_storage_path(WALLET_DB_PATH)
```
kwek20: i think we can use the normal wallet path here as well

sdk/examples/wallet/multithread.rs
```diff
@@ -0,0 +1,198 @@
...
+                    recv_address, transaction.transaction_id
+                );
+
+                // Wait for transaction to get included
+                let block_id = account1_clone
+                    .retry_transaction_until_included(&transaction.transaction_id, None, None)
+                    .await?;
+
+                println!(
+                    "Transaction included: {}/block/{}",
```
kwek20: lets add thread index/num here as well

sdk/.env.example
Alex6323: yeah ... I think it's only some client examples that use those. All wallet examples use the mnemonics. Since this PR is already way too big I wanted to do that in a separate PR.

sdk/examples/wallet/accounts.rs
Alex6323: Hmm, okay. If you think that could be useful in more places, I think we should create an issue for that?!

sdk/Cargo.toml
```diff
@@ -587,20 +587,20 @@ path = "examples/wallet/output_consolidation.rs"
...
 
 [[example]]
 name = "recover_accounts"
 path = "examples/wallet/recover_accounts.rs"
 required-features = [ "wallet" ]
 
+[[example]]
+name = "spammer"
+path = "examples/wallet/spammer.rs"
+required-features = [ "wallet", "storage" ]
```
Thoralf-M:
```suggestion
required-features = [ "wallet" ]
```
Can we remove this and `.with_storage_path(&var("WALLET_DB_PATH").unwrap())` in the example?

sdk/examples/wallet/spammer.rs
```diff
@@ -0,0 +1,198 @@
...
+        .with_secret_manager(SecretManager::Mnemonic(secret_manager))
+        .with_storage_path(&var("WALLET_DB_PATH").unwrap())
+        .with_client_options(client_options)
+        .with_coin_type(SHIMMER_COIN_TYPE)
+        .finish()
+        .await?;
+
+    // Ensure there's some base coin balance
+    let account = get_or_create_account(&wallet, ACCOUNT_ALIAS).await?;
+    let addresses = generate_max_addresses(&account, NUM_CONCURRENT_TXS).await?;
```
Thoralf-M: No need to generate multiple addresses, this will only make syncing slower

sdk/examples/wallet/spammer.rs
```diff
@@ -0,0 +1,198 @@
...
+
+#[tokio::main(flavor = "multi_thread")]
+async fn main() -> Result<()> {
+    // This example uses secrets in environment variables for simplicity which should not be done in production.
+    dotenvy::dotenv().ok();
+
+    println!(
+        "{} cores detected. Running spammer with {NUM_CONCURRENT_TXS} concurrent transactions...",
+        num_cpus::get(),
+    );
```
Thoralf-M: What's the point of printing the amount of cores when we just use the NUM_CONCURRENT_TXS anyways?

sdk/examples/wallet/spammer.rs
```diff
@@ -0,0 +1,198 @@
...
+            .await?;
+
+        println!(
+            "Transaction included: {}/block/{}",
+            var("EXPLORER_URL").unwrap(),
+            block_id
+        );
+
+        account.sync(None).await?;
+    }
```
Thoralf-M: This should just be a single transaction with 127 (so we have 1 left for remainder) outputs to the first address

sdk/examples/wallet/split_funds.rs
```diff
@@ -1,92 +1,144 @@
...
     },
     types::block::output::{unlock_condition::AddressUnlockCondition, BasicOutputBuilder},
-    wallet::{ClientOptions, Result, Wallet},
+    wallet::{account::types::AccountAddress, Account, ClientOptions, Result, Wallet},
 };
 
+// The base coin amount to send
+const SEND_AMOUNT: u64 = 1_000_000;
+// The maximum number of addresses funds are distributed to
+const MAX_ADDRESSES_TO_SPLIT_FUNDS: usize = 15;
```
Thoralf-M: Why `MAX` can it be less?

sdk/examples/wallet/ledger_nano.rs
```diff
@@ -17,64 +24,78 @@ use iota_sdk::{
     wallet::{ClientOptions, Result, SendAmountParams, Wallet},
 };
 
+// The account alias used in this example
+const ACCOUNT_ALIAS: &str = "ledger";
+// The wallet database folder created in this example
+const WALLET_DB_PATH: &str = "./example.ledger_nano.walletdb";
```
Thoralf-M:
```suggestion
const WALLET_DB_PATH: &str = "./example-ledger_nano-walletdb";
```

sdk/examples/wallet/nft_collection/01_mint_collection_nft.rs
```diff
@@ -1,72 +1,96 @@
...
+// The location of the NFT.
+const NFT_URI: &str = "ipfs://wrongcVm9fx47YXNTkhpMEYSxCD3Bqh7PJYr7eo5Ywrong";
+// The name of the NFT
+const NFT_NAME: &str = "Shimmer OG NFT";
+// The description of the NFT
+const NFT_DESCRIPTION: &str = "The Shimmer OG NFT was handed out 1337 times by the IOTA Foundation to celebrate the official launch of the Shimmer Network.";
+// The issuer of the NFT
+const NFT_ISSUER_NAME: &str = "IOTA Foundation";
+// The name of the NFT collection
+const NFT_COLLECTION_NAME: &str = "Shimmer OG";
```
Thoralf-M: Do you really think this is a better representation? It's a JSON payload after all and there can be also other fields we didn't set here https://github.com/iotaledger/tips/blob/main/tips/TIP-0027/tip-0027.md#nft-schema

Alex6323: I think you're right. It should be a single JSON string constant, that can be modified in any desirable way.

sdk/examples/wallet/split_funds.rs
Alex6323: no, good point

sdk/examples/wallet/spammer.rs
Alex6323: ah, of course :+1:

sdk/examples/wallet/spammer.rs
Alex6323: hmm, I see your point. My intention was to let the user decide how many cores to use and at least give the information what  the reasonable max number COULD be used. You never know  what other stuff is currently running on the device this example is running on, so it propably shouldn't smoke the CPU. I have a better idea of doing this now. Good catch!

sdk/.env.example
Alex6323: created an issue: https://github.com/iotaledger/iota-sdk/issues/458 
Okay with this @kwek20 ?

sdk/examples/wallet/accounts.rs
Alex6323: Can you open an issue for that @kwek20 ?

sdk/examples/wallet/spammer.rs
Alex6323: Changed.

sdk/examples/wallet/14_burn_nft.rs
```diff
@@ -13,21 +20,24 @@ async fn main() -> Result<()> {
...
 
-    // Get the account we generated with `01_create_wallet`
-    let account = wallet.get_account("Alice").await?;
     // May want to ensure the account is synced before sending a transaction.
     let balance = account.sync(None).await?;
 
     // Get the first nft
     if let Some(nft_id) = balance.nfts().first() {
-        println!("Balance before burning:\n{balance:?}",);
+        println!("Balance BEFORE burning:\n{balance:?}",);
```
kwek20: lets display only nfts, like the foundry example

sdk/examples/wallet/15_destroy_foundry.rs
```diff
@@ -13,39 +21,43 @@ async fn main() -> Result<()> {
...
+            var("EXPLORER_URL").unwrap(),
             block_id
         );
 
         let balance = account.sync(None).await?;
-
-        println!("Balance after destroying:\n{balance:?}",);
+        let foundries_after = balance.foundries();
+        println!("Foundries AFTER destroying:\n{foundries_after:#?}",);
     }
```
kwek20, line 65: add log when no foundry available like in 14

sdk/examples/wallet/15_destroy_foundry.rs
```diff
@@ -13,39 +21,43 @@ async fn main() -> Result<()> {
...
+    let account = wallet.get_account(&var("ACCOUNT_ALIAS_1").unwrap()).await?;
 
-    // Get the account we generated with `01_create_wallet`
-    let account = wallet.get_account("Alice").await?;
     // May want to ensure the account is synced before sending a transaction.
     let balance = account.sync(None).await?;
 
-    // Get the first foundry
+    // We try to destroy the first foundry in the account
     if let Some(foundry_id) = balance.foundries().first() {
```
kwek20, line 35: Should we find a foundry that can be destroyed instead? Most likely the first one still has tokens

sdk/examples/wallet/14_burn_nft.rs
```diff
@@ -38,14 +48,16 @@ async fn main() -> Result<()> {
             .await?;
 
         println!(
-            "Block included: {}/block/{}",
-            std::env::var("EXPLORER_URL").unwrap(),
+            "Transaction included: {}/block/{}",
+            var("EXPLORER_URL").unwrap(),
             block_id
         );
```
kwek20, line 57: lets add here:println! ("Burned NFT {}", nft_id)

kwek20: same for destoy alias and foundry

thibault-martinez (changes requested):
Conflicts

sdk/examples/wallet/spammer.rs
```diff
@@ -57,22 +57,23 @@ async fn main() -> Result<()> {
         .with_coin_type(SHIMMER_COIN_TYPE)
         .finish()
         .await?;
-
-    // Ensure there's some base coin balance
     let account = get_or_create_account(&wallet, ACCOUNT_ALIAS).await?;
+
+    // Ensure there are enough addresses in the account.
     let addresses = ensure_enough_addresses(&account, num_simultaneous_txs).await?;
```
Thoralf-M: There is no point in having multiple addresses

sdk/examples/wallet/multithread.rs
Alex6323: this example actually didn't work as intended, so I deleted it. It really doesn't add much anyway when compared with the `spammer` example, and there ... the multithreading actually works.

sdk/Cargo.toml
Alex6323: Actually I use this in every example to create a wallet db with an `example-` prefix.

sdk/examples/wallet/spammer.rs
Alex6323: Done

sdk/examples/wallet/spammer.rs
Alex6323: changed

sdk/examples/wallet/logger.rs
```diff
@@ -15,51 +18,55 @@ use iota_sdk::{
...
+    // This example uses secrets in environment variables for simplicity which should not be done in production.
+    dotenvy::dotenv().ok();
+
+    // Initialize a logger that writes to the specified file
     let logger_output_config = fern_logger::LoggerOutputConfigBuilder::new()
-        .name("wallet.log")
+        .name(LOG_FILE_NAME)
         .target_exclusions(&["h2", "hyper", "rustls"])
-        .level_filter(log::LevelFilter::Debug);
+        .level_filter(LOG_LEVEL);
```
Thoralf-M: Can we change this back? Before one could just copy paste this part to have the logger in another example, with the consts further above this isn't that easy anymore

sdk/examples/wallet/spammer.rs
```diff
@@ -0,0 +1,241 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! In this example we will spam transactions from multiple threads simultaneously to our own addresses.
+//!
+//! Rename `.env.example` to `.env` first, then run the command:
+//! ```sh
+//! cargo run --release --all-features --example spammer
```
Thoralf-M:
```suggestion
//! cargo run --release --features=wallet --example spammer
```

sdk/examples/wallet/spammer.rs
```diff
@@ -0,0 +1,241 @@
...
+        let outputs = std::iter::repeat_with(|| {
+            BasicOutputBuilder::new_with_amount(output_amount)
+                .add_unlock_condition(AddressUnlockCondition::new(recv_address))
+                .finish_output(token_supply)
+                .unwrap()
+        })
+        .take(127)
+        .collect::<Vec<_>>();
+
+        let transaction = account.send(outputs, None).await?;
```
Thoralf-M: This can be just
```suggestion
        let transaction = account
            .send_amount(vec![SendAmountParams::new(recv_address, output_amount); 127], None)
            .await?;
```

sdk/examples/wallet/logger.rs
Alex6323: okay ... done

sdk/examples/wallet/spammer.rs
Alex6323: Applied

sdk/examples/wallet/spammer.rs
Alex6323: Applied

sdk/examples/wallet/14_burn_nft.rs
Alex6323: sounds good ... done

sdk/examples/wallet/15_destroy_foundry.rs
Alex6323: done

sdk/examples/wallet/15_destroy_foundry.rs
Alex6323: I would rather keep the example simple. We could mention the prerequisites of this example to actually work by mentioning it in the doc comments like we do with other examples... so ... make sure to burn all tokens with example `burn_native_token` example. Happy to here some suggestions. Finding a proper foundry id like you suggest is also possible, but there's no guarantee that's work either, so I kind of prefer just docs. Wdyt? Any suggestions?

sdk/examples/wallet/14_burn_nft.rs
Alex6323: okay...done

sdk/examples/wallet/nft_collection/01_mint_collection_nft.rs
Alex6323: Done (I think ... please check again)

sdk/examples/wallet/split_funds.rs
Alex6323: Done

sdk/examples/wallet/spammer.rs
Alex6323: Changed some time a ago already ... please resolve :trollface:

sdk/examples/wallet/ledger_nano.rs
Alex6323: Done

sdk/examples/wallet/accounts.rs
Alex6323: wanna resolve? :trollface:

kwek20: https://github.com/iotaledger/iota-sdk/issues/523

sdk/examples/wallet/15_destroy_foundry.rs
kwek20: yea ive realized finding one is more annoying. Ive made a related issue: https://github.com/iotaledger/iota-sdk/issues/524

sdk/examples/wallet/split_funds.rs
```diff
@@ -1,92 +1,144 @@
...
         secret::{mnemonic::MnemonicSecretManager, SecretManager},
     },
     types::block::output::{unlock_condition::AddressUnlockCondition, BasicOutputBuilder},
-    wallet::{ClientOptions, Result, Wallet},
+    wallet::{account::types::AccountAddress, Account, ClientOptions, Result, Wallet},
 };
 
+// The base coin amount to send
+const SEND_AMOUNT: u64 = 1_000_000;
+// The maximum number of addresses funds are distributed to
```
Thoralf-M:
```suggestion
// The number of addresses funds are distributed to
```

sdk/examples/wallet/nft_collection/01_mint_collection_nft.rs
```diff
@@ -79,3 +91,37 @@ async fn main() -> Result<()> {
...
+        "version":"v1.0",
+        "type":"video/mp4",
+        "uri":"ipfs://wrongcVm9fx47YXNTkhpMEYSxCD3Bqh7PJYr7eo5Ywrong",
+        "name":"Shimmer OG NFT #{index}",
+        "description":"The Shimmer OG NFT was handed out 1337 times by the IOTA Foundation to celebrate the official launch of the Shimmer Network.",
+        "issuerName":"IOTA Foundation",
+        "collectionId":"{issuer_nft_id}",
+        "collectionName":"Shimmer OG"
+    }}"#
+    )
```
Thoralf-M: Doesn't this include the newlines?
If so they should be replaced at the end, they just require more bytes and therefore a higher storage deposit which is not wanted

Alex6323: ah, I see, well ... I remove the newlines then.

Alex6323: Done

Thoralf-M: Hm will also add the empty spaces in between
```Rust
fn main() {
    let r = format!(
        r#"{{
        "standard":"IRC27",
        "version":"v1.0",
        "type":"video/mp4"
    }}"#
    )
    .replace('\n', "");

    println!("{r}");

    println!(
        "{}",
        serde_json::from_str::<serde_json::Value>(&r)
            .unwrap()
            .to_string()
    );
}
```
will print
```
{        "standard":"IRC27",        "version":"v1.0",        "type":"video/mp4"    }
{"standard":"IRC27","type":"video/mp4","version":"v1.0"}
```
so I suggest to use `serde_json::from_str::<serde_json::Value>()` (the replace() is also not needed with that)

Alex6323: ~well, okay, but there's only empty space in your example, not in the actual example. Your suggestion still makes sense though.~ no, you're right!

Alex6323: Done, thanks for being so thorough!!

sdk/examples/wallet/output_consolidation.rs
```diff
@@ -3,84 +3,87 @@
...
 
-    let mnemonic: &str = &std::env::var("NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC_1").unwrap();
-    let mnemonic_secret_manager = MnemonicSecretManager::try_from_mnemonic(mnemonic).unwrap();
-    let secret_manager = SecretManager::Mnemonic(mnemonic_secret_manager);
-
-    let client_options = ClientOptions::new().with_node(&std::env::var("NODE_URL").unwrap())?;
-
-    // Create the wallet
     let wallet = Wallet::builder()
-        .with_secret_manager(secret_manager)
```
DaughterOfMars, line 32: Are these the default now?

sdk/examples/wallet/participation.rs
```diff
@@ -1,206 +1,256 @@
...
-        .with_ignore_node_health();
-
-    let secret_manager =
-        MnemonicSecretManager::try_from_mnemonic(&std::env::var("NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC_1").unwrap())?;
-
     let wallet = Wallet::builder()
-        .with_secret_manager(SecretManager::Mnemonic(secret_manager))
-        .with_client_options(client_options)
-        .with_coin_type(SHIMMER_COIN_TYPE)
+        // TODO: is this needed?
```
DaughterOfMars: Is it?

sdk/examples/wallet/spammer.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    println!("Spammer set up to issue {num_simultaneous_txs} transactions simultaneously.");
+
+    // Restore wallet from a mnemonic phrase.
+    let client_options = ClientOptions::new().with_node(&std::env::var("NODE_URL").unwrap())?;
+    let secret_manager =
+        MnemonicSecretManager::try_from_mnemonic(&std::env::var("NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC_1").unwrap())?;
+    let wallet = Wallet::builder()
+        .with_secret_manager(SecretManager::Mnemonic(secret_manager))
+        // .with_storage_path(&var("WALLET_DB_PATH").unwrap())
```
DaughterOfMars: Commented?

sdk/examples/wallet/spammer.rs
```diff
@@ -0,0 +1,229 @@
...
+#[tokio::main(flavor = "multi_thread")]
+async fn main() -> Result<()> {
+    // This example uses secrets in environment variables for simplicity which should not be done in production.
+    dotenvy::dotenv().ok();
+
+    let num_simultaneous_txs = NUM_SIMULTANEOUS_TXS.min(num_cpus::get());
+
+    println!("Spammer set up to issue {num_simultaneous_txs} transactions simultaneously.");
+
+    // Restore wallet from a mnemonic phrase.
```
DaughterOfMars, line 45: Not a restore?

sdk/examples/wallet/spammer.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    // Ensure there are enough available funds for spamming.
+    let available_funds = ensure_enough_funds(&account, &recv_address, num_simultaneous_txs).await?;
+    account.sync(None).await?;
+
+    let split_amount = available_funds / 2;
+    let output_amount = split_amount / 127u64;
+
+    println!("Consolidating outputs...");
+    if let Ok(transaction) = account.consolidate_outputs(false, Some(127)).await {
```
DaughterOfMars: Can you explain this block?

sdk/examples/wallet/spammer.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    let num_unspent_basic_outputs = account
+        .unspent_outputs(FilterOptions {
+            output_types: Some(vec![3]),
+            ..Default::default()
+        })
+        .await?
+        .len();
+
+    if num_unspent_basic_outputs < 127 {
```
DaughterOfMars: shouldn't this be >?

sdk/examples/wallet/spammer.rs
Alex6323: yeah, it's not needed. I'll remove it.

sdk/examples/wallet/spammer.rs
Alex6323: Depends how we define "restore" I would say. Since we don't go through stronghold here (get access to the stored seed via a password), but instead provide the mnemonic/seed directly you could call it a restore. If you lose access to your stronghold snapshot, but still have the mnemonic this is exactly what you'ld have to do to "restore" your wallet.

sdk/examples/wallet/spammer.rs
Alex6323: yeah, I can try. So ... what I ran into was that if you don't consolidate the outputs but keep on splitting funds in the splitting transaction down below, then this can actually make the ISA error out, or at least I think it's the ISA. I got an error like (too many inputs or something). Just try it out and remove the consolidation code, and run and stop the spammer a few times. You'll probably run into it as well. I am not sure however why that's a correct behavior. @Thoralf-M might know. IIRC he wasn't absolutely sure either, though. But maybe that has changed in the meantime.

sdk/examples/wallet/spammer.rs
Alex6323: So the idea (actually Thoralf's idea) was to maximally split the funds into outputs to improve parallelization. The specific number comes from how many inputs are maximally allowed in the ISA. The 128th output will contain the remainder amount. This way the ISA has many unspent outputs to pick for assembling the transferred amount and we can create transactions in parallel where all have their unique unspent inputs. This line ensures that there are 127 unspent outputs at the beginning for the ISAs to choose from.

sdk/examples/wallet/participation.rs
Alex6323: hmm, it works without, which is why I put this there as reminder. If it's not required for the example I would remove it to keep the example as simple as possible. I think it's probably generally not a good idea to use a node that's not healthy for participating in an event, but @Thoralf-M or @thibault-martinez probably know better.

sdk/examples/wallet/output_consolidation.rs
Alex6323: It's assumed that you run the `create_account` example before which is fine because a lot of other examples also have that one as a dependency. Then all that's needed is the storage path to build the wallet.

sdk/examples/wallet/spammer.rs
DaughterOfMars: If this is necessary then there should be a comment about it probably.

sdk/examples/wallet/participation.rs
Thoralf-M: Not needed, you can remove

sdk/examples/wallet/spammer.rs
Thoralf-M: It depends also on the amount these outputs have and what you do want with them
Consolidating here to just split again later makes no sense

sdk/examples/wallet/participation.rs
Alex6323: Done

sdk/examples/wallet/spammer.rs
Alex6323: Done

sdk/examples/wallet/spammer.rs
DaughterOfMars: Okay, just making sure you didn't do a typo

sdk/examples/wallet/output_consolidation.rs
DaughterOfMars: I see.

DaughterOfMars (dismissed):

sdk/examples/wallet/spammer.rs
Alex6323: aight...I removed the consolidation part as you can use some other example do consolidate again in case of an error. There's no real reason anyway to change the SEND_AMOUNT so users won't probably run into this.

sdk/examples/how_tos/accounts_and_addresses/list_addresses.rs
```diff
@@ -3,17 +3,24 @@
...
+    // This example uses secrets in environment variables for simplicity which should not be done in production.
+    dotenvy::dotenv().ok();
 
-    // Get the account
-    let account = wallet.get_account("Alice").await?;
+    let wallet = Wallet::builder()
+        .with_storage_path(&var("WALLET_DB_PATH").unwrap())
+        .finish()
+        .await?;
+    let account = wallet.get_account(&var("ACCOUNT_ALIAS_1").unwrap()).await?;
```
thibault-martinez: I'm not sure how I feel about things like that tbh. It makes the example unnecessarily cryptic. It should be as straightforward as possible. A newcomer will understand `Alice` but will be completely lost with `&var("ACCOUNT_ALIAS_1").unwrap()`.

DaughterOfMars: I think it's better like this.

thibault-martinez: I think it's definitely worse. We have to understand these are not just random examples, these are the how tos that will be shown on the website for new users to learn. Things like that make it hard to process. But I guess I can live with it.

DaughterOfMars: After some thought I think you're probably right. For tests it makes sense to do this, but not so much for examples.

Alex6323: I am not sure ... a dev that sees like 50 examples (or more if we take all binding examples into account) and still doesn't feel like running away might be confident enough to decipher it!? :trollface: 

Jokes aside ... I don't really see an alternative. I felt similar which is why I at least imported the `var` function. I wouldn't enforce "Alice" as an account name, and even if we do I would create a let binding because sometimes we use the alias in a println.

Thoralf-M: An alternative could be to just get the first account by index `wallet.get_account(0).await?` and not care about any aliases

Alex6323: not sure about that ... numbers are cryptic to people as well ... I think users would prefer names tbh. I see Thibault's point he's making. If there are no objections I would remove the env vars for account aliases and just use `Alice` and `Bob`.

DaughterOfMars: I think that's for the best. We have to consider examples as reading material, not necessarily perfect code.

thibault-martinez: No objection :)

sdk/examples/how_tos/accounts_and_addresses/create_address.rs
```diff
@@ -1,31 +1,56 @@
 // Copyright 2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-//! In this example we will generate an address.
-//! Rename `.env.example` to `.env` first.
+//! In this example we will generate an address for an already existing wallet.
```
Thoralf-M:
```suggestion
//! In this example we will generate addresses for an already existing wallet.
```

sdk/examples/how_tos/simple_transaction/simple_transaction.rs
```diff
@@ -1,37 +1,48 @@
...
+        .await?;
+    let account = wallet.get_account(&var("ACCOUNT_ALIAS_1").unwrap()).await?;
 
-    // Get the account we generated with `01_create_wallet`
-    let account = wallet.get_account("Alice").await?;
     // May want to ensure the account is synced before sending a transaction.
     let balance = wallet.sync(None).await?;
 
-    if balance.base_coin().available() >= 1_000_000 {
+    if balance.base_coin().available() >= SEND_AMOUNT {
```
Thoralf-M: Do we really think this is good in the example?
`send_amount()` takes care of checking the amount and will return an error if not enough is there, the user doesn't need to check it himself before calling it

sdk/examples/wallet/wallet.rs
```diff
@@ -1,106 +1,168 @@
 // Copyright 2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-//! TODO: Example description
+//! In this example we will:
+//! * restore a wallet from a mnemonic phrase
```
Thoralf-M: Why restore? It just creates it

sdk/examples/wallet/wallet.rs
```diff
@@ -1,106 +1,168 @@
...
+    let transaction = account.send_amount(outputs, None).await?;
+    wait_for_inclusion(&transaction.transaction_id, &account).await?;
+
+    sync_print_balance(&account, false).await?;
+
+    println!("Example finished successfully");
+    Ok(())
+}
+
+async fn restore_wallet() -> Result<Wallet> {
```
Thoralf-M: Related to the comment above, don't see why this is restoring

sdk/examples/wallet/update_alias_output.rs
```diff
@@ -45,42 +53,65 @@ async fn main() -> Result<()> {
         .find_map(|output_data| match &output_data.output {
             Output::Alias(alias_output) => {
                 let output_alias_id = alias_output.alias_id_non_null(&output_data.output_id);
-                if output_alias_id == alias_id {
-                    Some(output_data)
-                } else {
-                    None
-                }
+                (output_alias_id == alias_id).then_some(output_data)
```
Thoralf-M: Can use `unspent_alias_output()` now

sdk/examples/wallet/spammer.rs
Thoralf-M: You could also just store the mnemonic in a new stronghold and use that, I also think the name restore here sounds a bit wrong

sdk/examples/wallet/spammer.rs
```diff
@@ -0,0 +1,214 @@
+// Copyright 2021 IOTA Stiftung
```
Thoralf-M:
```suggestion
// Copyright 2023 IOTA Stiftung
```

sdk/examples/wallet/spammer.rs
```diff
@@ -0,0 +1,214 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! In this example we will spam transactions from multiple threads simultaneously to our own addresses.
```
Thoralf-M:
```suggestion
//! In this example we will spam transactions from multiple threads simultaneously to our own address.
```

sdk/examples/wallet/nft_collection/00_mint_issuer_nft.rs
```diff
@@ -1,51 +1,51 @@
...
+    println!("Account synced!");
 
     // Set the stronghold password
     wallet
-        .set_stronghold_password(&std::env::var("STRONGHOLD_PASSWORD").unwrap())
+        .set_stronghold_password(&var("STRONGHOLD_PASSWORD").unwrap())
         .await?;
 
-    let nft_options = [MintNftParams::new()
+    // Issue the minting transaction and wait for it's inclusion
```
Thoralf-M:
```suggestion
    // Issue the minting transaction and wait for its inclusion
```

sdk/examples/wallet/getting_started.rs
```diff
@@ -47,15 +48,12 @@ async fn main() -> Result<()> {
     wallet.store_mnemonic(mnemonic).await?;
```
Thoralf-M: Should we print the mnemonic here? If not we can never get it anymore since it's not actually stored, but only the the derived seed (do we have a better name for `store_mnemonic()` which is more clear?)

Alex6323: 1x yeah, 1x maybe :trollface:

sdk/examples/wallet/wallet.rs
Alex6323: I am not going to fight over this word, but I used "restore" because if the mnemonic already exists, then probably the wallet bound to that mnemonic/seed was initially created the first time some time before. (I know that you can just create a mnemonic with the CLI wallet.)

sdk/examples/how_tos/accounts_and_addresses/list_addresses.rs
Alex6323: done

sdk/examples/wallet/wallet.rs
Alex6323: Changed

sdk/examples/wallet/wallet.rs
Alex6323: Changed

sdk/examples/how_tos/simple_transaction/simple_transaction.rs
Alex6323: you're right ... changed

sdk/examples/wallet/update_alias_output.rs
Alex6323: sweet ... done

sdk/examples/wallet/getting_started.rs
Alex6323: how about `bind_mnemonic` or `link_mnemonic`? It's an operation that binds or links the wallet to a certain mnemonic. It wouldn't claim that the mnemonic is stored as-is.

Alex6323: Done (it now prints the generated mnemonic)

Thoralf-M:
Nodejs needs formatting

bindings/nodejs/examples/wallet/16-destroy-alias.ts
```diff
@@ -14,20 +14,16 @@ import { getUnlockedWallet } from './common';
...
+        const account = await manager.getAccount('Alice');
 
         // May want to ensure the account is synced before sending a transaction.
         let balance = await account.sync();
 
         if (balance.aliases.length == 0) {
-            throw new Error(
-                `No Alias available in account '${process.env.ACCOUNT_ALIAS_1}'`,
-            );
+            throw new Error(`No Alias available in account ''`);
```
Thoralf-M:
```suggestion
            throw new Error(`No Alias available in account 'Alice'`);
```

Thoralf-M (dismissed):

bindings/nodejs/examples/wallet/16-destroy-alias.ts
```diff
@@ -14,20 +14,16 @@ import { getUnlockedWallet } from './common';
 async function run() {
     try {
         // Create the wallet
-        const wallet = await getUnlockedWallet();
+        const manager = await getUnlockedManager();
```
thibault-martinez: This is wrong

sdk/Cargo.toml
```diff
@@ -75,6 +75,7 @@ iota-sdk = { path = ".", default-features = false, features = [ "rand" ] }
 
 dotenvy = { version = "0.15.7", default-features = false }
 fern-logger = { version = "0.5.0", default-features = false }
+tokio = { version = "1.27.0", default-features = false, features = [ "macros", "rt-multi-thread", "time", "sync", "fs" ] }
```
thibault-martinez:
```suggestion
tokio = { version = "1.28.2", default-features = false, features = [ "macros", "rt-multi-thread", "time", "sync", "fs" ] }
```

sdk/examples/how_tos/accounts_and_addresses/check_balance.rs
```diff
@@ -2,26 +2,44 @@
...
-    // Get the account
+    let wallet = Wallet::builder()
+        .with_storage_path(&var("WALLET_DB_PATH").unwrap())
+        .finish()
+        .await?;
     let account = wallet.get_account("Alice").await?;
 
     // Sync and get the balance
-    let _balance = account.sync(None).await?;
+    let _ = account.sync(None).await?;
```
thibault-martinez: I would not even do the `let _` then, it's confusing. And edit the comment above.

bindings/nodejs/examples/wallet/16-destroy-alias.ts
Alex6323: what the hell? Must somehow have happened during the merge 🤔

sdk/examples/how_tos/outputs/unlock_conditions.rs
```diff
@@ -3,7 +3,9 @@
 
 //! In this example we will build outputs with all possible unlock conditions.
 //!
-//! `cargo run --release --example output_unlock_conditions`
```
thibault-martinez, line 6: You are not always putting your `Rename `.env.example` to `.env` first, then run the command:` message, any reason why?

Alex6323: think that was by another PR that got merged before :shrug:

thibault-martinez: Ah, makes sense, can you make sure it's everywhere?

sdk/examples/how_tos/accounts_and_addresses/check_balance.rs
Alex6323: we actually get the balance right after this line. Don't know this example didn't do it in one step already. Changed

sdk/examples/how_tos/outputs/unlock_conditions.rs
Alex6323: Done

sdk/examples/how_tos/accounts_and_addresses/check_balance.rs
Alex6323: Done

bindings/nodejs/examples/wallet/16-destroy-alias.ts
Alex6323: Fixed

bindings/nodejs/examples/wallet/16-destroy-alias.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright 2023 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-import { getUnlockedWallet } from './common';
+import { getUnlockedManager } from './account-manager';
```
Thoralf-M: Wrong change?

Alex6323: don't know what happend 🤷 😆

Thoralf-M (dismissed):

sdk/examples/wallet/offline_signing/3_send_transaction.rs
```diff
@@ -3,57 +3,54 @@
...
+    types::block::payload::transaction::TransactionId,
+    wallet::{Account, Result, Wallet},
+};
+use tokio::{
+    fs::File,
+    io::{AsyncReadExt, BufReader},
 };
 
-const SIGNED_TRANSACTION_FILE_NAME: &str = "examples/wallet/offline_signing/signed_transaction.json";
+// TODO: .env file?
```
DaughterOfMars: TODO

DaughterOfMars (dismissed):

sdk/examples/wallet/offline_signing/3_send_transaction.rs
Alex6323: Decided against putting those into `.env` files because despite putting those paths into consts at the top it's not actually something that makes a lot of sense for the user to change.

Thoralf-M (approved):

DaughterOfMars (approved):

PR #377, opened 2023-05-03
Add stronghold getter

abdulmth:
# Description of change

Adds a stronghold getter to enable access to the inner stronghold instance required by the identity library.
See https://github.com/iotaledger/identity.rs/pull/1157

Feel free to change the method name/documentation.

## Type of change

Method is used in https://github.com/iotaledger/identity.rs/pull/1157

## Change checklist

### DIFF
--- a/sdk/src/client/stronghold/mod.rs
+++ b/sdk/src/client/stronghold/mod.rs
@@ -60,7 +60,10 @@ use std::{
 use derive_builder::Builder;
 use iota_stronghold::{KeyProvider, SnapshotPath, Stronghold};
 use log::{debug, error, warn};
-use tokio::{sync::Mutex, task::JoinHandle};
+use tokio::{
+    sync::{Mutex, MutexGuard},
+    task::JoinHandle,
+};
 use zeroize::Zeroizing;
 
 use self::common::PRIVATE_DATA_CLIENT_PATH;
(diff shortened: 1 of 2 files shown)


Thoralf-M (changes requested):
Please add a changelog entry here https://github.com/iotaledger/iota-sdk/blob/develop/sdk/CHANGELOG.md#added

Thoralf-M (dismissed):

sdk/src/client/stronghold/mod.rs
```diff
@@ -513,6 +516,11 @@ impl StrongholdAdapter {
 
         Ok(())
     }
+
+    /// Acquire the stronghold lock.
+    pub async fn get_stronghold(&self) -> MutexGuard<'_, Stronghold> {
```
thibault-martinez: We don't usually put a `get_` prefix on our getters.
Also, `stronghold.get_stronghold()` may be a bit weird, I'd suggest just `inner` but let's see what the other think

abdulmth: I think `inner` also works fine.

kwek20: Id like to see .inner() instead as well

Thoralf-M (approved):

thibault-martinez (approved):

PR #385, opened 2023-05-04
Use consistent naming scheme for message params

### LINKED ISSUE #290: Define a naming scheme for recipient addresses and sent data
## Description

There are instances in the code where we send some data to a recipient address, and we therefore have many Key-Value structs to support these.

```rust
pub struct AddressWithAmount {
    address: String,
    amount: u64,
    return_address: Option<String>,
    expiration: Option<u32>,
}

pub struct AddressNativeTokens {
    pub address: String,
    pub native_tokens: Vec<(TokenId, U256)>,
    pub return_address: Option<String>,
    pub expiration: Option<u32>,
}

pub struct AddressAndNftId {
    pub address: String,
    pub nft_id: NftId,
}
```
This leads to fields with long, awkward names that still do not adequately describe their usage:
```rust
addresses_and_nft_ids: Vec<AddressAndNftId>,
```

I propose that we define a keyword for these Key-Value-like structs to unify them.

### Suggestions

- Package (ex. `BaseTokenPackage`, `NftPackage`)
- Parcel (same as above, but perhaps more associated with mail and sending)
- Packet
- Bundle

Please comment with any other suggestions you have.

DaughterOfMars:
# Description of change

Defined `Params` naming scheme for Send methods.

- Closes #290

## Type of change


bindings/nodejs/lib/wallet/Account.ts
```diff
@@ -716,21 +712,21 @@ export class Account {
...
+        mintNativeTokenParams: MintNativeTokenParams,
         transactionOptions?: TransactionOptions,
     ): Promise<MintTokenTransaction> {
         const response = await this.methodHandler.callAccountMethod(
             this.meta.index,
             {
                 name: 'mintNativeToken',
                 data: {
-                    nativeTokenOptions: nativeTokenOptions,
+                    mintNativeTokenParams: mintNativeTokenParams,
```
Thoralf-M: Shouldn't the field be only `params`? Also for the others below

DaughterOfMars: Yes, thank you. It's hard to find all the locations 😤

bindings/python/iota_sdk/types/common.py
```diff
@@ -52,9 +52,9 @@ def as_dict(self):
         return config
 
 
-class AddressWithAmount():
+class SendAmountParams():
     def __init__(self, address, amount):
-        """Initialise an AddressWithAmount
+        """Initialise a SendAmountParams
```
Thoralf-M:
```suggestion
        """Initialise SendAmountParams
```

.changes/send-params.md
```diff
@@ -0,0 +1,16 @@
+---
+"iota-sdk-bindings-core": patch
+"iota-sdk-nodejs": patch
+"iota-sdk-python": patch
```
Thoralf-M: We only need this for the wallet nodejs binding we release with covector
```suggestion
"wallet-nodejs-binding": patch
```

DaughterOfMars: Can you explain?

Thoralf-M: The .change files are used for covector so it knows what change should lead to which version bump and also for the changelog entry
But supported packages need to be configured here https://github.com/iotaledger/iota-sdk/blob/develop/.changes/config.json and we only have the wallet nodejs binding there (we want to get rid of covector alltogether very soon, then we will just do the changelogs as for the Rust crate and don't need the .change files anymore)

sdk/src/wallet/account/operations/transaction/high_level/send_nft.rs
```diff
@@ -13,10 +13,10 @@ use crate::{
     wallet::account::{operations::transaction::Transaction, Account, TransactionOptions},
 };
 
-/// Address and nft for `send_nft()`
+/// Params `send_nft()`
```
Thoralf-M:
```suggestion
/// Params for `send_nft()`
```
Or order changed?

sdk/src/wallet/bindings/nodejs/types/bridge/account.ts
```diff
@@ -215,39 +214,38 @@ export type __IncreaseNativeTokenSupplyMethod__ = {
...
+        params: MintNftParams[];
         options?: TransactionOptions;
     };
 };
 
 export type __PrepareOutputMethod__ = {
     name: 'prepareOutput';
     data: {
-        options: OutputOptions;
+        options: OutputParams;
```
Thoralf-M:
```suggestion
        params: OutputParams;
```

.changes/send-params.md
DaughterOfMars: Oh. Must have changed since I last looked at it. Thanks.

Thoralf-M (approved):

thibault-martinez (approved):

PR #386, opened 2023-05-04
Clean re-exports

### LINKED ISSUE #66: Review all re-exports

thibault-martinez:
Re-export crates whose types we use in public APIs to avoid version mismatches.
Fixes #66

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,9 +94,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.3.0"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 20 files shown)


DaughterOfMars (approved):

kwek20:
has there been an issue with U256? It seems like a pretty stable struct, so we wouldnt need to reexport i think

thibault-martinez:
> has there been an issue with U256? It seems like a pretty stable struct, so we wouldnt need to reexport i think

It doesn't really have anything to do with stability. It's just that since it's part of our public API, if consumers use a different version, even if seemingly compatible, it won't compile. So better use it through us so they always have the exact same version.

kwek20 (approved):

Issue #63, opened 2023-02-01
Ideas for client refactoring


Thoralf-M:
With the new changes, mainly the new ISA, we should check if we can change/remove the following things
- [x] Having both `ClientBlockBuilder` and `ClientBlockBuilderOptions` feels a bit weird
- [ ] Check if all DTOs are actually useful
- [x] Make `is_alias_transition()` public, so `alias_state_transition()` can be removed from wallet.rs?
- [x] Remove SecretManageExt?

DaughterOfMars:
- [x] Evaluate `Into` impls for types which take references

DaughterOfMars:
I think ClientBlockBuilder/Options are necessary.

DaughterOfMars:
As annoying as it can be using `Into` with references, the alternatives are even worse. There are probably places where we could do better with memory management but at the end of the day I think the effects are pretty minimal.

DaughterOfMars:
Given that there are now separate issues for DTO changes, I think it's fair to close this in favor of those.

PR #394, opened 2023-05-04
Restructure the client A BIT to reduce `Arc` usage

### LINKED ISSUE #299: Improve `Client` reference counting
## Description

The `Client` struct should be broken up into `Client` and `ClientInner` using a single `Arc` rather than one for each field.

## Motivation

Fewer reference counters is good for performance.

DaughterOfMars:
# Description of change

The main changes are moving the core client stuff into a `ClientInner` and wrapping it in a single `Arc`, then I restructured the relevant fns to pass the `Client` around instead of the internals. The change which blew up the diffs is making the `ClientBuilder::finish` fn async. It makes sense to do this because it means that it is not creating a new runtime if we are calling the fn from within an existing one. Speaking of runtimes, I removed it from the `Client` as it should not longer be necessary.

- #63 
- Closes #299

## Type of change


sdk/src/client/builder.rs
```diff
@@ -327,69 +326,77 @@ impl ClientBuilder {
...
-                    self.node_manager_builder.ignore_node_health,
-                );
-                (runtime, sync_handle)
-            })
-            .join()
-            .expect("failed to init node syncing process");
-            (Some(Arc::new(runtime)), Some(sync_handle))
-        };
+    #[cfg(not(target_family = "wasm"))]
+    pub async fn finish(self) -> Result<Client> {
```
thibault-martinez, line 330: Changelog 🙏🏻

sdk/CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - rename `Client::try_get_outputs()` into `Client::get_outputs_ignore_errors()`;
 - rename `Client::try_get_outputs_metadata()` into `Client::get_outputs_metadata_ignore_errors()`;
 - MQTT connections to a node using https will now use wss/tls with native certificates;
+- `ClientBuilder::finish` is now async.
```
thibault-martinez:
```suggestion
- `ClientBuilder::finish` is now async;
```

thibault-martinez (approved):

sdk/src/client/client.rs
```diff
@@ -32,74 +31,55 @@ use crate::{
 /// An instance of the client using HORNET or Bee URI
 #[derive(Clone)]
 pub struct Client {
-    #[allow(dead_code)]
+    pub(crate) inner: Arc<ClientInner>,
     #[cfg(not(target_family = "wasm"))]
-    pub(crate) runtime: Option<Arc<Runtime>>,
+    pub(crate) _sync_handle: Arc<SyncHandle>,
```
Thoralf-M, line 36: Why did you put an underscore here?

Thoralf-M (approved):

PR #397, opened 2023-05-04
Add simple transaction how-tos

Dr-Electron:
# Description of change

I added the simple transaction how-tos. 

Open Questions:

- [x] The current example account can't request any funds anymore. Does it matter? Do you normally update the env file from time to time?
- [x] Can somebody tell me how to run a nodes example? I couldn't get it to work
- [x] Should I try to also unify the examples if it's not too much work, as I prefer having examples over not having any ;)

## Links to any relevant issues

Fixes #362.

## Type of change

- Documentation Enhancement

## How the change has been tested

Run locally

## Change checklist

Tick the boxes that are relevant to your changes, and delete any items that are not.

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

PR #400, opened 2023-05-05
Add get_client() to Python binding

### LINKED ISSUE #326: Add get_client() to python wallet
## Description

Add get_client() to python wallet, requires https://github.com/iotaledger/iota-sdk/pull/307

## Motivation

Make the client from the wallet available to the user

## Requirements

1. Add get_client() to the wallet

Yes.

Thoralf-M:
# Description of change

Add get_client() to Python binding
Also added some error instead of panic

Closes #326 

## Type of change

Running the new example

## Change checklist

### DIFF
--- a/bindings/python/examples/wallet/get_client.py
+++ b/bindings/python/examples/wallet/get_client.py
@@ -1,0 +1,10 @@
+from iota_sdk import Wallet
+
+# This example gets a client from the wallet.
+
+wallet = Wallet('./alice-database')
+
+client = wallet.get_client()
+
+info = client.get_info()
+print(f'{info}')
--- a/bindings/python/iota_sdk/client/client.py
+++ b/bindings/python/iota_sdk/client/client.py
@@ -104,8 +104,8 @@ class Client(NodeCoreAPI, NodeIndexerAPI, HighLevelAPI, ClientUtils):
             client_config['remote_pow_timeout'] = {'secs': int(client_config['remote_pow_timeout'].total_seconds(
             )), 'nanos': get_remaining_nano_seconds(client_config['remote_pow_timeout'])}
 
-        # Set _client_handle to None here, because it can't be serialized
-        client_config["_client_handle"]=None
+        # Delete _client_handle, because it's not needed here and can't be serialized
+        del client_config["_client_handle"]
         client_config = humps.camelize(client_config)
         client_config = dumps(client_config)
 
--- a/bindings/python/iota_sdk/client/client.py
+++ b/bindings/python/iota_sdk/client/client.py
@@ -35,7 +35,8 @@ class Client(NodeCoreAPI, NodeIndexerAPI, HighLevelAPI, ClientUtils):
         user_agent=None,
         local_pow=None,
         fallback_to_local_pow=None,
-        pow_worker_count=None
+        pow_worker_count=None,
+        _client_handle=None
     ):
         """Initialize the IOTA Client.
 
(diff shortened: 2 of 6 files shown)


bindings/python/iota_sdk/client/client.py
```diff
@@ -103,11 +104,16 @@ def get_remaining_nano_seconds(duration: timedelta):
             client_config['remote_pow_timeout'] = {'secs': int(client_config['remote_pow_timeout'].total_seconds(
             )), 'nanos': get_remaining_nano_seconds(client_config['remote_pow_timeout'])}
 
+        # Set _client_handle to None here, because it can't be serialized
+        client_config["_client_handle"]=None
```
thibault-martinez: Does this actually skip serialisation yeah?

thibault-martinez (dismissed):

bindings/python/iota_sdk/client/client.py
Thoralf-M: Maybe not, but I just deleted it now which is even better

thibault-martinez (approved):

thibault-martinez (changes requested):
Sad CI noises

bindings/python/iota_sdk/client/client.py
```diff
@@ -35,7 +35,8 @@ def __init__(
         user_agent=None,
         local_pow=None,
         fallback_to_local_pow=None,
-        pow_worker_count=None
+        pow_worker_count=None,
+        _client_handle=None
```
Dr-Electron: Not sure if that's the proper pythonic way. 
I would remove the underscore.

Or add another constructor. Not sure what the proper way would be. Just don't like the current one 😆

Dr-Electron (changes requested):

bindings/python/iota_sdk/client/client.py
Thoralf-M: How to add another constructor? I haven't really seen an example that didn't reuse the init one

Dr-Electron: Yeah, not possible, sorry.
Why did you use the underscore. Is it somehow useful here? It doesn't really change anything, no?
Or does it than not show up in IDEs or something?

Dr-Electron: What if we make a special hidden function for passing an existing client?

Thoralf-M: > Why did you use the underscore. Is it somehow useful here? It doesn't really change anything, no?
Or does it than not show up in IDEs or something?

Initally I thought it could be hidden with two, but that didn't work and then I made it with one so it's visible that it's not a param like the others (and in Rust that's the way to ignore it :trollface:)

> What if we make a special hidden function for passing an existing client?

How is that possible? You still need the init at the end, no?

Dr-Electron: Yeah, true that...

So I guess just remove the underscore as it doesn't really help with anything and we are good 😆

thibault-martinez (dismissed):

bindings/python/iota_sdk/client/client.py
```diff
@@ -103,11 +104,17 @@ def get_remaining_nano_seconds(duration: timedelta):
             client_config['remote_pow_timeout'] = {'secs': int(client_config['remote_pow_timeout'].total_seconds(
             )), 'nanos': get_remaining_nano_seconds(client_config['remote_pow_timeout'])}
 
+        # Delete client_handle, because it's not needed here and can't be serialized
+        if "client_handle" in client_config:
+            del client_config["client_handle"]
```
Dr-Electron: Btw would it makes sense to move all deletes to the top? So they are all in one place?

Dr-Electron (changes requested):
Never

Dr-Electron (approved):
Jk. All good ;)

thibault-martinez (approved):

bindings/python/examples/wallet/get_client.py
```diff
@@ -0,0 +1,10 @@
+from iota_sdk import Wallet
```
kwek20, line 1: are we sure this example is needed? :D we already have so many :/

kwek20 (approved):
dunno bout that example but otherwise lgtm

bindings/python/examples/wallet/get_client.py
Thoralf-M: Many examples are good to have, we just need to structure them better but that's in progress already
If we have another example that shows this being used, for example the faucet one in https://github.com/iotaledger/iota-sdk/pull/397, I'm fine if we remove this again, but I would like to have at least one where this is shown

PR #403, opened 2023-05-05
Cleanup SecretManageExt and fix recursion

DaughterOfMars:
# Description of change

Cleans up `SecretManageExt` and renames it.

- #63 

## Type of change

### DIFF
--- a/sdk/src/client/secret/mod.rs
+++ b/sdk/src/client/secret/mod.rs
@@ -81,17 +81,13 @@ pub trait SecretManage: Send + Sync {
     }
 }
 
-/// An extension to [`SecretManager`].
-///
-/// This trait is automatically implemented for any type that implements [`SecretManager`] - it contains methods for
-/// internal use that are based on the methods in [`SecretManager`]. Secret managers don't implement this on their
-/// sides.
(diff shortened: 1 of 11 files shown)


sdk/src/client/secret/mod.rs
```diff
@@ -81,17 +81,13 @@ pub trait SecretManage: Send + Sync {
...
 
-/// An extension to [`SecretManager`].
-///
-/// This trait is automatically implemented for any type that implements [`SecretManager`] - it contains methods for
-/// internal use that are based on the methods in [`SecretManager`]. Secret managers don't implement this on their
-/// sides.
+/// Defines a type that can sign a transaction essence.
 #[async_trait]
-pub trait SecretManageExt: SecretManage {
+pub trait SignTransactionEssence: SecretManage {
```
thibault-martinez, line 86: So you think we should keep it, yeah?

DaughterOfMars: There's not a good way to remove it and keep the default impl as-is, due to the associated error type.

thibault-martinez (approved):

Thoralf-M (approved):

PR #424, opened 2023-05-09
Add check_existing_db test

Thoralf-M:
# Description of change

Add check_existing_db test

## Type of change

Running it

## Change checklist


sdk/tests/wallet/check_existing_db_test/000097.sst
thibault-martinez, line 1: Can you put the db in the fixtures folder pls?

Thoralf-M: Done

sdk/tests/wallet/mod.rs
```diff
@@ -20,3 +20,5 @@ mod syncing;
 mod transactions;
 #[allow(clippy::module_inception)]
 mod wallet;
+#[cfg(not(target_os = "windows"))]
```
thibault-martinez, line 23: Is that because the database itself would be different?

sdk/tests/wallet/wallet_storage.rs
```diff
@@ -0,0 +1,80 @@
...
+    assert_eq!(pending_transactions.len(), 1);
+
+    let incoming_transactions = account.incoming_transactions().await?;
+    assert_eq!(incoming_transactions.len(), 1);
+
+    let unspent_outputs = account.unspent_outputs(None).await?;
+    assert_eq!(unspent_outputs.len(), 9);
+
+    // balance depends on the network
+    if &account.client().get_network_name().await? == "private_tangle1" {
```
thibault-martinez, line 64: But should we also assert that it's that? Otherwise we may silently not check the balance?

sdk/tests/wallet/mod.rs
Thoralf-M: Yes for some reason a db created with linux is broken on windows

sdk/tests/wallet/wallet_storage.rs
Thoralf-M: The client returns the default network name if the localhost node isn't available and default is "shimmer" and then the balance is 0, only when the private tangle is running it passes

thibault-martinez: Is this test running with the private tangle then? Shouldn't it be `ignore`?

Thoralf-M: Not sure if it will use it when running at the same time, but I also don't think it's important, the balance isn't stored in the db and will only use the outputs, so it's actually not that interesting to have in this test alltogether

And if we ignore it then we will probably only notice in the CI and not locally if it breaks

thibault-martinez (approved):

sdk/tests/wallet/wallet_storage.rs
```diff
@@ -0,0 +1,80 @@
...
+async fn check_existing_db() -> Result<()> {
+    let storage_path = "check_existing_db_test";
+    setup(storage_path)?;
+    // Copy db so the original doesn't get modified
+    copy_folder("./tests/wallet/fixtures/check_existing_db_test", storage_path).unwrap();
+
+    let manager = Wallet::builder().with_storage_path(storage_path).finish().await?;
+
+    // Test if setting stronghold password still works
+    manager.set_stronghold_password("STRONGHOLD_PASSWORD").await?;
```
kwek20, line 21: lets move this password to `constants.rs`

also maybe "Alice" and the copy_folder?

Thoralf-M: But it's specific do this db, why would you want it somewhere else?

kwek20: every example also uses that password, so i thought it should count as common, but im fine with it either way

kwek20 (approved):

PR #425, opened 2023-05-09
Move client from account to wallet

### LINKED ISSUE #423: Have only one Client in Wallet instead of all Account?

DaughterOfMars:
# Description of change

Moves the `Client` from each account to the wallet, so it is accessible via that in the account. This required adding a lot of locking mechanisms to the client fields, but unless the fields are being updated frequently this will not result in any contention.

- Closes #423

## Type of change


thibault-martinez:
Can be rebased now

thibault-martinez:
So overall, you think this is worth it?

DaughterOfMars:
Yeah, I think so. It's cleaner for the most part.

sdk/src/wallet/wallet/operations/address_generation.rs
```diff
@@ -124,16 +124,7 @@ impl Wallet {
...
-                self.client_options
-                    .read()
-                    .await
-                    .clone()
-                    .finish()
-                    .await?
-                    .get_bech32_hrp()
-                    .await?
-            }
+            None => self.client().get_bech32_hrp().await?,
```
thibault-martinez, line 127: 😍

thibault-martinez: Wait, was there a reason it was like that? Why wasn't it using the client in the first place?

sdk/src/wallet/account/builder.rs
```diff
@@ -84,11 +84,6 @@ impl AccountBuilder {
             }
         }
 
-        let client = match accounts.first() {
-            Some(account) => account.client.clone(),
-            None => self.wallet.client_options.read().await.clone().finish().await?,
-        };
```
thibault-martinez, line 90: Love that

sdk/src/wallet/account/update.rs
```diff
@@ -276,9 +275,8 @@ impl Account {
 
     // Should only be called from the Wallet so all accounts are on the same state
     // Will update the addresses with a possible new Bech32 HRP and clear the inaccessible_incoming_transactions.
-    pub(crate) async fn update_account_with_new_client(&mut self, client: Client) -> crate::wallet::Result<()> {
-        self.client = client;
-        let bech32_hrp = self.client.get_bech32_hrp().await?;
+    pub(crate) async fn update_account_with_new_client(&mut self) -> crate::wallet::Result<()> {
```
thibault-martinez: Just checking, the name suggest a new client, I don't exactly know where it was coming from. Is it actually right to update with the wallet client?

sdk/src/wallet/wallet/operations/address_generation.rs
DaughterOfMars: I guess because this is a Wallet fn not an Account fn.

sdk/src/wallet/account/update.rs
DaughterOfMars: The name is obviously kind of wrong now, but basically this is called after we have updated the shared client ref in a few places. If you have a better name please suggest it.

sdk/src/wallet/wallet/operations/address_generation.rs
thibault-martinez: Right

thibault-martinez: Could have used an account client

sdk/src/wallet/account/update.rs
kwek20: `update_account_with_updated_client` :)

thibault-martinez (changes requested):
Conflicts

sdk/src/wallet/wallet/operations/client.rs
```diff
@@ -141,30 +132,21 @@ impl Wallet {
...
             self.storage_manager
                 .lock()
                 .await
-                .save_wallet_data(&wallet_builder)
+                .save_wallet_data(&WalletBuilder::from_wallet(self).await)
                 .await?;
         }
 
-        let new_client = new_client_options.finish().await?;
+        self.client.update_node_manager(node_manager_builder).await?;
```
thibault-martinez: Why is this taking a builder?

DaughterOfMars: Changed it 👍

sdk/src/wallet/account/update.rs
DaughterOfMars: Maybe `update_account_with_bech32_hrp` is better, since that seems to be the focus?

thibault-martinez: `with_` sounds like it's a parameter, maybe just `update_account_bech32_hrp`

sdk/src/client/node_manager/syncing.rs
```diff
@@ -116,10 +116,46 @@ impl ClientInner {
...
+impl Client {
+    #[cfg(not(target_family = "wasm"))]
+    pub async fn update_node_manager(&self, node_manager: NodeManager) -> crate::wallet::Result<()> {
+        let node_sync_interval = node_manager.node_sync_interval;
+        let ignore_node_health = node_manager.ignore_node_health;
+        let nodes = node_manager
+            .primary_node
+            .iter()
+            .chain(node_manager.nodes.iter())
+            .map(|node| node.clone().into())
```
thibault-martinez: Doesn't compile

thibault-martinez (dismissed):

sdk/src/wallet/account/mod.rs
```diff
@@ -191,7 +189,7 @@ impl Account {
 
     // Get the Client
     pub fn client(&self) -> &Client {
-        &self.client
+        &self.wallet.client
```
kwek20, line 192: should we call the client() here instead?

sdk/src/wallet/account/builder.rs
```diff
@@ -128,7 +123,7 @@ impl AccountBuilder {
                 let bech32_hrp = {
                     match bech32_hrp {
                         Some(bech32_hrp) => bech32_hrp,
-                        None => client.get_bech32_hrp().await?,
+                        None => self.wallet.client.get_bech32_hrp().await?,
```
kwek20: and here

kwek20 (dismissed):
work of art! not sure on the consensus of directly calling client vs .client(), but thats just a nit

sdk/src/wallet/account/builder.rs
kwek20: and a few others where we call direcy, ill tag em if we want to change

sdk/src/wallet/account/mod.rs
DaughterOfMars: That is the `client()` fn?

sdk/src/wallet/account/builder.rs
DaughterOfMars: This is called on the AccountBuilder not Account, so that method do not exist.

DaughterOfMars: Nevermind, I'm wrong lol

thibault-martinez (dismissed):

kwek20:
WHoeps, doenst compile now :D My bad

thibault-martinez (approved):

kwek20 (approved):

PR #450, opened 2023-05-15
Error instead of panic

### LINKED ISSUE #374: Return error from bindings if wallet is destroyed instead of panic
## Description

Return error from bindings if wallet is destroyed and don't panic

## Motivation

Might not completely break applications right away

## Requirements

1. Return error in call wallet method if wallet is destroyed

Maybe.

Thoralf-M:
# Description of change

Return an error instead of panic in bindings where possible

Closes #374 

## Type of change

Tests and running an example

## Change checklist

### DIFF
--- a/bindings/nodejs/lib/wallet/WalletMethodHandler.ts
+++ b/bindings/nodejs/lib/wallet/WalletMethodHandler.ts
@@ -38,7 +38,11 @@ export class WalletMethodHandler {
             this.methodHandler,
         ).catch((error: Error) => {
             try {
-                error = JSON.parse(error.toString()).payload;
+                if (error.message !== undefined) {
+                    error = JSON.parse(error.message).payload;
+                } else {
+                    error = JSON.parse(error.toString()).payload;
+                }
             } catch (e) {
                 console.error(e);
             }
@@ -71,6 +75,19 @@ export class WalletMethodHandler {
     }
 
     async getClient(): Promise<Client> {
-        return new Client(await getClientFromWallet(this.methodHandler));
+        return new Promise((resolve, reject) => {
+            getClientFromWallet(this.methodHandler).then((result: any) => {
+                if (result.message !== undefined) {
+                    if (result.message !== undefined) {
+                        result = JSON.parse(result.message).payload;
+                    } else {
+                        result = JSON.parse(result.toString()).payload;
+                    }
+                    reject(result);
+                } else {
+                    resolve(new Client(result));
+                }
+            });
+        });
     }
 }
--- a/bindings/nodejs/lib/wallet/WalletMethodHandler.ts
+++ b/bindings/nodejs/lib/wallet/WalletMethodHandler.ts
@@ -78,12 +78,7 @@ export class WalletMethodHandler {
         return new Promise((resolve, reject) => {
             getClientFromWallet(this.methodHandler).then((result: any) => {
                 if (result.message !== undefined) {
-                    if (result.message !== undefined) {
-                        result = JSON.parse(result.message).payload;
-                    } else {
-                        result = JSON.parse(result.toString()).payload;
-                    }
-                    reject(result);
+                    reject(JSON.parse(result.message).payload);
                 } else {
                     resolve(new Client(result));
                 }
--- a/bindings/nodejs/src/wallet.rs
+++ b/bindings/nodejs/src/wallet.rs
@@ -101,6 +101,7 @@ pub fn call_wallet_method(mut cx: FunctionContext) -> JsResult<JsUndefined> {
     let method_handler = Arc::clone(&&cx.argument::<JsBox<WalletMethodHandlerWrapper>>(1)?.0);
     let callback = cx.argument::<JsFunction>(2)?.root(&mut cx);
 
+    let (sender, receiver) = std::sync::mpsc::channel();
     crate::RUNTIME.spawn(async move {
         if let Some(method_handler) = &*method_handler.read().await {
             let (response, is_error) = method_handler.call_method(method).await;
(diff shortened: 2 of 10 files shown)


thibault-martinez (dismissed):

bindings/nodejs/lib/wallet/WalletMethodHandler.ts
```diff
@@ -71,6 +75,19 @@ export class WalletMethodHandler {
     }
 
     async getClient(): Promise<Client> {
-        return new Client(await getClientFromWallet(this.methodHandler));
+        return new Promise((resolve, reject) => {
+            getClientFromWallet(this.methodHandler).then((result: any) => {
+                if (result.message !== undefined) {
+                    if (result.message !== undefined) {
```
Alex6323: ugh, what? brain not like this (why does this have to be checked twice?)

bindings/nodejs/src/wallet.rs
```diff
@@ -101,6 +101,7 @@ pub fn call_wallet_method(mut cx: FunctionContext) -> JsResult<JsUndefined> {
     let method_handler = Arc::clone(&&cx.argument::<JsBox<WalletMethodHandlerWrapper>>(1)?.0);
     let callback = cx.argument::<JsFunction>(2)?.root(&mut cx);
 
+    let (sender, receiver) = std::sync::mpsc::channel();
```
Alex6323, line 104: sync channel in async context? is this on purpose? wouldn't it be better to use an async one? If on purpose then I would add a comment I think.

qrayven: If this is only to set a flag, I would propose using atomic bool, which is also good for async

thibault-martinez: But how do you check the flag then?

bindings/nodejs/src/wallet.rs
```diff
@@ -122,10 +123,15 @@ pub fn call_wallet_method(mut cx: FunctionContext) -> JsResult<JsUndefined> {
                 Ok(())
             });
         } else {
-            panic!("Wallet got destroyed")
+            sender.send("Wallet got destroyed").unwrap();
```
Alex6323: undocumented unwrap :shrug: (is it ensured that the receiver will always be instantiated at the time of sending?)

also ... you could probably just send a `()` and create the string "Wallet got destroyed" on the receiving end because it's probably the only thing that gets sent (just used as a signal)?

bindings/python/src/lib.rs
```diff
@@ -26,10 +31,10 @@ pub(crate) fn block_on<C: futures::Future>(cb: C) -> C::Output {
     runtime.lock().unwrap().block_on(cb)
 }
 
-/// Init the logger of wallet library.
+/// Init the Rust logger.
 #[pyfunction]
-pub fn init_logger(config: String) -> PyResult<()> {
-    rust_init_logger(config).expect("failed to init logger");
+pub fn init_logger(config: String) -> Result<()> {
+    rust_init_logger(config).map_err(|err| Error::from(format!("{:?}", err).as_str()))?;
```
Alex6323: We probably should add a `From<String>` impl for `Error`. We have a string here (format!) only to pass it as a `&str` to then turn it into a `String` again.

bindings/python/src/wallet.rs
```diff
@@ -96,7 +96,13 @@ pub fn get_client_from_wallet(wallet: &Wallet) -> Result<Client> {
             .await
             .as_ref()
             .map(|w| w.client().clone())
-            .ok_or_else(|| Error::from("wallet got destroyed"))
+            .ok_or_else(|| {
+                Error::from(
+                    serde_json::to_string(&Response::Panic("wallet got destroyed".into()))
+                        .unwrap()
```
Alex6323: Should I keep pointing out undocumented `unwrap`s? Not sure about our policy here, but it would make reviewing a bit easier I think if one can see it has been thought about it :smile: . 

Also ... above I think you use `.expect("json to string error")`. Might do that here as well?

bindings/nodejs/src/wallet.rs
qrayven: ok I looked closer, atomic bool isn't the solution here, because we need to wait for the execution of async block
edit: also assuming we want to wait for async block to be executed in non-async ctx, we could use `block_on` from runtime

Thoralf-M: The point of this is to get something out of the async block, because neon cx (CallContext) is not Send and therefore can't be used inside of the async block

bindings/nodejs/src/wallet.rs
Thoralf-M: Not sure about your first question, receiver is created the same time as the sender, so it always exists

bindings/nodejs/lib/wallet/WalletMethodHandler.ts
Thoralf-M: That happens when you quickly copy paste and it works fine so you don't look at it anymore :trollface:

bindings/nodejs/src/wallet.rs
Alex6323: theoretically(!!) `receiver.recv()` could return `Err`, right? then `call_wallet_method` would exit/return with `Ok(cx.undefined())` deallocating `receiver`, while the spawn task would still be alive.

Alex6323: ignore me... docs say  this can't happen as long as the sender is alive.

Alex6323: oops. resolved too quickly ... the 2nd point still stands ... no need to send this `&str`.

Thoralf-M: I changed that too?

Thoralf-M: https://github.com/iotaledger/iota-sdk/blob/3954255ddd9cd4c7a69a89fb7ddebfe82bba82f6/bindings/nodejs/src/wallet.rs#L128

Alex6323: rilly? then I resolve

Alex6323 (approved):

thibault-martinez (approved):

PR #451, opened 2023-05-15
Add recreated types to make migration static

DaughterOfMars:
# Description of change

Migrations must not rely on types outside of their definition lest they change.

## Type of change


sdk/src/wallet/migration/migrate_0.rs
```diff
@@ -169,39 +136,205 @@ trait Convert {
...
+
+    use crate::{impl_id, string_serde_impl, types::block::Error};
+
+    impl_id!(
+        pub TransactionId,
+        32,
+        "A transaction identifier, the BLAKE2b-256 hash of the transaction bytes. See <https://www.blake2.net/> for more information."
+    );
+
+    string_serde_impl!(TransactionId);
```
Alex6323: are you sure we can use those macros? what if they change?

sdk/src/wallet/migration/migrate_0.rs
```diff
@@ -169,39 +136,205 @@ trait Convert {
...
+            index.copy_from_slice(&self.index.to_le_bytes());
+            write!(f, "{}", prefix_hex::encode(buffer))
+        }
+    }
+
+    string_serde_impl!(OutputId);
+
+    #[derive(Serialize, Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct OutputMetadata {
```
Thoralf-M, line 295: Should we maybe add a version suffix or have own types for each version?
Since this is a new type, it could be used in a future migration as the old type and then it would be wasted to duplicate it
Without any versioning it's probably hard to keep track of what was being used where

sdk/src/wallet/migration/migrate_0.rs
thibault-martinez: I wouldn't worry about that tbh, it's just doing to_string/from_str, not sure how that would ever change

Alex6323: I don't know ... most of the derives and trait impls really seem like they won't ever change, maybe with the exception of `Packable`. It should just be made clear (maybe even by adding a comment to the macros themselves, that those 2 macros are pretty much locked since they are used in the migration logic)

sdk/src/wallet/migration/migrate_0.rs
Alex6323: but aren't those types essentially versioned by living inside `migrate_0` module? The `types` module should probably be `pub(crate)` though to have access to them from other `migrate_x` modules.

Thoralf-M: > but aren't those types essentially versioned by living inside migrate_0 module?

If you just look at OutputMetadata or OutputMetadataDto you don't know which one is the old or new version

sdk/src/wallet/migration/migrate_0.rs
DaughterOfMars: It's a fair point. I will suggest that in this case we should probably leave it and we can change it in the future if they ever change. We should get errors from our tests alerting us to the problem in that case.

sdk/src/wallet/migration/migrate_0.rs
DaughterOfMars: I don't think it's a good idea to share the internal types between migrations.

sdk/src/wallet/migration/migrate_0.rs
DaughterOfMars: I looked at it more, and removing the macros wasn't very difficult so I went ahead and did it.

sdk/src/wallet/migration/migrate_0.rs
Thoralf-M: Why not if they're the same?

DaughterOfMars: Sure, if they are the same then it's fine. We can always make it more public in the future.

Alex6323: > > but aren't those types essentially versioned by living inside migrate_0 module?
> 
> If you just look at OutputMetadata or OutputMetadataDto you don't know which one is the old or new version

that's true ... I personally would just fully qualify in that case, but I can see your point. Since it's just migration types I don't care about the type name, so I'ld be fine with the suffix tbh.

Thoralf-M: Yeah alright, fine if we change when really needed

sdk/src/wallet/migration/migrate_0.rs
```diff
@@ -169,39 +136,258 @@ trait Convert {
...
+        }
+    }
+
+    string_serde_impl!(TransactionId);
+
+    #[derive(Serialize, Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Transaction {
+        pub payload: TransactionPayload,
+        pub block_id: Option<serde_json::Value>,
```
Thoralf-M, line 212: What's the reason to use `serde_json::Value` instead of `String` here?

sdk/src/wallet/migration/migrate_0.rs
```diff
@@ -169,39 +136,258 @@ trait Convert {
...
-                build_transaction_from_payload_and_inputs(tx_id, tx_payload, inputs)?,
-            );
+            let types::TransactionEssence::Regular(tx_essence) = &tx_payload.essence;
+            let txn = types::Transaction {
+                network_id: tx_essence.network_id,
+                payload: tx_payload,
+                block_id: inputs
+                    .first()
+                    .map(|i: &types::OutputWithMetadataResponse| i.metadata.block_id.clone()),
+                inclusion_state: types::InclusionState::Confirmed,
```
thibault-martinez, line 352: Looks sus

thibault-martinez (approved):

sdk/src/wallet/migration/migrate_0.rs
DaughterOfMars: Consistency?

sdk/src/wallet/migration/migrate_0.rs
DaughterOfMars: This is how the original was implemented. Is that incorrect?

Thoralf-M: Yes, incoming transaction are always confirmed, otherwise we wouldn't get them

Thoralf-M (approved):

PR #452, opened 2023-05-15
Remove some DTOs

thibault-martinez:


DaughterOfMars (approved):
Looks like there's more that can be done, but good start

qrayven (approved):

PR #457, opened 2023-05-16
Remove NativeTokenDto

thibault-martinez:


sdk/src/wallet/migration/migrate_0.rs
```diff
@@ -11,6 +11,77 @@ use crate::wallet::Error;
 
 pub struct Migrate;
 
+fn migrate_native_token(output: &mut serde_json::Value) {
```
DaughterOfMars, line 14: This may be a problem in the case where we are migrating a storage without a version set. Part of the purpose of the `Convert` trait is checking whether we already migrated.

```rust
 if serde_json::from_value::<Self::New>(value.clone()).is_err()
```

If we don't do this it's conceivable that we will have already migrated, but since we don't have the version in storage we try to do it again. Your code would panic then.

thibault-martinez: I'm checking if the `token_id` key exists though, wouldn't that be enough? What would panic still?

DaughterOfMars: Ah, nevermind. I thought the structure was changing but it's just a rename.

Thoralf-M (approved):

sdk/src/types/block/output/native_token.rs
```diff
@@ -20,6 +20,7 @@ use crate::types::block::{output::TokenId, Error};
 #[packable(unpack_error = Error)]
 pub struct NativeToken {
     // Identifier of the native token.
+    #[serde(rename = "id")]
```
kwek20, line 23: cant we migrate this to keep it token_id? rename should be avoided as much as possible id say

Thoralf-M: It's only id in the node API

thibault-martinez: Yeah it's the protocol

thibault-martinez: And there is no rename, it was already `id`, I just move the serde rename to the main type instead of the Dto

kwek20: wellshould we jsut call it id as well then?

thibault-martinez: To be a bit precise, the protocol TIP call it token_id and the API call it id, which is exactly what we have atm, so I'd rather commit to the TIPs 🤷🏻‍♂️

thibault-martinez: I know it's weird / not ideal but that's how it is

kwek20: ah okay, yea ofcourse the tip should be adhered, perfect

kwek20 (approved):

PR #462, opened 2023-05-17
Allow SecretManage to be extensible

### LINKED ISSUE #407: Use Trait in favor of enum for SecretManager
## Description

(Copied over from Discord [here](https://discord.com/channels/397872799483428865/443602228813627392/1103320427843039242) )

We're trying to switch from Chrysalis to Shimmer (for our identity management product) and we're having an issue:

We would like to store our keys not in Stronghold but in a PKCS#11 compatible HSM. However, currently, all relevant APIs in iota_client require an instance of the `SecretManager` (https://github.com/iotaledger/iota-sdk/blob/develop/sdk/src/client/secret/mod.rs#L102) enum. None of the enums variants seem to fit our current usecase.

## Motivation

Only allowing Stronghold or Ledger Nano as key storages doesn't really work for industry^^

## Requirements

From our perspective, it would make sense to
  - Either add a generic variant to `SecretManager` that contains a `T: SecretManage` (aka a type that implements `SecretManage` (https://github.com/iotaledger/iota-sdk/blob/develop/sdk/src/client/secret/mod.rs#L59);
  - Or let APIs in iota_client that require `SecretManager` instead require `SecretManage` (which is implemented by `SecretManager` currently anyways (https://github.com/iotaledger/iota-sdk/blob/develop/sdk/src/client/secret/mod.rs#L227))
( - Or implement the PKCS#11 interface as an enum variant, but while this is probably the quickest fix, it really is also an ugly workaround and wouldn't really solve our issue since we're forking cryptoki anyways since they're missing EDDSA (https://github.com/parallaxsecond/rust-cryptoki/pull/113))

## Open questions (optional)

Any help/input would be much appreciated 🙂

Maybe?

DaughterOfMars:
# Description of change

Updates the `SecretManage` trait so that it can be extended by users and provided to a `Wallet` and `Account`, thereby enabling its usage via rust. This does not add the ability to pass a custom Secret Manager to the client methods that accept one, however, it may enable us to refactor those methods later to allow passing one in.

- Closes #407

## Type of change

## Open Questions
- I tried to make the types default to use the `SecretManager` enum, but this only seems to affect the impl blocks. So, I created a new type definition for the default and exported it from the top level. Not sure this is the best way.


sdk/src/wallet/wallet/operations/mod.rs
```diff
@@ -8,6 +8,8 @@ pub(crate) mod client;
 pub(crate) mod get_account;
 #[cfg(feature = "ledger_nano")]
 pub(crate) mod ledger_nano;
+
```
thibault-martinez:
```suggestion
```

sdk/src/wallet/wallet/operations/get_account.rs
```diff
@@ -1,17 +1,20 @@
...
+use crate::{
+    client::secret::SecretManage,
+    wallet::{
+        account::{types::AccountIdentifier, Account},
+        Wallet,
+    },
 };
 
-impl Wallet {
+impl<S: SecretManage> Wallet<S> {
```
thibault-martinez, line 12: Don't need the error bound here ?

DaughterOfMars: Apparently not lol

sdk/src/wallet/wallet/operations/storage.rs
```diff
@@ -0,0 +1,88 @@
...
+                }
+                Ok(Some(builder))
+            } else {
+                Ok(None)
+            }
+        }
+    }
+
+    #[async_trait]
+    impl SaveLoadWallet for WalletBuilder<MnemonicSecretManager> {
```
Thoralf-M, line 99: This is not needed with the generic version above?

sdk/src/wallet/wallet/operations/stronghold.rs
```diff
@@ -64,3 +67,46 @@ impl Wallet {
         }
     }
 }
+
+impl Wallet<StrongholdAdapter> {
```
Thoralf-M, line 76: What's this for? It seems to work just fine with this removed

sdk/src/wallet/wallet/operations/stronghold_backup/mod.rs
```diff
@@ -178,6 +173,167 @@ impl Wallet {
...
+            for account in accounts.iter() {
+                account.save(None).await?;
+            }
+        }
+
+        Ok(())
+    }
+}
+
+impl Wallet<StrongholdSecretManager> {
```
Thoralf-M, line 209: Same here, seems to work just fine with this whole block removed

sdk/src/wallet/wallet/operations/storage.rs
DaughterOfMars: It is very much needed. The point of `SaveLoadWallet` is to unify the types, some of which impl `SecretManagerConfig` and some of which don't.

sdk/src/wallet/wallet/operations/stronghold.rs
DaughterOfMars: Usability. You can use the enum wrapper `StorageManager` OR the underlying types and both will work.

sdk/src/wallet/wallet/operations/stronghold_backup/mod.rs
DaughterOfMars: Same answer as above.

sdk/src/wallet/wallet/operations/stronghold.rs
Thoralf-M: But we don't use it, so who does it help?

DaughterOfMars: Users of the lib. These are all publicly accessible.

Thoralf-M: Now I got it, took me some time to understand that this is additional to the SecretManager

Thoralf-M (dismissed):

sdk/src/wallet/wallet/operations/stronghold_backup/mod.rs
```diff
@@ -179,6 +174,169 @@ impl Wallet {
...
+                        .storage_options
+                        .storage_path
+                        .clone()
+                        .into_os_string()
+                        .into_string()
+                        .expect("can't convert os string"),
+                )
+                .with_client_options(self.client_options().await)
+                .with_coin_type(self.coin_type.load(Ordering::Relaxed));
+            // drop secret manager, otherwise we get a deadlock in save_wallet_data
```
Thoralf-M:
```suggestion
```

sdk/src/wallet/wallet/operations/stronghold_backup/mod.rs
```diff
@@ -179,6 +174,169 @@ impl Wallet {
...
+            }
+        });
+
+        // Update Wallet with read data
+        if ignore_if_coin_type_mismatch.is_none() {
+            if let Some(read_client_options) = read_client_options {
+                // If the nodes are from the same network as the current client options, then extend it
+                self.set_client_options(read_client_options).await?;
+            }
+        }
```
Thoralf-M: This needs to be moved to L309

Thoralf-M (dismissed):

sdk/src/wallet/wallet/operations/storage.rs
thibault-martinez, line 1: Why not in the storage module?

sdk/src/wallet/account/mod.rs
```diff
@@ -140,10 +143,19 @@ pub struct AccountDetails {
 }
 
 /// A thread guard over an account, so we can lock the account during operations.
-#[derive(Debug, Clone)]
-pub struct Account {
+#[derive(Debug)]
+pub struct Account<S: SecretManage = SecretManager> {
```
thibault-martinez, line 147: Changelog

sdk/src/wallet/account/builder.rs
```diff
@@ -15,16 +15,19 @@ use crate::{
 };
 
 /// The AccountBuilder
-pub struct AccountBuilder {
+pub struct AccountBuilder<S: SecretManage = SecretManager> {
```
thibault-martinez, line 18: Changelog

sdk/src/wallet/wallet/mod.rs
```diff
@@ -22,47 +22,62 @@ use crate::wallet::events::{
...
     wallet::account::{builder::AccountBuilder, operations::syncing::SyncOptions, types::AccountBalance, Account},
 };
 
 /// The wallet, used to create and get accounts. One wallet can hold many accounts, but they should
 /// all share the same secret_manager type with the same seed/mnemonic.
-#[derive(Debug, Clone)]
-pub struct Wallet {
-    pub(crate) inner: Arc<WalletInner>,
+#[derive(Debug)]
+pub struct Wallet<S: SecretManage = SecretManager> {
```
thibault-martinez, line 35: Do we need a type alias in lib.rs if we already have a default here?

sdk/src/wallet/wallet/builder.rs
```diff
@@ -25,19 +26,35 @@ use crate::wallet::{
...
-    client::secret::SecretManager,
+    client::secret::{SecretManage, SecretManager},
     wallet::{wallet::WalletInner, Account, ClientOptions, Wallet},
 };
 
-#[derive(Debug, Serialize, Deserialize, Default)]
+#[derive(Debug, Serialize, Deserialize)]
 /// Builder for the wallet.
-pub struct WalletBuilder {
+pub struct WalletBuilder<S: SecretManage = SecretManager> {
```
thibault-martinez, line 35: changelog

sdk/src/wallet/wallet/mod.rs
thibault-martinez, line 35: Do we need a type alias in lib.rs if we already have a default here?

thibault-martinez, line 35: Changelog

DaughterOfMars: This, unfortunately, is only applied in impl blocks.

sdk/src/wallet/wallet/operations/storage.rs
DaughterOfMars: Well that module is feature gated, and this relies on `not(feature = "storage")`.

DeppLearning:
Thanks a lot for your work on this! Do I understand correctly that #407 won't be resolved by this as per

> This does not add the ability to pass a custom Secret Manager to the client methods that accept one, however, it may enable us to refactor those methods later to allow passing one in

If so,does this mean that all the type signatures of the clients methods that are used by the identity.rs crates and currently take a `SecretManager` struct, will be refactored to be generic over types that implement the corresponding trait? Is there  already some roadmap for these changes? For us to use this without forking, we also need those changes to be incorporated into the identity.rs crates.

DaughterOfMars:
> Do I understand correctly that #407 won't be resolved by this

Maybe. It depends on whether using our bindings with the custom secret manager is a strong requirement. To clarify, custom bindings can be written for just the pieces needed by this (basically copy/pasted from ours) with either a custom `SecretManager` enum or the direct custom type.

> If so,does this mean that all the type signatures of the clients methods that are used by the identity.rs crates and currently take a `SecretManager` struct, will be refactored to be generic over types that implement the corresponding trait?

If there are functions in the client that cannot be moved to the secret manager, then likely they will be adapted to accept a generic secret manager rather than sending it as part of the message.

> Is there already some roadmap for these changes? For us to use this without forking, we also need those changes to be incorporated into the identity.rs crates.

The current implementation for secret manager functions is usable generically. For the client methods which accept a secret manager, custom bindings will need to be created. These can be extremely minimal, and only used to handle those messages.

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -398,6 +374,54 @@ impl SignTransactionEssence for LedgerSecretManager {
...
+    fn to_config(&self) -> Option<Self::Config> {
+        Some(self.is_simulator)
+    }
+
+    fn from_config(config: &Self::Config) -> Result<Self, Self::Error> {
+        Ok(Self::new(*config))
+    }
+}
+
+/// needs_blind_signing
```
Alex6323: remove?

DaughterOfMars: Done

sdk/src/client/secret/placeholder.rs
```diff
@@ -3,68 +3,7 @@
...
-use super::{GenerateAddressOptions, SecretManage, SignTransactionEssence};
-use crate::{
-    client::{secret::PreparedTransactionData, Error},
-    types::block::{address::Ed25519Address, signature::Ed25519Signature, unlock::Unlocks},
-};
-
 /// Secret manager that is only useful to prevent accidental address generation in a wallet
 /// that has an offline counterpart for address generation and signing.
+#[derive(Debug)]
 pub struct PlaceholderSecretManager;
```
thibault-martinez, line 8: Do we actually need this at all? It looks like just a unit variant of the enum would be enough.

DaughterOfMars: It's required by some bounds checks.

thibault-martinez: I tried removing it locally and it's fine 🤷🏻‍♂️

DaughterOfMars: 🤨

DaughterOfMars: RIght, it's because I removed the trait impl.

sdk/src/wallet/account/operations/balance.rs
```diff
@@ -19,10 +20,13 @@ use crate::{
...
 
-impl Account {
-    /// Get the balance of the account.
-    pub async fn balance(&self) -> crate::wallet::Result<Balance> {
-        log::debug!("[BALANCE] balance");
+impl<S: 'static + SecretManage> Account<S>
+where
+    Error: From<S::Error>,
+{
+    /// Get the AccountBalance
```
thibault-martinez:
```suggestion
    /// Get the balance of the account.
```
AccountBalance is no more

sdk/src/client/secret/placeholder.rs
thibault-martinez: You don't have to remove it now, we can do in a separate PR because maybe it touches the bindings as well idk

DaughterOfMars: No, I got it.

thibault-martinez: What do you mean I got it? Are you removing it now or later or not at all?

DaughterOfMars: I already did it yesterday 😆

thibault-martinez: Well it's still here ?

DaughterOfMars: You must be looking at an out of date file...

EDIT: I am dumb

thibault-martinez: https://github.com/iotaledger/iota-sdk/issues/573

sdk/src/wallet/wallet/operations/address_generation.rs
```diff
@@ -109,13 +112,173 @@ impl Wallet {
...
+            )
+            .await?;
+
+        Ok(*address
+            .first()
+            .ok_or(crate::wallet::Error::MissingParameter("address"))?)
     }
+}
+
+impl Wallet<MnemonicSecretManager> {
```
Thoralf-M: Is there actually a good reason why anyone should use the Wallet with a SecretManager not via the enum, if there is a variant for it? Because to me it looks like an option nobody will use but just adds a lot more code

DaughterOfMars: Well, I'd rather have some duplicated code than completely remove the ability to use the wallet like this.

Thoralf-M: Why "remove" the ability? It was never there and you can still use a MnemonicSecretManager in the enum without this

Thoralf-M: Also the function is available via the MnemonicSecretManager directly, no need for the wallet (currently it's just there for old bindings and for ledger nano because of the events)

DaughterOfMars: Hmm, fair point. Okay, I can remove it then.

DaughterOfMars: Done

sdk/src/wallet/wallet/operations/address_generation.rs
```diff
@@ -109,13 +112,137 @@ impl Wallet {
...
+        };
+
+        Ok(*address
+            .first()
+            .ok_or(crate::wallet::Error::MissingParameter("address"))?)
+    }
+}
+
+#[cfg(feature = "stronghold")]
+impl Wallet<crate::client::secret::stronghold::StrongholdSecretManager> {
```
Thoralf-M: This and LedgerNano should also be removed I think

DaughterOfMars: Done

sdk/src/wallet/wallet/operations/stronghold.rs
Thoralf-M: Sorry coming back to this again, but since the user has access to the secret manager and can just call the methods on it, this is not needed and has no real benefit for any users?

DaughterOfMars: Do you think the whole file should be removed?

Thoralf-M: No, the old implementation should stay for convenience and because it's used in the bindings
I'm just thinking that anyone who wants to use Stronghold should just use the enum variant and then has access to the partially more convenient functions as well, therefore there is no need to have them here "duplicated"

DaughterOfMars: I don't agree that you should be forced to use the enum if you want to access a particular function. The argument for only supporting backwards-compatable fns that will be removed in the future is fair, but for fns that have no plans to remove it doesn't make sense.

Thoralf-M (dismissed):

kwek20 (approved):

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -151,13 +151,17 @@ impl SecretManage for LedgerSecretManager {
...
             // and generate a single address that is shown to the user
-            ledger.get_addresses(true, bip32, address_indexes.len())?
+            ledger
+                .get_addresses(true, bip32, address_indexes.len())
+                .map_err(Error::from)?
         } else {
-            ledger.get_addresses(false, bip32, address_indexes.len())?
+            ledger
+                .get_addresses(false, bip32, address_indexes.len())
+                .map_err(Error::from)?
```
Alex6323, line 164: I think this can be DRYed by putting `options.ledger_nano_prompt` as the first argument of `.get_addresses(..)`?

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -329,7 +303,7 @@ impl SignTransactionEssence for LedgerSecretManager {
                                 }
                             } else {
                                 log::debug!("[LEDGER] unsupported output");
-                                return Err(Error::MiscError);
+                                return Err(Error::MiscError.into());
```
Alex6323, line 306: it may be a bit uncommon but I think it's worth mentioning at least that you could just use `?` to get rid of the explicit `.into()` calls. Taking this line as an example this would reduce to:
```rust
Err(Error::MiscError)?
```

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -398,6 +374,53 @@ impl SignTransactionEssence for LedgerSecretManager {
...
+            for output in essence.outputs().iter() {
+                // only basic outputs allowed
+                if let Output::Basic(output) = output {
+                    if output.simple_deposit_address().is_some() {
+                        continue;
+                    }
+                }
+                // not fine, return
+                return true;
+            }
```
Alex6323, line 414: this loop could be replaced with:
```rust
            if !essence
                .outputs()
                .iter()
                .all(|o| matches!(o, Output::Basic(output) if output.simple_deposit_address().is_some()))
            {
                return true;
            }
```
It's debatable whether that's more readable, but IMHO it is.

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -398,6 +374,53 @@ impl SignTransactionEssence for LedgerSecretManager {
...
+        }
+    }
+    // check if essence + bip32 indices fit into the buffer of the device
+    let essence_bytes = prepared_transaction.essence.pack_to_vec();
```
Alex6323, line 418: can you not just use `.packed_len()`? seems like you don't need the `essence_bytes` for anything else.

sdk/src/client/secret/mod.rs
```diff
@@ -373,99 +410,103 @@ impl SecretManager {
...
+    crate::client::Error: From<M::Error>,
+{
+    log::debug!("[sign_transaction] {:?}", prepared_transaction_data);
+    let current_time = unix_timestamp_now().as_secs() as u32;
 
-        validate_transaction_payload_length(&tx_payload)?;
+    let unlocks = secret_manager
+        .sign_transaction_essence(&prepared_transaction_data, Some(current_time))
+        .await?;
+    let tx_payload = TransactionPayload::new(prepared_transaction_data.essence.clone(), unlocks)?;
```
Alex6323, line 500: Can't we destructure `PreparedTransactionData`, so we don't need to clone the essence here?

sdk/src/wallet/account/operations/balance.rs
```diff
@@ -19,10 +20,13 @@ use crate::{
...
-impl Account {
+impl<S: 'static + SecretManage> Account<S>
+where
+    Error: From<S::Error>,
+{
     /// Get the balance of the account.
-    pub async fn balance(&self) -> crate::wallet::Result<Balance> {
-        log::debug!("[BALANCE] balance");
+    pub async fn balance(&self) -> Result<Balance> {
+        log::debug!("[BALANCE] get balance");
```
Alex6323, line 29: I am okay with adding `get` in general, but it's not done consistently atm (see `addresses_balance` fn). Don't know how much effort it would be to make it consistent across the whole code base. Maybe different PR that fixes some logging?

sdk/src/wallet/account/operations/output_consolidation.rs
```diff
@@ -143,28 +153,32 @@ impl Account {
...
-                            let available_buffer_size_for_inputs =
-                                buffer_size - ESSENCE_SIZE_WITHOUT_IN_AND_OUTPUTS - MIN_OUTPUT_SIZE_IN_ESSENCE;
-                            (available_buffer_size_for_inputs / INPUT_SIZE) as u16
-                        })
-                        .unwrap_or(INPUT_COUNT_MAX)
-                }
+        #[cfg(feature = "ledger_nano")]
+        let max_inputs = if let Some(ledger) = ((&*self.wallet.secret_manager.read().await)
+            as &(dyn std::any::Any + Send + Sync))
+            .downcast_ref::<crate::client::secret::ledger_nano::LedgerSecretManager>()
```
Alex6323, line 159: This downcasting is done in several places, so it might be worth moving it into a method 
`async fn get_ledger_secret_manager(&self) -> Option<_>`, maybe attributed with an inline or smth. Might improve readability of the code a bit.

edit: or maybe just a `async fn is_ledger_secret_manager(&self) -> bool` so we don't hold the read lock for longer than we actually need to. Wdyt?

sdk/src/wallet/error.rs
```diff
@@ -112,6 +109,12 @@ pub enum Error {
     /// Transaction not found
     #[error("transaction {0} not found")]
     TransactionNotFound(TransactionId),
+    // TODO more precise error
```
Alex6323, line 112: TODO (want to create an issue?)

thibault-martinez: It's irrelevant to this PR and there is already an issue to handle TODOs

Alex6323: okidoki

sdk/src/wallet/wallet/operations/storage.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    #[async_trait]
+    pub trait SaveLoadWallet {
+        async fn save_data(
+            &self,
+            storage: &impl StorageAdapter<Error = crate::wallet::Error>,
+        ) -> crate::wallet::Result<()>;
+
+        async fn get_data(
```
Alex6323, line 50: Considering the trait is named **SaveLoad**Wallet, maybe `load_data` is the more expected name for this fn? Also maybe `save` and `load` are fine names already?

sdk/src/wallet/wallet/operations/stronghold_backup/mod.rs
```diff
@@ -185,6 +181,170 @@ impl Wallet {
...
+        });
+
+        if !ignore_backup_values {
+            if let Some(read_coin_type) = read_coin_type {
+                self.coin_type.store(read_coin_type, Ordering::Relaxed);
+            }
+        }
+
+        if let Some(mut read_secret_manager) = read_secret_manager {
+            read_secret_manager.snapshot_path = new_snapshot_path.clone().into_os_string().to_string_lossy().into();
```
Alex6323, line 293: Is `.into_os_string` necessary? I've seen that we sometimes call it and sometimes we don't.

sdk/src/client/secret/mnemonic.rs
```diff
@@ -113,6 +119,21 @@ impl SecretManage for MnemonicSecretManager {
...
+    ) -> Result<Unlocks, Self::Error> {
+        super::default_sign_transaction_essence(self, prepared_transaction_data, time).await
+    }
+
+    async fn sign_transaction(
+        &self,
+        prepared_transaction_data: PreparedTransactionData,
+    ) -> Result<Payload, Self::Error> {
+        super::default_sign_transaction(self, prepared_transaction_data).await
+    }
```
thibault-martinez, line 136: Shouldn't these be a default in the trait itself instead?

sdk/src/wallet/account/operations/output_consolidation.rs
thibault-martinez: I think I would also do cast methods, probably `as_*`

sdk/src/wallet/wallet/operations/storage.rs
thibault-martinez: +1

sdk/src/client/secret/ledger_nano.rs
DaughterOfMars: True, though that fn just calls pack_to_vec().len() 😅

sdk/src/client/secret/mnemonic.rs
DaughterOfMars: They cannot be, because the associated error type is not known there, and the default impl returns a `Result<_, client::Error>`.

thibault-martinez: sad reacts only

sdk/src/client/secret/ledger_nano.rs
Alex6323: damn :joy:

thibault-martinez: Not really true, it internally uses a specific packer that doesn't allocate

thibault-martinez (approved):

sdk/src/client/secret/mod.rs
DaughterOfMars: This one was also resolved by the other PR

sdk/src/client/secret/ledger_nano.rs
DaughterOfMars: I don't like this change 🤷‍♀️

Alex6323: fair enough, but in some file you (then probably accidentally) did it yourself. Or maybe it wasn't you.

sdk/src/client/secret/mod.rs
Alex6323: you're right, missed it

sdk/src/client/secret/ledger_nano.rs
DaughterOfMars: Shh, I'm allowed to be a hypocrite

Alex6323 (approved):

Issue #129, opened 2023-03-09
QA: Make sure all bindings are properly typed


thibault-martinez:
- [x] Python
- [x] Nodejs

Dr-Electron:
Python: 

- [x] generate_addresses_options  in consolidate_funds https://github.com/iotaledger/iota-sdk/pull/683
- [x] TransactionOptions for different functions https://github.com/iotaledger/iota-sdk/pull/640
- [x] All Params of different send or prepare functions https://github.com/iotaledger/iota-sdk/pull/783
- [x] client_options in Wallet constructor https://github.com/iotaledger/iota-sdk/pull/805
- [x] In some places we don't "typehint" the secret_manager with `secret_manager: Optional[LedgerNanoSecretManager | MnemonicSecretManager | StrongholdSecretManager]` https://github.com/iotaledger/iota-sdk/pull/683
- [x] sync options https://github.com/iotaledger/iota-sdk/pull/707

Nodejs seems fine, so I guess was should be too. 
Java is not a thing anymore and swift won't be in our repo, right?

Thoralf-M:
> In some places we don't "typehint" the secret_manager with secret_manager: Optional[LedgerNanoSecretManager | MnemonicSecretManager | StrongholdSecretManager]

Can this be done with SecretManagerType alias or something like that? We have a `SecretManager` class which is different, so that name we can't use :thinking:

Dr-Electron:
> > In some places we don't "typehint" the secret_manager with secret_manager: Optional[LedgerNanoSecretManager | MnemonicSecretManager | StrongholdSecretManager]
> 
> Can this be done with SecretManagerType alias or something like that? We have a `SecretManager` class which is different, so that name we can't use 🤔

You could create an Alias for all types. But why would you want to do that? Can't we just use this in every function where a secret_manager can be passed?

So something like:
```python
function_with_secret_manager(secret_manager: Optional[LedgerNanoSecretManager | MnemonicSecretManager | StrongholdSecretManager])
```

Thoralf-M:
Sure we can do that, just is very large to always have all

Dr-Electron:
Yeah true. But I prefer it, as it clearly shows all options 🤔

thibault-martinez:
Python:

These needs to be objects, not dicts
- [x] The Transaction returned by send functions https://github.com/iotaledger/iota-sdk/pull/783
[...]

PR #463, opened 2023-05-19
Python: add type hints

Thoralf-M:
# Description of change

Add https://peps.python.org/pep-0484 type hints to some parts
Other parts should be done new PRs so it's easier to review

#129 

## Type of change

mypi

## Change checklist

### DIFF
--- a/bindings/python/iota_sdk/utils.py
+++ b/bindings/python/iota_sdk/utils.py
@@ -1,9 +1,10 @@
 # Copyright 2023 IOTA Stiftung
 # SPDX-License-Identifier: Apache-2.0
 
-from iota_sdk import call_utils_method, OutputId
+from iota_sdk import call_utils_method
+from iota_sdk.types.output_id import OutputId
 from json import dumps, loads
-from typing import Dict, Any
+from typing import Any, Dict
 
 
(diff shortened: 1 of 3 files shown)


bindings/python/iota_sdk/utils.py
```diff
@@ -43,7 +44,7 @@ def nft_id_to_bech32(nft_id, bech32_hrp):
         })
 
     @staticmethod
-    def hex_public_key_to_bech32_address(hex, bech32_hrp=None):
+    def hex_public_key_to_bech32_address(hex: str, bech32_hrp=None) -> str:
```
thibault-martinez: Is there no type int when there is also a default value?

Thoralf-M: There is, would be `bech32_hrp: Optional[str]=None`, but was actually wrong since it's not optional here
Updated it

thibault-martinez (approved):

DaughterOfMars (approved):
:o you can do that?

Thoralf-M:
> :o you can do that?

Yes, but it has no impact on when you run the code, it's only for the devs

PR #468, opened 2023-05-22
feat(types)!: introduce `Password` type that is `ZeroizeOnDrop`

### LINKED ISSUE #384: Add a password type that is ZeroizeOnDrop?

Alex6323:
# Description of change

Introduces the `Password` type that takes care of zeroizing this sensitive data (i.e. the memory) when the instance gets dropped.

I tried to make it as less breaking as possible, but instead of a reference `StrongholdSecretManager` builder now takes ownership when providing the password, which is no problem because in most cases we have ownership in the first place and can just move it.

Closes #384 

## Type of change

Unit tests.


sdk/src/wallet/wallet/operations/stronghold_backup/mod.rs
```diff
@@ -92,9 +93,11 @@ impl Wallet {
             PathBuf::from("wallet.stronghold")
         };
 
+        let stronghold_password = stronghold_password.into();
```
Thoralf-M: Let's say you provide a String and then call .into() to make it a Password, then the provided string will not be zeroized, no?

Alex6323: yeah, now the ownership of the String will be transferred to the `Password` type, so `ZeroizeOnDrop` will take care of it.

Alex6323: the `impl<T: AsRef<str>> From<T> for Password` can't be used unfortunately. I realized it and removed it. Too bad because it kept the convencience what we had before, but then if you pass it a `String s` it would call `s.as_ref().to_owned` which would essentially create a clone, and `s` would not be zeroized.

Alex6323: As a pre-caution I did not impl `From<&str>` or `From<&String`, because that could lead to someone forgetting to zeroize the original manually. That causes some manual clones, but I think safety has a higher priority here.

sdk/src/client/stronghold/mod.rs
```diff
@@ -554,7 +555,7 @@ mod tests {
 
         let stronghold_path = "test_clear_key.stronghold";
         let mut adapter = StrongholdAdapter::builder()
-            .password("drowssap")
```
kwek20, line 567: never noticed this, what does it stand for :D

sdk/src/client/secret/types.rs
```diff
@@ -32,6 +32,19 @@ pub struct StrongholdDto {
     /// The path for the Stronghold file
     pub snapshot_path: String,
 }
+
+/// Represents the Stronghold password.
```
kwek20: Do we need to mention Stronghold? could also be used for the SEED_1 we still have somewhere. Its just a password wrapper

sdk/src/wallet/wallet/operations/stronghold.rs
```diff
@@ -3,26 +3,29 @@
...
-        current_password: &str,
-        new_password: &str,
+        current_password: impl Into<Password> + Send,
+        new_password: impl Into<Password> + Send,
     ) -> crate::wallet::Result<()> {
         if let SecretManager::Stronghold(stronghold) = &mut *self.secret_manager.write().await {
-            stronghold.set_password(current_password).await?;
-            stronghold.change_password(new_password).await?;
+            stronghold.set_password(current_password.into()).await?;
+            stronghold.change_password(new_password.into()).await?;
```
Thoralf-M: If `set_password()` returns an error, then `new_password` would never be zeroized?

sdk/src/client/stronghold/mod.rs
Thoralf-M: reversed?

sdk/src/wallet/wallet/operations/stronghold.rs
Alex6323: good catch ... it may not!

sdk/src/client/secret/types.rs
Alex6323: hmm, but isn't a Password something else than a Seed? You can derive addresses from a seed but not necessarily from a password, right?

sdk/src/client/stronghold/mod.rs
Alex6323: reversed? no, no. It's bad english for "drowning in sap/juice" :yum: :trollface:

kwek20: damn mind=blown

thibault-martinez (changes requested):
Conflicts

Thoralf-M (approved):

Thoralf-M (changes requested):
CI fails

sdk/src/client/secret/types.rs
Alex6323: Done

DaughterOfMars (dismissed):

cli/src/command/wallet.rs
```diff
@@ -110,9 +110,9 @@ pub async fn backup_command(storage_path: &Path, snapshot_path: &Path, backup_pa
 
 pub async fn change_password_command(storage_path: &Path, snapshot_path: &Path) -> Result<Wallet, Error> {
     let password = get_password("Stronghold password", !snapshot_path.exists())?;
```
thibault-martinez, line 119: Surely this should return a `Password`?

Alex6323: makes much more sense, I had this at one point and for some reason I can't remember undid it. (undid? What a weird word)

Alex6323: done

sdk/src/client/stronghold/migration.rs
```diff
@@ -12,17 +12,20 @@ use iota_stronghold::{Client, SnapshotPath, Stronghold};
...
-use crate::client::stronghold::{check_or_create_snapshot, storage::insert as v3_insert, Error as StrongholdError};
+use crate::client::{
+    secret::types::Password,
+    stronghold::{check_or_create_snapshot, storage::insert as v3_insert, Error as StrongholdError},
+};
 
 impl StrongholdAdapter {
     /// Migrates a snapshot from version 2 to version 3.
-    pub fn migrate_snapshot_v2_to_v3<P: AsRef<Path>>(
+    pub fn migrate_snapshot_v2_to_v3<P: AsRef<Path>, PW: Into<Password>>(
```
DaughterOfMars: Using a single generic forces the two passwords to be the same type, FYI

Alex6323: yeah, that's good to know. I actually ran into the problem at first, then realized why it was done this way with the path.

DaughterOfMars (dismissed):

kwek20 (dismissed):

kwek20 (dismissed):

sdk/examples/client/stronghold.rs
```diff
@@ -15,7 +15,7 @@ use iota_sdk::client::{
 #[tokio::main]
 async fn main() -> Result<()> {
     let stronghold_secret_manager = StrongholdSecretManager::builder()
-        .password("some_hopefully_secure_password")
+        .password("some_hopefully_secure_password".to_owned())
```
thibault-martinez, line 18: Maybe impl `From<&str>` for `Password` so the UX isn't degraded?

DaughterOfMars: The problem is that this encourages passing in referenced data which will not be zeroized. Tbh the whole thing is highly dependent on the user managing their data properly, but I think forcing owned data here is better.

thibault-martinez: But these are string literals they'll end up in the binary and loaded in memory so not zeroized anyway

DaughterOfMars: Right but my point is references should never be passed to this fn (with the exception of our tests and examples which are not secure obviously)

Alex6323: I thought about that, but considered it too much of a risk just to make the examples and tests more convenient. In actual code the user would never have a `&static str` for a password, but instead get an owned `String` from somewhere. If we allow `&str`, then the user might make the error of providing just a reference to the String, and forget about zeroing the original on his own. He could also just clone the String and also forget about zeroing the original, but I consider that to be acceptable, because we did what we could to protect his secret.

Alex6323: oh, shoot, browser didn't update the discussion ... exactly what Alex has said is my point as well.

sdk/src/client/stronghold/migration.rs
```diff
@@ -12,17 +12,20 @@ use iota_stronghold::{Client, SnapshotPath, Stronghold};
...
         current_path: P,
-        current_password: &str,
-        salt: &str,
+        current_password: PW,
+        mut salt: String,
```
thibault-martinez: This is not really required tbh, it's the same for everyone...

sdk/CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Multiple prepare methods returning `PreparedMintTokenTransaction`: `prepare_mint_native_token` and `prepare_increase_native_token_supply`;
 - Stronghold snapshot migration from v2 to v3;
 - `SecretManage::sign_evm`;
+- `Password` type which derives `Zeroize` and `ZeroizeOnDrop`;
 
 ### Changed
```
thibault-martinez, line 43: I would mention that some function now takes an `impl Into<Password>` since it's breaking change

sdk/src/client/stronghold/mod.rs
```diff
@@ -286,7 +287,7 @@ impl StrongholdAdapter {
     /// data, provide a list of keys in `keys_to_re_encrypt`, as we have no way to list and iterate over every
     /// key-value in the Stronghold store - we'll attempt on the ones provided instead. Set it to `None` to skip
     /// re-encryption.
-    pub async fn change_password(&self, new_password: &str) -> Result<(), Error> {
+    pub async fn change_password(&self, new_password: Password) -> Result<(), Error> {
```
thibault-martinez: Not Into?

sdk/src/wallet/wallet/operations/stronghold.rs
```diff
@@ -3,26 +3,34 @@
...
+    wallet::Wallet,
+};
 
 impl Wallet {
     /// Sets the Stronghold password
-    pub async fn set_stronghold_password(&self, password: &str) -> crate::wallet::Result<()> {
+    pub async fn set_stronghold_password(&self, password: impl Into<Password> + Send) -> crate::wallet::Result<()> {
         if let SecretManager::Stronghold(stronghold) = &mut *self.secret_manager.write().await {
-            stronghold.set_password(password).await?;
+            stronghold.set_password(password.into()).await?;
```
thibault-martinez: I think we should be extra careful and do the .into() FIRST thing in the function. What if we don't go in the if?

sdk/src/client/stronghold/common.rs
```diff
@@ -29,7 +31,7 @@ pub(super) const PRIVATE_DATA_CLIENT_PATH: &[u8] = b"iota_seed";
 pub(super) const USERDATA_STORE_KEY_RECORD_PATH: &[u8] = b"userdata-store-key";
 
 /// Hash a password, deriving a key, for accessing Stronghold.
-pub(super) fn key_provider_from_password(password: &str) -> KeyProvider {
+pub(super) fn key_provider_from_password(password: Password) -> KeyProvider {
```
thibault-martinez, line 34: Into? Or not, because it's private ?

sdk/src/wallet/wallet/operations/stronghold.rs
thibault-martinez: So yeah, if you could do them all at the beginning that would be best 🙏🏻

sdk/src/client/secret/types.rs
```diff
@@ -32,6 +31,17 @@ pub struct StrongholdDto {
     /// The path for the Stronghold file
     pub snapshot_path: String,
 }
+
+/// A password wrapper that takes care of zeroing the memory when being dropped.
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Zeroize, ZeroizeOnDrop, derive_more::From)]
+pub struct Password(String);
```
thibault-martinez: I kinda think this should be in `utils`. I understand why `secret` sounds like an interesting place to have such a type but in reality, it has absolutely nothing to do with the content of this module.

sdk/src/client/stronghold/common.rs
Alex6323: Yeah, that's the reason. Another one is that the function name also suggests that we already have a `Password`.

thibault-martinez: Fair enough

sdk/src/client/secret/types.rs
Alex6323: I agree. However, I wouldn't expect such a type to live in `utils` when there's a module `secret`. We also have a `mnemonic` module, so for consistency it might make sense to also have a `password` module. I decided against that in the first place because it wouldn't contain much except for this type and the tests associated with it. But `utils` doesn't sound right to me tbh.

thibault-martinez: Mnemonic is there because it contains a `SecretManage` implementation which is the purpose of the `secret` module. Nothing in `secret` requires `Password` to be there so it shouldn't be there.

sdk/src/wallet/wallet/operations/stronghold.rs
Alex6323: you are right...that's so subtle and not trivially solved by just calling it before the if. If we don't go in the if there's no guarantee that LLVM will execute this line at all, because it will just see that the converted type is not going to be used.

thibault-martinez: If it's before the if, I'm pretty sure we have the guarantee it'll be executed.

Alex6323: no, I am afraid we don't have that guarantee as LLVM is allowed to reorder instructions as it pleases in release mode. But there's an easy fix how we can make sure that the given argument is zeroized. Pushing it in a bit.

thibault-martinez: Well anyway, surely it's better at the top of the function

Alex6323: I realized that IMHO it's much easier to call `.into()` on the caller side than to take an `impl` as an argument when it comes to zeroizable data like this, because if in some path `.into()` is NOT called, then WE would be responsible for zeroizing the argument which leads to ugly `if ... else`s and mutable arguments.

thibault-martinez: Are you really 100% sure into wouldn't be called? Because I'm not.

Alex6323: I am almost certain that compilers nowadays are very good at detecting which lines are redundant, and even if they wouldn't be yet there's no guarantee they won't in the future. @Alexandcoats What's your take on this?

Alex6323: I'll push what I changed and you can take another look, Thibault.

thibault-martinez: Well you already know what I'm going to say and I have to say it, but it makes UX even worse than it already was 😬

sdk/src/client/secret/types.rs
Alex6323: Okay. I agree. So `utils` it is. Are you okay with a `client/utils/password.rs`?

sdk/src/client/stronghold/migration.rs
Alex6323: removed

sdk/CHANGELOG.md
Alex6323: ~`impl Into<Password>` has been removed for the possible security issues you pointed out.~

sdk/src/client/secret/types.rs
thibault-martinez: We have `sdk/utils` and `sdk/client/utils.rs`, no need ti create a new module

sdk/src/client/stronghold/mod.rs
Alex6323: This is probably outdated. It's best to not use `Into<Password>` no matter how convenient it is, bc the problem is that it might be the case that `.into()` is not called in certain circumstances which would then leak the secret. To protect against that we would need to make those arguments mutable, add the Zeroize trait bound, and zeroize them manually. So instead it's best to enforce the caller to provide the `Password` instance already.

sdk/src/wallet/wallet/operations/stronghold.rs
Alex6323: yeah, I know ... the good thing is though, that passwords are usually provided once, right, and then the user is done with it. So it's a one-time thing basically and we also protected their secret better which probably is/should be their main concern.

sdk/src/client/secret/types.rs
Alex6323: oh, damn, looked at the wrong utils. Moving ...

Alex6323: Done

sdk/src/client/stronghold/migration.rs
```diff
@@ -12,17 +12,20 @@ use iota_stronghold::{Client, SnapshotPath, Stronghold};
...
+};
 
 impl StrongholdAdapter {
     /// Migrates a snapshot from version 2 to version 3.
     pub fn migrate_snapshot_v2_to_v3<P: AsRef<Path>>(
         current_path: P,
-        current_password: &str,
-        salt: &str,
+        current_password: Password,
+        salt: String,
```
thibault-martinez: Revert the salt, or at least make it `Into<String>`

sdk/CHANGELOG.md
Alex6323: Updated

sdk/src/client/stronghold/mod.rs
Alex6323: outdated ... can be closed ... I think we made sure that `.into()` is called whenever we have an `impl Into<Password>`

sdk/src/wallet/wallet/operations/stronghold.rs
Alex6323: can be closed IMO

sdk/src/client/secret/types.rs
Alex6323: close?

thibault-martinez (changes requested):
CI fails

sdk/src/wallet/wallet/operations/stronghold.rs
```diff
@@ -17,11 +20,13 @@ impl Wallet {
     /// Change the Stronghold password to another one and also re-encrypt the values in the loaded snapshot with it.
     pub async fn change_stronghold_password(
         &self,
-        current_password: &str,
-        new_password: &str,
+        current_password: impl Into<Password> + Send,
+        new_password: impl Into<Password> + Send,
```
thibault-martinez, line 26: Need to call into for these 2

bindings/python/src/lib.rs
```diff
@@ -51,18 +51,18 @@ pub fn call_utils_method(method: String) -> Result<String> {
...
+    salt: &str,
     rounds: u32,
     new_path: Option<String>,
     new_password: Option<String>,
 ) -> Result<()> {
     Ok(StrongholdAdapter::migrate_snapshot_v2_to_v3(
         &current_path,
-        &current_password,
-        &salt,
+        current_password.into(),
```
thibault-martinez, line 61: Any reason not to take an Into for this function?

Alex6323: if you feel strongly about it I can change that too. My reasoning was that `migration` is special in the sense that it's not repeatedly used (right?), hence the convenience aspect is kind of irrelevant. So I wanted to discuss whether it's worth making the function signature more complicated by adding another generic for no true benefit.

sdk/src/wallet/wallet/operations/stronghold.rs
Alex6323: Done

bindings/python/src/lib.rs
thibault-martinez: I don't, was just wondering, fair enough

sdk/src/wallet/wallet/operations/stronghold.rs
```diff
@@ -17,11 +23,17 @@ impl Wallet {
...
     ) -> crate::wallet::Result<()> {
+        // Note: we need to enforce the conversions here to make sure sensitive data is zeroed.
+        let current_password = current_password.into();
+        let new_password = new_password.into();
+
         if let SecretManager::Stronghold(stronghold) = &mut *self.secret_manager.write().await {
-            stronghold.set_password(current_password).await?;
+            if let Err(err) = stronghold.set_password(current_password).await {
+                return Err(err.into());
```
thibault-martinez: revert?

thibault-martinez (dismissed):

sdk/src/client/stronghold/mod.rs
```diff
@@ -286,7 +292,9 @@ impl StrongholdAdapter {
     /// data, provide a list of keys in `keys_to_re_encrypt`, as we have no way to list and iterate over every
     /// key-value in the Stronghold store - we'll attempt on the ones provided instead. Set it to `None` to skip
     /// re-encryption.
-    pub async fn change_password(&self, new_password: &str) -> Result<(), Error> {
+    pub async fn change_password(&self, new_password: impl Into<Password> + Send) -> Result<(), Error> {
+        let new_password = new_password.into();
```
Thoralf-M, line 294: Also put this here?
```suggestion
        // Note: we deliberately put the password conversion right here. Only change if you know what you're doing.
        let new_password = new_password.into();
```

sdk/src/wallet/wallet/operations/stronghold_backup/mod.rs
```diff
@@ -6,21 +6,29 @@ pub(crate) mod stronghold_snapshot;
...
 impl Wallet {
     /// Backup the wallet data in a Stronghold file
     /// stronghold_password must be the current one when Stronghold is used as SecretManager.
-    pub async fn backup(&self, backup_path: PathBuf, mut stronghold_password: String) -> crate::wallet::Result<()> {
+    pub async fn backup(
+        &self,
+        backup_path: PathBuf,
+        stronghold_password: impl Into<Password> + Send,
+    ) -> crate::wallet::Result<()> {
+        let stronghold_password = stronghold_password.into();
```
Thoralf-M, line 30: Same here?

sdk/src/wallet/wallet/operations/stronghold_backup/mod.rs
```diff
@@ -67,10 +73,12 @@ impl Wallet {
     pub async fn restore_backup(
         &self,
         backup_path: PathBuf,
-        mut stronghold_password: String,
+        stronghold_password: impl Into<Password> + Send,
         ignore_if_coin_type_mismatch: Option<bool>,
         ignore_if_bech32_hrp_mismatch: Option<Hrp>,
     ) -> crate::wallet::Result<()> {
+        let stronghold_password = stronghold_password.into();
```
Thoralf-M, line 80: And here?

sdk/src/client/stronghold/mod.rs
```diff
@@ -149,7 +149,10 @@ fn check_or_create_snapshot(
 /// Extra / custom builder method implementations.
 impl StrongholdAdapterBuilder {
     /// Use an user-input password string to derive a key to use Stronghold.
-    pub fn password(mut self, password: &str) -> Self {
+    pub fn password(mut self, password: impl Into<Password>) -> Self {
+        // Note: we deliberately put the password conversion right here. Only change if you know what you're doing.
```
DaughterOfMars: This is one of those comments that is both critically important and profoundly unhelpful.

sdk/src/client/utils.rs
```diff
@@ -192,3 +193,40 @@ impl Client {
...
+pub struct Password(String);
+
+impl Password {
+    pub fn as_bytes(&self) -> &[u8] {
+        self.0.as_bytes()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
DaughterOfMars: These tests feel...pointless.

thibault-martinez: I thought the same, testing some derivations...

sdk/src/client/stronghold/mod.rs
DaughterOfMars: I'd rather just remove these comments. They aren't helpful.

kwek20 (dismissed):

sdk/src/client/utils.rs
Alex6323: tbh ... I agree with you, but they at least don't hurt anybody either.

sdk/src/client/stronghold/mod.rs
Alex6323: agreed

sdk/src/wallet/wallet/operations/stronghold_backup/mod.rs
Alex6323: can you make it a suggestion as well? then I'll commit it as a batch through GH.

sdk/src/client/stronghold/mod.rs
Thoralf-M: Also fine for me to remove everywhere

sdk/src/wallet/wallet/operations/stronghold_backup/mod.rs
Thoralf-M: No, I don't want to have it added :trollface:

sdk/src/client/stronghold/mod.rs
Alex6323: I always hate those kind of comments which is why I would have preferred the slight inconvenience in the special case of zeroable passwords/mnemonics. Just doesn't feel right to me to rely on no one accidentally moving this code and causing secrets to leak.

sdk/src/client/utils.rs
Alex6323: maybe @kwek20 wants to chime in ... are you fine with removing the `zeroize` test again? We kind of rely on them not messing things up anyway. The `from` test I am happy to remove too.

sdk/src/client/stronghold/mod.rs
Alex6323: the comments were supposed to prevent changes like this in the future:
```
let mut key_provider_guard = self.key_provider.lock().await?;
let key_provider = self::common::key_provider_from_password(password.into());
```
but yeah, maybe a bit over-cautious on my side.

Alex6323: removed all those comments

sdk/src/client/utils.rs
Alex6323: removed the tests

Thoralf-M (approved):

thibault-martinez (approved):

DaughterOfMars (approved):

PR #479, opened 2023-05-23
Add and use HexStr

Thoralf-M:
# Description of change

Add and use HexStr

https://github.com/iotaledger/iota-sdk/issues/129

## Type of change

mypy doesn't complain

### DIFF
--- a/bindings/python/iota_sdk/types/burn.py
+++ b/bindings/python/iota_sdk/types/burn.py
@@ -5,6 +5,7 @@ from __future__ import annotations  # Allow reference to Burn in Burn class
 from dataclasses import dataclass
 from typing import List, Optional, Dict, Any
 from iota_sdk.types.native_token import NativeToken
+from iota_sdk.types.common import HexStr
 
 
 @dataclass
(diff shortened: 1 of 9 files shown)


bindings/python/iota_sdk/types/burn.py
```diff
@@ -23,33 +24,34 @@ class Burn:
...
         return self
 
-    def add_foundry(self, foundry: str) -> Burn:
+    def add_foundry(self, foundry: HexStr) -> Burn:
         if self.foundries is None:
             self.foundries = []
         self.foundries.append(foundry)
         return self
 
     def add_native_token(self, native_token: NativeToken) -> Burn:
```
thibault-martinez, line 50: Is it a python thing that params are snake_case and fields camelCase?

thibault-martinez (approved):

bindings/python/iota_sdk/types/burn.py
Thoralf-M: Probably not, I think @qrayven did it like this because in the serialized version it needs to be camelCase

thibault-martinez: Ah, right

Alex6323 (approved):

PR #486, opened 2023-05-24
Add more type hints, remove old methods

Thoralf-M:
# Description of change

Add more type hints
Remove old methods, they're only available from the client now

#129 

## Type of change

mypy and running examples

## Change checklist

### DIFF
--- a/bindings/python/iota_sdk/client/client.py
+++ b/bindings/python/iota_sdk/client/client.py
@@ -7,10 +7,16 @@ from iota_sdk.client._node_core_api import NodeCoreAPI
 from iota_sdk.client._node_indexer_api import NodeIndexerAPI
 from iota_sdk.client._high_level_api import HighLevelAPI
 from iota_sdk.client._utils import ClientUtils
-from iota_sdk.types.common import Node
+from iota_sdk.types.common import HexStr, Node
+from iota_sdk.types.feature import Feature
+from iota_sdk.types.native_token import NativeToken
+from iota_sdk.types.token_scheme import TokenScheme
+from iota_sdk.types.unlock_condition import UnlockCondition
 from json import dumps, loads
 import humps
 from datetime import timedelta
+from typing import Any, Dict, List, Optional
+
 
 class ClientError(Exception):
     """client error"""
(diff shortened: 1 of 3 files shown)


bindings/python/iota_sdk/wallet/wallet.py
```diff
@@ -51,15 +51,15 @@ def get_client(self):
         return Client(client_handle=get_client_from_wallet(self.handle))
 
     @_call_method_routine
-    def _call_method(self, name: str, data=None) -> Dict[str, str]:
+    def _call_method(self, name: str, data=None):
```
thibault-martinez, line 54: What does this mean? It does return something, right?

Thoralf-M: Because of the `@_call_method_routine` it returns `Any` which is the same as not specifying anything
Also this function shouldn't be used by users anyways

bindings/python/iota_sdk/wallet/account.py
```diff
@@ -5,16 +5,12 @@
...
-from iota_sdk.types.feature import Feature
 from iota_sdk.types.output_id import OutputId
-from iota_sdk.types.token_scheme import TokenScheme
-from typing import List, Optional, Union
+from typing import List, Optional
 
 
 class Account:
-    def __init__(self, account_id: Union[str, int], handle):
+    def __init__(self, account_id: str|int, handle):
```
Alex6323:
```suggestion
    def __init__(self, account_id: str | int, handle):
```

bindings/python/iota_sdk/client/client.py
```diff
@@ -493,47 +511,47 @@ def build_and_post_block(self,
...
         return self._call_method('getTipsInterval')
 
-    def get_local_pow(self):
+    def get_local_pow(self) -> bool:
         """Returns if local pow should be used or not.
         """
         return self._call_method('getLocalPow')
 
-    def get_fall_back_to_local_pow(self):
+    def get_fall_back_to_local_pow(self) -> bool:
```
Alex6323: I feel like this should either be named `get_fallback_` as in `fallback` as a noun, or just `fall_back_` used as a verb, but I am not entirely sure.

Thoralf-M: Yeah agree

thibault-martinez (approved):

qrayven (approved):

PR #497, opened 2023-05-29
feat: improve constructions of `*Params` structures for `send*` methods 

### LINKED ISSUE #470: Improve `Send` struct creation
Just thinking out loud: is there no way to make this a bit better UX wise? What if we have a `SendNftParams` constructor that takes a Bech32AddressLike? It would make it fallible but we already have a fallible thing here, I don't think it would matter much but would make the overall thing a bit better?

_Originally posted by @thibault-martinez in https://github.com/iotaledger/iota-sdk/pull/366#discussion_r1198732382_

We can add builder patterns for these as a convenience.

qrayven:
# Description of change

UX imporovement for creation of `SendNftParams`

closes #470

## Type of change

## Change checklist


sdk/src/wallet/account/operations/transaction/high_level/send_nft.rs
```diff
@@ -23,6 +23,19 @@ pub struct SendNftParams {
     pub nft_id: NftId,
```
thibault-martinez: I'd make them private + getters now

qrayven: fixed. They're still public in the crate tough, as they're pretty helpful when destructuring

thibault-martinez:
Is this ready? If yes, it doesn't really close #470 yet

sdk/src/wallet/account/operations/transaction/high_level/minting/mint_nfts.rs
```diff
@@ -53,40 +53,40 @@ impl MintNftParams {
 
     pub fn with_address(
         mut self,
-        address: impl TryInto<Bech32Address, Error = impl Into<crate::wallet::Error>>,
+        address: Option<impl TryInto<Bech32Address, Error = impl Into<crate::wallet::Error>>>,
```
DaughterOfMars: Should be just `impl Bech32AddressLike`

sdk/src/wallet/account/operations/transaction/high_level/minting/mint_nfts.rs
```diff
@@ -53,40 +53,40 @@ impl MintNftParams {
...
     ) -> Result<Self, crate::wallet::error::Error> {
-        self.address = Some(address.try_into().map_err(Into::into)?);
+        self.address = address.map(|v| v.try_into().map_err(Into::into)).transpose()?;
         Ok(self)
     }
 
     pub fn with_sender(
         mut self,
-        sender: impl TryInto<Bech32Address, Error = impl Into<crate::wallet::Error>>,
+        sender: Option<impl TryInto<Bech32Address, Error = impl Into<crate::wallet::Error>>>,
```
DaughterOfMars: These should not accept options imo

thibault-martinez: Why? Not even `impl Into<Option<_>>` ?

DaughterOfMars: It's just not ergonomic and you can't mix `impl Into<Option<_>>` and `impl Bech32Like`.

thibault-martinez: I thought you were all in for `impl Into<Option<_>>`

thibault-martinez: But yeah, `impl Bech32Like` is definitely more important here

qrayven: so we stay with `Option<impl Bech32Like>` or `impl Bech32Like` ?

DaughterOfMars: Trust me, if I could have both I would

sdk/CHANGELOG.md
```diff
@@ -71,6 +71,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Address::try_from_bech32_with_hrp` refactored to `try_from_bech32`;
 - `{MetadataFeature, TagFeature}::new` take an `impl Into<Vec<u8>>` param;
 - Made certain `prepare_` methods public: `prepare_mint_nfts`, `prepare_send_native_tokens`, `prepare_send_nft` and `prepare_create_alias_output`;
+- Add constructors for `SendNftParams`, `SendAmountParams`, `SendNativeTokensParams`, `MintNftParams`
```
Thoralf-M:
```suggestion
- Add constructors for `SendNftParams`, `SendAmountParams`, `SendNativeTokensParams`, `MintNftParams`;
```

sdk/src/types/block/macro.rs
```diff
@@ -49,6 +49,23 @@ macro_rules! impl_id {
             }
         }
 
+        #[cfg(feature="std")]
```
Thoralf-M: This shouldn't be required?

sdk/src/wallet/account/operations/transaction/high_level/minting/mint_nfts.rs
qrayven: you're right fixed

sdk/src/wallet/account/operations/transaction/high_level/minting/mint_nfts.rs
qrayven: ok, as has been agreed, I've removed the `Option` from the `with_*` methods

sdk/src/types/block/macro.rs
qrayven: do you refer to the feature flag? The `String` belongs to the std and requires mem allocation. Without it, the no_std CI check fails.

DaughterOfMars (commented):
This is fine, but I think we could make it better. The non-address fns can take `impl Into<Option<_>>`. We can then split the address setters into `try_with_x(impl Bech32Like) -> Result` and `with_x(impl Into<Option<Bech32Address>>)`.

This way we can use whichever is most convenient, and we can avoid the weirdness of the results when we already have a valid `Bech32Address` type.

qrayven:
> This is fine, but I think we could make it better. The non-address fns can take `impl Into<Option<_>>`. We can then split the address setters into `try_with_x(impl Bech32Like) -> Result` and `with_x(impl Into<Option<Bech32Address>>)`.
> 
> This way we can use whichever is most convenient, and we can avoid the weirdness of the results when we already have a valid `Bech32Address` type.

Ok, I've introduced the `try_with*` for optional addresses

sdk/src/types/block/macro.rs
DaughterOfMars: It requires `alloc` not `std`.

sdk/src/wallet/error.rs
```diff
@@ -135,6 +136,13 @@ impl Serialize for Error {
     }
 }
 
+// Implements conversion from Infallible. This never happens, but is required by the compiler.
```
DaughterOfMars, line 139: What needs this? Feels like a signal that some other design is wrong.

sdk/src/types/block/macro.rs
qrayven: yeah, alloc is reexported by std, so I've made the direct reference on `alloc` only.

sdk/src/wallet/error.rs
qrayven: https://github.com/qrayven/iota-sdk/blob/d59e291c78f6ec3b51240656dcd662f37ad79fe6/sdk/src/wallet/account/operations/transaction/high_level/send_nft.rs#L32 This part (`Error=Into<crate::wallet::error`> This is necessary when we want to parametrize the error. In the case when infallible `NftID` is provided, the compiler uses the 'Infailable' enum, because it cannot recognize the error type. I could remove the `TryInto`, but we could lose the ability to use `str` as nft id

DaughterOfMars: Why are you using `TryInto` there rather than `Into`?

qrayven: because conversion from `str`  to  `NftID` is failable.  We don't have implementation `From<str>` for `NftId`, only `TryFrom`. (as I said, we could also require providing `NftId` type, so we could remove the `TryInto`)

DaughterOfMars: Hm. I guess it's fine.

sdk/src/wallet/account/operations/transaction/high_level/send_nft.rs
```diff
@@ -1,26 +1,41 @@
...
+    #[getset(get = "pub")]
+    address: Bech32Address,
     /// Nft id
-    pub nft_id: NftId,
+    #[getset(get = "pub")]
+    nft_id: NftId,
+}
+
+impl SendNftParams {
+    /// Creates a  new instance of [`SendNftParams`]
```
Thoralf-M:
```suggestion
    /// Creates a new instance of [`SendNftParams`]
```

sdk/src/wallet/account/operations/transaction/high_level/send_nft.rs
```diff
@@ -1,26 +1,41 @@
...
+    pub fn new(
+        address: impl Bech32AddressLike,
+        nft_id: impl TryInto<NftId, Error = impl Into<crate::wallet::Error>>,
```
Alex6323, line 32: Should we also have a (generic?)`*Like` trait for IDs. Seems a bit inconsistent to me that we use both conversion traits here? Probably a separate issue tho.

qrayven: Yes, I was thinking about it yesterday. Additionally, technically it would be possible to eliminate the implementation for `Infalliable` here https://github.com/qrayven/iota-sdk/blob/aaa7e49559e9e2ee7384a41cfe833c40a2f18515/sdk/src/wallet/error.rs#L140, mentioned by @Alexandcoats. But, yeah, maybe a separate issue for that?

Alex6323: Yeah, that `From<Infallible> for Error` is really weird. Would be nice to get rid of it.

DaughterOfMars: On the other hand, we can't implement a trait for every type we have. At some point the code load outweighs the benefit.

qrayven: ok, so as we decided we can try to use universal trait `ConvertTo<T>` to solve this issue

DaughterOfMars (approved):

kwek20 (approved):

Alex6323 (approved):

PR #517, opened 2023-05-31
Map errors to wallet error variant

### LINKED ISSUE #480: Consistent error is thrown when account cannot cover storage deposit
## Description

When Firefly tries to claim transactions that cannot be claimed due to no outputs available for covering the storage deposit, different type of errors are thrown per output. As an example.

If I have two incoming NFT Outputs that need claiming, but I only have one output with SMR, then I get the following error if I try to claim the transaction:

`{type: 'insufficientFunds', error: 'insufficient funds 0/70500 available'}`

If I have two incoming Basic Outputs with SD requirements in the same scenario, the following error is thrown:

`{type: 'block', error: 'insufficient output amount for storage deposit: 1 (should be at least 42600)'}`

If I want to attach extra storage deposit that I don't have due to pending transaction, the following error is thrown:

`{type: 'client', error: 'insufficient amount: found 125000, required 133000'}`

## Motivation

Consistent Error Handling, if  a transaction cannot be confirmed for the same reason

## Requirements

1. Consistently throw one error: our preferred one is `{type: 'insufficientFunds', error: 'insufficient funds 0/70500 available'}`
2. Check whether there are other places where a similar error should be thrown that we missed

No.

Thoralf-M:
# Description of change

Map errors to wallet error variant 

Fixes https://github.com/iotaledger/iota-sdk/issues/480

## Type of change

Added test

## Change checklist

### DIFF
--- a/sdk/src/wallet/account/operations/output_claiming.rs
+++ b/sdk/src/wallet/account/operations/output_claiming.rs
@@ -162,7 +162,24 @@ impl Account {
     pub async fn claim_outputs(&self, output_ids_to_claim: Vec<OutputId>) -> crate::wallet::Result<Transaction> {
         log::debug!("[OUTPUT_CLAIMING] claim_outputs");
         let basic_outputs = self.get_basic_outputs_for_additional_inputs().await?;
-        self.claim_outputs_internal(output_ids_to_claim, basic_outputs).await
+        self.claim_outputs_internal(output_ids_to_claim, basic_outputs)
+            .await
+            .map_err(|error| {
+                // Map InsufficientStorageDepositAmount error here because it's the result of InsufficientFunds in this
+                // case and then easier to handle
+                match error {
+                    crate::wallet::Error::Block(block_error) => match *block_error {
+                        crate::types::block::Error::InsufficientStorageDepositAmount { amount, required } => {
+                            crate::wallet::Error::InsufficientFunds {
+                                available: amount,
+                                required,
+                            }
+                        }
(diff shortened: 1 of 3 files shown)


sdk/src/wallet/account/operations/output_claiming.rs
```diff
@@ -162,7 +162,24 @@ impl Account {
...
         log::debug!("[OUTPUT_CLAIMING] claim_outputs");
         let basic_outputs = self.get_basic_outputs_for_additional_inputs().await?;
-        self.claim_outputs_internal(output_ids_to_claim, basic_outputs).await
+        self.claim_outputs_internal(output_ids_to_claim, basic_outputs)
+            .await
+            .map_err(|error| {
+                // Map InsufficientStorageDepositAmount error here because it's the result of InsufficientFunds in this
+                // case and then easier to handle
+                match error {
+                    crate::wallet::Error::Block(block_error) => match *block_error {
```
DaughterOfMars, line 171: Maybe it would be better to have a manual `From` impl for the error types? Unless this should only ever happen here of course

Thoralf-M: Initially I did in on the error type, but I think it's better to have it only here. In other places we want to keep the information that it's about the storage deposit since there it's not about the available funds, but in the claiming this is the case

DaughterOfMars (approved):

thibault-martinez (approved):

PR #526, opened 2023-06-01
Feat/convert trait

### LINKED ISSUE #522: Replace `XLike` traits with a single conversion trait
## Description

We can replace all the various `XLike` traits with a single one that can be used instead of `TryInto`, using the crate `Result` rather than an associated error type.

Something like this:

```rust
trait ConvertTo<T> {
  fn convert(self) -> crate::Result<T>;

  fn convert_unchecked(self) -> T;
}
```

DaughterOfMars:
# Description of change

This replaces various `XLike` traits with a single unified `ConvertTo`.

- Closes #522

## Type of change


qrayven (dismissed):

sdk/src/types/mod.rs
```diff
@@ -8,3 +8,4 @@ extern crate std;
 
 pub mod api;
 pub mod block;
+pub mod convert;
```
thibault-martinez: This should probably be in `sdk/utils` instead?

sdk/src/types/convert.rs
kwek20, line 1: Should we implement like Rc, Box and Option here? Or leave for when we need it

sdk/src/wallet/error.rs
```diff
@@ -136,13 +135,6 @@ impl Serialize for Error {
     }
 }
 
-// Implements conversion from Infallible. This never happens, but is required by the compiler.
-impl From<Infallible> for Error {
```
kwek20, line 140: 💯

kwek20 (commented):
idk im jsut thinking that maybe the trait should impl Into and TryInto, and we use those methods?

sdk/src/types/convert.rs
DaughterOfMars: Probably just wait until we need it.

thibault-martinez (changes requested):
Changelog 😬

thibault-martinez (approved):

qrayven (approved):

PR #529, opened 2023-06-02
Vec cleanup

### LINKED ISSUE #528: Implement `IntoIter` for send function
## Description

We could implement `IntoIter` for outputs in the send function to make it less verbose

No.

DaughterOfMars:
# Description of change

- Removes a LOT of unnecessary `vec!` calls.
- Allows many more functions to accept arrays.
- Fixes bugs with batched tasks where a task was being spawned for each item rather than each batch.
- Other reorganization for performance when many async calls can be concurrent.

- Closes #528 

## Type of change


sdk/src/client/api/block_builder/input_selection/core/mod.rs
```diff
@@ -165,11 +165,12 @@ impl InputSelection {
 
     /// Creates a new [`InputSelection`].
     pub fn new(
-        available_inputs: Vec<InputSigningData>,
-        outputs: Vec<Output>,
-        addresses: Vec<Address>,
+        available_inputs: impl Into<Vec<InputSigningData>>,
+        outputs: impl Into<Vec<Output>>,
```
qrayven, line 169: Is there a specific reason, why we use `Into<Vec<>>`  for outputs, but `IntoIterator<>` for address? `IntoIterator` seems to be more generic and universal. `Vec` is always heap-based.

thibault-martinez: I don't think it really depends on whether it's outputs or address, but whether if at some point we really need a Vec or not. And in this specific case we're storing the Vec in a struct.

sdk/src/wallet/storage/manager.rs
```diff
@@ -147,18 +148,14 @@ impl StorageManager {
...
-
-        Ok(accounts)
+        futures::stream::iter(&self.account_indexes)
+            .filter_map(|account_index| async {
+                let account_index = *account_index;
+                let key = format!("{ACCOUNT_INDEXATION_KEY}{account_index}");
+                self.get(&key).await.transpose()
+            })
+            .try_collect::<Vec<_>>()
+            .await
```
thibault-martinez, line 158: Very nice 👍🏻

sdk/src/client/api/block_builder/input_selection/core/mod.rs
Alex6323: But you can always collect into a Vec if and when you actually need it. The API would become more flexible.

qrayven: Yeah, and `collect()` seems (not 100% sure) optimized by the compilator when you provide a vector.

DaughterOfMars: I could believe that collecting an owned vec is optimized. If that's the case, we probably should use it rather than `Into<Vec<_>>`, but I wasn't sure. Can you prove that this case is optimized?

qrayven: image  
This is a source code I used:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=258ad1b9427a99d8c434a6797e28b5a6

DaughterOfMars: I'm not sure how convincing this is, because I'm not familiar with benchmarking using criterion. It seems odd to me that `Into<Vec<_>>` could ever be slower than the collect. At best they should be identical, and at worst the collect will be slower because it is not optimized.

qrayven: I wouldn't pay attention to these small differences. This is probably a matter of order of execution. Ok, so we could investigate this another time then.

qrayven (approved):

thibault-martinez (approved):

Issue #505, opened 2023-05-30
Add sign EVM to the SecretManage trait


thibault-martinez:
Also expose to nodejs and python bindings

PR #531, opened 2023-06-02
Add new bindings for `sign_evm`

DaughterOfMars:
# Description of change

Adds the new bindings for `sign_evm`.

- #505 

## Type of change

- Enhancement (a non-breaking change which adds functionality)**


bindings/core/src/response.rs
```diff
@@ -105,6 +105,9 @@ pub enum Response {
     /// - [`SignEd25519`](crate::method::SecretManagerMethod::SignEd25519)
     Ed25519Signature(Ed25519SignatureDto),
     /// Response for:
+    /// - [`SignEvm`](crate::method::SecretManagerMethod::SignEvm)
+    EvmSignature { public_key: String, signature: String },
```
Thoralf-M, line 110: I think we need `#[serde(rename_all = "camelCase")]` here for public_key

thibault-martinez: Is it not enough that it's on the enum?

bindings/nodejs/lib/secretManager/SecretManager.ts
```diff
@@ -117,6 +117,23 @@ export class SecretManager {
         return JSON.parse(response).payload;
     }
 
+    /**
+     * Signs a message with an Evm private key.
+     */
+    async signEvm(
+        message: HexEncodedString,
+        chain: IBip32Chain,
+    ): Promise<IEd25519Signature> {
```
thibault-martinez: Wrong type

bindings/core/src/response.rs
Thoralf-M: No, unfortunately not, this only takes care of the variant name

DaughterOfMars: You're right, thanks

bindings/core/src/method/secret_manager.rs
```diff
@@ -40,6 +40,13 @@ pub enum SecretManagerMethod {
         /// Chain to sign the essence hash with
         chain: Vec<u32>,
     },
+    /// Signs a message with an Evm private key.
+    SignEvm {
+        /// The message to sign, hex encoded String
+        message: String,
+        /// Chain to sign the essence hash with
```
thibault-martinez:
```suggestion
        /// Chain to sign the message with
```

bindings/core/src/response.rs
thibault-martinez: Allowing myself to resolve as Thoralf is on holidays.

thibault-martinez (dismissed):

bindings/nodejs/lib/types/client/constants.ts
```diff
@@ -7,6 +7,8 @@ export const SHIMMER_TESTNET_BECH32_HRP = 'rms';
 export enum CoinType {
     IOTA = 4218,
     Shimmer = 4219,
+    Ether = 60,
 }
 
 export const HD_WALLET_TYPE = 44;
+export const HARDEN_MASK = (1 << 31) >>> 0;
```
thibault-martinez, line 14: What does >>> 0 mean?

DaughterOfMars: It's a right shift that does not shift in a sign bit, so it effectively converts the signed value to an unsigned one.

thibault-martinez: Thanks

bindings/nodejs/examples/how_tos/sign_evm/sign_evm.ts
```diff
@@ -0,0 +1,73 @@
...
+        // Store the mnemonic in the Stronghold snapshot, this needs to be done only the first time.
+        // The mnemonic can't be retrieved from the Stronghold file, so make a backup in a secure place!
+        await secretManager.storeMnemonic(
+            process.env.NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC_1,
+        );
+
+        const bip32Chain = [
+            (HD_WALLET_TYPE | HARDEN_MASK) >>> 0,
+            (CoinType.Ether | HARDEN_MASK) >>> 0,
+            (ACCOUNT_INDEX | HARDEN_MASK) >>> 0,
```
thibault-martinez, line 59: Is that the best way to cast to unsigned though? Looks rather odd to do a simple operation like that

DaughterOfMars: I dunno, couldn't find a better one.

bindings/nodejs/examples/how_tos/sign_evm/sign_evm.ts
```diff
@@ -0,0 +1,73 @@
...
+// yarn run-example ./how_tos/sign_evm/sign_evm.ts
+
+const FOUNDRY_METADATA = {
+    standard: 'IRC30',
+    name: 'NativeToken',
+    description: 'A native token',
+    symbol: 'NT',
+    decimals: 6,
+    logoUrl: 'https://my.website/nativeToken.png',
+};
```
Thoralf-M, line 25: I think this should be some other "random" message here, signing the foundry metadata is interesting with ed25519 because it's needed for the token registry https://github.com/iotaledger/token-registry, but an EVM address can't be used for that

DaughterOfMars: Does it really matter though?

Thoralf-M: It's just not a good example I would say, but we can also keep it until we have something that actually makes sense

Thoralf-M (approved):

bindings/nodejs/lib/types/secretManager/secretManager.ts
```diff
@@ -31,3 +33,14 @@ export type SecretManagerType =
...
+
+export interface EvmSignature {
+    /**
+     * The public key.
+     */
+    publicKey: HexEncodedString;
+    /**
+     * The signature.
+     */
+    signature: HexEncodedString;
```
Tuditi, line 45: Can we get this aligned with ledger, so that it respectively returns the `v`, `r`, `s`  values? What's the reason for including the publicKey?

thibault-martinez: This is aligned with Ed25519 signatures in the protocol, they also contain the public key so you can get back the address associated with it

Tuditi (changes requested):
Is it possible to align this with the method documented here: https://www.npmjs.com/package/@ledgerhq/hw-app-eth#signtransaction

thibault-martinez:
> Is it possible to align this with the method documented here: https://www.npmjs.com/package/@ledgerhq/hw-app-eth#signtransaction

What are v, r, s? They don't explain in your link

Tuditi:
> > Is it possible to align this with the method documented here: https://www.npmjs.com/package/@ledgerhq/hw-app-eth#signtransaction
> 
> What are v, r, s? They don't explain in your link

v is the signature prefix value that indicates two things: the chain ID and the recovery identifier to help the ECDSArecover function check the signature.

r & s are the signature values.

Reference: https://github.com/ethereumbook/ethereumbook/blob/develop/06transactions.asciidoc#transaction-signing-in-practice

thibault-martinez:
> > > Is it possible to align this with the method documented here: https://www.npmjs.com/package/@ledgerhq/hw-app-eth#signtransaction
> > 
> > 
> > What are v, r, s? They don't explain in your link
> 
> v is the signature prefix value that indicates two things: the chain ID and the recovery identifier to help the ECDSArecover function check the signature.
> 
> r & s are the signature values.
> 
> Reference: https://github.com/ethereumbook/ethereumbook/blob/develop/06transactions.asciidoc#transaction-signing-in-practice

Then no, it's not possible to align atm. Why do you need these values anyway? This seems rather low level, if we provide `sign` and `verify`, why do you needs them?

Tuditi:
> > > > Is it possible to align this with the method documented here: https://www.npmjs.com/package/@ledgerhq/hw-app-eth#signtransaction
> > > 
> > > 
> > > What are v, r, s? They don't explain in your link
> > 
> > 
> > v is the signature prefix value that indicates two things: the chain ID and the recovery identifier to help the ECDSArecover function check the signature.
> > r & s are the signature values.
> > Reference: https://github.com/ethereumbook/ethereumbook/blob/develop/06transactions.asciidoc#transaction-signing-in-practice
> 
> Then no, it's not possible to align atm. Why do you need these values anyway? This seems rather low level, if we provide `sign` and `verify`, why do you needs them?

That might be problematic... we need to create the transaction format ourselves, use the sdk to get the signature and submit the signed transaction to the wasp node. The wasp node exposes an RPC endpoint that is compatible with that of Ethereum. The library we are using requires these values to be able to correctly create a transaction.

Alternatively, we can get away with it if the method can be combined with the `web3.eth.sendSignedTransaction()` method. Documented [here](https://web3js.readthedocs.io/en/v1.10.0/web3-eth.html#sendsignedtransaction).

thibault-martinez:
> > > > > Is it possible to align this with the method documented here: https://www.npmjs.com/package/@ledgerhq/hw-app-eth#signtransaction
> > > > 
> > > > 
> > > > What are v, r, s? They don't explain in your link
> > > 
> > > 
> > > v is the signature prefix value that indicates two things: the chain ID and the recovery identifier to help the ECDSArecover function check the signature.
> > > r & s are the signature values.
> > > Reference: https://github.com/ethereumbook/ethereumbook/blob/develop/06transactions.asciidoc#transaction-signing-in-practice
> > 
> > 
> > Then no, it's not possible to align atm. Why do you need these values anyway? This seems rather low level, if we provide `sign` and `verify`, why do you needs them?
> 
> That might be problematic... we need to create the transaction format ourselves, use the sdk to get the signature and submit the signed transaction to the wasp node. The wasp node exposes an RPC endpoint that is compatible with that of Ethereum. The library we are using requires these values to be able to correctly create a transaction.
> 
> Alternatively, we can get away with it if the method can be combined with the `web3.eth.sendSignedTransaction()` method. Documented [here](https://web3js.readthedocs.io/en/v1.10.0/web3-eth.html#sendsignedtransaction).

@semenov-vladyslav Can we get v,r,s from public key + signature?

semenov-vladyslav:
Yes, most likely. The (recoverable) signature is 65 bytes with the first 32 bytes representing scalar r, the next 32 bytes -- scalar s, the last byte (two lowest bits) is value v. You don't need public key here.

thibault-martinez:
> Yes, most likely. The (recoverable) signature is 65 bytes with the first 32 bytes representing scalar r, the next 32 bytes -- scalar s, the last byte (two lowest bits) is value v. You don't need public key here.

@Tuditi Looks like we're already returning v,r,s then!

thibault-martinez (approved):

bindings/nodejs/lib/types/secretManager/secretManager.ts
thibault-martinez: We'll go ahead and merge it as we're already returning v,r,s in some form. Let us know if it works when you do some use cases. These are the new bindings anyway, we've already merge the old bindings in the same shape.

PR #548, opened 2023-06-06
fix(cli): fix wallet initialization

### LINKED ISSUE #361: [Bug]: Disallow account creation before storing a valid mnemonic on first run
### Issue description

If on first run one aborts during the prompt where the user is supposed to enter a mnemonic and then restart the wallet one can get passed this step and immediately create new account.

### Version

1.0.0-rc.2

### Expected behaviour

The user should - again - be requested to enter a mnemonic before proceeding to the initial account creation step.

### Actual behaviour

The user is allowed to create an initial account, but then runs into:
```sh
wallet error: `no mnemonic has been stored into the Stronghold vault`
```

Yes

_No response_

### Errors

```sh
wallet error: `no mnemonic has been stored into the Stronghold vault`
```

### Duplicate declaration

Alex6323:
# Description of change

This fixes a bug with initialization of a new wallet. It ensures that a stronghold snapshot file is only created when a mnemonic has been successfully stored. Prior to that PR you could abort the dialoguer interaction with CTRL-C during mnemonic generation, and this would end up with a Stronghold file that has no mnemonic stored. An alternative would be to check whether the stronghold file contains a mnemnoic on start and re-request the user to provide a mnemonic, but the solution provided with this PR fixes the issue with minimal code changes.

Fixes #361 

## Type of change

Running the cli-wallet binary.

### DIFF
--- a/cli/src/command/wallet.rs
+++ b/cli/src/command/wallet.rs
@@ -140,10 +140,7 @@ pub async fn init_command(
             snapshot_path.display()
         )));
     }
-    // Set up password.
     let password = get_password("Stronghold password", true)?;
-
-    // Set up mnemonic.
     let mnemonic = match parameters.mnemonic_file_path {
         Some(path) => import_mnemonic(&path).await?,
         None => enter_or_generate_mnemonic().await?,
--- a/cli/src/command/wallet.rs
+++ b/cli/src/command/wallet.rs
@@ -128,32 +128,41 @@ pub async fn init_command(
     snapshot_path: &Path,
     parameters: InitParameters,
 ) -> Result<Wallet, Error> {
-    let password = get_password("Stronghold password", !snapshot_path.exists())?;
-    let secret_manager = SecretManager::Stronghold(
-        StrongholdSecretManager::builder()
-            .password(&password)
-            .build(snapshot_path)?,
-    );
-    let wallet = Wallet::builder()
-        .with_secret_manager(secret_manager)
-        .with_client_options(ClientOptions::new().with_node(parameters.node_url.as_str())?)
-        .with_storage_path(storage_path.to_str().expect("invalid unicode"))
-        .with_coin_type(parameters.coin_type)
-        .finish()
-        .await?;
+    if storage_path.exists() {
+        return Err(Error::Miscellaneous(format!(
+            "cannot initialize: {} already exists",
+            storage_path.display()
+        )));
+    }
+    if snapshot_path.exists() {
+        return Err(Error::Miscellaneous(format!(
+            "cannot initialize: {} already exists",
+            snapshot_path.display()
+        )));
+    }
+    // Set up password.
+    let password = get_password("Stronghold password", true)?;
 
+    // Set up mnemonic.
     let mnemonic = match parameters.mnemonic_file_path {
         Some(path) => import_mnemonic(&path).await?,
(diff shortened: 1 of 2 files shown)


cli/src/command/wallet.rs
```diff
@@ -128,32 +128,41 @@ pub async fn init_command(
...
 
-    if let SecretManager::Stronghold(secret_manager) = &mut *wallet.get_secret_manager().write().await {
-        secret_manager.store_mnemonic(mnemonic).await?;
-    } else {
-        panic!("cli-wallet only supports Stronghold-backed secret managers at the moment.");
-    }
+    let secret_manager = StrongholdSecretManager::builder()
+        .password(&password)
+        .build(snapshot_path)?;
+    // TODO: allow providing the mnemonic in the builder?
```
thibault-martinez: We have an issue to remove TODOs so let's not add some more. Open an issue if you want to discuss that.

cli/src/command/wallet.rs
```diff
@@ -128,32 +128,41 @@ pub async fn init_command(
...
+            storage_path.display()
+        )));
+    }
+    if snapshot_path.exists() {
+        return Err(Error::Miscellaneous(format!(
+            "cannot initialize: {} already exists",
+            snapshot_path.display()
+        )));
+    }
+    // Set up password.
```
thibault-martinez: Doesn't look like a very useful comment tbh

Alex6323: removed

cli/src/command/wallet.rs
Alex6323: Done

cli/src/command/wallet.rs
```diff
@@ -128,32 +128,37 @@ pub async fn init_command(
...
+            "cannot initialize: {} already exists",
+            storage_path.display()
+        )));
+    }
+    if snapshot_path.exists() {
+        return Err(Error::Miscellaneous(format!(
+            "cannot initialize: {} already exists",
+            snapshot_path.display()
+        )));
+    }
```
thibault-martinez, line 142: I'm a bit confused how this actually fixes #361, what about `The user should - again - be requested to enter a mnemonic before proceeding to the initial account creation step.` ?

Alex6323: this can no longer happen. If you fully run through `init` the stronghold will contain a mnemonic. What I describe in the issue is an alternative way to go about it. I explain in the PR description why I decided to do it differently (basically not allowing a stronghold file with a faulty state)

Alex6323: I verified that this PR fixes the issue. To reproduce it:
1. delete your wallet-db and stronghold snapshot
2. run the cli wallet witthout any args or with `--init`
3. you will be prompted to set up a password and a mnemonic
4. during the dialoguer "interrogation" hit CTRL-C
5. no wallet-db/snapshot file should be created
6. run the cli-wallet again (you should be requested to do all the steps again)

thibault-martinez: Ok, I trust you

Alex6323: The "thing" that actually fixes this issue is moving the `mnemonic` creation/importing step before the secret manager build!

thibault-martinez (approved):

qrayven (approved):

Thoralf-M (approved):

PR #553, opened 2023-06-07
Make LedgerNanoStatus locked an Option

### LINKED ISSUE #491: [Bug]: Ethereum app locked status returns wrong value when connected to the sdk.
### Issue description

Firefly tries to get the ledger status from the Ethereum Ledger app through `getLedgerNanoStatus` on the AccountManager object. The issue is that even if the app is locked, we get a `locked: false` property returned by the sdk. Could you guys look into this?

### Version

rc.35

### Expected behaviour

Whenever the Ethereum app is locked, we should get a `locked: true` response

### Actual behaviour

Locked Ethereum app returns a `locked: false` state
image

Yes

Steps to reproduce
1. Opt.: Set your ledger timeout to 1 min
2. Open the Ethereum app on Ledger
3. Wait until the ledger locks
4. The following status is returned from the sdk:

### Errors

None

### Duplicate declaration

thibault-martinez:
Fixes https://github.com/iotaledger/iota-sdk/issues/491

### DIFF
--- a/sdk/src/client/secret/types.rs
+++ b/sdk/src/client/secret/types.rs
@@ -103,7 +103,7 @@ pub struct LedgerNanoStatus {
     /// Ledger is available and ready to be used.
     pub(crate) connected: bool,
     /// Ledger is connected and locked.
-    pub(crate) locked: bool,
+    pub(crate) locked: Option<bool>,
     /// Ledger blind signing enabled
     pub(crate) blind_signing_enabled: bool,
     /// Ledger opened app.
@@ -120,7 +120,7 @@ impl LedgerNanoStatus {
         self.connected
     }
     /// Ledger is connected and locked.
-    pub fn locked(&self) -> bool {
+    pub fn locked(&self) -> Option<bool> {
         self.locked
     }
     /// Ledger blind signing enabled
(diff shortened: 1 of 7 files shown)


sdk/src/client/secret/types.rs
```diff
@@ -103,7 +103,7 @@ pub struct LedgerNanoStatus {
     /// Ledger is available and ready to be used.
     pub(crate) connected: bool,
     /// Ledger is connected and locked.
```
Thoralf-M: Probably good to update the comments that it's only Some() for Shimmer and IOTA app

thibault-martinez: Wanted to do this and forgot, thanks

thibault-martinez: Done

sdk/src/client/secret/types.rs
```diff
@@ -120,7 +120,7 @@ impl LedgerNanoStatus {
         self.connected
     }
     /// Ledger is connected and locked.
```
Thoralf-M: Same here

thibault-martinez: Done

Thoralf-M (approved):

qrayven (approved):

Alex6323 (approved):

PR #471, opened 2023-05-22
feat(types)!: use `Mnemonic` type from `crypto.rs`

Alex6323:
# Description of change

Consistent with the new `Password` type in #468 this PR adds a `Mnemonic` type, that makes sure sensitive data gets zeroed when the owner is being dropped.

## Links to any relevant issues

None

## Type of change

- Enhancement (a non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How the change has been tested

Unit tests

PR #564, opened 2023-06-07
Use owned data over refs for dto conversions

### LINKED ISSUE #563: Do DTO conversions really need a ref?

DaughterOfMars:
# Description of change

Do DTO conversions need to use refs? No. No they do not.

- Closes #563

## Type of change


bindings/core/src/lib.rs
```diff
@@ -49,7 +49,7 @@ pub struct WalletOptions {
 }
 
 impl WalletOptions {
-    pub async fn build_manager(&self) -> iota_sdk::wallet::Result<Wallet> {
+    pub async fn build_manager(self) -> iota_sdk::wallet::Result<Wallet> {
```
thibault-martinez: Just seeing this now but this should have been renamed a while ago. Can you rename to either `build_wallet` or simply `build` please? (+ changelog)

sdk/src/types/block/block.rs
```diff
@@ -287,36 +287,40 @@ pub mod dto {
     }
 
     impl Block {
-        fn _try_from_dto(value: &BlockDto) -> Result<BlockBuilder, Error> {
```
thibault-martinez, line 290: Any technical reason to remove the factorisation or you just didn't like it? For this one it's alright but some some it's a lot of repeated code

DaughterOfMars: Same reason you had to `std::mem::take` before. The values are moved into the inner fn and cannot then be used afterward. We also can't do it the other way around, because we would have partially moved the value. I could have parameterized the inner fields but I felt that it wasn't worth it for a tiny bit of duplicated code. However, I am open to suggestions.

thibault-martinez: Nah that's fine, was just wondering, thanks

sdk/src/client/secret/mod.rs
```diff
@@ -154,12 +153,12 @@ impl FromStr for SecretManager {
     type Err = Error;
 
     fn from_str(s: &str) -> crate::client::Result<Self> {
-        Self::try_from(&serde_json::from_str::<SecretManagerDto>(s)?)
+        Self::try_from(serde_json::from_str::<SecretManagerDto>(s)?)
     }
 }
 
 /// DTO for secret manager types with required data.
-#[derive(Debug, Clone, Serialize, Deserialize, ZeroizeOnDrop)]
```
Thoralf-M, line 162: Why did you remove `ZeroizeOnDrop`?

sdk/src/client/secret/mod.rs
```diff
@@ -182,35 +181,37 @@ pub enum SecretManagerDto {
...
-    fn try_from(value: &SecretManagerDto) -> crate::client::Result<Self> {
+    fn try_from(value: SecretManagerDto) -> crate::client::Result<Self> {
         Ok(match value {
             #[cfg(feature = "stronghold")]
             SecretManagerDto::Stronghold(stronghold_dto) => {
                 let mut builder = StrongholdSecretManager::builder();
 
-                if let Some(password) = &stronghold_dto.password {
+                if let Some(password) = stronghold_dto.password {
                     // `SecretManagerDto` is `ZeroizeOnDrop` so it will take care of zeroizing the original.
```
Thoralf-M: This comment can be removed then

sdk/src/types/block/payload/milestone/option/parameters.rs
```diff
@@ -99,10 +99,10 @@ pub mod dto {
         }
     }
 
-    impl TryFrom<&ParametersMilestoneOptionDto> for ParametersMilestoneOption {
+    impl TryFrom<ParametersMilestoneOptionDto> for ParametersMilestoneOption {
         type Error = Error;
 
-        fn try_from(value: &ParametersMilestoneOptionDto) -> Result<Self, Self::Error> {
+        fn try_from(value: ParametersMilestoneOptionDto) -> Result<Self, Self::Error> {
```
Thoralf-M, line 105: Do we want to change it when it's not needed? Same for other changes

sdk/src/client/secret/mod.rs
DaughterOfMars: Because it means that you cannot move out of the type, and it doesn't actually make sense for this type to be zeroize, only the password.

sdk/src/types/block/payload/milestone/option/parameters.rs
DaughterOfMars: What do you mean?

sdk/src/client/secret/mod.rs
thibault-martinez: What about the mnemonic and seed ?

DaughterOfMars: Good point.

DaughterOfMars: I had to impl a lot of changes that will probably be overwritten by #471 to fix it, just FYI

sdk/src/types/block/payload/milestone/option/parameters.rs
Thoralf-M: What's the benefit of not using a ref here, when we don't clone anything anyways?

DaughterOfMars: Uniformity?

thibault-martinez (approved):

sdk/src/types/block/payload/milestone/option/parameters.rs
Thoralf-M: In other places we change something to prevent a clone, but whatever

Thoralf-M (approved):

PR #590, opened 2023-06-13
with_alias takes an impl Into<String>

thibault-martinez:
I feel like this `impl Into` is more important than the Option one. Let me know if you prefer the Option.


Thoralf-M (approved):

kwek20 (approved):

DaughterOfMars (approved):
I was going to do this at some point.

PR #606, opened 2023-06-14
Cleanup after #471

DaughterOfMars:
# Description of change

Just some cleanup, separated from the big PR for ease of review.

## Type of change


sdk/src/client/secret/ledger_nano.rs
```diff
@@ -400,24 +393,19 @@ impl SecretManagerConfig for LedgerSecretManager {
...
-                // not fine, return
-                return true;
-            }
-        }
+    let TransactionEssence::Regular(essence) = &prepared_transaction.essence;
+
+    if essence
+        .outputs()
+        .iter()
+        .any(|output| !matches!(output, Output::Basic(o) if o.simple_deposit_address().is_some()))
```
Alex6323: ah ... you prefer this version ... any specific reason why?

sdk/src/client/secret/mod.rs
```diff
@@ -369,6 +369,16 @@ impl SecretManage for SecretManager {
...
 
+pub trait DowncastSecretManager: SecretManage {
+    fn downcast<'a, T: 'static + SecretManage>(&'a self) -> Option<&'a T>;
+}
+
+impl<S: 'static + SecretManage + Send + Sync> DowncastSecretManager for S {
+    fn downcast<'a, T: 'static + SecretManage>(&'a self) -> Option<&'a T> {
+        (self as &(dyn std::any::Any + Send + Sync)).downcast_ref::<T>()
+    }
+}
```
Alex6323: that's fancy!!

sdk/src/client/secret/ledger_nano.rs
DaughterOfMars: No, I just glanced at your comment before writing it so I happened to do the opposite. They should be functionally identical.

sdk/src/client/secret/mod.rs
DaughterOfMars: Traits 😎

sdk/src/client/secret/ledger_nano.rs
Alex6323: yeah, they are. Had this one, too, but then switched to `.all`.

Alex6323 (dismissed):

thibault-martinez (approved):

Alex6323 (approved):

Issue #389, opened 2023-05-04
QA: test all CLI operations


thibault-martinez:
- Test all CLI commands with all possible parameter combinations
- Look for inconvenient patterns and possible UX improvements
- Read the CLI wiki, make sure it's up to date, well explained and easy entry level

qrayven:
- [x] Would be great to have the ability to read the node URL. We can only set one. Maybe subcommand `node-url`.
- [x] First run (without initialized wallet): Is it possible to list all accounts with their aliases attached to the wallet? It would be nice to have a command like `list-accounts`. #658 
- [x] First run (without initialized wallet): This error should make a suggestion that the wallet needs to be initialized first: `./wallet sync` -> wallet error: missing parameter: client_options.
- [x] Shouldn't `new` command be replaced with `new-account`? `new` suggests that we want to create a new wallet, as the wallet is the main ‘actor’ for the CLI.
- [ ] ~~There is a flag `log-level` set to debug, but there are no debug messages at all. Maybe the wallet should be more talkative when this level is enabled.~~
- [ ] `*.stronghold` files have invalid permissions; they should be 600.
- [x] `*.mnemonic` files have invalid permission; they should be 600.
- [ ] ~~Recommendation for root flag: use `wallet-db-dir-path` and `stronghold-snaphshot-file-path` instead of `--wallet-db-path` and `--stronghold-snapshot-path`~~.
- [x] When we want to migrate from non-existent path, there is a generic I/O error. It would be better to say that file is not found: `wt migrate-stronghold-snapshot-v2-to-v3 sss`.
- [x] `./wallet restore [invalid_path]`: I get the prompt about password first, which is not logical. We should be asked for a password when the file is available.
- [x] When doing `./wallet backup`, we have "Wallet has been backed up to "aaa/bbb"". When doing `./wallet restore`, we have nothing, so the user doesn’t know if the status is a success or failed.
- [x] `./wallet change-password`: it should be information that password has been successfully changed. We get nothing.
- [x] When doing `./wallet init` with `–-stronghold-snapshot-file`, it should be information like "Creating a new vault, please provide a new password" because it's confusing which password I should provide when doing init for another mnemonic.
- [ ] ~~When doing the second time init with a different mnemonic and the same `wallet-db-path`, it passes. Should it? In theory, init creates also the `db-path`, but `db-path` cannot be shared between two mnemonics, right?~~
[...]

PR #611, opened 2023-06-15
fix: change default permissions for mnemonic file

qrayven:
# Description of change

The file containing the mnemonic contains sensitive content. According to layered security, default permission should be minimal `(0600)` 

Part of #389 

## Type of change

manually

## Change checklist

### DIFF
--- a/cli/src/helper.rs
+++ b/cli/src/helper.rs
@@ -197,8 +197,9 @@ pub async fn import_mnemonic(path: &str) -> Result<String, Error> {
 async fn write_mnemonic_to_file(path: &str, mnemonic: &str) -> Result<(), Error> {
     let mut open_options = OpenOptions::new().create(true).append(true).to_owned();
 
-    #[cfg(unix)]
-    open_options.mode(0o600);
+    if cfg!(unix) {
+        open_options.mode(0o600);
+    }
 
     let mut file = open_options.open(path).await?;
     file.write_all(format!("{mnemonic}\n").as_bytes()).await?;
--- a/cli/src/helper.rs
+++ b/cli/src/helper.rs
@@ -195,7 +195,8 @@ pub async fn import_mnemonic(path: &str) -> Result<String, Error> {
 }
 
 async fn write_mnemonic_to_file(path: &str, mnemonic: &str) -> Result<(), Error> {
-    let mut open_options = OpenOptions::new().create(true).append(true).to_owned();
+    let mut open_options = OpenOptions::new();
+    open_options.create(true).append(true);
 
     #[cfg(unix)]
     open_options.mode(0o600);
--- a/cli/src/helper.rs
+++ b/cli/src/helper.rs
@@ -195,7 +195,12 @@ pub async fn import_mnemonic(path: &str) -> Result<String, Error> {
 }
 
 async fn write_mnemonic_to_file(path: &str, mnemonic: &str) -> Result<(), Error> {
-    let mut file = OpenOptions::new().create(true).append(true).open(path).await?;
+    let mut file = OpenOptions::new()
+        .create(true)
+        .append(true)
+        .mode(0o600)
+        .open(path)
+        .await?;
     file.write_all(format!("{mnemonic}\n").as_bytes()).await?;
 
     Ok(())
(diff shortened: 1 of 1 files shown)


thibault-martinez (dismissed):

cli/src/helper.rs
```diff
@@ -195,7 +195,12 @@ pub async fn import_mnemonic(path: &str) -> Result<String, Error> {
 }
 
 async fn write_mnemonic_to_file(path: &str, mnemonic: &str) -> Result<(), Error> {
-    let mut file = OpenOptions::new().create(true).append(true).open(path).await?;
+    let mut file = OpenOptions::new()
+        .create(true)
+        .append(true)
+        .mode(0o600)
```
Alex6323: what happens on Windows?

thibault-martinez: Good point, I thought they would have abstracted this away but it seems Unix and Windows both have their own `OpenOptionsExt`

thibault-martinez: Oh actually this is the tokio version, it doesn't even exist on Windows my bad, just put a cfg

Alex6323: okay, then we need to have specific code depending on the platform we're running on.

qrayven: Good point, I was hoping the `tokio` has some mappings from posix. Apprently windows have a different set of methods.

thibault-martinez: Put it under a cfg like they do in this example https://doc.rust-lang.org/std/os/unix/fs/trait.OpenOptionsExt.html#examples-1

qrayven: fixed

cli/src/helper.rs
```diff
@@ -195,7 +195,13 @@ pub async fn import_mnemonic(path: &str) -> Result<String, Error> {
 }
 
 async fn write_mnemonic_to_file(path: &str, mnemonic: &str) -> Result<(), Error> {
-    let mut file = OpenOptions::new().create(true).append(true).open(path).await?;
+    let mut open_options = OpenOptions::new().create(true).append(true).to_owned();
+
+    if cfg!(unix) {
```
DaughterOfMars: Use a cfg tag to avoid compiling this on non-unix devices.

cli/src/helper.rs
```diff
@@ -195,7 +195,12 @@ pub async fn import_mnemonic(path: &str) -> Result<String, Error> {
 }
 
 async fn write_mnemonic_to_file(path: &str, mnemonic: &str) -> Result<(), Error> {
-    let mut file = OpenOptions::new().create(true).append(true).open(path).await?;
+    let mut open_options = OpenOptions::new().create(true).append(true).to_owned();
```
thibault-martinez: You could separate the create and append to avoid the to_owned

cli/src/helper.rs
qrayven: Fixed, I had to revert, because previous commit was correct

cli/src/helper.rs
qrayven: fixed

cli/src/helper.rs
DaughterOfMars: So what about windows?

cli/src/helper.rs
qrayven: This is a good question. I've been trying to figure out what would be best for Windows. But honestly, I haven't used this system for many years and their permission model is unknown to me,  and don't have the chance to test it. I would suggest creating another story for Windows as it requires additional research. But if it's necessary I could try to look at this closer.

cli/src/helper.rs
qrayven: This is a good question. I've been trying to figure out what would be best for Windows. But honestly, I haven't used this system for many years and their permission model is unknown to me, and don't have the chance to test it. I would suggest creating another story for Windows as it requires additional research. But if it's necessary I could try to look at this closer.

thibault-martinez: Can you open an issue please?

qrayven: https://github.com/iotaledger/iota-sdk/issues/624

DaughterOfMars (approved):

thibault-martinez (approved):

PR #615, opened 2023-06-15
Remove NativeTokensBalanceDto and BalanceDto

thibault-martinez:


qrayven (dismissed):

sdk/src/wallet/account/types/balance.rs
```diff
@@ -156,59 +114,34 @@ pub struct NativeTokensBalance {
...
-    pub(crate) metadata: Option<MetadataFeature>,
     /// Total amount
     #[getset(get_copy = "pub")]
     pub(crate) total: U256,
     /// Balance that can currently be spent
     #[getset(get_copy = "pub")]
     pub(crate) available: U256,
+    /// Token foundry immutable metadata
+    #[getset(get = "pub")]
+    pub(crate) metadata: Option<String>,
```
DaughterOfMars: Wouldn't it be better to keep the concrete type?

thibault-martinez: Added back, with serde support for option_string

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -31,6 +31,14 @@ impl TryFrom<Vec<u8>> for MetadataFeature {
     }
 }
 
+impl FromStr for MetadataFeature {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Self::new(prefix_hex::decode::<Vec<u8>>(s).map_err(Error::Hex)?)
```
DaughterOfMars, line 38: Could change `new` to take `impl Into<Box<[u8]>>` while you're in here.

sdk/src/wallet/account/types/balance.rs
```diff
@@ -122,7 +122,8 @@ pub struct NativeTokensBalance {
     pub(crate) available: U256,
     /// Token foundry immutable metadata
     #[getset(get = "pub")]
-    pub(crate) metadata: Option<String>,
+    #[serde(with = "crate::utils::serde::option_string")]
```
DaughterOfMars, line 125: shouldn't this use the serde impl of `MetadataFeature`?

DaughterOfMars: and `serde(default)`

thibault-martinez: No, that would print a packable BoxedSlicePrefix

sdk/src/types/block/output/feature/metadata.rs
thibault-martinez: And also change the TryFrom then?

DaughterOfMars: I don't know about that

sdk/src/wallet/account/types/balance.rs
DaughterOfMars: Any reason `MetadataFeature` can't use a prefix hex serde impl?

sdk/src/types/block/output/feature/metadata.rs
thibault-martinez: Doing the box breaks some nice usecase like `MetadataFeature::new("Hello, World!")`

sdk/src/wallet/account/types/balance.rs
thibault-martinez: It's just that most of the time we use the DTO instead, so we'd do the custom serde impl just for this use case

sdk/src/types/block/output/feature/metadata.rs
DaughterOfMars: We could just add a `impl From<&str> for MetadataFeature` and call `MetadataFeature::from("Hello, World!")` instead.

sdk/src/wallet/account/types/balance.rs
DaughterOfMars: That DTO can't be deleted?

thibault-martinez: No, it serialises the type field as well

DaughterOfMars: So then shouldn't this convert to the DTO and serialize with that impl?

thibault-martinez: Not sure what you mean but we do NOT want the DTO serialised form here

DaughterOfMars: Oh, I see.

sdk/src/types/block/output/feature/metadata.rs
thibault-martinez: Can't, needs to be fallible because of the size

DaughterOfMars (approved):

Alex6323 (approved):

PR #623, opened 2023-06-16
Fix optional transaction note

### LINKED ISSUE #569: Fix optional transaction note
## Description

Fix optional transaction note, currently it's just always set to None https://github.com/iotaledger/iota-sdk/blob/bf7e2a8c9895d3fb88d83ffdf2ce94b1f778c406/sdk/src/wallet/account/operations/transaction/mod.rs#L179
instead it should use what's provided in the TransactionOptions https://github.com/iotaledger/iota-sdk/blob/bf7e2a8c9895d3fb88d83ffdf2ce94b1f778c406/sdk/src/wallet/account/operations/transaction/options.rs#L31

A test should be added which sends a transaction with a note and then checks if the note is set

thibault-martinez:
Fixes https://github.com/iotaledger/iota-sdk/issues/569

### DIFF
--- a/sdk/tests/wallet/transactions.rs
+++ b/sdk/tests/wallet/transactions.rs
@@ -162,3 +162,37 @@ async fn send_nft() -> Result<()> {
 
     tear_down(storage_path)
 }
+
+#[ignore]
+#[tokio::test]
+async fn send_with_note() -> Result<()> {
+    let storage_path = "test-storage/send_with_note";
+    setup(storage_path)?;
+
+    let wallet = make_wallet(storage_path, None, None).await?;
+
+    let account_0 = &create_accounts_with_funds(&wallet, 1).await?[0];
+    let account_1 = wallet.create_account().finish().await?;
+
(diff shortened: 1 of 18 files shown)


sdk/tests/wallet/transactions.rs
```diff
@@ -162,3 +162,37 @@ async fn send_nft() -> Result<()> {
...
+                amount,
+            )?],
+            Some(TransactionOptions {
+                note: Some(String::from("send_with_note")),
+                ..Default::default()
+            }),
+        )
+        .await?;
+
+    account_0
```
kwek20: retry isnt needed here, makes the test longer, but consistent i guess?

thibault-martinez: I just copy pasted a test, I can remove it

thibault-martinez: Done

kwek20 (approved):

DaughterOfMars (approved):

Issue #622, opened 2023-06-16
Non-trivially removable DTOs


thibault-martinez:
- [ ] MigratedFundsEntryDto
- [x] OutputMetadataDto
- [x] ProtocolParametersDto
- [ ] AccountDetailsDto
- [ ] TransactionDto
- [ ] OutputDataDto
- [ ] PreparedMintTokenTransactionDto
- [ ] MintTokenTransactionDto
- [ ] TransactionOptionsDto
- [ ] BlockDto
- [ ] InputSigningDataDto
- [ ] StrongholdDto
- [ ] BurnDto
- [ ] RemainderDataDto
- [ ] SignedTransactionDataDto
- [ ] PreparedTransactionDataDto
- [x] NetworkInfoDto

DaughterOfMars:
I think everything has been examined and the only one left that I think could be removed is `FeatureDto`. Not sure if it's worth it.

PR #626, opened 2023-06-16
🔪 DTOs 🩸 🩸 

DaughterOfMars:
# Description of change

Make output DTOs extinct

- #622 

## Type of change

## Open Questions

- The changes assume that enums will use internal tags, but the DTOs used adjacent tags. It's possible to keep that, but it will require slightly more complicated serde impls. Migration would still be necessary.


DaughterOfMars:
I may need some help with the bindings tests/examples. Something is wrong with the fixtures I think but I'm not sure what.

Alex6323:
Gotta love this PR title :joy:

sdk/src/types/block/address/mod.rs
```diff
@@ -214,107 +211,65 @@ impl From<&Self> for Address {
...
-                Address::Nft(a) => Self::Nft(a.into()),
-            }
+#[cfg(feature = "serde")]
+impl<'de> Deserialize<'de> for Address {
+    fn deserialize<D: Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+        #[derive(Deserialize)]
+        struct TypedAddress {
+            #[serde(rename = "type")]
+            kind: u8,
+            data: serde_json::Value,
```
Thoralf-M, line 221: This key can be pubKeyHash, aliasId or nftId but not data

sdk/src/types/block/output/feature/mod.rs
```diff
@@ -145,6 +145,74 @@ impl Feature {
...
 }
 
+#[cfg(feature = "serde")]
+impl<'de> Deserialize<'de> for Feature {
+    fn deserialize<D: serde::Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+        #[derive(Deserialize)]
+        struct TypedFeature {
+            #[serde(rename = "type")]
+            kind: u8,
+            data: serde_json::Value,
```
Thoralf-M, line 155: This is also not limited to `data`, can be be example also `tag`

sdk/src/types/block/address/mod.rs
Thoralf-M: Maybe I'm missing something, because most tests still pass

sdk/src/client/node_api/core/routes.rs
```diff
@@ -301,13 +293,13 @@ impl ClientInner {
 
     /// Get the metadata for a given `OutputId` (TransactionId + output_index).
     /// GET /api/core/v2/outputs/{outputId}/metadata
-    pub async fn get_output_metadata(&self, output_id: &OutputId) -> Result<OutputMetadataDto> {
+    pub async fn get_output_metadata(&self, output_id: &OutputId) -> Result<OutputMetadata> {
```
Thoralf-M, line 296: This fails now with `Error: Node(Reqwest(reqwest::Error { kind: Decode, source: Error("missing field `outputId`", line: 1, column: 300) }))` because in the response the outputId is split into transactionId and outputIndex

sdk/src/types/block/address/mod.rs
DaughterOfMars: I changed the naming to be unified. However, it can be reverted with a little effort.

Thoralf-M: I missed the `serde(tag = "type", content = "data")` since it was only at `Address` and not the Dto, should be fine this way

sdk/src/types/block/output/feature/mod.rs
Thoralf-M: Same here, missed the serde(tag = "type", content = "data")

PR #627, opened 2023-06-19
UX minor improvements for wallet CLI

qrayven:
# Description of change

- different message when a new password is set 
- information that password has been changed
- aliases are visible in help
-  parser accepts multiple whitespaces when we are in account mode i.e `account "main": mint-nft-token  addr    100   500`
- when use `sync` or other top level command without initialization the following error happens: `wallet error: missing parameter client_options`, now its:
 ```
wallet sync
Provide a new Stronghold password: [hidden]
Please make sure the wallet is initialized.
wallet error: missing parameter: client_options
```

part of #389 

## Type of change

### CODE REVIEW

File: cli/src/helper.rs [hunk 1]
```diff
@@ -27,6 +27,7 @@ pub fn get_password(prompt: &str, confirmation: bool) -> Result<Password, Error>
     password.with_prompt(prompt);
 
     if confirmation {
+        password.with_prompt("Provide a new Stronghold password");
         password.with_confirmation("Confirm password", "Password mismatch");
```
thibault-martinez, line 33: Isn't that going to ask twice?

qrayven: no, it's like this: 
image

Thoralf-M: `with_confirmation()` should ask for the password twice, otherwise the name makes no sense

qrayven: yes,  `with_confirmation` asks for confirmation. I was thinking that `with_prompt`  causes another one prompt.

thibault-martinez: And we are sure that confirmation ALWAYS mean new password? There are no other case where we ask for confirmation than initialisation? Otherwise the message would be weird

qrayven: I think so, I couldn't find any other case. The confirmation is done when we either change the password or create a new password to protected db.

Thoralf-M (approved):

thibault-martinez (approved):

PR #651, opened 2023-06-21
Add commitment routes

thibault-martinez:


sdk/src/client/node_api/core/routes.rs
```diff
@@ -349,6 +348,56 @@ impl ClientInner {
...
+    pub async fn get_slot_commitment_by_id_index(&self, slot_index: &SlotIndex) -> Result<Vec<u8>> {
+        let path = &format!("api/core/v3/commitments/by-index/{slot_index}");
+
+        self.node_manager
+            .read()
+            .await
+            .get_request_bytes(path, None, self.get_timeout().await)
+            .await
+    }
+
```
Thoralf-M, line 400: What about `/api/core/v3/commitments/by-index/{index}/utxo-changes` and`/api/core/v3/commitments/{commitmentId}/utxo-changes`?

thibault-martinez: Will do another PR

Thoralf-M (dismissed):

sdk/src/client/node_api/core/routes.rs
```diff
@@ -349,6 +348,56 @@ impl ClientInner {
...
+        self.node_manager
+            .read()
+            .await
+            .get_request::<SlotCommitment>(path, None, self.get_timeout().await, false, true)
+            .await
+    }
+
+    /// Gets the slot commitment, as raw bytes, by the given slot index.
+    /// /api/core/v3/commitments/by-index/{index}
+    pub async fn get_slot_commitment_by_id_index(&self, slot_index: &SlotIndex) -> Result<Vec<u8>> {
```
qrayven: shouldn't be `by_index_raw`?

thibault-martinez: Good catch

sdk/src/client/node_api/core/routes.rs
```diff
@@ -349,6 +348,56 @@ impl ClientInner {
...
     }
 
+    // Commitments routes.
+
+    /// Gets the slot commitment by the given slot commitment id.
+    /// GET /api/core/v3/commitments/{commitmentId}
+    pub async fn get_slot_commitment_by_id(&self, slot_commitment_id: &SlotCommitmentId) -> Result<SlotCommitment> {
+        let path = &format!("api/core/v3/commitments/{slot_commitment_id}");
+
+        self.node_manager
```
qrayven, line 358: can we just reuse the `get_slot_commitment_by_id_raw`?

thibault-martinez: We could but I'd rather not, not sure we have the guarantee the raw version will exist on all nodes, etc...

qrayven: aa ok makes sense

sdk/src/client/node_api/core/routes.rs
thibault-martinez: Fixed

sdk/src/client/node_api/core/routes.rs
```diff
@@ -349,6 +348,56 @@ impl ClientInner {
...
+            .read()
+            .await
+            .get_request::<SlotCommitment>(path, None, self.get_timeout().await, false, true)
+            .await
+    }
+
+    /// Gets the slot commitment, as raw bytes, by the given slot index.
+    /// /api/core/v3/commitments/by-index/{index}
```
Alex6323: totally nit-picky...so ignore if you want ... but here you missed the `GET`, also you use `index` in the doc, but `slot_index` would be better I think as you do in the actual route.

thibault-martinez: I'm adding the GET but I'd rather keep the rest identical to the API spec

sdk/src/types/block/slot/index.rs
```diff
@@ -10,6 +10,12 @@ use derive_more::{Deref, From};
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct SlotIndex(u64);
 
+impl core::fmt::Display for SlotIndex {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
```
Alex6323: you can just `derive_more` this I think

sdk/src/client/node_api/core/routes.rs
thibault-martinez: Done

sdk/src/types/block/slot/index.rs
thibault-martinez: Done

qrayven (approved):

Thoralf-M (approved):

Alex6323 (approved):

kwek20 (approved):

PR #668, opened 2023-06-22
Some routes changes

thibault-martinez:


sdk/tests/client/node_api.rs
```diff
@@ -245,16 +245,7 @@ async fn test_get_milestone_by_id() {
...
-            &node_info
-                .node_info
-                .status
-                .latest_milestone
-                .milestone_id
-                .unwrap()
-                .parse()
-                .unwrap(),
-        )
+        .get_milestone_by_id(&node_info.node_info.status.latest_milestone.milestone_id.unwrap())
```
Alex6323, line 248: Maybe use this opportunity to document the unwrap? why is this an `Option` again? How can a `latest_milestone` instance not have an ID?

sdk/tests/client/node_api.rs
```diff
@@ -284,16 +275,7 @@ async fn test_get_utxo_changes_by_id() {
...
-            &node_info
-                .node_info
-                .status
-                .latest_milestone
-                .milestone_id
-                .unwrap()
-                .parse()
-                .unwrap(),
-        )
+        .get_utxo_changes_by_id(&node_info.node_info.status.latest_milestone.milestone_id.unwrap())
```
Alex6323, line 278: here also

sdk/tests/client/node_api.rs
thibault-martinez: Honestly, these are tests, there are thousands of undocumented unwraps in tests, it's not worth it...

Thoralf-M (approved):

Alex6323 (approved):

PR #671, opened 2023-06-22
StrongParents, WeakParents and ShallowLikeParents

thibault-martinez:


sdk/src/types/block/parent.rs
```diff
@@ -70,3 +73,7 @@ fn verify_parents<const VERIFY: bool>(parents: &[BlockId], _: &()) -> Result<(),
         Ok(())
     }
 }
+
+pub type StrongParents = Parents<1, 8>;
+pub type WeakParents = Parents<0, 8>;
+pub type ShallowLikeParents = Parents<0, 8>;
```
Alex6323, line 79: I feel like those are likely to become wrapper types as their names seem to carry very different semantics, and this could become interesting for (at least) some Tangle metrics.

sdk/src/types/block/rand/block.rs
```diff
@@ -22,8 +24,8 @@ pub fn rand_block_ids(len: usize) -> Vec<BlockId> {
 }
 
 /// Generates a random block with given parents.
-pub fn rand_block_with_parents(parents: Parents) -> Block {
-    BlockBuilder::new(parents)
```
Alex6323, line 26: probably the builder should have `.with_strong_parents`, `.with_weak_parents`, ... methods?

thibault-martinez: As said on Slack, this is not the purpose of this PR. I will add them when I actually update the Block model.

Alex6323: maybe also `Parents` should be come an enum now:
```rust
enum Parents {
    Strong(_),
    Weak(_),
    ShallowLike(_)
}
```

thibault-martinez: I don't think so, these are 3 distinct fields so there is no need for an enum, would make everything much more complex for no real reason

Alex6323: okay

sdk/src/types/block/parent.rs
thibault-martinez: Type alias here is very handy, I'll have a look if wrappers make sense 👍🏻

thibault-martinez: I've come to the conclusion that it's not very desirable to have wrappers

sdk/src/types/block/error.rs
```diff
@@ -182,8 +185,8 @@ impl fmt::Display for Error {
             Self::InvalidOutputAmount(amount) => write!(f, "invalid output amount: {amount}"),
             Self::InvalidOutputCount(count) => write!(f, "invalid output count: {count}"),
             Self::InvalidOutputKind(k) => write!(f, "invalid output kind: {k}"),
-            Self::InvalidParentCount(count) => {
-                write!(f, "invalid parents count: {count}")
+            Self::InvalidParentCount => {
```
Thoralf-M, line 188: Why remove the count?

thibault-martinez: https://github.com/iotaledger/iota-sdk/pull/671/files/50728389683378681d33ff3777199b4de01fce0c#diff-334f5ed764795a38029cdd22cd2ec408970f26ec1379048c18b4e68fe186634dR62

Thoralf-M (approved):

qrayven (approved):

Alex6323 (approved):

kwek20 (approved):
arguably you should remove weak and shallow as theyre not used in this pr. but sure!

PR #679, opened 2023-06-26
Feat: Added client destroy

### LINKED ISSUE #520: [Bug]: Open handle on NodeJS Client
### Issue description

running a wallet, and calling wallet.destroy() leaves an open handle when built with walletOptions.

I believe the handle is references in bindings/nodejs/src/client.rs:65 (create_client)

Potential fix is a destroy client or clean that handle somehow (?)

### Version

0.1.0 nodejs binding

### Expected behaviour

NO open handle

### Actual behaviour

Open handle

Yes

`yarn test` inside bindings/nodejs

### Errors

```
Jest has detected the following 1 open handle potentially keeping Jest from exiting:

  ●  neon threadsafe function

      12 |         // The rust client object is not extensible
      13 |         if (Object.isExtensible(options)) {
    > 14 |             this.methodHandler = createClient(JSON.stringify(options));
         |                                              ^
      15 |         } else {
      16 |             this.methodHandler = options as ClientMethodHandler;
      17 |         }

      at new ClientMethodHandler (lib/client/ClientMethodHandler.ts:14:46)
      at new Client (lib/client/Client.ts:63:30)
      at Object.<anonymous> (tests/client/utilityMethods.spec.ts:11:23)

error Command failed with exit code 1.
```

### Duplicate declaration

kwek20:
# Description of change

client had an open handle, now you can close it!

fixes #520 

## Type of change

added it to test fixtures, no longer does jest complain! :carlos:

## Change checklist


bindings/nodejs/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `Client::destroy` to close an open handle;
```
thibault-martinez: Remove

Thoralf-M:
In the tests it still prints `Force exiting Jest: Have you considered using `--detectOpenHandles` to detect async operations that kept running after all tests finished?`
Do we need the same for the SecretManager or does the destroy just need to be called in some other tests as well?

kwek20:
Question: Setting to undefined is useful at all? I have no idea

bindings/nodejs/lib/client/client-method-handler.ts
```diff
@@ -22,14 +22,15 @@ export class ClientMethodHandler {
...
+    async destroy() {
+        await destroyClient(this.methodHandler);
+        this.methodHandler = undefined;
     }
 
     async callMethod(method: __ClientMethods__): Promise<string> {
         return callClientMethodAsync(
             JSON.stringify(method),
-            this.methodHandler,
+            this.methodHandler!,
```
Thoralf-M: Forbidden non-null assertion

kwek20: i know, see my question above ^
I can change them all to ? i suppose, but not sure

Thoralf-M: Ah wasn't sure about what you mean with the question, but I think it's not useful, at least haven't seen anywhere that this should help

bindings/nodejs/lib/wallet/wallet.ts
```diff
@@ -79,7 +79,8 @@ export class Wallet {
      * Destroy the Wallet and drop its database connection.
      */
     async destroy(): Promise<void> {
-        await this.methodHandler.destroy();
+        await (await this.getClient()).destroy();
```
Thoralf-M:
```suggestion
```
Makes no sense to create a ClientMethodHandlerWrapper just to destroy the Client inside again

kwek20: i wasnt sure about this, but youre right

bindings/nodejs/lib/wallet/wallet.ts
```diff
@@ -79,7 +79,7 @@ export class Wallet {
      * Destroy the Wallet and drop its database connection.
      */
     async destroy(): Promise<void> {
-        await this.methodHandler.destroy();
+        return this.methodHandler.destroy();
```
thibault-martinez, line 82: For my own understanding, can you explain?

thibault-martinez: I mean, I'm assuming this is very similar to rust and this would have returned void instead of the future?

bindings/nodejs/lib/client/client.ts
```diff
@@ -63,6 +63,10 @@ export class Client {
         this.methodHandler = new ClientMethodHandler(options);
     }
 
+    async destroy() {
+        return this.methodHandler?.destroy();
```
thibault-martinez: What's that `?` and why doesn't the wallet destroy method have one?

bindings/nodejs/package.json
```diff
@@ -14,7 +14,7 @@
         "prebuild-arm64": "prebuild --runtime napi --target 6 --prepack scripts/neon-build.js --strip --arch arm64",
         "rebuild": "node scripts/neon-build && tsc && node scripts/strip.js",
         "install": "prebuild-install --runtime napi --tag-prefix='iota-sdk-nodejs-v' && tsc || npm run rebuild",
-        "test": "jest --forceExit"
```
Thoralf-M, line 17: Just checking, this is not needed anymore because it will use the config by default?

bindings/nodejs/lib/wallet/wallet.ts
kwek20: yea its confused me as well a bit, and is essentially the same. This explanation helped me:
https://jakearchibald.com/2017/await-vs-return-vs-return-await/

bindings/nodejs/package.json
kwek20: yes

bindings/nodejs/lib/client/client.ts
kwek20: ? means execute the statement after this if the object isnt null/undefined. it wasnt needed

bindings/nodejs/src/client.rs
```diff
@@ -6,21 +6,26 @@ use std::sync::Arc;
...
 use neon::prelude::*;
+use tokio::sync::RwLock;
 
 type JsCallback = Root<JsFunction<JsObject>>;
 
+// Wrapper so we can destroy the ClientMethodHandler
+pub type ClientMethodHandlerWrapperInner = Arc<RwLock<Option<ClientMethodHandler>>>;
+// Wrapper because we can't impl Finalize on ClientMethodHandlerWrapperInner
+pub struct ClientMethodHandlerWrapper(pub ClientMethodHandlerWrapperInner);
 pub struct ClientMethodHandler {
```
thibault-martinez, line 20: This kind of pattern can usually be done with 2 types, why do we need 3 🤔 ?

kwek20: Inner is a type not a struct. i didnt try rewriting it though, this is how it is in wallet

Thoralf-M (dismissed):

thibault-martinez (changes requested):
Conflicts

bindings/nodejs/src/client.rs
thibault-martinez: I guess I was confused because usually it would be the other way around. The type actually being used (here `ClientMethodHandlerWrapper`) should not have any suffix. So I'd say `ClientMethodHandler` should be `ClientMethodHandlerInner`, `ClientMethodHandlerWrapper` should be `ClientMethodHandler` and `ClientMethodHandlerWrapperInner` something else or not even exist.

bindings/nodejs/CHANGELOG.md
```diff
@@ -21,11 +21,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## 1.0.0-rc.2 - 2023-0x-xx
 
+<<<<<<< HEAD
+### Added
+
+- `Client::destroy` to close an open handle;
+=======
```
thibault-martinez: :carlos:

bindings/nodejs/src/client.rs
kwek20: Sure we can open a ticket for it

kwek20: https://github.com/iotaledger/iota-sdk/issues/699

DaughterOfMars: Definitely feels like the type def isn't needed here.

Thoralf-M (approved):

thibault-martinez (approved):

PR #694, opened 2023-06-29
Add support for sign_ed25519 to ledger nano

thibault-martinez:


sdk/src/client/secret/ledger_nano.rs
```diff
@@ -172,8 +172,55 @@ impl SecretManage for LedgerSecretManager {
         Err(Error::UnsupportedOperation.into())
     }
 
-    async fn sign_ed25519(&self, _msg: &[u8], _chain: &Chain) -> Result<Ed25519Signature, Self::Error> {
-        Err(Error::UnsupportedOperation.into())
+    async fn sign_ed25519(&self, msg: &[u8], chain: &Chain) -> Result<Ed25519Signature, Self::Error> {
+        if msg.len() != 32 {
+            return Err(Error::UnsupportedOperation.into());
```
Thoralf-M, line 178: Other error would be good or at least have a comment for the function that it's limited to 32 bytes

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -254,14 +301,14 @@ impl SecretManage for LedgerSecretManager {
...
-                        let remainder_bip32_indices: Vec<u32> = match &a.chain {
+                        let remainder_bip32_indices = match &a.chain {
                             Some(chain) => {
                                 chain
                                     .segments()
                                     .iter()
                                     // XXX: "ser32(i)". RTFSC: [crypto::keys::slip10::Segment::from_u32()]
                                     .map(|seg| u32::from_be_bytes(seg.bs()))
-                                    .collect()
+                                    .collect::<Vec<_>>()
```
Alex6323, line 312: looks like this could be a utility method on a chain? The same logic is repeated 3 times.

sdk/src/client/secret/ledger_nano.rs
thibault-martinez: Tbh I think this ones fits, signing anything else than 32 bytes is an unsupported operation. But yeah I'll add a comment.

sdk/src/client/secret/ledger_nano.rs
thibault-martinez: Yeah, I'm not doing it on purpose because `Chain` already disappeared in the latest crypto.rs and we need to update.

sdk/src/client/secret/ledger_nano.rs
thibault-martinez: Added

Alex6323 (approved):

Thoralf-M (approved):

PR #703, opened 2023-06-30
Add MinimumStorageDeposit and remove some TODOs

thibault-martinez:


sdk/src/types/block/output/rent.rs
```diff
@@ -140,3 +148,44 @@ impl<T: Rent, const N: usize> Rent for [T; N] {
...
+
+    pub fn with_storage_deposit_return(mut self, return_address: Address) -> Result<Self, Error> {
+        self.builder = self
+            .builder
+            .add_unlock_condition(StorageDepositReturnUnlockCondition::new(
+                return_address,
+                Output::AMOUNT_MIN,
+                self.token_supply,
+            )?)
+            .add_unlock_condition(ExpirationUnlockCondition::new(return_address, 1)?);
```
thibault-martinez: Should we add it automatically or have another with_ method ?

Thoralf-M: Should be separated, it's not always wanted

sdk/src/types/block/output/rent.rs
```diff
@@ -140,3 +148,44 @@ impl<T: Rent, const N: usize> Rent for [T; N] {
         self.iter().map(|elem| elem.weighted_bytes(config)).sum()
     }
 }
+
+pub struct MinimumStorageDeposit {
+    config: RentStructure,
+    token_supply: u64,
+    builder: BasicOutputBuilder,
```
Thoralf-M: Maybe call the whole struct `MinimumStorageDepositBasicOutput`? For other output types it's different
Or it should be possible to provide the type of output? (probably not worth it right now if we don't use it anywhere)

thibault-martinez: Are we ever interested in non basic though? We only use this for returns

Thoralf-M: From the name it sounds like a generic thing and not only for this specific case
If we don't want it generic, then I would make it pub(crate) and maybe add a comment above that it's used for storage deposit returns

sdk/src/types/block/output/rent.rs
thibault-martinez: Done

sdk/src/types/block/output/rent.rs
```diff
@@ -140,3 +148,50 @@ impl<T: Rent, const N: usize> Rent for [T; N] {
...
+            .builder
+            .add_unlock_condition(StorageDepositReturnUnlockCondition::new(
+                return_address,
+                Output::AMOUNT_MIN,
+                self.token_supply,
+            )?);
+        Ok(self)
+    }
+
+    pub fn with_expiration(mut self, return_address: Address) -> Result<Self, Error> {
```
Thoralf-M: Why accept an address here and not also do `Address::from(Ed25519Address::from([0; Ed25519Address::LENGTH]))`?

thibault-martinez: Yeah you are right, blind copy paste, will change

thibault-martinez: Done

sdk/src/types/block/output/rent.rs
thibault-martinez: I'll just rename it

thibault-martinez: Done

sdk/src/types/block/output/rent.rs
```diff
@@ -140,3 +148,51 @@ impl<T: Rent, const N: usize> Rent for [T; N] {
...
+    token_supply: u64,
+    builder: BasicOutputBuilder,
+}
+
+impl MinimumStorageDepositBasicOutput {
+    pub fn new(config: RentStructure, token_supply: u64) -> Self {
+        Self {
+            config,
+            token_supply,
+            builder: BasicOutputBuilder::new_with_amount(Output::AMOUNT_MIN).add_unlock_condition(
```
DaughterOfMars, line 163: Isn't there a `new_with_min_storage_deposit` variant?

thibault-martinez: There is, but in this case the result is the same and `new_with_min_storage_deposit` induces a clone of the output

DaughterOfMars: boo

bindings/python/examples/wallet/destroy_alias.py
Thoralf-M, line 1: Actually can you delete this file?
We have how_tos/alias/destroy.py already

thibault-martinez: Yes sir

thibault-martinez: Done

Thoralf-M (approved):

DaughterOfMars (approved):

Alex6323 (approved):

PR #710, opened 2023-07-03
Add StakingFeature

thibault-martinez:


sdk/src/types/block/output/feature/mod.rs
```diff
@@ -376,6 +419,7 @@ pub mod dto {
                 T2(&'a IssuerFeatureDto),
                 T3(&'a MetadataFeatureDto),
                 T4(&'a TagFeatureDto),
+                T6(&'a StakingFeatureDto),
```
Thoralf-M, line 402: Why not T5?

thibault-martinez: It'll be BlockIssuerFeature, I have another PR for it but it's blocked by crypto stuff for now

sdk/src/types/block/output/feature/staking.rs
```diff
@@ -0,0 +1,66 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+/// Stakes IOTA coins to become eligible for committee selection, validate the network and receive Mana rewards.
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, packable::Packable)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct StakingFeature {
+    /// The amount of IOTA coins that are locked and staked in the containing account.
```
Thoralf-M:
```suggestion
/// Stakes coins to become eligible for committee selection, validate the network and receive Mana rewards.
#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, packable::Packable)]
#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
pub struct StakingFeature {
    /// The amount of coins that are locked and staked in the containing account.
```
Not sure if we should do that if it's also used in Shimmer

thibault-martinez: Yeah I copy pasted from the TIP

sdk/src/types/block/output/feature/staking.rs
```diff
@@ -0,0 +1,66 @@
...
+        self.end_epoch
+    }
+}
+
+#[allow(missing_docs)]
+pub mod dto {
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+    pub struct StakingFeatureDto {
```
Thoralf-M, line 58: Maybe don't add the Dto and just add the `#[serde(rename = "type")]` to the StakingFeature?

thibault-martinez: StakingFeature has no `kind` field

Thoralf-M: We could do some other serde stuff, but probably not worth it

thibault-martinez: Yeah I didn't want to touch this depending how Alex's PR evolves

Thoralf-M: Btw why doesn't it has a `kind` field and only the const instead?

thibault-martinez: because there is no point storing it, it's always the same value, just takes space for nothing

Thoralf-M (dismissed):

sdk/src/types/block/output/feature/mod.rs
```diff
@@ -143,6 +153,21 @@ impl Feature {
...
+
+    /// Gets the feature as an actual [`StakingFeature`].
+    /// PANIC: do not call on a non-staking feature.
+    pub fn as_staking(&self) -> &StakingFeature {
+        if let Self::Staking(feature) = self {
+            feature
+        } else {
+            panic!("as_staking called on a non-staking feature");
+        }
+    }
```
Alex6323: out of interest ... why do we panic for those `as_` methods instead of returning a `Result` or even just an `Option`?

sdk/src/types/block/output/feature/mod.rs
```diff
@@ -264,6 +290,15 @@ impl Features {
...
         }
     }
+
+    /// Gets a reference to a [`StakingFeature`], if any.
+    pub fn staking(&self) -> Option<&StakingFeature> {
+        if let Some(Feature::Staking(staking)) = self.get(StakingFeature::KIND) {
+            Some(staking)
+        } else {
+            None
+        }
```
Alex6323: map?

kwek20 (dismissed):

sdk/src/types/block/output/feature/mod.rs
thibault-martinez: Convenience

sdk/src/types/block/output/feature/staking.rs
```diff
@@ -0,0 +1,66 @@
...
+/// Stakes coins to become eligible for committee selection, validate the network and receive Mana rewards.
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, packable::Packable)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct StakingFeature {
+    /// The amount of coins that are locked and staked in the containing account.
+    staked_amount: u64,
+    /// The fixed cost of the validator, which it receives as part of its Mana rewards.
+    fixed_cost: u64,
+    /// The epoch index in which the staking started.
+    start_epoch: u64,
```
Alex6323, line 13: can this be in the future like the `end_epoch`?

sdk/src/types/block/output/feature/mod.rs
Alex6323: ah, okay. Is that something we should reconsider doing because for the user of the library it may be more convenient if it's possible to recover.

thibault-martinez: I wouldn't, there is a comment saying it panics and there is an `is_` method to check before if they really want. There are plenty of situations where we know for sure the type and it's just unnecessary to have to unwrap something.

sdk/src/types/block/output/feature/staking.rs
thibault-martinez: Not yet clear to me atm

sdk/src/types/block/output/feature/mod.rs
thibault-martinez: What do you want to map?

Alex6323: `self.get(StakingFeature::KIND).map(|f| f.as_staking())`

thibault-martinez: Yeah I guess that could work 👍🏻

thibault-martinez: I'll do it on develop, rebase and do it here as well

Alex6323: awesome :+1:

sdk/src/types/block/output/feature/mod.rs
```diff
@@ -143,6 +153,21 @@ impl Feature {
...
+        matches!(self, Self::Staking(_))
+    }
+
+    /// Gets the feature as an actual [`StakingFeature`].
+    /// PANIC: do not call on a non-staking feature.
+    pub fn as_staking(&self) -> &StakingFeature {
+        if let Self::Staking(feature) = self {
+            feature
+        } else {
+            panic!("as_staking called on a non-staking feature");
```
DaughterOfMars: I feel the messaging could be better. The doc should say something like

```rust
/// NOTE: Will panic if the feature is not a `StakingFeature`.
```

and the panic message like


```rust
panic!("invalid downcast of non-staking feature");
```

thibault-martinez: I'll change that in develop as well

sdk/src/types/block/output/feature/mod.rs
thibault-martinez: Done

sdk/src/types/block/output/feature/mod.rs
thibault-martinez: Done

kwek20 (dismissed):

DaughterOfMars (commented):
LGTM, other than the failing CI

Thoralf-M (dismissed):

Thoralf-M (dismissed):

kwek20 (approved):

Thoralf-M (approved):

Alex6323 (approved):

qrayven (approved):

DaughterOfMars (approved):

Issue #42, opened 2023-03-09
Make sure all bindings are properly typed


thibault-martinez:
- [ ] Java
- [ ] Wasm
- [ ] Python
- [ ] Nodejs

PR #720, opened 2023-07-04
Python: add NodeInfo class

Thoralf-M:
# Description of change

Python: add NodeInfo class

#42 

## Type of change

Running the get_info example (modified also with get_node_info())

## Change checklist

### DIFF
--- a/bindings/python/iota_sdk/types/node_info.py
+++ b/bindings/python/iota_sdk/types/node_info.py
@@ -40,7 +40,7 @@ class NodeInfoMetrics:
 
 @dataclass
 class RentStructure:
-    """Rent structure for the storage deposit
+    """Rent structure for the storage deposit.
     """
     vByteCost: int
     vByteFactorData: int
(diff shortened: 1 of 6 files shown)


bindings/python/iota_sdk/types/node_info.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+
+    blocksPerSecond: float
+    referencedBlocksPerSecond: float
+    referencedRate: float
+
+
+@dataclass
+class RentStructure:
+    """Rent structure for the storage deposit
```
Alex6323:
```suggestion
    """Rent structure for the storage deposit.
```
:stuck_out_tongue_winking_eye:

Alex6323 (approved):

qrayven (approved):

kwek20 (approved):
Do you think its worth making a types structure like in nodejs? python seems to be way less type-heavy, so less code and less files. but still would be convenient to have a similar structure

thibault-martinez (approved):

Thoralf-M:
@kwek20 yes probably makes sense to align even if we have less files :+1:

Issue #586, opened 2023-06-12
Refactor send functions


Dr-Electron:
## Description

I see multiple issues with our send functions. Let's use this issue to discuss them and eventually split it up in multiple issues if needed:

1) Our basic send function is called `send_amount`. And our "advanced" or send output function is called `send` (although in the bindings we call it send_outputs/sendOutputs). I think it should be the other way around:
You want to send something without any fancy stuff? Use `send`. 
You need more advanced outputs? Use `send_outputs`

2) SendAmountParams:
a) Do we really need that as it's own object, class struct? Can't the send function be `send(amount, address, options)`?
b) If we keep it, can we rename the var to something else than `output` in the examples? It's confusing 😆 . Let's call it params or something

## Are you planning to do it yourself in a pull request?

No.

thibault-martinez:
> 1. Our basic send function is called `send_amount`. And our "advanced" or send output function is called `send` (although in the bindings we call it send_outputs/sendOutputs). I think it should be the other way around:
>    You want to send something without any fancy stuff? Use `send`.
>    You need more advanced outputs? Use `send_outputs`

We agree, we'll change it.

> 2. SendAmountParams:
>    a) Do we really need that as it's own object, class struct? Can't the send function be `send(amount, address, options)`?
>    b) If we keep it, can we rename the var to something else than `output` in the examples? It's confusing 😆 . Let's call it params or something

We don't really agree, there is more to `SendAmountParams` than just amount and address. Also these functions take an array of `SendAmountParams` and not a single amount/address tuple.

Dr-Electron:
> We don't really agree, there is more to `SendAmountParams` than just amount and address. Also these functions take an array of `SendAmountParams` and not a single amount/address tuple.

But why can't the other options be part of "options"?
In most cases people probably just want to do something like `send(10, "rms...")` and for the 10% edge cases we provide something like `send(10, "rms...", Options::new().with_expiration())`

A function should always be optimised for the 90% of the use cases 🤔

Alex6323:
Please have a look at #725 and comment on it whether you'ld be okay with it. I don't think merging the other params with the transaction options is a good idea, though.

Thoralf-M:
Before closing this issue bindings still need to be updated with this https://github.com/iotaledger/iota-sdk/pull/725

PR #725, opened 2023-07-05
feat(account)!: simplify send(); introduce send_with_params()

Alex6323:
# Description of change

Addresses point 2) as described in #586. Let's discuss whether we like it, or not.

#586 

## Type of change

### DIFF
--- a/sdk/src/wallet/account/operations/transaction/high_level/send.rs
+++ b/sdk/src/wallet/account/operations/transaction/high_level/send.rs
@@ -31,13 +31,13 @@ use crate::{
 /// Parameters for `send()`
 #[derive(Debug, Clone, Serialize, Deserialize, Getters)]
 pub struct SendParams {
-    /// Bech32 encoded address
-    #[getset(get = "pub")]
-    address: Bech32Address,
     /// Amount
     #[serde(with = "crate::utils::serde::string")]
     #[getset(get = "pub")]
     amount: u64,
+    /// Bech32 encoded address
+    #[getset(get = "pub")]
+    address: Bech32Address,
     /// Bech32 encoded return address, to which the storage deposit will be returned if one is necessary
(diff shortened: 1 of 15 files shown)


sdk/src/wallet/account/operations/transaction/high_level/send.rs
```diff
@@ -31,13 +31,13 @@ use crate::{
...
-    /// Bech32 encoded address
-    #[getset(get = "pub")]
-    address: Bech32Address,
     /// Amount
     #[serde(with = "crate::utils::serde::string")]
     #[getset(get = "pub")]
     amount: u64,
+    /// Bech32 encoded address
+    #[getset(get = "pub")]
+    address: Bech32Address,
```
Thoralf-M, line 40: Why change the order?

Alex6323: consistency (with the `new` ctor arg list which I also changed to reflect the proposed order by Kevin in #586 )

Thoralf-M: @Dr-Electron do you think amount should be first or was it just an example?
Asking because in the cli wallet we also always accept the address first

Dr-Electron: I don't have a strong opinion on it. But if I'm allowed to decide I prefer having the amount first.

Looks nicer if you `send` `x` token to `address`.

thibault-martinez: I think I actually prefer amount first, more consistent with the way it is in outputs

DaughterOfMars (approved):

Thoralf-M (approved):
Bindings in another PR?

Issue #387, opened 2023-05-04
QA: reach 75% coverage


thibault-martinez:
1.3% each :)

- [x] Thibault https://coveralls.io/builds/61233706 (0.2%), https://coveralls.io/builds/61240890 (0.2%), https://coveralls.io/builds/61274869 (0.2%), https://coveralls.io/builds/61320781 (0.3%), https://coveralls.io/builds/61327310 (0,07%), https://coveralls.io/builds/61409496 (0.4%)
- [x] AlexC https://coveralls.io/builds/61223133 (1.2%), https://coveralls.io/builds/61351390 (0.1%)
- [ ] AlexS
- [x] Thoralf https://coveralls.io/builds/60917230 (1.0%), https://coveralls.io/builds/61019584 (0.2%), https://coveralls.io/builds/61247677 (0.02%) https://coveralls.io/builds/61277690 (0.1%)
- [ ] Pawel  https://coveralls.io/builds/61490033 (0.4%)  https://coveralls.io/builds/61494944 (0.02%)
- [ ] Brord https://coveralls.io/builds/61418775 (0.3%)

PR #737, opened 2023-07-07
Add test for incoming transaction sync

DaughterOfMars:
# Description of change

Just a little test to cover some missing code.

- #387 

## Type of change

### DIFF
--- a/sdk/tests/wallet/syncing.rs
+++ b/sdk/tests/wallet/syncing.rs
@@ -151,6 +151,65 @@ async fn sync_only_most_basic_outputs() -> Result<()> {
 
 #[ignore]
 #[tokio::test]
+async fn sync_incoming_transactions() -> Result<()> {
+    let storage_path = "test-storage/sync_incoming_transactions";
+    setup(storage_path)?;
+
+    let wallet = make_wallet(storage_path, None, None).await?;
+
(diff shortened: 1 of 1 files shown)


sdk/tests/wallet/syncing.rs
```diff
@@ -149,6 +149,65 @@ async fn sync_only_most_basic_outputs() -> Result<()> {
...
+            *output_data
+                .output
+                .unlock_conditions()
+                .unwrap()
+                .address()
+                .unwrap()
+                .address(),
+            account_1_address
+        );
+    });
```
Thoralf-M: This only checks for unspent outputs which you also get without `sync_incoming_transactions: true`
There should be at least something like
```
    let incoming_transactions = account_1.incoming_transactions().await;
    assert_eq!(incoming_transactions.len(), 1);
```
And you can also check that the input for this transaction was synced as well (since we sync right away it can't be pruned yet and must be available)
```
    let incoming_tx = account_1.get_incoming_transaction(&tx.transaction_id).await?;
    assert_eq!(incoming_tx.inputs.len(), 1);
```

DaughterOfMars: Thanks!

Thoralf-M (approved):

thibault-martinez (approved):

PR #738, opened 2023-07-07
chore(examples): some fixes

Alex6323:
`process.env.MNEMONIC` was still only used in some places ... and now it's consistently used in the whole SDK.


bindings/nodejs/examples/exchange/1-create-account.ts
```diff
@@ -30,8 +30,8 @@ async function run() {
                 '.env STRONGHOLD_PASSWORD is undefined, see .env.example',
             );
         }
-        if (!process.env.MNEMONIC) {
-            throw new Error('.env MNEMONIC is undefined, see .env.example');
+        if (!process.env.NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC_1) {
```
Thoralf-M: In the exchange guide it was extra made `MNEMONIC` for https://wiki.iota.org/next/iota-sdk/how-tos/exchange-guide/#1-generate-a-mnemonic
Would look weird to use `NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC_1` there

Alex6323: ah, makes sense ... but then we need to change `.env.example` or have a separate for them, right?

Alex6323: tbh ... I would call it `MNEMONIC` everywhere, it's simpler, and we already say in the docs of `.env.example` that those credentials should never be used for anything serious.

Thoralf-M: The guide says to create one with the values

kwek20: yea i agree, its already weird that we have specific examples for exchange (even though theyre not so different).can at least have the code inside reused

thibault-martinez: I would also rename everything to `MNEMONIC`

Alex6323: Done

Thoralf-M (changes requested):
nodejs fmt

Thoralf-M (approved):

thibault-martinez (approved):

DaughterOfMars:
Do we really want to change the explicit name here?

thibault-martinez:
> Do we really want to change the explicit name here?

I think it's fine tbh, we're being explicit enough with all the warnings around dotenv. And it's a bit inconsistent, if we really wanted to do that we should do it with anything related to seed/mnemonic/password everywhere is examples and we don't.

DaughterOfMars (approved):

PR #742, opened 2023-07-10
Custom serde test

thibault-martinez:


Thoralf-M (approved):

sdk/tests/utils/serde.rs
```diff
@@ -0,0 +1,81 @@
...
+
+#[test]
+fn serde_string() {
+    let block_id = BlockId::from_str(BLOCK_ID).unwrap();
+    let test_1 = SerdeStringTest { block_id };
+    let str = serde_json::to_string(&test_1).unwrap();
+    let test_2 = serde_json::from_str(&str).unwrap();
+
+    assert_eq!(test_1, test_2);
+    assert!(str.contains(BLOCK_ID));
```
Alex6323, line 40: isn't that already implied with the first `assert_eq`?

thibault-martinez: No, it could be doing something completely different if we messed up the serde impl, I'm checking that this specific string is actually there

Alex6323 (approved):

PR #749, opened 2023-07-10
Add test with custom input

Thoralf-M:
# Description of change

Add test with custom input

#387 

## Type of change

Running it with a private tangle

## Change checklist

### DIFF
--- a/sdk/tests/client/transactions.rs
+++ b/sdk/tests/client/transactions.rs
@@ -7,6 +7,7 @@ use iota_sdk::{
     client::{api::GetAddressesOptions, node_api::indexer::query_parameters::QueryParameter, Result},
     types::block::{
         address::ToBech32Ext,
+        input::{Input, UtxoInput},
         output::{unlock_condition::AddressUnlockCondition, BasicOutputBuilder, OutputId},
         payload::{transaction::TransactionEssence, Payload},
     },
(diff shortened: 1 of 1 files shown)


sdk/tests/client/transactions.rs
```diff
@@ -68,3 +69,69 @@ async fn send_basic_output() -> Result<()> {
...
+        .await?[0];
+
+    let bech32_hrp = client.get_bech32_hrp().await?;
+
+    // First create multiple outputs on our address, so there is even the possibility to select another input
+    let output = BasicOutputBuilder::new_with_amount(1_000_000)
+        .add_unlock_condition(AddressUnlockCondition::new(address))
+        .finish_output(token_supply)?;
+    let block = client
+        .block()
```
DaughterOfMars, line 91: Unrelated to this PR, but I feel like this should be `build_block()`, right??

sdk/tests/client/transactions.rs
```diff
@@ -68,3 +69,69 @@ async fn send_basic_output() -> Result<()> {
...
+    let output_ids_response = client
+        .basic_output_ids([
+            QueryParameter::Address(address.to_bech32(bech32_hrp)),
+            QueryParameter::HasExpiration(false),
+            QueryParameter::HasTimelock(false),
+            QueryParameter::HasStorageDepositReturn(false),
+        ])
+        .await?;
+
+    // Use not the first one, because the client could also always just do that when selecting inputs automatically
```
DaughterOfMars: Strange phrasing here

Thoralf-M: Is `Don't use the first one, because the client could also always just do that when automatically selecting inputs` better? If not please write something better

sdk/tests/client/transactions.rs
Thoralf-M: We already had `send_transaction_message, send_indexation_message`, `send`, `message`,  `block` now :trollface:
The thing is that it's not only building it, but also sending
Also in 2.0 we most probably remove this completely (the transaction part there is already removed), so I would just keep it as it is now

sdk/tests/client/transactions.rs
DaughterOfMars: How about "The first output would be picked when using automatic input selection, so use the second one instead"

sdk/tests/client/transactions.rs
DaughterOfMars: But you have to call `finish` right? So this fn just gives you the builder.

sdk/tests/client/transactions.rs
Thoralf-M:
```suggestion
    // The first output would be picked when using automatic input selection, so use the second one instead
```

Thoralf-M: Thanks

sdk/tests/client/transactions.rs
Thoralf-M: True, I don't care that much how we name it, open an issue if you want

DaughterOfMars (approved):

thibault-martinez (approved):

PR #750, opened 2023-07-10
Python: add balance classes

Thoralf-M:
# Description of change

Python: add balance classes

#42 

## Type of change

Running examples

## Change checklist

### DIFF
--- a/bindings/python/examples/how_tos/accounts_and_addresses/check_balance.py
+++ b/bindings/python/examples/how_tos/accounts_and_addresses/check_balance.py
@@ -17,4 +17,4 @@ _balance = account.sync()
 
 # Just calculate the balance with the known state
 balance = account.get_balance()
-print(f'Balance {json.dumps(balance, indent=4)}')
+print(f'Balance {json.dumps(balance.as_dict(), indent=4)}')
--- a/bindings/python/iota_sdk/types/balance.py
+++ b/bindings/python/iota_sdk/types/balance.py
@@ -1,0 +1,58 @@
+# Copyright 2023 IOTA Stiftung
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+from dataclasses import dataclass
+from typing import List, Optional
+from iota_sdk.types.common import HexStr
+
+
+@dataclass
+class BaseCoinBalance:
+    """Base coin fields for Balance.
+    """
+    total: str
+    available: str
+
+
+@dataclass
+class RequiredStorageDeposit:
+    """Required storage deposit for the outputs in the account.
+    """
+    alias: str
+    basic: str
+    foundry: str
+    nft: str
+
+
+@dataclass
+class NativeTokensBalance:
+    """Native tokens fields for Balance.
+    """
+    tokenId: HexStr
+    total: HexStr
+    available: HexStr
+    metadata: Optional[HexStr]
+
+
+@dataclass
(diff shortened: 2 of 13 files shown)


bindings/python/examples/how_tos/accounts_and_addresses/check_balance.py
```diff
@@ -17,4 +17,4 @@
 
 # Just calculate the balance with the known state
 balance = account.get_balance()
-print(f'Balance {json.dumps(balance, indent=4)}')
+print(f'Balance {json.dumps(balance.as_dict(), indent=4)}')
```
DaughterOfMars, line 20: What's the difference between this and `__dict__`?

bindings/python/iota_sdk/types/balance.py
```diff
@@ -0,0 +1,58 @@
...
+    basic: str
+    foundry: str
+    nft: str
+
+
+@dataclass
+class NativeTokensBalance:
+    """Native tokens fields for Balance.
+    """
+    tokenId: HexStr
```
DaughterOfMars, line 32: No type for this?

bindings/python/iota_sdk/types/balance.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@dataclass
+class Balance:
+    """The balance of an account.
+    """
+    baseCoin: BaseCoinBalance
+    requiredStorageDeposit: RequiredStorageDeposit
+    nativeTokens: List[NativeTokensBalance]
+    nfts: List[HexStr]
```
DaughterOfMars, line 45: No types for these IDs either?

bindings/python/examples/how_tos/accounts_and_addresses/check_balance.py
Thoralf-M: `as_dict()` needs to be manually defined and can for example convert something, `__dict__` is automatically derived and will not do any conversions

bindings/python/iota_sdk/types/balance.py
Thoralf-M: No

bindings/python/iota_sdk/types/balance.py
Thoralf-M: No

bindings/python/iota_sdk/types/balance.py
DaughterOfMars: Ok 👍

bindings/python/examples/how_tos/accounts_and_addresses/check_balance.py
DaughterOfMars: Should this be able to use `__dict__` then?

DaughterOfMars (approved):

bindings/python/examples/how_tos/accounts_and_addresses/check_balance.py
Thoralf-M: Not in `json.dumps()`, because there are other classes inside of this one and then it returns an error `Object of type BaseCoinBalance is not JSON serializable` unless we would add a custom encoder
It's possible to print without the `json.dumps()`/`.as_dict()`, but then the output wouldn't be the same JSON format as we have also in the other languages

thibault-martinez (approved):

Issue #506, opened 2023-05-30
Listen to MQTT. Shimmer transactions https://wiki.iota.org/shimmer/iota.rs/how_tos/mqtt/


abcen7:
## Description

Need to release that: https://wiki.iota.org/shimmer/iota.rs/how_tos/mqtt/

## Motivation

Bivreost is need this feature to get external transactions for the wallet

## Requirements

Create a Client which will connect to the [Shimmer Testnet](https://api.testnet.shimmer.network/).
Subscribe to a set of topics.
Listen to the topic defined in step 2 and log the output.

## Are you planning to do it yourself in a pull request?

No.

Dr-Electron:
@thibault-martinez guess this should still be in v1.0 as we need to make sure the bindings have the same features

abcen7:
HELLO GUYS, HOW IS YOUR WORK ABOUT THIS PROBLEM, PLEASE RELEASE THAT, WE REALLY NEED THIS FEATURE TO OUR PRODUCTION WALLET!!!!!!!!

PR #754, opened 2023-07-10
Add MQTT to Python binding

Thoralf-M:
# Description of change

Add MQTT to Python binding

#506 

## Type of change

Running the added example

## Change checklist

### DIFF
--- a/bindings/nodejs/lib/client/client.ts
+++ b/bindings/nodejs/lib/client/client.ts
@@ -985,7 +985,7 @@ export class Client {
     /**
      * Listen to MQTT topics.
      */
-    async listen(
+    async listenMqtt(
         topics: string[],
         callback: (error: Error, result: string) => void,
     ): Promise<void> {
@@ -995,7 +995,7 @@ export class Client {
     /**
      * Stop listening for provided MQTT topics.
      */
-    async clearListeners(topics: string[]): Promise<void> {
+    async clearMqttListeners(topics: string[]): Promise<void> {
         await this.methodHandler.callMethod({
             name: 'clearListeners',
             data: {
--- a/bindings/python/examples/client/mqtt.py
+++ b/bindings/python/examples/client/mqtt.py
@@ -21,10 +21,10 @@ def callback(event):
 
 
 # Topics can be found here https://studio.asyncapi.com/?url=https://raw.githubusercontent.com/iotaledger/tips/main/tips/TIP-0028/event-api.yml
-client.listen(["blocks"], callback)
+client.listen_mqtt(["blocks"], callback)
 
 # Exit after 10 received events
 while True:
     if received_events > 10:
-        client.clear_listeners(["blocks"])
+        client.clear_mqtt_listeners(["blocks"])
         exit()
--- a/bindings/python/examples/client/mqtt.py
+++ b/bindings/python/examples/client/mqtt.py
@@ -26,4 +26,5 @@ client.listen(["blocks"], callback)
 # Exit after 10 received events
 while True:
     if received_events > 10:
+        client.clear_listeners(["blocks"])
         exit()
--- a/bindings/python/examples/client/mqtt.py
+++ b/bindings/python/examples/client/mqtt.py
@@ -29,6 +29,6 @@ def callback(event):
 client.listen_mqtt(["blocks"], callback)
 
 # Exit after 10 received events
-while received_10_events.wait():
-    client.clear_mqtt_listeners(["blocks"])
-    exit()
+received_10_events.wait()
+client.clear_mqtt_listeners(["blocks"])
+exit()
--- a/bindings/python/examples/client/mqtt.py
+++ b/bindings/python/examples/client/mqtt.py
@@ -1,4 +1,9 @@
-from iota_sdk import Client, init_logger
+# Copyright 2023 IOTA Stiftung
+# SPDX-License-Identifier: Apache-2.0
+
+# This example shows how to listen to MQTT events of a node.
+
+from iota_sdk import Client
 from dotenv import load_dotenv
 import os
 import threading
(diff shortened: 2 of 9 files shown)


bindings/python/examples/client/mqtt.py
```diff
@@ -0,0 +1,29 @@
...
+    print(event_dict)
+    global received_events
+    received_events += 1
+
+
+# Topics can be found here https://studio.asyncapi.com/?url=https://raw.githubusercontent.com/iotaledger/tips/main/tips/TIP-0028/event-api.yml
+client.listen(["blocks"], callback)
+
+# Exit after 10 received events
+while True:
```
DaughterOfMars: Is....is this for real the best way?

bindings/python/iota_sdk/client/client.py
```diff
@@ -517,3 +517,8 @@ def submit_payload(self, payload) -> List[HexStr|Block]:
         })
         result[1] = Block.from_dict(result[1])
         return result
+
+    def listen(self, topics: List[str], handler):
```
thibault-martinez: Should we also call this `listen_mqtt` to avoid confusion with the other `listen`, even if it's in the wallet?

bindings/python/examples/client/mqtt.py
Thoralf-M: Probably not the best way, but it's a simple one, don't know what's better for an example

bindings/python/iota_sdk/client/client.py
Thoralf-M: Done

bindings/nodejs/lib/client/client.ts
```diff
@@ -985,7 +985,7 @@ export class Client {
     /**
      * Listen to MQTT topics.
      */
-    async listen(
+    async listenMqtt(
```
thibault-martinez, line 988: changelog

bindings/python/examples/client/mqtt.py
DaughterOfMars: Well we certainly don't want to have an example with a really bad method in it. Is the callback executed concurrently? Is there a sync primitive we can use to do this properly?

bindings/nodejs/lib/client/client.ts
Thoralf-M: Added

bindings/python/examples/client/mqtt.py
thibault-martinez: But this loop looks like a 100% CPU until we have 10 events

bindings/python/iota_sdk/client/client.py
```diff
@@ -509,11 +510,23 @@ def sign_transaction(self, secret_manager, prepared_transaction_data):
...
         result[1] = Block.from_dict(result[1])
         return result
+
+    def listen_mqtt(self, topics: List[str], handler):
+        """Listen to MQTT events.
+        """
+        listen_mqtt(self.handle, topics, handler)
+
+    def clear_mqtt_listeners(self, topics: List[str]):
+        """Removes all listeners for the provided topics.
```
thibault-martinez:
```suggestion
        """Removes all listeners for the provided MQTT topics.
```

bindings/python/examples/client/mqtt.py
Thoralf-M: I used `threading.Event()` now which seems to be an efficient way of doing it

thibault-martinez (dismissed):

bindings/python/examples/client/mqtt.py
```diff
@@ -0,0 +1,34 @@
...
+    received_events += 1
+    if received_events > 10:
+        received_10_events.set()
+
+
+# Topics can be found here https://studio.asyncapi.com/?url=https://raw.githubusercontent.com/iotaledger/tips/main/tips/TIP-0028/event-api.yml
+client.listen_mqtt(["blocks"], callback)
+
+# Exit after 10 received events
+while not received_10_events.wait():
```
DaughterOfMars: This seems backwards?

Thoralf-M: What do you mean?

DaughterOfMars: It may be the way this works is not obvious, but shouldn't we just call `received_10_events.wait()` without the while loop? Doesn't it just return when set by the other thread?

Thoralf-M: Yeah you're right, it just worked before as well because it will always exit, fixed

Thoralf-M: Yes you're right, it will wait and the while isn't required at all
Updated

thibault-martinez (dismissed):

bindings/python/examples/client/mqtt.py
```diff
@@ -0,0 +1,33 @@
+from iota_sdk import Client, init_logger
```
Alex6323: add an example description?

bindings/python/iota_sdk/client/client.py
```diff
@@ -20,6 +20,7 @@
 from typing import Any, Dict, List, Optional
 from dacite import from_dict
 
+
```
Alex6323, line 23: weird Python formatter :trollface:

bindings/python/examples/client/mqtt.py
Thoralf-M: Done

bindings/python/iota_sdk/client/client.py
Thoralf-M: Yeah maybe we should have our own config that doesn't do it like that :trollface:

Alex6323 (approved):

thibault-martinez (approved):

PR #763, opened 2023-07-11
Add tests for raw API endpoints

thibault-martinez:


Thoralf-M (approved):

Thoralf-M (changes requested):
Test fails

Thoralf-M (approved):

sdk/tests/client/node_api.rs
```diff
@@ -287,13 +287,17 @@ async fn test_get_milestone_by_id_raw() {
...
     let milestone = client.get_milestone_by_id(&latest_milestone_id).await.unwrap();
-    let milestone_raw = MilestonePayload::unpack_verified(
+    let milestone_raw = Payload::unpack_verified(
         client.get_milestone_by_id_raw(&latest_milestone_id).await.unwrap(),
         &client.get_protocol_parameters().await.unwrap(),
     )
     .unwrap();
 
-    assert_eq!(milestone, milestone_raw);
+    if let Payload::Milestone(milestone_raw) = milestone_raw {
```
DaughterOfMars, line 296: So one of these returns a `MilestonePayload` and one returns `Payload`? Why is that?

thibault-martinez: Comes from a difference between the SDK and iota.go I guess, they don't have a payload enum so the type is on the variant

thibault-martinez: I don't think we should modify the bytes returned by the API to remove the first one

DaughterOfMars: Alright 🤷‍♀️

DaughterOfMars (approved):

PR #776, opened 2023-07-12
Use `BigInt` instead of hex encoded numbers in Node.js

### LINKED ISSUE #490: NodeJS to accept numbers when it applies
## Description

Sometimes we ask for hexencoded numbers in nodejs.
This makes usage less smooth, we should switch to number.

We could use a [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) for this

## Motivation

UX

## Requirements

1. `HexEncodedAmount` to change to number externally

## Open questions (optional)

What do do about `HexEncodedString`
- Update OutputId and friends to use HexEncoded when possible

Yes/No.

abdulmth:
# Description of change

Use `BigInt` values instead of
* Hex encoded numbers for native tokens.
* Decimal numbers as strings for output amount.

The default serialization for `bigint` values is set to be a hex string in `bindings.ts`.
```typescript
(BigInt.prototype as any).toJSON = function () {
    return bigIntToHex(this);
};
```

 Thus, serialization to decimal must be done manually.

fixes #490 

## Type of change


bindings/nodejs/examples/client/14-build-foundry-output.ts
```diff
@@ -33,9 +33,12 @@ async function run() {
             '0xff311f59790ccb85343a36fbac2f06d233734794404142b308c13f2c616935b5';
 
         const foundryOutput = await client.buildFoundryOutput({
-            serialNumber: 0,
-            // 10 hex encoded
-            tokenScheme: new SimpleTokenScheme('0xa', '0x0', '0xa'),
+            serialNumber: 1,
```
abdulmth, line 36: With 0 the example fails.

bindings/nodejs/examples/client/14-build-foundry-output.ts
```diff
@@ -33,9 +33,12 @@ async function run() {
...
+            tokenScheme: new SimpleTokenScheme(
+                BigInt(10),
+                BigInt(0),
+                BigInt(10),
+            ),
             amount: '1000000',
```
Thoralf-M: Shouldn't this also be a BigInt now?

thibault-martinez: I thought only native tokens would be Bigint?

Thoralf-M: No, I think we want all Bigint, so it's easier to use (you can use the + * - % ** operators which will not work properly with strings)
Only native tokens get hex encoded when serialized, amount gets converted to a string

abdulmth: done

abdulmth: done

thibault-martinez (changes requested):
CHANGELOG entries

bindings/nodejs/CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Rename `Client::listen` to `listenMqtt`, `Client::clearListeners` to `clearMqttListeners`;
 - Moved `minimumRequiredStorageDeposit()` from `Account` to `Client`;
+- Use `BigInt` instead of strings for token amounts.
```
```suggestion
- Use `BigInt` instead of strings for token amounts;
```

bindings/nodejs/lib/bindings.ts
```diff
@@ -99,6 +99,11 @@ const callWalletMethodAsync = (
         });
     });
 
+// Allow bigint to be serialized as hex string.
+(BigInt.prototype as any).toJSON = function () {
+    return bigIntToHex(this);
+};
```
Thoralf-M, line 109: This works because the amount that should be string encoded will always be string encoded before, right?
If so, maybe extend the comment with that

bindings/nodejs/lib/types/wallet/account.ts
```diff
@@ -56,8 +56,8 @@ export interface RequiredStorageDeposit {
 export interface NativeTokenBalance {
```
Thoralf-M, line 56: Can't comment further above, shouldn't `BaseCoinBalance` fields also be bigint?

bindings/nodejs/lib/bindings.ts
abdulmth: I'm not sure what do you mean exactly, but the bindings expect a hex encoded string for the native tokens, if you keep it as of type `bigint`, you get an error that it doesn't know what to do with a `bigint` since it calles `JSON.stringify` on the object that contains the amount.
for example 
```javascript
let a = {a: BigInt(10)}
JSON.stringify(a)
```
will output 
```
Uncaught TypeError: Do not know how to serialize a BigInt
    at JSON.stringify (<anonymous>)
    at <anonymous>:1:6
```
but if the `.toJSON` is added to `BigInt` then it will serialize it to hex string as defined in `bigIntToHex` 
`'{"a":"0xa"}'`

bindings/nodejs/lib/types/wallet/account.ts
abdulmth: right, probably also `RequiredStorageDeposit` .

bindings/nodejs/lib/bindings.ts
Thoralf-M: Yeah this part I understand, but for the amount we also have bigint and there it should be serialized as string, so this distinction is made earlier?

bindings/nodejs/lib/types/wallet/account.ts
abdulmth: done

abdulmth: Maybe we also should change the voting power, but probably better in another PR for better review.

bindings/nodejs/lib/types/wallet/transaction-options.ts
```diff
@@ -46,9 +45,9 @@ export type CustomAddress = {
 export interface CreateNativeTokenParams {
     aliasId?: string;
     /** Hex encoded number */
-    circulatingSupply: HexEncodedAmount;
+    circulatingSupply: bigint;
```
thibault-martinez, line 48: You sometimes have `bigint | string` and sometimes just `bigint` for params, what's the reasoning?

bindings/nodejs/lib/types/block/output/output.ts
```diff
@@ -31,9 +31,13 @@ abstract class Output /*implements ICommonOutput*/ {
 
     private type: OutputType;
 
-    constructor(type: OutputType, amount: string) {
+    constructor(type: OutputType, amount: bigint | string) {
         this.type = type;
-        this.amount = amount;
+        if (typeof amount == 'bigint') {
+            this.amount = amount.toString(10);
```
thibault-martinez, line 37: Does that mean the field type is just a string? Should it be a bigint instead? For other types, the field has been changed as well

bindings/nodejs/lib/types/block/output/token-scheme.ts
```diff
@@ -26,39 +26,59 @@ abstract class TokenScheme {
...
-        this.mintedTokens = mintedTokens;
-        this.meltedTokens = meltedTokens;
-        this.maximumSupply = maximumSupply;
+        if (typeof mintedTokens === 'bigint') {
+            this.mintedTokens = mintedTokens;
+        } else if (mintedTokens) {
+            this.mintedTokens = hexToBigInt(mintedTokens);
+        } else {
+            this.mintedTokens = BigInt(0);
+        }
```
Alex6323, line 45: out-of-interest (not a TS/JS guy): the `constructor` does not enforce `bigint` type already? So that's why an actual type check with `typeof` is still necessary and one could also pass in a String? Or should this be `bigint | string`?

abdulmth: If the bindings construct this object, it will be of type `string`, in runtime, the TS types aren't relevant, so if the user wants to construct this object, they can use `bigint`, but if it's constructed through the bindings, it's gonna be of type `string` and the `else if` clause will deal with that case. We could make it `bigint | string` but I think the bindings don't care about the type as I understood.

bindings/nodejs/lib/types/block/output/output.ts
abdulmth: In the other type, the fields were serialized to hex which will be done automatically, but this one needs to be a decimal string, hence it's stored internally as a decimal string. If we make it `bigint` it will be serialized as hex and the values will be wrong.

bindings/nodejs/lib/types/wallet/transaction-options.ts
abdulmth: If the serialization is done manually, for example in `bindings/nodejs/lib/client/client.ts`  the resulted property is converted from `bigint` to `string` type. If we want to make it take `bigint` only, we could make the entire object of `any` type, but then we don't have type safety internally.

bindings/nodejs/lib/bindings.ts
abdulmth: Done, I added extra comment.

bindings/nodejs/lib/types/block/output/token-scheme.ts
Alex6323: okay

Thoralf-M (dismissed):

bindings/nodejs/examples/how_tos/native_tokens/melt.ts
```diff
@@ -40,7 +40,7 @@ async function run() {
             );
         }
 
-        console.log(`Balance before melting:`, parseInt(token.available));
+        console.log(`Balance before melting:`, token.available.toString());
```
kwek20: can you put it like the burn example inside `${token.available}` ?

bindings/nodejs/examples/how_tos/native_tokens/mint.ts
```diff
@@ -39,8 +39,7 @@ async function run() {
                 `Couldn't find native token '${tokenId}' in the account`,
             );
         }
-
-        console.log(`Balance before minting:`, parseInt(token.available));
+        console.log(`Balance before minting:`, token.available.toString());
```
kwek20: same here

bindings/nodejs/examples/how_tos/native_tokens/send.ts
```diff
@@ -51,7 +51,7 @@ async function run() {
                     `Couldn't find native token '${tokenId}' in the account`,
                 );
             }
-            console.log(`Balance before sending:`, parseInt(token.available));
+            console.log(`Balance before sending:`, token.available.toString());
```
kwek20: and here :D

bindings/nodejs/examples/wallet/06-send-micro-transaction.ts
```diff
@@ -30,7 +30,7 @@ async function run() {
         await account.sync();
 
         console.log(
-            `Sending '${SEND_MICRO_AMOUNT}' coin(s) to '${RECV_ADDRESS}'...`,
+            `Sending '${SEND_MICRO_AMOUNT.toString()}' coin(s) to '${RECV_ADDRESS}'...`,
```
kwek20: yousure this was needed? now im confused 😢

bindings/nodejs/lib/types/block/output/token-scheme.ts
```diff
@@ -26,39 +26,59 @@ abstract class TokenScheme {
...
+
+        if (typeof meltedTokens === 'bigint') {
```
kwek20, line 47: SHould we make this a helper function? we have it a few times around

bindings/nodejs/lib/types/client/build-block-options.ts
```diff
@@ -31,5 +31,5 @@ export interface IBuildBlockOptions {
 /** Address with base coin amount */
 export interface IClientBlockBuilderOutputAddress {
     address: string;
-    amount: string;
+    amount: bigint | string;
```
kwek20, line 34: should this be hexencodedamount, or jsut string is fine?

bindings/nodejs/lib/types/utils/hex-encoding.ts
```diff
@@ -0,0 +1,30 @@
...
+export declare type HexEncodedString = string;
+/**
+ * Hex encoded U256.
+ */
+export declare type HexEncodedAmount = string;
+
+/**
+ * Converts `bigint` value to hexadecimal string representation prefixed with "0x".
+ */
+export function bigIntToHex(value: bigint): string {
```
kwek20, line 16: return HexEncodedAmoutn?

bindings/nodejs/lib/types/client/output_builder_params/basic-output-params.ts
```diff
@@ -9,7 +9,7 @@ export interface BasicOutputBuilderParams {
     /**
      * If not provided, minimum storage deposit will be used
      */
-    amount?: string;
+    amount?: bigint | string;
```
kwek20, line 12: should this be hexencodedamount, or jsut string is fine?

bindings/nodejs/lib/types/wallet/transaction-options.ts
kwek20: yea we should make it at least bigint | hexencodedamount   (if that works)

bindings/nodejs/examples/wallet/06-send-micro-transaction.ts
abdulmth: You're right, it's not needed, the `console.log` function calls it by default, I was afraid that it will add `n` to the number if `toString()` was not called, but it actually doesn't.

bindings/nodejs/lib/types/client/build-block-options.ts
abdulmth: This is a decimal string, not hex, should we actually make a type for that too? Something like `NumericString`?

bindings/nodejs/lib/types/client/output_builder_params/basic-output-params.ts
abdulmth: Here too it's decimal, not hex.

bindings/nodejs/lib/types/block/output/token-scheme.ts
abdulmth: what exactly? the `typeof meltedTokens === 'bigint'` part?

bindings/nodejs/lib/types/client/build-block-options.ts
kwek20: i think we should yea, to avoid confusion as currently its still not very clear what is what NumericString sounds great

bindings/nodejs/lib/types/block/output/token-scheme.ts
kwek20: the `if {} elseif {} else {}`

bindings/nodejs/lib/types/client/build-block-options.ts
abdulmth: ✅ Will do in separate PR

bindings/nodejs/lib/types/block/output/token-scheme.ts
abdulmth: this one is not really duplicated anywhere else, but there's 
```typescript 
        if (params.amount && typeof params.amount === 'bigint') {
            params.amount = params.amount.toString(10);
        }
```
duplicated in a couple of places, but maybe it's better to keep it as is, since it's not much and and we would avoid modifying the object by reference in a function.

Thoralf-M (approved):

bindings/nodejs/lib/types/block/output/token-scheme.ts
kwek20: i mean like a method:
```
function coerceBigInt(item) -> BigInt
       if (typeof item=== 'bigint') {
           return item;
        } else if (item) {
            return hexToBigInt(item);
        } else {
            return BigInt(0);
        }
```

thibault-martinez (approved):

PR #807, opened 2023-07-18
refactor: passphrase arg

Alex6323:


sdk/src/client/stronghold/secret.rs
```diff
@@ -325,19 +325,14 @@ impl SecretManagerConfig for StrongholdAdapter {
 /// Private methods for the secret manager implementation.
 impl StrongholdAdapter {
     /// Execute [Procedure::BIP39Recover] in Stronghold to put a mnemonic into the Stronghold vault.
-    async fn bip39_recover(
-        &self,
-        mnemonic: Mnemonic,
-        passphrase: Option<String>,
-        output: Location,
-    ) -> Result<(), Error> {
+    async fn bip39_recover(&self, mnemonic: Mnemonic, passphrase: Passphrase, output: Location) -> Result<(), Error> {
```
Thoralf-M, line 328: Why not optional anymore?

Alex6323: because it wasn't really optional. A password is used in any case ... it's just that if you passed in `None` it would use `Password::default()`.

Thoralf-M (approved):

thibault-martinez (approved):

PR #808, opened 2023-07-18
Update bindings with send split

### LINKED ISSUE #586: Refactor send functions

DaughterOfMars:
# Description of change

This PR changes the bindings to match the previous send split PR #725.

- Closes #586 

## Type of change

### DIFF
--- a/bindings/nodejs/examples/how_tos/advanced_transactions/send_micro_transaction.ts
+++ b/bindings/nodejs/examples/how_tos/advanced_transactions/send_micro_transaction.ts
@@ -33,7 +33,7 @@ async function run() {
             'rms1qpszqzadsym6wpppd6z037dvlejmjuke7s24hm95s9fg9vpua7vluaw60xu';
         const amount = BigInt(1);
 
-        const transaction = await account.send(address, amount, {
+        const transaction = await account.send(amount, address, {
             allowMicroAmount: true,
         });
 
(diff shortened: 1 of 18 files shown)


bindings/nodejs/examples/how_tos/advanced_transactions/send_micro_transaction.ts
```diff
@@ -33,17 +33,9 @@ async function run() {
...
-                    address,
-                    amount,
-                },
-            ],
-            {
-                allowMicroAmount: true,
-            },
-        );
+        const transaction = await account.send(address, amount, {
+            allowMicroAmount: true,
```
Thoralf-M, line 37: This option is only possible to set with `sendWithParams()`, no?

Thoralf-M (approved):

Thoralf-M (changes requested):
CI fails

thibault-martinez (dismissed):

thibault-martinez (approved):

Thoralf-M (approved):

PR #813, opened 2023-07-19
Remove some TODOs

thibault-martinez:
https://github.com/iotaledger/iota-sdk/issues/401

### DIFF
--- a/sdk/src/types/block/output/token_id.rs
+++ b/sdk/src/types/block/output/token_id.rs
@@ -3,7 +3,7 @@
 
 use crate::types::block::output::FoundryId;
 
-impl_id!(pub TokenId, 38, "TODO.");
+impl_id!(pub TokenId, 38, "Unique identifiers of native tokens. The TokenId of native tokens controlled by a specific foundry is the same as the FoundryId.");
 
 #[cfg(feature = "serde")]
 string_serde_impl!(TokenId);
--- a/sdk/src/types/block/output/token_id.rs
+++ b/sdk/src/types/block/output/token_id.rs
@@ -3,7 +3,7 @@
 
 use crate::types::block::output::FoundryId;
 
-impl_id!(pub TokenId, 38, "Unique identifiers of native tokens. The TokenId of native tokens controlled by a specific foundry is the same as the FoundryId.");
+impl_id!(pub TokenId, 38, "Unique identifiers of native tokens. The TokenId of native tokens minted by a specific foundry is the same as the FoundryId.");
 
 #[cfg(feature = "serde")]
 string_serde_impl!(TokenId);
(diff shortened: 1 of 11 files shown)


sdk/src/types/block/output/token_id.rs
```diff
@@ -3,7 +3,7 @@
 
 use crate::types::block::output::FoundryId;
 
-impl_id!(pub TokenId, 38, "TODO.");
+impl_id!(pub TokenId, 38, "Unique identifiers of native tokens. The TokenId of native tokens controlled by a specific foundry is the same as the FoundryId.");
```
Thoralf-M:
```suggestion
impl_id!(pub TokenId, 38, "Unique identifiers of native tokens. The TokenId of native tokens minted by a specific foundry is the same as the FoundryId.");
```

sdk/src/wallet/error.rs
```diff
@@ -115,6 +115,10 @@ pub enum Error {
     #[cfg_attr(docsrs, doc(cfg(feature = "participation")))]
     #[error("voting error {0}")]
     Voting(String),
+    #[cfg(feature = "participation")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "participation")))]
+    #[error("invalid voting power")]
+    InvalidVotingPower,
```
Thoralf-M, line 121: Changelog :trollface:

Thoralf-M (approved):

abdulmth (approved):

sdk/src/wallet/account/operations/participation/voting_power.rs
```diff
@@ -58,10 +58,10 @@ where
             Some(current_output_data) => {
                 let output = current_output_data.output.as_basic();
 
-                // TODO checked addition
+                let new_amount = output.amount().checked_add(amount).ok_or(Error::InvalidVotingPower)?;
```
kwek20, line 61: Maybe more descriptive? like how much is missing or what we tried to add

thibault-martinez: Not much we can do, it can overflow so we'd have to store the addition in an u128, and for the subtraction it gets even weirder

thibault-martinez: And if we want to only store what we tried to add/sub, we'd need two variants? Otherwise it's confusing

Thoralf-M: I also think without value it's fine, if the code is not wrong, it shouldn't return an error at all
So users should never see this error and if they do, then they can't do anything about it and we have to fix

DaughterOfMars (approved):

PR #815, opened 2023-07-19
Changelog as release description

thibault-martinez:


.github/workflows/bindings-nodejs-publish.yml
```diff
@@ -246,6 +246,11 @@ jobs:
...
-      - name: Upload prebuild to GitHub release
-        run: npx prebuild --upload-all ${{ secrets.GITHUB_TOKEN }} --tag-prefix iota-sdk-nodejs-v
-        working-directory: bindings/nodejs
+      - name: Upload prebuild to Github release
+        uses: softprops/action-gh-release@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          body: https://github.com/iotaledger/iota-sdk/blob/develop/bindings/nodejs/CHANGELOG.md
+          files: bindings/nodejs/prebuilds/@iota/*
+          tag_name: iota-sdk-nodejs-v${npm pkg get version --workspaces=false | tr -d \"}
```
thibault-martinez: not fully sure about that

kwek20: works for me

kwek20 (dismissed):

.github/workflows/bindings-nodejs-publish.yml
Thoralf-M: This is good stuff, copied from my working code :trollface: 
At least the `npm pkg get version --workspaces=false | tr -d \"`
Don't we need double curly brackets ${{ }} in the workflow?

thibault-martinez: Yeah that's what I was referring to, I don't really know

Thoralf-M: Actually might need to change more https://github.com/softprops/action-gh-release/issues/360

.github/workflows/bindings-nodejs-publish.yml
```diff
@@ -224,6 +224,11 @@ jobs:
           security -v list-keychains -s signing.keychain
           security find-identity -vp codesigning
 
+      - name: Prepare Github release
+        id: prepare_release
+        run: |
+          echo "tag_name=iota-sdk-nodejs-v$(npm pkg get version --workspaces=false | tr -d \")" >> "$GITHUB_OUTPUT"
```
thibault-martinez, line 230: Not sure if we can touch the name L239, maybe we could reuse this. But don't know how it related to the npm releases.

Thoralf-M: It reads the version from the package.json so should be all good, but not sure if's worth it to reuse

Thoralf-M (approved):

qrayven (approved):

abdulmth (approved):

PR #823, opened 2023-07-19
Clean up lots of serde field impls and remove unnecessary derives

DaughterOfMars:
# Description of change

This PR removes a lot of pointless fields from serialized json, removed serde impls for types with Dtos, and generally tidies up.

## Type of change

### DIFF
--- a/sdk/CHANGELOG.md
+++ b/sdk/CHANGELOG.md
@@ -25,10 +25,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Migrated storage types field casing to uniform camelCase;
 - Replaced inappropriate serde impls from `AccountDetails` with fn impls for conversion;
+- `MqttPayload` now uses Dtos;
+- `NodeManagerBuilder` node fields are no longer wrapped in `Option`;
+- Prefix-hex string values replaced with boxed slices in Dtos;
+- `MilestoneEssence::new` now takes generic metadata for convenience;
+- `ParametersMilestoneOption::new` now accepts boxed slice;
 
 ### Removed
 
 - `ProtocolParametersDto`, `NetworkInfoDto`, `OutputMetadataDto` in favor of base types;
+- `serde` derives for types with explicit Dtos;
+- More fields that are considered empty are no longer serialized;
 
 ### Fixed
 
--- a/sdk/src/utils/serde.rs
+++ b/sdk/src/utils/serde.rs
@@ -54,24 +54,50 @@ pub mod option_string {
     }
 }
 
-pub mod option_prefix_hex_vec {
-    use alloc::{string::String, vec::Vec};
+pub mod prefix_hex_bytes {
+    use alloc::string::String;
 
+    use prefix_hex::{FromHexPrefixed, ToHexPrefixed};
     use serde::{de, Deserialize, Deserializer, Serializer};
 
-    pub fn serialize<S>(value: &Option<Vec<u8>>, serializer: S) -> Result<S::Ok, S::Error>
+    pub fn serialize<S, T>(value: &T, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
+        for<'a> &'a T: ToHexPrefixed,
+    {
+        serializer.serialize_str(&prefix_hex::encode(value))
+    }
+
+    pub fn deserialize<'de, D, T>(deserializer: D) -> Result<T, D::Error>
+    where
+        D: Deserializer<'de>,
+        T: FromHexPrefixed,
+    {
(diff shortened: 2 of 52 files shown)


Thoralf-M (dismissed):
Good job cleaning this all up

Alex6323 (dismissed):

thibault-martinez (changes requested):
.

sdk/CHANGELOG.md
```diff
@@ -25,10 +25,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Migrated storage types field casing to uniform camelCase;
 - Replaced inappropriate serde impls from `AccountDetails` with fn impls for conversion;
+- `MqttPayload` now uses Dtos;
+- `NodeManagerBuilder` node fields are no longer wrapped in `Option`;
+- Prefix-hex string values replaced with boxed slices in Dtos;
```
thibault-martinez, line 30: Why?

thibault-martinez: Less conversions I guess

thibault-martinez: Oh was that because Strings were never actually empty because of 0x?

sdk/src/utils/serde.rs
```diff
@@ -54,24 +54,50 @@ pub mod option_string {
...
     }
 
-    pub fn deserialize<'de, D>(deserializer: D) -> Result<Option<Vec<u8>>, D::Error>
+    pub fn deserialize<'de, D, T>(deserializer: D) -> Result<Option<T>, D::Error>
     where
         D: Deserializer<'de>,
+        T: FromHexPrefixed,
     {
         Option::<String>::deserialize(deserializer)?
             .map(|string| prefix_hex::decode(string).map_err(de::Error::custom))
```
thibault-martinez, line 103: Can we not also use super::deserialize here ?

sdk/src/wallet/migration/migrate_1.rs
```diff
@@ -37,7 +37,9 @@ impl Migration<crate::wallet::storage::Storage> for Migrate {
         }
 
         if let Some(mut wallet) = storage.get::<serde_json::Value>(WALLET_INDEXATION_KEY).await? {
-            migrate_client_options(&mut wallet["client_options"])?;
+            if let Some(client_options) = wallet.get_mut("client_options") {
```
thibault-martinez, line 40: Would this have crashed?

sdk/src/wallet/migration/migrate_4.rs
```diff
@@ -0,0 +1,260 @@
...
+    }
+    Ok(())
+}
+
+fn migrate_transaction(transaction: &mut serde_json::Value) -> Result<()> {
+    let transaction = transaction
+        .as_object_mut()
+        .ok_or(Error::Storage("malformatted transaction".to_owned()))?;
+    check_omitted_str("note", transaction, "");
+    check_omitted_str("blockId", transaction, "");
```
thibault-martinez: Shouldn't it be "0x"?

sdk/CHANGELOG.md
DaughterOfMars: Yes, and we would have to add the check for emptiness in our conversion code rather than in serde.

sdk/src/utils/serde.rs
DaughterOfMars: I tried but the problem is the two-stage process. Though I suppose I could use Option<Value> instead. Up to you.

sdk/src/wallet/migration/migrate_1.rs
DaughterOfMars: Yeah, I think so lol

sdk/src/wallet/migration/migrate_4.rs
thibault-martinez: Oh but it'll go in the null check, not the string check, I guess

sdk/src/utils/serde.rs
DaughterOfMars: Okay, it works but it's not much of an improvement. It's now a three-step process too.

thibault-martinez: I was just being curious if it was just forgotten, you do as you want.

sdk/src/wallet/migration/migrate_4.rs
DaughterOfMars: Actually it should be the other method for `Options` since blockId can't be empty. Changed it.

sdk/src/utils/serde.rs
DaughterOfMars: I think I'd rather leave it.

thibault-martinez (approved):

Thoralf-M (approved):

PR #828, opened 2023-07-20
Sign transaction returns a TransactionPayload

thibault-martinez:


bindings/nodejs/lib/client/client.ts
```diff
@@ -293,7 +293,7 @@ export class Client {
             },
         });
 
-        return parsePayload(JSON.parse(response).payload);
+        return JSON.parse(response).payload;
```
thibault-martinez: Is this correct?

kwek20: No, as TransactionPayload has fields that can have multiple types (Subclasses) it needs to be 
```
const parsed = JSON.parse(response) as Response<TransactionPayload>;
return plainToInstance(TransactionPayload, parsed.payload);
```

thibault-martinez: Ok, was already like that for this type somewhere else so was probably wrong

thibault-martinez: Changed

DaughterOfMars (dismissed):

Thoralf-M (changes requested):
Conflict

Thoralf-M (approved):

kwek20 (approved):

DaughterOfMars (approved):

PR #830, opened 2023-07-20
Added dependency to python package

anistark:
# Description of change

Currently, the python package in pypi doesn't install needed dependencies especially:
```toml
dacite = "^1.8.1"
pyhumps = "^3.8.0"
```
Hence added them to `setup.py` -> `install_requires`

Also, updated examples link which is broken when it appears in [pypi page](https://pypi.org/project/iota-sdk/1.0.0rc0/).

## Type of change

- Documentation Fix

## Change checklist


Thoralf-M (dismissed):

bindings/python/setup.py
```diff
@@ -41,8 +53,5 @@ def get_py_version_cfgs():
     ],
     include_package_data=True,
     zip_safe=False,
-    install_requires=[
-        '<pyhumps> >= <3.8.0>',
-        '<dacite>'
-    ]
+    install_requires=load_requirements("requirements-dev.txt")
```
Thoralf-M: This doesn't work for pypi? At least in the released packages I don't see requirements-dev.txt included

bindings/python/setup.py
```diff
@@ -41,4 +53,5 @@ def get_py_version_cfgs():
     ],
     include_package_data=True,
     zip_safe=False,
+    install_requires=load_requirements("requirements.txt")
```
Thoralf-M: You know if this file would get published? Because the requirements.dev.txt doesn't

Maybe we should do it as described here https://github.com/PyO3/maturin#python-metadata?

anistark: Ideally it should build and paste in the array like format. But since right now there's only 2, can do it directly.
Updating.

Thoralf-M (dismissed):

bindings/python/README.md
```diff
@@ -4,14 +4,16 @@ Python binding to the [iota-sdk library](/README.md).
...
+  - [Table of contents](#table-of-contents)
+  - [Requirements](#requirements)
+  - [Getting Started](#getting-started)
     - [Install the IOTA SDK](#install-the-iota-sdk)
-- [Client](#client-usage)
-- [Wallet](#wallet-usage)
-- [Examples](#examples)
-- [API Reference](#api-reference)
-- [Learn More](#learn-more)
+  - [Client Usage](#client-usage)
```
thibault-martinez: Pretty sure we had Usage in the past and removed it because we didn't like it

thibault-martinez: I think it was @Alexandcoats that removed it

thibault-martinez (approved):

bindings/python/README.md
anistark: My bad. Markdown auto docs or some plugin added it. Removed it now.

Thoralf-M (approved):

PR #832, opened 2023-07-20
Make public enums `non_exhaustive` where appropriate

DaughterOfMars:
# Description of change

This PR will help avoid some breaking changes when adding variants to enums. This still does not guarantee we don't break functionality, but lib user code will still compile. I avoided adding the tags to Error enums, as I feel that changes to those should always be breaking, but if you believe otherwise we can discuss it.

## Type of change


sdk/src/types/block/payload/transaction/essence/mod.rs
```diff
@@ -33,6 +33,18 @@ impl TransactionEssence {
...
     }
+
+    /// Checks whether the essence is a [`RegularTransactionEssence`].
+    pub fn is_regular(&self) -> bool {
+        matches!(self, Self::Regular(_))
+    }
+
+    /// Gets the essence as an actual [`RegularTransactionEssence`].
+    /// PANIC: do not call on a non-regular essence.
+    pub fn as_regular(&self) -> &RegularTransactionEssence {
```
thibault-martinez, line 44: changelog

DaughterOfMars: done

Thoralf-M (dismissed):

thibault-martinez (approved):

Thoralf-M (approved):

PR #833, opened 2023-07-20
deny unreachable pub

### LINKED ISSUE #65: Review of all visibility attributes

DaughterOfMars:
# Description of change

This PR enables the deny tag for `unreachable_pub`, and fixes the visibility.

Closes #65

## Type of change


thibault-martinez (changes requested):
No point going too much into details but maybe put something generic about visibility in the changelog

bindings/core/src/lib.rs
```diff
@@ -50,9 +50,31 @@ pub struct WalletOptions {
 }
 
 impl WalletOptions {
+    pub fn with_storage_path(mut self, storage_path: impl Into<Option<String>>) -> Self {
```
thibault-martinez, line 53: Were these meant to change the visibility of fields ?

bindings/core/tests/secrets_debug.rs
```diff
@@ -43,12 +43,7 @@ fn method_interface_secrets_debug() {
     let response = Response::GeneratedMnemonic("mnemonic".to_string());
     assert_eq!(format!("{:?}", response), "GeneratedMnemonic(<omitted>)");
 
-    let wallet_options = WalletOptions {
-        storage_path: None,
-        client_options: None,
-        coin_type: None,
-        secret_manager: Some(SecretManagerDto::Placeholder),
-    };
+    let wallet_options = WalletOptions::default().with_secret_manager(SecretManagerDto::Placeholder);
```
thibault-martinez, line 46: nice

sdk/src/wallet/mod.rs
```diff
@@ -9,7 +9,7 @@ pub mod account;
 /// The core module.
 pub mod core;
 #[cfg(any(feature = "stronghold", feature = "storage"))]
-pub(crate) mod migration;
+pub mod migration;
```
thibault-martinez, line 12: Why is migration public?

sdk/tests/client/node_api/mqtt.rs
```diff
@@ -53,6 +53,11 @@ async fn test_mqtt() {
             } => {
                 panic!("mqtt disconnected");
             }
+            _ = async {
+                tokio::time::sleep(instant::Duration::from_secs(120)).await;
+            } => {
+                panic!("timed out after 120s");
```
thibault-martinez: We keeping this ?

bindings/core/src/lib.rs
DaughterOfMars: no

sdk/src/wallet/mod.rs
DaughterOfMars: some types are, like the version struct

sdk/tests/client/node_api/mqtt.rs
DaughterOfMars: not sure. could maybe use nextest to timeout the whole test instead

DaughterOfMars:
> No point going too much into details but maybe put something generic about visibility in the changelog

Since all the types were already unreachable, it's not actually a breaking change. Still want it?

bindings/core/src/lib.rs
thibault-martinez: Why can't they be private if we have that ? We only use `WalletOptions` to build a `Wallet` in this module

sdk/src/client/builder.rs
```diff
@@ -351,6 +351,33 @@ pub struct NetworkInfo {
     pub latest_milestone_timestamp: Option<u32>,
 }
 
+impl NetworkInfo {
+    pub fn with_protocol_parameters(mut self, protocol_parameters: impl Into<ProtocolParameters>) -> Self {
+        self.protocol_parameters = protocol_parameters.into();
+        self
+    }
+
+    pub fn with_local_pow(mut self, local_pow: bool) -> Self {
```
thibault-martinez, line 360: There are instances in the code base where we set fields directly, should they start using these setters?

sdk/src/wallet/mod.rs
thibault-martinez: And out of curiosity why is the version struct public? Seems pretty self contained to me

bindings/core/src/lib.rs
DaughterOfMars: It's not about whether something could be private. Private fields should only be used when you cannot allow people to build your type from fields (i.e. OutputBuilders), otherwise you are limiting access for no good reason and making the API difficult to use.

sdk/src/wallet/mod.rs
DaughterOfMars: Just seems like something users may want to be able to deserialize themselves.

sdk/src/client/builder.rs
DaughterOfMars: Doesn't really matter

bindings/core/src/lib.rs
thibault-martinez: I don't think it makes the API difficult to use, you actually made it easier by adding the into setters so I don't see a good reason to keep the fields public. But as you want, I'll resolve.

sdk/src/client/builder.rs
thibault-martinez: Same answer, not sure why we're adding setters if we keep the fields public, they are basically useless.

sdk/src/wallet/mod.rs
thibault-martinez: IMO users shouldn't even be aware of migrations but alright.

sdk/tests/client/node_api/mqtt.rs
thibault-martinez: As you want

thibault-martinez (dismissed):

thibault-martinez (dismissed):

sdk/src/wallet/mod.rs
kwek20: yea i agree, all migration related code should be internal only.
as a user id never want to touch it :D

thibault-martinez (approved):

Thoralf-M (approved):

PR #834, opened 2023-07-20
Various tests

thibault-martinez:


qrayven (approved):

sdk/tests/types/parents.rs
```diff
@@ -10,6 +11,13 @@ use iota_sdk::types::block::{
 };
 use packable::{bounded::TryIntoBoundedU8Error, error::UnpackError, prefix::VecPrefix, PackableExt};
 
+#[test]
+fn len() {
+    for i in 1..=8 {
+        assert_eq!(Parents::from_vec(rand_block_ids(i)).unwrap().len(), i);
```
Thoralf-M, line 17: If `rand_block_ids()` ever returns the same id twice this would fail :trollface:

sdk/tests/types/unlock/alias.rs
```diff
@@ -0,0 +1,57 @@
...
+#[test]
+fn new_valid_max_index() {
+    assert_eq!(AliasUnlock::new(126).unwrap().index(), 126);
+}
+
+#[test]
+fn new_invalid_more_than_max_index() {
+    assert!(matches!(
+        AliasUnlock::new(128),
+        Err(Error::InvalidAliasIndex(InvalidBoundedU16(128)))
```
Thoralf-M, line 26: Use 127 already?
```suggestion
        AliasUnlock::new(127),
        Err(Error::InvalidAliasIndex(InvalidBoundedU16(127)))
```

Thoralf-M: Actually the UNLOCK_INDEX_RANGE is wrong?

sdk/tests/types/parents.rs
thibault-martinez: You have more chances to be hit by a meteorite, twice

sdk/tests/types/unlock/alias.rs
thibault-martinez: Hmm no ? 127 is valid, 128 is not?

Thoralf-M: How can 127 be valid if 127 is the highest possible unlock? It would be a self reference

thibault-martinez: 127 is allowed by the TIP, but I agree with you, it would fail during validation of `Unlocks`, but not here 🤷🏻

Thoralf-M: Then the TIP should also get updated? :carlos:

thibault-martinez: Maybe yeah, it's just a bit weird but in the end it's all validated correctly

Thoralf-M: Opened a PR to update the TIP https://github.com/iotaledger/tips/pull/134

thibault-martinez: Can we still merde this until they accept it? This is going to break more tests anyway so I'd do another PR

Thoralf-M (approved):

PR #838, opened 2023-07-21
Remove TODOs in participation 

Thoralf-M:
# Description of change

Remove/address TODOs in participation

#401 

## Type of change

- Documentation Fix

### DIFF
--- a/sdk/src/wallet/account/operations/participation/mod.rs
+++ b/sdk/src/wallet/account/operations/participation/mod.rs
@@ -284,7 +284,6 @@ where
             .get_participation_events(account_index)
             .await?;
 
-        // TODO try to remove this clone
         for participation in participations.participations.clone().iter() {
             if let Some(event_with_nodes) = events.get(&participation.event_id) {
                 if event_with_nodes.data.milestone_index_end() < &latest_milestone_index {
--- a/sdk/src/wallet/account/operations/participation/voting.rs
+++ b/sdk/src/wallet/account/operations/participation/voting.rs
@@ -64,8 +64,6 @@ where
             }
         }
 
-        // TODO check if answers match the questions ?
-
         let voting_output = self
             .get_voting_output()
             .await?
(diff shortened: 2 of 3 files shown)


sdk/src/wallet/account/operations/participation/voting.rs
```diff
@@ -157,11 +153,9 @@ where
 
                 let length_before = participations.participations.len();
 
-                // TODO use remove return when merged
```
Thoralf-M, line 160: Remove doesn't return anything, so can't do that

sdk/src/wallet/account/operations/participation/voting.rs
```diff
@@ -64,8 +64,6 @@ where
             }
         }
 
-        // TODO check if answers match the questions ?
```
Thoralf-M, line 67: This wouldn't allow to participate without registering an event before, don't think it's worth it

sdk/src/wallet/account/operations/participation/mod.rs
```diff
@@ -284,7 +284,6 @@ where
             .get_participation_events(account_index)
             .await?;
 
-        // TODO try to remove this clone
```
Thoralf-M, line 287: I tried and couldn't find a better way since it contains async code we can't use retain

sdk/src/wallet/account/operations/participation/voting_power.rs
```diff
@@ -118,7 +118,6 @@ where
             .ok_or_else(|| crate::wallet::Error::Voting("No unspent voting output found".to_string()))?;
         let output = current_output_data.output.as_basic();
 
-        // TODO what is amount > output.amount() ?
```
Thoralf-M, line 121: This is now handled `.checked_sub(amount).ok_or(Error::InvalidVotingPower)?;`

sdk/src/wallet/account/operations/participation/voting_power.rs
```diff
@@ -169,7 +168,6 @@ where
 
             participation_bytes
         } else {
-            // TODO participation bytes are incorrect, should we really just ignore?
```
Thoralf-M, line 172: I would ignore, otherwise we could get in a state where users can't do anything with their wallet anymore

sdk/src/wallet/account/operations/participation/mod.rs
thibault-martinez: Yeah I tried a lot as well and found nothing satisfying

thibault-martinez (approved):

qrayven (approved):

PR #839, opened 2023-07-21
Constructor to create `UTXOInput` from `OutputId`

### LINKED ISSUE #760: Reorganize `Utils` functions
### Move util functions to concrete types

```
const nftAddress: Address = new NftAddress(Utils.nftIdToBech32(nftId, networkName));
```

Consider making the `nftIdToBech32` function internal to `NftAddress`, e.g.:

```
const nftAddress: Address = NftAddress.fromNftId(nftId, networkName));
```

- The same applies to
    - `Utils.aliasIdToBech32`.
    - `Utils.computeTokenId`. It already returns a `TokenId` so is just a constructor for `TokenId` without living on the type itself.
    - `Utils.computeOutputId`

### Conversion function from Output ID to UTXOInput

Add a constructor for `UTXOInput` that has the functionality of `TransactionHelper.inputFromOutputId` from iota.js.

### Move top-level functions to Utils namespace

The `utf8ToBytes`, `utf8ToHex` and `hexToUtf8` functions are directly in the top-level namespace of the sdk. Consider moving thos "util" functions to the `Utils` namespace.

### Nits

- `Utils.computeNftId` returns a `string` instead of the type alias `NftId`.

## Motivation

The address util functions take a `string` and return one and it's not immediately obvious what that string represents, especially for new users, I imagine. So having them directly on the stronger-typed classes like `AliasAddress` seems both clearer and more convenient.
Constructors should live on the classes that they construct rather than in `Utils`. A user checking the API docs for how to construct a certain type would check that type's docs and rarely consider that a constructor is in `Utils`.

The constructor `new UTXOInput` requires a transaction ID and output index, but it's not easily possible to extract those from some `outputID`, because that is just a string. There should be a convenient way to convert.

No.

abdulmth:
# Description of change

Adds the ability to create a `UTXOInput` from an `OutputId`.

fixes #760

## Type of change

### DIFF
--- a/bindings/core/src/method/utils.rs
+++ b/bindings/core/src/method/utils.rs
@@ -154,5 +154,5 @@ pub enum UtilsMethod {
     },
     /// Creates a UTXOInput from outputId.
     #[serde(rename_all = "camelCase")]
-    OutputIdToUTXOInput { output_id: String },
+    OutputIdToUtxoInput { output_id: String },
 }
--- a/bindings/core/src/method/utils.rs
+++ b/bindings/core/src/method/utils.rs
@@ -152,4 +152,7 @@ pub enum UtilsMethod {
         signature: String,
         message: String,
     },
+    /// Creates a UTXOInput from outputId.
+    #[serde(rename_all = "camelCase")]
+    OutputIdToUTXOInput { output_id: String },
 }
--- a/bindings/nodejs/CHANGELOG.md
+++ b/bindings/nodejs/CHANGELOG.md
@@ -19,6 +19,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 1.0.0-rc.4 - 2023-MM-DD
+
+### Added
+- `fromOutputId` constructor for `UTXOInput`;
+
 ## 1.0.0-rc.3 - 2023-07-21
 
 ### Added
@@ -26,7 +31,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `TransactionProgressType` export;
 - Optional `CreateAccountPayload::addresses` field;
 - Optional `FilterOptions::{aliasIds, foundryIds, nftIds}` fields;
-- `fromOutputId` constructor for `UTXOInput`;
 
 ### Changed
 
--- a/bindings/nodejs/CHANGELOG.md
+++ b/bindings/nodejs/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `TransactionProgressType` export;
 - Optional `CreateAccountPayload::addresses` field;
 - Optional `FilterOptions::{aliasIds, foundryIds, nftIds}` fields;
+- `fromOutputId` constructor for `UTXOInput`;
 
 ### Changed
 
--- a/bindings/nodejs/lib/types/block/input/input.ts
+++ b/bindings/nodejs/lib/types/block/input/input.ts
@@ -1,6 +1,7 @@
 // Copyright 2023 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+import { callUtilsMethod } from '../../../bindings';
 import { HexEncodedString } from '../../utils';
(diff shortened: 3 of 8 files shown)


bindings/nodejs/lib/types/block/input/input.ts
```diff
@@ -62,6 +63,28 @@ class UTXOInput extends Input {
         this.transactionId = transactionId;
         this.transactionOutputIndex = transactionOutputIndex;
     }
+
+    /**
+     * Creates a `UTXOInput` from an output id.
+     */
+    static fromOutputId(outputId: OutputId): UTXOInput {
```
abdulmth, line 71: Maybe we can implement this in Rust and expose it? I found this easier, although things like `TRANSACTION_ID_LENGTH` are not used in node so far.

thibault-martinez (changes requested):
Changelog

bindings/nodejs/lib/types/block/input/input.ts
DaughterOfMars: Yes, this should be implemented in rust

abdulmth: So we need to Create a `UTXOInput` in Rust and convert it to back to JS? or create a utility function? I'm not sure how that can be done in an idiomatic way.

thibault-martinez: That would have to go through the message interface

DaughterOfMars (dismissed):
Much better

bindings/nodejs/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `TransactionProgressType` export;
 - Optional `CreateAccountPayload::addresses` field;
 - Optional `FilterOptions::{aliasIds, foundryIds, nftIds}` fields;
+- `fromOutputId` constructor for `UTXOInput`;
```
thibault-martinez: In a new section, RC3 has been released

DaughterOfMars (dismissed):

bindings/core/src/method/utils.rs
```diff
@@ -152,4 +152,7 @@ pub enum UtilsMethod {
         signature: String,
         message: String,
     },
+    /// Creates a UTXOInput from outputId.
+    #[serde(rename_all = "camelCase")]
+    OutputIdToUTXOInput { output_id: String },
```
Thoralf-M: Should this be `OutputIdToUtxoInput`?
I think everywhere else we named it with `txo` small
`GetUtxoChangesById`, `GetUtxoChangesByIndex`, `UtxoInputDto`

abdulmth: I changed it, the TS class name is `UTXOInput`, this is why I used it this way.

DaughterOfMars: (Please don't resolve comments other people started)

DaughterOfMars (approved):

Thoralf-M (approved):

bindings/core/src/method/utils.rs
```diff
@@ -152,4 +152,7 @@ pub enum UtilsMethod {
         signature: String,
         message: String,
     },
+    /// Creates a UTXOInput from outputId.
+    #[serde(rename_all = "camelCase")]
+    OutputIdToUtxoInput { output_id: String },
```
thibault-martinez, line 157: Param should/can be an OutputId?

DaughterOfMars: Good point

bindings/nodejs/CHANGELOG.md
```diff
@@ -19,6 +19,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 1.0.0-rc.4 - 2023-MM-DD
+
+### Added
+- `fromOutputId` constructor for `UTXOInput`;
```
thibault-martinez, line 25:
```suggestion

- `fromOutputId` constructor for `UTXOInput`;
```

thibault-martinez:
Closing in favour of https://github.com/iotaledger/iota-sdk/pull/847

PR #841, opened 2023-07-21
Create nextest CI profile

DaughterOfMars:
# Description of change

Creates a config file with a CI profile for `nextest`. This adds two flags that weren't used before:

- `failure-output = "immediate-final"` (this will print failed test output at the bottom of the logs in addition to above, for convenience)
- `slow-timeout = { period = "60s", terminate-after = 2 }`

## Type of change

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,8 @@
+[profile.ci]
+# Print out output for failing tests as soon as they fail, and also at the end
+# of the run (for easy scrollability).
+failure-output = "immediate-final"
+fail-fast = false
+retries = 2
+test-threads = "num-cpus"
+slow-timeout = { period = "60s", terminate-after = 2 }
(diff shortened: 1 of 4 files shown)


.config/nextest.toml
```diff
@@ -0,0 +1,8 @@
+[profile.ci]
```
kwek20, line 1: i wouldnt mind just making it the default like this, we would also not need to mention it anymore.
Or do you think its better to specify for readability?

DaughterOfMars: 🤷‍♀️ I can see the reasons for both cases.

Thoralf-M (approved):

.config/nextest.toml
kwek20: sure, lets leave it

kwek20 (approved):

thibault-martinez (approved):

PR #847, opened 2023-07-21
Alternate Output Id to UTXO Input

### LINKED ISSUE #760: Reorganize `Utils` functions

DaughterOfMars:
See #839 
Fixes #760

### DIFF
--- a/bindings/nodejs/lib/types/utils/bridge/utils.ts
+++ b/bindings/nodejs/lib/types/utils/bridge/utils.ts
@@ -195,4 +195,3 @@ export interface __OutputIdToUTXOInput__ {
         outputId: string;
     };
 }
-
--- a/bindings/nodejs/lib/types/utils/bridge/utils.ts
+++ b/bindings/nodejs/lib/types/utils/bridge/utils.ts
@@ -188,3 +188,11 @@ export type __FaucetMethod__ = {
         address: string;
     };
 };
+
+export interface __OutputIdToUTXOInput__ {
+    name: 'outputIdToUTXOInput';
+    data: {
+        outputId: string;
+    };
+}
+
--- a/bindings/nodejs/lib/types/utils/bridge/utils.ts
+++ b/bindings/nodejs/lib/types/utils/bridge/utils.ts
@@ -189,8 +189,8 @@ export type __FaucetMethod__ = {
     };
 };
 
-export interface __OutputIdToUTXOInput__ {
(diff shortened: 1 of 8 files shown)


Thoralf-M (approved):

bindings/nodejs/lib/types/utils/bridge/utils.ts
```diff
@@ -188,3 +188,10 @@ export type __FaucetMethod__ = {
         address: string;
     };
 };
+
+export interface __OutputIdToUtxoInput__ {
+    name: 'outputIdToUtxoInput';
+    data: {
+        outputId: string;
```
thibault-martinez, line 195: Does this need to be changed as well?

Thoralf-M: There is an OutputId type but that's also just a string, so doesn't really make a difference

thibault-martinez (approved):

PR #893, opened 2023-07-24
Clean up by privatizing more DTOs

DaughterOfMars:
# Description of change

This PR makes a lot of Dtos private and moves their serde impls to the base type via the private Dto definition. This is also possible for many of the types that impl `TryFromDto` using the non-validating conversion, but I left them as-is for now. I also messed around with some types and added more serde tags to reduce our boilerplate code.

## Type of change


sdk/src/types/block/signature/ed25519.rs
```diff
@@ -165,4 +165,27 @@ pub(crate) mod dto {
             )
         }
     }
+
+    impl<'de> Deserialize<'de> for Ed25519Signature {
+        fn deserialize<D: serde::Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+            let dto = Ed25519SignatureDto::deserialize(d)?;
+            if dto.kind != Self::KIND {
+                return Err(serde::de::Error::custom(format!(
+                    "invalid account unlock type: expected {}, found {}",
```
Thoralf-M:
```suggestion
                    "invalid signature type: expected {}, found {}",
```

sdk/src/types/block/output/unlock_condition/mod.rs
```diff
@@ -56,12 +56,12 @@ pub enum UnlockCondition {
...
     fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
-        self.kind().partial_cmp(&other.kind())
+        Some(self.cmp(other))
     }
 }
 impl Ord for UnlockCondition {
     fn cmp(&self, other: &Self) -> core::cmp::Ordering {
-        self.partial_cmp(other).unwrap()
+        self.kind().cmp(&other.kind())
     }
```
Thoralf-M: Why is it changing like this?

DaughterOfMars: clippy

sdk/src/types/block/output/unlock_condition/mod.rs
```diff
@@ -498,162 +498,41 @@ mod test {
...
 
-    #[derive(Clone, Debug, Eq, PartialEq, From)]
+    #[derive(Clone, Debug, Eq, PartialEq, From, Serialize, Deserialize)]
+    #[serde(untagged)]
     pub enum UnlockConditionDto {
         /// An address unlock condition.
-        Address(AddressUnlockConditionDto),
+        Address(AddressUnlockCondition),
         /// A storage deposit return unlock condition.
         StorageDepositReturn(StorageDepositReturnUnlockConditionDto),
```
Alex6323, line 513: does this have to remain the dto type?

sdk/src/types/block/signature/ed25519.rs
```diff
@@ -17,7 +17,7 @@ use packable::{
 use crate::types::block::{address::Ed25519Address, Error};
 
 /// An Ed25519 signature.
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
```
Alex6323, line 20: Probably moving this type - if possible - is cheaper?

sdk/src/types/block/unlock/mod.rs
```diff
@@ -34,10 +34,12 @@ pub(crate) type UnlockIndex = BoundedU16<{ *UNLOCK_INDEX_RANGE.start() }, { *UNL
 #[derive(Clone, Eq, PartialEq, Hash, From, Packable)]
 #[packable(unpack_error = Error)]
 #[packable(tag_type = u8, with_error = Error::InvalidUnlockKind)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize), serde(untagged))]
 pub enum Unlock {
     /// A signature unlock.
     #[packable(tag = SignatureUnlock::KIND)]
-    Signature(SignatureUnlock),
+    #[from(ignore)]
```
Alex6323, line 41: Can you use `#[from(forward)]` here?

sdk/src/types/block/output/unlock_condition/mod.rs
DaughterOfMars: Yeah, unfortunately. The reason is that the conversion from `Dto` -> Base type can be validated with token supply.

sdk/src/types/block/signature/ed25519.rs
DaughterOfMars: What do you mean?

sdk/src/types/block/unlock/mod.rs
DaughterOfMars: Probably! I'll try it

DaughterOfMars: Turns out, I cannot. It causes the other impls to break.

sdk/src/types/block/signature/ed25519.rs
Alex6323: This type is about 100 bytes in length, so I was wondering whether we should make it `Copy`.

DaughterOfMars: I'm not particularly concerned about it 🤷‍♀️

Alex6323: Okay. Any more opinions from the team?

thibault-martinez: It probably makes 0 difference

Thoralf-M (dismissed):

sdk/tests/types/input/utxo.rs
```diff
@@ -78,60 +72,6 @@ fn from_str() {
...
-    assert_eq!(utxo_dto.transaction_output_index, output_id.index());
-
-    let input = Input::from(utxo_input);
-    let dto = InputDto::from(&input);
-
-    assert_eq!(dto, InputDto::Utxo(utxo_dto));
-}
-
-#[test]
-fn dto_roundtrip() {
```
thibault-martinez, line 98: Can't we convert all these tests to a serde roundrip? The DTOs are not public anymore but we should still test their implementation ?

DaughterOfMars: I'd rather just move these to the unit

sdk/src/types/block/output/feature/staking.rs
```diff
@@ -50,23 +49,70 @@ impl StakingFeature {
...
+                value.end_epoch,
+            )
+        }
+    }
+
+    impl<'de> Deserialize<'de> for StakingFeature {
+        fn deserialize<D: serde::Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+            let dto = StakingFeatureDto::deserialize(d)?;
+            if dto.kind != Self::KIND {
+                return Err(serde::de::Error::custom(format!(
```
Thoralf-M: Check `no_std` compatability
cannot find macro `format` in this scope

DaughterOfMars: dangit

sdk/tests/types/input/utxo.rs
DaughterOfMars: changed my mind

Thoralf-M (dismissed):

sdk/src/types/block/output/basic.rs
```diff
@@ -373,26 +369,27 @@ pub(crate) mod dto {
         #[serde(rename = "type")]
         pub kind: u8,
         // Amount of IOTA tokens held by the output.
-        pub amount: String,
+        #[serde(with = "string")]
```
kwek20, line 372: lets stick with 1 or the other way to import this. i prefer the full qualifier

sdk/src/types/block/output/delegation.rs
```diff
@@ -413,29 +413,34 @@ fn verify_unlock_conditions<const VERIFY: bool>(unlock_conditions: &UnlockCondit
...
     };
 
     #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
     #[serde(rename_all = "camelCase")]
     pub struct DelegationOutputDto {
         #[serde(rename = "type")]
         pub kind: u8,
-        pub amount: String,
-        pub delegated_amount: String,
+        #[serde(with = "string")]
```
kwek20, line 435: so a find all on this line i guess :)

sdk/src/types/block/output/feature/issuer.rs
```diff
@@ -26,15 +26,55 @@ impl IssuerFeature {
...
+        }
+    }
+
+    impl From<IssuerFeatureDto> for IssuerFeature {
+        fn from(value: IssuerFeatureDto) -> Self {
+            Self(value.address)
+        }
+    }
+
+    impl<'de> Deserialize<'de> for IssuerFeature {
```
kwek20: this happens a few times, can we derive the code? or prefer to keep it manual for each for readability? 
(For example that `serde_derive_impl!()`

sdk/src/types/block/payload/transaction/essence/regular.rs
```diff
@@ -344,11 +345,11 @@ pub(crate) mod dto {
                 kind: RegularTransactionEssence::KIND,
                 network_id: value.network_id().to_string(),
                 creation_time: value.creation_time(),
-                inputs: value.inputs().iter().map(Into::into).collect::<Vec<_>>(),
+                inputs: value.inputs().to_vec(),
                 inputs_commitment: value.inputs_commitment().to_string(),
                 outputs: value.outputs().iter().map(Into::into).collect::<Vec<_>>(),
                 payload: match value.payload() {
-                    Some(Payload::TaggedData(i)) => Some(PayloadDto::TaggedData(Box::new(i.as_ref().into()))),
+                    Some(p @ Payload::TaggedData(_)) => Some(p.into()),
```
kwek20, line 352: huh this is new to me! Very handy

sdk/src/types/block/signature/ed25519.rs
kwek20: Eh move is always better, but thats a super optimizing thing. make a ticket for it id say :)

sdk/src/types/block/signature/ed25519.rs
```diff
@@ -165,4 +165,27 @@ pub(crate) mod dto {
             )
         }
     }
+
+    impl<'de> Deserialize<'de> for Ed25519Signature {
+        fn deserialize<D: serde::Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+            let dto = Ed25519SignatureDto::deserialize(d)?;
+            if dto.kind != Self::KIND {
+                return Err(serde::de::Error::custom(format!(
+                    "invalid signature type: expected {}, found {}",
```
kwek20: all other errors like htis mention thespecific type (i.e.: Invalid Ed25519 signature type)

sdk/src/types/block/signature/mod.rs
```diff
@@ -15,19 +13,14 @@ use crate::types::block::Error;
...
 ///
 /// RFC: <https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md#signature-unlock-block>
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, packable::Packable)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, packable::Packable, From)]
 #[packable(unpack_error = Error)]
 #[packable(tag_type = u8, with_error = Error::InvalidSignatureKind)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize), serde(untagged))]
 pub enum Signature {
     /// An Ed25519 signature.
     #[packable(tag = Ed25519Signature::KIND)]
```
kwek20, line 22: so we do untagged but then also set a kind? Were sure this is correct?

kwek20: ah i guess we want the number, a bit annoying but i see

sdk/src/types/block/output/basic.rs
DaughterOfMars: I prefer an import

sdk/src/types/block/output/feature/issuer.rs
DaughterOfMars: I suppose I could

sdk/tests/wallet/events.rs
```diff
@@ -24,7 +24,7 @@ const ED25519_ADDRESS: &str = "0x52fdfc072182654f163f5f0f9a621d729566c74d10037c4
 const TRANSACTION_ID: &str = "0x52fdfc072182654f163f5f0f9a621d729566c74d10037c4d7bbb0407d1e2c649";
 
 fn assert_serde_eq(event_0: WalletEvent) {
-    let json = serde_json::to_string(&event_0).unwrap();
+    let json = serde_json::to_string_pretty(&event_0).unwrap();
```
kwek20: whats up with this? :) I like pretty too but is it needed

sdk/src/types/block/output/basic.rs
kwek20: Then update the one below this line please, it has a full import

Thoralf-M (approved):

kwek20 (approved):

qrayven (approved):

abdulmth (approved):

PR #933, opened 2023-07-26
Fix `BigInt` serialization in Wasm.

### LINKED ISSUE #924: [Bug]: TypeError: Do not know how to serialize a BigInt
### Issue description

Building an output with a native token throws an error: `TypeError: Do not know how to serialize a BigInt`

### Version

Wasm bindings built from `6a856129bfd57a71a3567f05d7811b2b9cc30a43`.

### Expected behaviour

The example succeeds.

### Actual behaviour

The example fails.

Yes

Run this example to reproduce:

```typescript
const client = new Client({
        primaryNode: API_ENDPOINT,
        localPow: true,
    });

    const address = Utils.parseBech32Address("tst1qzkly6hllhscsaqy2le6f4dfymanx6nze2anf2qhlsy4m2zc5rrs5sy83ms")
    const tokenId: string = Utils.computeTokenId("0xf000000000000000000000000000000000000000000000000000000000000000", 1, TokenSchemeType.Simple);

    client.buildBasicOutput({
      unlockConditions: [new AddressUnlockCondition(address)],
      nativeTokens: [
        
          {
              id: tokenId,
              amount: BigInt(500_000),
          },
      
      ]
    })
```

### Errors

```
~/git/iota-sdk/bindings/wasm/out/lib/client/client-method-handler.ts:31
            JSON.stringify(method),
                 ^
TypeError: Do not know how to serialize a BigInt
    at JSON.stringify (<anonymous>)
    at ClientMethodHandler.callMethod (~/git/iota-sdk/bindings/wasm/out/lib/client/client-method-handler.ts:31:18)
    at Client.buildBasicOutput (~/git/iota-sdk/bindings/wasm/out/lib/client/client.ts:933:51)
```

### Duplicate declaration

abdulmth:
# Description of change

Fixes serialization error in Wasm since `bindings.ts` was modified in node.js but not in Wasm.

Also maybe worth noting that we patch a built-in object to allow the serialization, although it's for `BigInt` "[not explicitly discouraged](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json)"  it might conflict with user patching `BigInt` too (I'm not sure how common is that, probably pretty rare), the solution could be to convert `bigint` values manually to a hex string for the native token amounts everywhere it's used. 

### Note
The issue can be resolved on the current `1.0.1` release by the user manually running: 

```
(BigInt.prototype as any).toJSON = function() {
    return '0x' + this.toString(16);
};
```

fixes #924

## Type of change

### DIFF
--- a/bindings/nodejs/lib/bindings.ts
+++ b/bindings/nodejs/lib/bindings.ts
@@ -77,7 +77,7 @@ const listenWalletAsync = (
 ): Promise<void> => {
     listenWallet(
         eventTypes,
-        function (err: any, data: string) {
+        function(err: any, data: string) {
             const parsed = JSON.parse(data);
             callback(err, new Event(parsed.accountIndex, parsed.event));
         },
@@ -99,15 +99,6 @@ const callWalletMethodAsync = (
             }
         });
     });
-
-// Allow bigint to be serialized as hex string.
-//
-// Note:
-// Serializing `bigint` to a different format, e.g. to decimal number string
-// must be done manually.
-(BigInt.prototype as any).toJSON = function () {
-    return bigIntToHex(this);
-};
 
 export {
     initLogger,
--- a/bindings/nodejs/lib/bindings.ts
+++ b/bindings/nodejs/lib/bindings.ts
@@ -77,7 +77,7 @@ const listenWalletAsync = (
 ): Promise<void> => {
     listenWallet(
         eventTypes,
-        function(err: any, data: string) {
+        function (err: any, data: string) {
             const parsed = JSON.parse(data);
             callback(err, new Event(parsed.accountIndex, parsed.event));
         },
--- a/bindings/nodejs/lib/bindings.ts
+++ b/bindings/nodejs/lib/bindings.ts
@@ -4,7 +4,7 @@
 import type { WalletEventType } from './types/wallet';
 import { Event } from './types/wallet';
 import type { WalletMethodHandler } from './wallet/wallet-method-handler';
-import { bigIntToHex, __UtilsMethods__ } from './types/utils';
+import { __UtilsMethods__ } from './types/utils';
 import type { SecretManagerMethodHandler } from './secret_manager/secret-manager-method-handler';
 import type { ClientMethodHandler } from './client/client-method-handler';
 
--- a/bindings/wasm/CHANGELOG.md
+++ b/bindings/wasm/CHANGELOG.md
@@ -19,6 +19,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 1.0.2 - 2023-MM-DD
+
+Same changes as https://github.com/iotaledger/iota-sdk/blob/develop/bindings/nodejs/CHANGELOG.md.
+
+### Fixed
+
+- Runtime serializtion error of `BigInt` when working with native token amounts;
+
 ## 1.0.1 - 2023-07-25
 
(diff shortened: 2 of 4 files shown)


Thoralf-M (dismissed):

bindings/wasm/lib/bindings.ts
```diff
@@ -20,6 +20,15 @@ const callUtilsMethod = (method: __UtilsMethods__): any => {
     }
 };
 
+// Allow bigint to be serialized as hex string.
+//
+// Note:
+// Serializing `bigint` to a different format, e.g. to decimal number string
+// must be done manually.
```
thibault-martinez: Are there not currently places where this would then have to be changed?

thibault-martinez: Or maybe another question, this isn't inherited from the nodejs bindings ?

abdulmth: Yes, only `bindings.ts` is different between Wasm and Node.js. In other places it's inherited from Node.js

thibault-martinez: So if this was somewhere else than in `bindings.ts`, we wouldn't have to duplicate it ? Shouldn't we just move it then?

abdulmth: I moved it to `index.ts`, probably `bindings.ts` fits better, but in `index.ts` we don't need to duplicate.

bindings/nodejs/lib/bindings.ts
```diff
@@ -100,15 +100,6 @@ const callWalletMethodAsync = (
...
 
-// Allow bigint to be serialized as hex string.
-//
-// Note:
-// Serializing `bigint` to a different format, e.g. to decimal number string
-// must be done manually.
-(BigInt.prototype as any).toJSON = function () {
-    return bigIntToHex(this);
-};
-
```
thibault-martinez, line 111: Is it breaking change to move it? Was it directly accessible through bindings.ts?

abdulmth: No it's not breaking, it just adds the function `toJSON` to `BigInt`, no matter where it's called from.

bindings/wasm/CHANGELOG.md
```diff
@@ -19,6 +19,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 1.0.2 - 2023-MM-DD
+
+Same changes as https://github.com/iotaledger/iota-sdk/blob/develop/bindings/nodejs/CHANGELOG.md.
+
+### Fixed
+
+- Runtime serializtion error of `BigInt` when working with native token amounts;
```
thibault-martinez, line 28: Not exactly sure what's best, but technically this hasn't been done in this crate anymore

abdulmth: But it was broken in Wasm and now it's fixed 🤷🏻

thibault-martinez: Yeah but I mean, not because of a change in the wasm bindings, let me think about it

abdulmth: Technically it is, it was copied from somewhere else, but it's a part of the Wasm bindings.

Thoralf-M (approved):

bindings/wasm/CHANGELOG.md
thibault-martinez: Alright 👍🏻

thibault-martinez (approved):

PR #937, opened 2023-07-26
Add PrivateKeySecretManager

thibault-martinez:


sdk/src/client/secret/private_key.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    async fn generate_evm_addresses(
+        &self,
+        _coin_type: u32,
+        _account_index: u32,
+        _address_indexes: Range<u32>,
+        _options: impl Into<Option<GenerateAddressOptions>> + Send,
+    ) -> Result<Vec<EvmAddress>, Self::Error> {
+        panic!()
```
thibault-martinez: There is no proper Error I can use and the enum is not non_exhaustive so I can't add a new variant, not sure what to do...

Thoralf-M: What about `Err(Error::UnsupportedOperation.into())`?

Thoralf-M: Ok that's behind the ledger nano feature flag, but what about `SecretManagerMismatch`? Not exactly the same, but better than panic

sdk/src/client/secret/private_key.rs
```diff
@@ -0,0 +1,124 @@
...
+        super::default_sign_transaction(self, prepared_transaction_data).await
+    }
+}
+
+impl PrivateKeySecretManager {
+    /// Create a new [`PrivateKeySecretManager`] from a base 58 encoded private key.
+    pub fn try_from_b58<T: AsRef<[u8]>>(b58: T) -> Result<Self, Error> {
+        let mut bytes = [0u8; ed25519::SecretKey::LENGTH];
+
+        if bs58::decode(b58.as_ref()).onto(&mut bytes).unwrap() != ed25519::SecretKey::LENGTH {
```
Thoralf-M: Here we could maybe use `iota_crypto::Error::PrivateKeyError`

sdk/tests/client/secret_manager/mod.rs
```diff
@@ -0,0 +1,7 @@
+// Copyright 2022 IOTA Stiftung
```
Thoralf-M:
```suggestion
// Copyright 2023 IOTA Stiftung
```

sdk/src/client/secret/private_key.rs
thibault-martinez: Changed

sdk/src/client/secret/private_key.rs
thibault-martinez: Changed

sdk/src/client/secret/private_key.rs
```diff
@@ -0,0 +1,127 @@
...
```
Thoralf-M: The unwrap is safe?

abdulmth: can't we return the same error on the `unwrap`? can't an invalid character cause a panic? 🤔

thibault-martinez: Forgot about it, changed

thibault-martinez: Done

abdulmth (dismissed):

Thoralf-M (dismissed):

DaughterOfMars (dismissed):

kwek20 (approved):

Thoralf-M (approved):

abdulmth (approved):

Issue #549, opened 2023-06-06
Allow providing a mnemonic in the StrongholdSecretManager builder


Alex6323:
## Description

I have this use-case in the cli-wallet `init_command` fn:

```rust
let secret_manager = StrongholdSecretManager::builder()
    .password(&password)
    .build(snapshot_path)?;
// TODO: allow providing the mnemonic in the builder?
secret_manager.store_mnemonic(mnemonic).await?;
let secret_manager = SecretManager::Stronghold(secret_manager);
```

## Motivation

Ergonomics.

## Open questions (optional)

I haven't investigated this deeply enough as to why this doesn't exist already. There might be a good reason for it.

## Are you planning to do it yourself in a pull request?

I would do it.

DaughterOfMars:
After thorough investigation, I do not believe this is feasible without making the build function `async`, which I think would worsen the ergonomics in general.

Alex6323:
Yeah, I don't think we're going to switch to sync locks for this. Also there are other problems like the order of how you'ld call the builder methods (password must be set before mnemonic and so on ...) which makes me realize that it's too much work to redesign this and it's probably not going to be worth the effort anyway. However, I still would have liked if setting up a *new* stronghold could be possible with providing everything atomically in one builder call, which would then ensure that a "built" stronghold instance was always something in a complete and valid state.

So ... I am fine with closing the issue.

PR #940, opened 2023-07-26
Simplify `StrongholdAdapterBuilder`

DaughterOfMars:
# Description of change

This PR came about as a result of investigating #549. I found that the handling of this builder type was overly complicated and could be simplified.

## Type of change

### DIFF
--- a/sdk/src/client/stronghold/mod.rs
+++ b/sdk/src/client/stronghold/mod.rs
@@ -52,14 +52,14 @@ mod migration;
 mod secret;
 mod storage;
 
+use alloc::sync::Weak;
 use std::{
     path::{Path, PathBuf},
     sync::Arc,
     time::Duration,
 };
 
-use derive_builder::Builder;
-use iota_stronghold::{KeyProvider, SnapshotPath, Stronghold};
+use iota_stronghold::{ClientError, KeyProvider, SnapshotPath, Stronghold};
 use log::{debug, error, warn};
 use tokio::{
     sync::{Mutex, MutexGuard},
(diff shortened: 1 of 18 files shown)


sdk/src/client/stronghold/mod.rs
```diff
@@ -74,11 +74,9 @@ use super::{storage::StorageAdapter, utils::Password};
 /// A wrapper on [Stronghold].
 ///
 /// See the [module-level documentation](self) for more details.
-#[derive(Builder, Debug)]
```
thibault-martinez, line 77: Really happy about this, I've always hated it

sdk/src/client/stronghold/mod.rs
```diff
@@ -52,14 +52,14 @@ mod migration;
...
 mod storage;
 
+use alloc::sync::Weak;
 use std::{
     path::{Path, PathBuf},
     sync::Arc,
     time::Duration,
 };
 
-use derive_builder::Builder;
```
thibault-martinez, line 61: Can the dependency be removed? I think it was the only use

sdk/src/client/stronghold/mod.rs
thibault-martinez, line 1: There is a comment about derive_builder L167

sdk/src/client/stronghold/mod.rs
DaughterOfMars: I think it would tell me if it was unused.

DaughterOfMars: Oh, misread

thibault-martinez: Idk, udeps is notoriously full or misses and false positives. It really looks unused to me

thibault-martinez (approved):

Thoralf-M (approved):

PR #946, opened 2023-07-27
Return error on secret manager mismatch

Alex6323:
# Description of change

In some cases we just return an `Ok(())` on some stronghold related wallet methods even if the internally used secret manager is something else than a stronghold secret manager. This I think might be confusing to the user. This PR changes this behavior and will consistently return an error instead if a stronghold-only operation is tried to be called on some other secret manager.

None

## Type of change

### DIFF
--- a/.github/workflows/bindings-nodejs.yml
+++ b/.github/workflows/bindings-nodejs.yml
@@ -2,7 +2,7 @@ name: Nodejs bindings checks
 
 on:
   push:
-    branches: [develop, production]
+    branches: [develop, production, 1.1]
     paths:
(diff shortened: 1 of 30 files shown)


thibault-martinez (changes requested):
Rebase

Thoralf-M (dismissed):

thibault-martinez (changes requested):
Changelog

DaughterOfMars (dismissed):

Thoralf-M (approved):

thibault-martinez (approved):

PR #950, opened 2023-07-27
feat(cli): Add cli command to list wallet accounts

### LINKED ISSUE #658: cli: list-accounts command
## Description

Currently, when creating a new account (with optional alias) we are unable to list all created accounts associated with the wallet.
The `list-accounts` would be helpful to see accounts created in the past, about which users could simply forget. 

## Motivation

improved UX.  Users can see all created accounts with associated aliases

## Requirements

1. new command `list-accounts` that would list accounts with aliases

Yes/No.

Alex6323:
# Description of change

Nice little usability enhancement to the CLI to show all available accounts in the wallet.

Closes #658 

## Type of change

Used the CLI wallet.

### DIFF
--- a/cli/src/account.rs
+++ b/cli/src/account.rs
@@ -21,14 +21,14 @@ use crate::{
     },
     error::Error,
     helper::{bytes_from_hex_or_file, print_account_help},
-    println_log_error, println_log_info,
+    println_log_error,
 };
 
 // loop on the account prompt
 pub async fn account_prompt(wallet: Wallet, account: String) -> Result<(), Error> {
     let mut history = AccountHistory::default();
     loop {
-        match account_prompt_internal(wallet.clone(), account.clone(), &mut history).await {
+        match account_prompt_internal(&wallet, &account, &mut history).await {
             Ok(true) => {
                 return Ok(());
             }
@@ -42,8 +42,8 @@ pub async fn account_prompt(wallet: Wallet, account: String) -> Result<(), Error
 
 // loop on the account prompt
 pub async fn account_prompt_internal(
-    wallet: Wallet,
-    account: String,
+    wallet: &Wallet,
+    account: &str,
     history: &mut AccountHistory,
 ) -> Result<bool, Error> {
     let account = wallet.get_account(account).await?;
@@ -62,10 +62,10 @@ pub async fn account_prompt_internal(
             // Clear console
             let _ = std::process::Command::new("clear").status();
         }
-        "l" | "list" => {
+        "accounts" => {
             // List all accounts
             let aliases = wallet.get_account_aliases().await?;
-            println_log_info!("{aliases:?}");
+            println!("{aliases:?}");
         }
         _ => {
             // Prepend `Account: ` so the parsing will be correct
--- a/cli/src/account.rs
+++ b/cli/src/account.rs
@@ -64,7 +64,11 @@ pub async fn account_prompt_internal(
         "accounts" => {
             // List all accounts
             let accounts = wallet.get_account_ids().await?;
-            println!("{accounts:?}");
+
+            println!("INDEX\tALIAS");
+            for (index, alias) in accounts {
+                println!("{index}\t{alias}");
+            }
         }
         _ => {
             // Prepend `Account: ` so the parsing will be correct
--- a/cli/src/account.rs
+++ b/cli/src/account.rs
@@ -4,7 +4,7 @@
 use clap::Parser;
 use colored::Colorize;
 use dialoguer::Input;
-use iota_sdk::wallet::Wallet;
+use iota_sdk::wallet::{Account, Wallet};
 
 use crate::{
     account_completion::ACCOUNT_COMPLETION,
@@ -25,10 +25,10 @@ use crate::{
 };
 
 // loop on the account prompt
-pub async fn account_prompt(wallet: Wallet, account: String) -> Result<(), Error> {
+pub async fn account_prompt(wallet: &Wallet, account: &Account) -> Result<(), Error> {
     let mut history = AccountHistory::default();
     loop {
-        match account_prompt_internal(&wallet, &account, &mut history).await {
+        match account_prompt_internal(wallet, account, &mut history).await {
             Ok(true) => {
                 return Ok(());
             }
@@ -43,10 +43,9 @@ pub async fn account_prompt(wallet: Wallet, account: String) -> Result<(), Error
 // loop on the account prompt
 pub async fn account_prompt_internal(
     wallet: &Wallet,
-    account: &str,
+    account: &Account,
     history: &mut AccountHistory,
 ) -> Result<bool, Error> {
-    let account = wallet.get_account(account).await?;
     let alias = {
         let account = account.details().await;
         account.alias().clone()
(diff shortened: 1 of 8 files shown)


thibault-martinez (changes requested):
You are bringing some changes from 1.0

cli/src/account.rs
```diff
@@ -52,11 +53,16 @@ pub async fn account_prompt_internal(account: Account, history: &mut AccountHist
         .completion_with(&ACCOUNT_COMPLETION)
         .interact_text()?;
     match command.as_str() {
-        "h" => print_account_help(),
-        "clear" => {
+        "h" | "help" => print_account_help(),
+        "c" | "clear" => {
```
qrayven, line 60: Probably I miss something but don't know yet :D.  Wouldn't be easier to extend `WalletCommand`? https://github.com/iotaledger/iota-sdk/blob/8c0819f40cdf4fb1b41c8be2070a2b7d8b32cea3/cli/src/wallet.rs#L24

Alex6323: This is about the "list" match arm that I added, right? I can totally imagine it being a `WalletCommand` as well or even instead. Do you think it's enough to get the account aliases at the command-line (non-interactive), or is it better to get them when you're already logged into an account (interactive) like it is now?

qrayven: I can imagine a use case when someone forgot the names of accounts. So they can use `wallet list-accounts`, find the name and the `wallet main` So probably, non-interactive mode would be better?

cli/src/account.rs
```diff
@@ -52,11 +53,16 @@ pub async fn account_prompt_internal(account: Account, history: &mut AccountHist
...
             // Clear console
             let _ = std::process::Command::new("clear").status();
         }
+        "l" | "list" => {
```
thibault-martinez: Just call it `accounts`, list is way too broad

cli/src/account.rs
```diff
@@ -41,7 +41,8 @@ pub async fn account_prompt(account: Account) -> Result<(), Error> {
 }
 
 // loop on the account prompt
-pub async fn account_prompt_internal(account: Account, history: &mut AccountHistory) -> Result<bool, Error> {
+pub async fn account_prompt_internal(wallet: Wallet, account: String, history: &mut AccountHistory) -> Result<bool, Error> {
```
thibault-martinez: Why take ownership of the Wallet ?

thibault-martinez: Forces you to clone for no reason

cli/src/account.rs
Alex6323: yeah, expected this comment, fine with me.

cli/src/command/wallet.rs
```diff
@@ -55,6 +55,8 @@ pub enum WalletCommand {
     ChangePassword,
     /// Initialize the wallet.
     Init(InitParameters),
+    /// List all accounts.
+    ListAccounts,
```
thibault-martinez: Just call it `Accounts`, we had List prefix in the past and we removed them

thibault-martinez (changes requested):
Changelog + documentation

cli/src/account.rs
Alex6323: you're right ... kind of expected we would need it since we claimed ownership for `Account` before this PR and that wasn't necessary either.

cli/src/command/wallet.rs
```diff
@@ -166,6 +168,16 @@ pub async fn init_command(
         .await?)
 }
 
+pub async fn list_accounts_command(storage_path: &Path, snapshot_path: &Path) -> Result<(), Error> {
+    let password = get_password("Stronghold password", false)?;
+    let wallet = unlock_wallet(storage_path, snapshot_path, password).await?;
+
+    let accounts = wallet.get_account_aliases().await?;
+    println!("{accounts:?}");
```
DaughterOfMars: Can we print indexes too?

Alex6323: We certainly could (you mean the account index I assume?), but not sure how that would be useful tbh. If there's no alias set, then we assign the account index to it.

DaughterOfMars: Sure, but you can access accounts by index so the list of accounts seems like it should include it.

cli/src/command/wallet.rs
```diff
@@ -168,7 +168,7 @@ pub async fn init_command(
         .await?)
 }
 
-pub async fn list_accounts_command(storage_path: &Path, snapshot_path: &Path) -> Result<(), Error> {
+pub async fn accounts_command(storage_path: &Path, snapshot_path: &Path) -> Result<(), Error> {
```
DaughterOfMars: I feel like `list_accounts` is better...

cli/src/command/wallet.rs
Alex6323: hmm, fair point.

cli/src/command/wallet.rs
Alex6323: Honestly, me too. It's much clearer to have Verb+Noun in the variant and in the corresponding fn. However, for some reasons we decided to call the variant `Accounts`, so the name of this fn is kinda pre-determined if we want be consistent with the rest of the file.

cli/src/command/wallet.rs
DaughterOfMars: Do you mind formatting it? Perhaps with a tab between the index and Alias? Column headers would be good too.

cli/src/command/wallet.rs
DaughterOfMars: Can we change both? I may have missed a discussion but I feel like both should be `list_accounts`.

cli/src/command/wallet.rs
Alex6323: sure

cli/src/command/wallet.rs
Alex6323: @thibault-martinez ^

thibault-martinez: I wouldn't, no. None of the other commands that list have a List prefix, not sure why we'd do it for this one. Happy to discuss it for 2.0 though.

DaughterOfMars (dismissed):

sdk/src/wallet/core/mod.rs
```diff
@@ -94,14 +94,15 @@ where
         Ok(self.accounts.read().await.clone())
     }
 
-    /// Get all account aliases
-    pub async fn get_account_aliases(&self) -> crate::wallet::Result<Vec<String>> {
+    /// Get all account identifiers
+    pub async fn get_account_ids(&self) -> crate::wallet::Result<Vec<(u32, String)>> {
```
thibault-martinez: Breaking change? Revert

cli/src/command/wallet.rs
```diff
@@ -166,6 +168,19 @@ pub async fn init_command(
         .await?)
 }
 
+pub async fn accounts_command(storage_path: &Path, snapshot_path: &Path) -> Result<(), Error> {
```
thibault-martinez: You've put it in order in the enum but not here :(

sdk/src/wallet/core/mod.rs
Alex6323: shiiiiit

thibault-martinez: You can add another method if really needed but don't touch this one

Alex6323: okay, I am going to add another fn that returns the indexes for the time being until we want to make a breaking release.

Alex6323: nah, this has problems, too ... I am going to have to use `get_accounts`.

DaughterOfMars (dismissed):

DaughterOfMars (approved):

cli/src/command/wallet.rs
```diff
@@ -107,6 +109,20 @@ impl Default for InitParameters {
     }
 }
 
+pub async fn accounts_command(storage_path: &Path, snapshot_path: &Path) -> Result<(), Error> {
+    let password = get_password("Stronghold password", false)?;
+    let wallet = unlock_wallet(storage_path, snapshot_path, password).await?;
+    let accounts = wallet.get_accounts().await?;
+
+    println!("INDEX\tALIAS");
```
thibault-martinez, line 117: Why scream

thibault-martinez (approved):

cli/src/command/wallet.rs
Alex6323: I tried both ... and having it all small letters doesn't make it distinguishable enough for eye to separate it from actual aliases, like:
```
index    alias
0        alice
```
compared to
```
INDEX    ALIAS
0        alice
```
which I find easier to parse.

thibault-martinez: fair enough

Alex6323: obviously that would be my preference, but I don't care so much tbh. Using all-caps for headers is not uncommon and not regarded as SCREAMING like it would in a regular sentence.

qrayven (approved):

PR #951, opened 2023-07-27
Fix `SenderFeature.getSender()` and `IssuerFeature.getIssuer()` in Node.js

abdulmth:
# Description of change

Both `SenderFeature.getSender()` and `IssuerFeature.getIssuer()` don't return the correct type. 

The Following code 

```typescript

    const client = new Client({
        primaryNode: NODE,
        localPow: true,
    });
    const outputResponse = await client.getOutput("0xeec64141ac0f06f3f810be2e916cb9e4955a4b9ba722744ea8ab0cc1f34ee57c0000");
    const output: NftOutput = outputResponse.output as NftOutput;
    let features = output.getImmutableFeatures();
    let issuer = (features![0] as IssuerFeature).getIssuer().getType();
```

throws an Error because `getType()` is not a function. The reason for that is `getIssuer()` returns an `Object` and not the proper type.

Be sure to reference any related issues by adding `fixes #issue_number`.

## Type of change

- Bug fix 

Code above runs without errors.

### DIFF
--- a/bindings/nodejs/lib/types/block/output/feature.ts
+++ b/bindings/nodejs/lib/types/block/output/feature.ts
@@ -1,7 +1,8 @@
 // Copyright 2023 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-import { Address } from '../address';
+import { Type } from 'class-transformer';
+import { Address, AddressDiscriminator } from '../address';
 
 /**
  * All of the feature block types.
@@ -29,6 +30,9 @@ abstract class Feature {
  * Sender feature.
  */
 class SenderFeature extends Feature {
+    @Type(() => Address, {
+        discriminator: AddressDiscriminator,
+    })
     private address: Address;
     constructor(sender: Address) {
         super(FeatureType.Sender);
@@ -45,6 +49,9 @@ class SenderFeature extends Feature {
  * Issuer feature.
  */
 class IssuerFeature extends Feature {
+    @Type(() => Address, {
+        discriminator: AddressDiscriminator,
+    })
     private address: Address;
     constructor(issuer: Address) {
         super(FeatureType.Issuer);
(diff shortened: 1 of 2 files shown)


qrayven (approved):

kwek20 (approved):

bindings/nodejs/lib/types/block/output/feature.ts
```diff
@@ -29,6 +30,9 @@ abstract class Feature {
  * Sender feature.
  */
 class SenderFeature extends Feature {
+    @Type(() => Address, {
+        discriminator: AddressDiscriminator,
+    })
```
thibault-martinez, line 35: I have a hard time wrapping my head around this. Why is this not attached to `getSender`? This seems a bit weird just here in the SenderFeature

abdulmth: `getSender` isn't really an issue, you can imagine that `address` is public, so when` SenderFeature` is constructed, `address` was assigned to an instance of `Object`. But with this annotation, now it's assigned to one of the `Address` classes. `getSender` just returns whatever is stored in `address`.

thibault-martinez: This wouldn't have worked if there was more fields than just address, right?

abdulmth: I think in that case, each field must have its own annotation.

abdulmth: here is an example of multiple fields https://github.com/iotaledger/iota-sdk/blob/develop/bindings/nodejs/lib/types/block/payload/transaction/essence.ts#L36

thibault-martinez: Thanks, that helped

thibault-martinez (approved):

PR #952, opened 2023-07-27
Release CLI wallet 1.0

thibault-martinez:


abdulmth (approved):

qrayven (approved):

.github/workflows/cli-publish.yml
```diff
@@ -24,20 +24,8 @@ jobs:
...
-
-            <!---  ## Deprecated -->
-
-            <!---  ## Removed -->
-
-            <!---  ## Fixed -->
-
-            # Checksums
-
+            https://github.com/iotaledger/iota-sdk/blob/develop/cli/CHANGELOG.md
```
kwek20, line 27: Can we make this link to the specific tag? Otherwise on newer release the link wont "work" for old ones

thibault-martinez: Not sure that's really worth it, a direct link would have to include tag+date which is not vert convenient

thibault-martinez: We can create an issue for it if you want. I wouldn't do it here because we have the same for the other releases.

thibault-martinez: I'll go ahead and merge this, please open an issue if you want this investigated

PR #967, opened 2023-07-28
blockid

kwek20:
# Description of change

BlockId is 40: https://github.com/iotaledger/tips-draft/blob/tip46/tips/TIP-0046/tip-0046.md#block-id

Is this correct?

### DIFF
--- a/sdk/src/types/block/block_id.rs
+++ b/sdk/src/types/block/block_id.rs
@@ -4,7 +4,11 @@
 impl_id!(
     pub BlockId,
     40,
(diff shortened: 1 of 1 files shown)


sdk/src/types/block/block_id.rs
```diff
@@ -3,7 +3,7 @@
 
 impl_id!(
     pub BlockId,
-    32,
+    40,
     "A block identifier, the BLAKE2b-256 hash of the block bytes. See <https://www.blake2.net/> for more information."
```
Thoralf-M: Wrong comment now

PR #982, opened 2023-08-01
Make `last_sync` a field in `ClientInner`

PhilippGackstatter:
# Description of change

When using a private tangle with HRP `tst` and multiple client instances in Wasm, the HRP that is returned from `get_network_info` is `smr` on any of the client instances but the first one that calls it.

The issue with the current approach for storing the last sync time is that it is a `static`. It is thus the same across different client instances. For instance, when instantiating `client1` and `client2` and running `client1.get_network_info()`, `client1` will have the fetched protocol params from the node set. It will also update the last synced time to only fetch the params again 60s from now. However, `client2` will not have the fetched protocol params set, but because the last synced time is shared with `client1`, it will also not fetch them on the first call to `get_network_info`. Thus `client2` will return the default protocol parameters from the client, which is `smr` for the HRP.

This PR changes the last sync time from a static to a field in the client, fixing the issue.

n/a

## Type of change

Tested locally in concert with examples in identity where the issue occurred.

## Change checklist

### DIFF
--- a/sdk/src/client/core.rs
+++ b/sdk/src/client/core.rs
@@ -54,6 +54,8 @@ pub struct ClientInner {
     pub(crate) pow_worker_count: RwLock<Option<usize>>,
     #[cfg(feature = "mqtt")]
     pub(crate) mqtt: MqttInner,
+    #[cfg(target_family = "wasm")]
+    pub(crate) last_sync: std::sync::Mutex<Option<u32>>,
 }
 
 #[derive(Default)]
@@ -104,11 +106,8 @@ impl ClientInner {
         // create invalid transactions/blocks.
         #[cfg(target_family = "wasm")]
         {
-            lazy_static::lazy_static! {
-                static ref LAST_SYNC: std::sync::Mutex<Option<u32>> = std::sync::Mutex::new(None);
-            };
             let current_time = crate::utils::unix_timestamp_now().as_secs() as u32;
-            if let Some(last_sync) = *LAST_SYNC.lock().unwrap() {
+            if let Some(last_sync) = *self.last_sync.lock().unwrap() {
                 if current_time < last_sync {
                     return Ok(self.network_info.read().await.clone());
                 }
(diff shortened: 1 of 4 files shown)


thibault-martinez (changes requested):
Changelog

sdk/src/client/core.rs
```diff
@@ -54,6 +54,8 @@ pub struct ClientInner {
     pub(crate) pow_worker_count: RwLock<Option<usize>>,
     #[cfg(feature = "mqtt")]
     pub(crate) mqtt: MqttInner,
+    #[cfg(target_family = "wasm")]
+    pub(crate) last_sync: std::sync::Mutex<Option<u32>>,
```
DaughterOfMars: We should use tokios `Mutex` here.

PhilippGackstatter: Agreed, changed it.

DaughterOfMars (approved):

Thoralf-M (dismissed):

sdk/src/client/core.rs
```diff
@@ -104,20 +106,16 @@ impl ClientInner {
         // create invalid transactions/blocks.
         #[cfg(target_family = "wasm")]
         {
-            lazy_static::lazy_static! {
```
thibault-martinez, line 107: For the team: do we have other instances of similar things that wouldn't play nicely with more than one instance?

thibault-martinez (dismissed):

sdk/src/client/core.rs
DaughterOfMars: Doesn't look like it, but we do have this weird lazy_static macro that seems to have been a custom replacement for the lib:

```rust
#[cfg(feature = "mqtt")]
macro_rules! lazy_static {
    ($init:expr => $type:ty) => {{
        static mut VALUE: Option<$type> = None;
        static INIT: std::sync::Once = std::sync::Once::new();

        INIT.call_once(|| unsafe { VALUE = Some($init) });
        unsafe { VALUE.as_ref() }.expect("failed to get lazy static value")
    }};
}
```

DaughterOfMars (dismissed):
We can actually remove the lazy_static dependency if we want to.

PhilippGackstatter:
If you could publish this as a new sdk-wasm version after its merged, that'd be great!

thibault-martinez:
> If you could publish this as a new sdk-wasm version after its merged, that'd be great!

Can you prepare a release for it in this PR then? As in, put an actual date + bump package.json

thibault-martinez (approved):

DaughterOfMars (approved):

PR #983, opened 2023-08-01
Feat/cli: added mnemonic gen from console only

### LINKED ISSUE #652: cli : generate mnemonic without interactive mode
## Description

Generation of mnemonic without interaction mode
example: 
```
/wallet mnemonic –output-file-name file-name.txt or ./wallet mnemonic –output-stdout.
```

## Motivation

CLI is many times used for simple shell automation scripts. 

## Requirements

1.  ./wallet mnemonic --output-file-name  filename.txt  will generate a mnemonic to the  file
2. ./wallet mnemonic --output-stdout will generate a mnemonic to the stdout
3. 

## Open questions (optional)

name of flags

Yes/No.

kwek20:
# Description of change

Allows getting mnemonic through just command line args

fixes [#652](https://github.com/iotaledger/iota-sdk/issues/652)

## Type of change

Test cases:
- `./wallet mnemonic --output-stdout no` -> Write to default file **DO WE WANT THIS OR BACK TO USER CHOICE**
- `./wallet mnemonic --output-stdout yes` -> Write to console
- `./wallet mnemonic --output-stdout` -> Write to console
- `./wallet mnemonic --output-file-name filename.txt` -> Writes to file
- `./wallet mnemonic --output-file-name filename.txt --output-stdout` -> Writes to file and console
- `./wallet mnemonic` -> Prompts user

### DIFF
--- a/cli/src/command/wallet.rs
+++ b/cli/src/command/wallet.rs
@@ -3,7 +3,7 @@
 
 use std::path::Path;
 
-use clap::{Args, Parser, Subcommand};
+use clap::{builder::BoolishValueParser, Args, Parser, Subcommand};
 use iota_sdk::{
     client::{
         constants::SHIMMER_COIN_TYPE,
@@ -63,7 +63,14 @@ pub enum WalletCommand {
         path: Option<String>,
     },
     /// Generate a random mnemonic.
-    Mnemonic,
+    Mnemonic {
+        // Output the mnemonic to the specified file.
+        #[arg(long)]
+        output_file_name: Option<String>,
+        // Output the mnemonic to the stdout.
+        #[arg(long, num_args = 0..2, default_missing_value = Some("true"), value_parser = BoolishValueParser::new())]
+        output_stdout: Option<bool>,
+    },
     /// Create a new account.
     NewAccount {
         /// Account alias, next available account index if not provided.
@@ -194,9 +201,8 @@ pub async fn migrate_stronghold_snapshot_v2_to_v3_command(path: Option<String>) 
     Ok(())
 }
 
-pub async fn mnemonic_command() -> Result<(), Error> {
-    generate_mnemonic().await?;
-
+pub async fn mnemonic_command(output_file_name: Option<String>, output_stdout: Option<bool>) -> Result<(), Error> {
+    generate_mnemonic(output_file_name, output_stdout).await?;
     Ok(())
 }
 
--- a/cli/src/command/wallet.rs
+++ b/cli/src/command/wallet.rs
@@ -68,7 +68,7 @@ pub enum WalletCommand {
         #[arg(long)]
         output_file_name: Option<String>,
         // Output the mnemonic to the stdout.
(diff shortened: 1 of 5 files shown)


thibault-martinez:
Not 2.0?

kwek20:
oh :D
didnt even realize

cli/src/command/wallet.rs
```diff
@@ -63,7 +63,14 @@ pub enum WalletCommand {
         path: Option<String>,
     },
     /// Generate a random mnemonic.
-    Mnemonic,
+    Mnemonic {
+        // Output the mnemonic to the specified file.
+        #[arg(long)]
+        output_file_name: Option<String>,
+        // Output the mnemonic to the stdout.
+        #[arg(long, num_args = 0..2, default_missing_value = Some("true"), value_parser = BoolishValueParser::new())]
```
DaughterOfMars: Does this still allow using the flag without args, like so?
``` cli --output-stdout ```

File: cli/src/helper.rs [hunk 2]
```diff
@@ -121,36 +121,54 @@ pub async fn enter_or_generate_mnemonic() -> Result<Mnemonic, Error> {
...
-        .with_prompt("Select how to proceed with it")
-        .items(&choices)
-        .default(0)
-        .interact_on(&Term::stderr())?;
+    let file_path = output_file_name
+        .clone()
+        .unwrap_or(DEFAULT_MNEMONIC_FILE_PATH.to_string());
+
+    let selected_choice = match (output_file_name, output_stdout) {
+        (None, None) => {
```
DaughterOfMars: shouldn't this match `(None, Some(false))` too?

cli/src/command/wallet.rs
kwek20: yes, its one of the test cases i did, the 0 num args allows this if there is a default

cli/src/helper.rs
kwek20: I chose not to, as when you provide no to console, then the user picking section still allows picking console, kinda weird.
But i left the question open in test case 1

cli/src/command/wallet.rs
DaughterOfMars: What about the other instance above then?

DaughterOfMars: Also can you use `0..=1` instead? This range is a little less clear to read

thibault-martinez (changes requested):
Changelog

cli/src/helper.rs
DaughterOfMars: Yeah, it is weird but it also doesn't really make sense to write to file when the user doesn't specify they want to

kwek20: lets see what the others think

cli/src/command/wallet.rs
kwek20: ah yea that works better

Thoralf-M (dismissed):
I'm fine with this, will also approve again if you change the match to `(None, Some(false))`

Thoralf-M (approved):

DaughterOfMars (approved):

PR #986, opened 2023-08-02
Fix ledger nano downcasting and add test

DaughterOfMars:
# Description of change

This PR fixes ledger nano secret manager variants not correctly downcasting and thus not sending events.

## Type of change

Added a new private tangle test

### DIFF
--- a/sdk/tests/wallet/transactions.rs
+++ b/sdk/tests/wallet/transactions.rs
@@ -273,12 +273,6 @@ async fn prepare_transaction_ledger() -> Result<()> {
     let account_1 = wallet.create_account().finish().await?;
 
     let amount = 1_000_000;
-    let tx = account_0
-        .send_with_params(
-            [SendParams::new(amount, *account_1.addresses().await?[0].address())?],
-            None,
-        )
-        .await?;
 
     let prepared_event = Arc::new(Mutex::new(None));
     let prepared_event_clone = prepared_event.clone();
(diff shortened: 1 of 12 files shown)


sdk/tests/wallet/transactions.rs
```diff
@@ -255,3 +255,54 @@ async fn conflicting_transaction() -> Result<()> {
...
+
+    wallet
+        .listen([WalletEventType::TransactionProgress], move |event| {
+            if let WalletEvent::TransactionProgress(progress) = &event.event {
+                *prepared_event_clone.lock().unwrap() = Some(progress.clone());
+            } else {
+                panic!("expected TransactionProgress")
+            }
+        })
+        .await;
```
Thoralf-M: This needs to be before `send_with_params()`, because the event will be emitted inside of that

sdk/tests/wallet/transactions.rs
```diff
@@ -255,3 +255,55 @@ async fn conflicting_transaction() -> Result<()> {
...
+
+    let amount = 1_000_000;
+
+    let prepared_event = Arc::new(Mutex::new(None));
+    let prepared_event_clone = prepared_event.clone();
+
+    wallet
+        .listen([WalletEventType::TransactionProgress], move |event| {
+            if let WalletEvent::TransactionProgress(progress) = &event.event {
+                *prepared_event_clone.lock().unwrap() = Some(progress.clone());
```
Thoralf-M: I think here it also need to match on the progress `TransactionProgressEvent`, because there are multiple events and the latest will be `TransactionProgressEvent::Broadcasting` wich would then overwrite the `PreparedTransaction` one before

Thoralf-M (approved):

qrayven (approved):

PR #989, opened 2023-08-03
Make Allotments unique sorted

DaughterOfMars:
# Description of change

This PR fixes the `Allotments` type so that it is unique and sorted by `AccountId` like other similar types.

## Type of change

### DIFF
--- a/sdk/src/types/block/mana/allotment.rs
+++ b/sdk/src/types/block/mana/allotment.rs
@@ -21,6 +21,18 @@ pub struct Allotment {
     pub(crate) mana: u64,
 }
 
+impl PartialOrd for Allotment {
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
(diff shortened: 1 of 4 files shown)


Thoralf-M (approved):

sdk/src/types/block/mana/allotment.rs
```diff
@@ -21,6 +21,18 @@ pub struct Allotment {
     pub(crate) mana: u64,
 }
 
+impl PartialOrd for Allotment {
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        Some(self.cmp(other))
```
qrayven, line 26: FOr `PartialOrd` you compare a whole`Allotment` but for `Ord` `account_id`. Is this correct?

DaughterOfMars: Partial just calls the Ord impl

abdulmth (approved):

qrayven (approved):

PR #928, opened 2023-07-25
Split block into basic and validation types

DaughterOfMars:
# Description of change

This splits blocks into two possible types.

## Links to any relevant issues

- Closes #868 
- Closes #874 
- Closes #877

PR #992, opened 2023-08-04
Implement new block ID

### LINKED ISSUE #894: Make BlockID a 40 bytes array
https://github.com/iotaledger/tips-draft/blob/tip46/tips/TIP-0046/tip-0046.md#block-id

### LINKED ISSUE #895: Adapt Block::id method
Blocked by https://github.com/iotaledger/iota-sdk/issues/894

https://github.com/iotaledger/tips-draft/blob/tip46/tips/TIP-0046/tip-0046.md#block-id

DaughterOfMars:
# Description of change

New PR because old one got messed up. 

This PR implements the new 40 byte `BlockId`.

Closes #894

## Blockers

#928


bindings/core/src/method/client.rs
```diff
@@ -369,4 +369,9 @@ pub enum ClientMethod {
         /// The address for request funds
         address: Bech32Address,
     },
+    /// Returns a block ID (Blake2b256 hash of block bytes) from a block
```
Thoralf-M: Outdated comment (also in the utils enum)

DaughterOfMars: Suggestions for a better description?

Thoralf-M: Maybe just remove `(Blake2b256 hash of block bytes)` since this is not correct anymore
If that's then not enough information we could link to the TIP https://github.com/iotaledger/tips-draft/blob/tip46/tips/TIP-0046/tip-0046.md#block-id (would need to be adjusted when it's an actual TIP)

sdk/src/types/block/block_id.rs
```diff
@@ -1,11 +1,104 @@
...
+        Self::new([0u8; Self::LENGTH])
+    }
+
+    /// Checks if the [`BlockId`] is null.
+    pub fn is_null(&self) -> bool {
+        self.as_ref().iter().all(|&b| b == 0)
+    }
+
+    /// Returns the [`BlockId`]'s slot index part.
+    pub fn slot_index(&self) -> SlotIndex {
```
Thoralf-M, line 51: Maybe also add a method to get the BlockHash?

DaughterOfMars: I don't think we should unless there's a good reason. Its a public field for the crate if we need it.

sdk/src/types/block/block_id.rs
```diff
@@ -1,11 +1,104 @@
...
 
-impl_id!(
-    pub BlockId,
-    32,
-    "A block identifier, the BLAKE2b-256 hash of the block bytes. See <https://www.blake2.net/> for more information."
-);
+use super::{slot::SlotIndex, ConvertTo};
+use crate::types::block::Error;
 
+impl_id!(pub BlockHash, 32, "The hash of a [`Block`].");
```
Thoralf-M, line 7: In the context of the BlockId how it's used now, this is called `ID` in the TIP
`ID` is weird, but `BlockHash` is also used in the context of signature with this definition:
```
`Block Hash` be the BLAKE2b-256 hash over the serialized content of the _Block_ field.
````
So I think for this hash here we should come up with a different name

File: sdk/src/types/block/block_id.rs (hunk 2 above)
Comment (Thoralf-M) [reply sdk/src/types/block/block_id.rs:51]: Ok, currently not sure if it could be useful in certain cases, we can still add it then if needed :+1:

File: sdk/src/types/block/block_id.rs (hunk 3 above)
Comment (DaughterOfMars) [reply sdk/src/types/block/block_id.rs:7]: I dunno, it feel appropriate

Comment (Thoralf-M) [reply sdk/src/types/block/block_id.rs:7]: How about BlockWrapperHash? Then `BlockHash` can be returned from `block_hash()` instead of `[u8; 32]`

Comment (DaughterOfMars) [reply sdk/src/types/block/block_id.rs:7]: But the type isn't useful there, we would always just consume it

File: sdk/src/types/block/block_id.rs [hunk 4]
```diff
@@ -1,11 +1,99 @@
...
+
+impl BlockHash {
+    pub fn with_slot_index(self, slot_index: SlotIndex) -> BlockId {
+        BlockId { hash: self, slot_index }
+    }
+}
+
+/// A block identifier, the BLAKE2b-256 hash of the block bytes. See <https://www.blake2.net/> for more information.
```
Thoralf-M:
```suggestion
/// A block identifier.
```

sdk/src/types/block/block_id.rs
Thoralf-M: Ok, but even if the type isn't needed there, from the name it would alone one could expect that it's the one there
But it's a bit of a mess with the naming of the blocks anyways

sdk/src/types/block/core.rs
```diff
@@ -254,13 +256,37 @@ impl<B> BlockWrapper<B> {
...
     /// The maximum number of bytes in a block.
     pub const LENGTH_MAX: usize = 32768;
+    /// The length of the block header.
+    pub const HEADER_LENGTH: usize = size_of::<u8>()
+        + 2 * size_of::<u64>()
+        + size_of::<SlotCommitmentId>()
+        + size_of::<SlotIndex>()
+        + size_of::<IssuerId>();
+    /// The length of the block signature.
+    pub const SIGNATURE_LENGTH: usize = Ed25519Signature::PUBLIC_KEY_LENGTH + Ed25519Signature::SIGNATURE_LENGTH;
```
Thoralf-M: Isn't this missing the type byte?

DaughterOfMars: Indeed

Thoralf-M (dismissed):

Thoralf-M (dismissed):

qrayven (dismissed):

abdulmth (dismissed):

Thoralf-M (dismissed):

sdk/src/wallet/account/operations/syncing/transactions.rs
```diff
@@ -126,7 +126,7 @@ where
                                         confirmed_unknown_output = true;
                                         updated_transaction_and_outputs(
                                             transaction,
-                                            Some(included_block.id()),
+                                            Some(included_block.id(&self.client().get_protocol_parameters().await?)),
```
kwek20: move the params to outside of the for loop?

Thoralf-M (dismissed):

sdk/src/types/block/block_id.rs
```diff
@@ -1,11 +1,112 @@
...
+    }
+}
+
+/// A block identifier.
+#[derive(Clone, Copy, Eq, PartialEq, Hash, Ord, PartialOrd, Debug, packable::Packable)]
+#[packable(unpack_error = Error)]
+#[repr(C)]
+pub struct BlockId {
+    pub(crate) hash: BlockHash,
+    // INPORTANT: On big-endian systems this value is misrepresented because it is transmuted directly
```
abdulmth:
```suggestion
    // IMPORTANT: On big-endian systems this value is misrepresented because it is transmuted directly
```

abdulmth (dismissed):

sdk/tests/types/block_id.rs
kwek20, line 1: This we should test with a non-zero slot index to make sure le and be work as expected

sdk/src/types/block/block_id.rs
```diff
@@ -1,11 +1,112 @@
...
+    /// Returns the [`BlockId`]'s slot index part.
+    #[cfg(target_endian = "big")]
+    pub fn slot_index(&self) -> SlotIndex {
+        self.slot_index.to_le().into()
+    }
+}
+
+impl AsRef<[u8]> for BlockId {
+    fn as_ref(&self) -> &[u8] {
+        unsafe { core::mem::transmute::<_, &[u8; Self::LENGTH]>(self) }
```
kwek20, line 58: so this would technically go wrong on be systems.
maybe set a feature flag warn when its called?

sdk/src/wallet/account/operations/syncing/transactions.rs
DaughterOfMars: done

sdk/tests/types/block_id.rs
DaughterOfMars: good call

sdk/src/types/block/block_id.rs
DaughterOfMars: no, this is correct on big or little endian because the underlying memory is always the same

sdk/tests/types/block_id.rs
DaughterOfMars: added

Thoralf-M (approved):

abdulmth (approved):

kwek20 (approved):

PR #995, opened 2023-08-07
Refactor block signature

### LINKED ISSUE #878: Add Block signing - rust
Blocked by https://github.com/iotaledger/iota-sdk/issues/877

https://github.com/iotaledger/tips-draft/blob/tip46/tips/TIP-0046/tip-0046.md#signature

DaughterOfMars:
# Description of change

After reading through the TIPs, I made some changes to how the block builder works to allow for calculating the block signature from the header and data hashes. I added a trait to allow random signing for tests, and likely a similar trait could simplify the UX for the block builder as well.

- #878


sdk/src/client/api/block_builder/mod.rs
```diff
@@ -4,21 +4,27 @@
...
+    pub async fn finish_basic_block_builder<S: SecretManage>(
         &self,
         issuer_id: IssuerId,
-        signature: Ed25519Signature,
         issuing_time: Option<u64>,
         strong_parents: Option<StrongParents>,
         payload: Option<Payload>,
-    ) -> Result<Block> {
+        coin_type: u32,
+        secret_manager: &S,
```
Thoralf-M, line 23: I don't think that's what we want here

Thoralf-M: I would prefer not to accept a secret manager here (unless maybe if we split finish_basic_block_builder(), so one function prepares the final block, just without the signature and then the signing happens in another transaction)
Also wherever we do, we shouldn't just accept the coin_type, but `Bip44`, since a user will probably not always want the public key from the default path (account and address index 0)

DaughterOfMars: seems kind of pointless to split it up, but the other point is fair

sdk/src/types/block/core.rs
```diff
@@ -508,17 +555,10 @@ impl Block {
...
-            Self::Validation(b) => {
-                bytes.push(ValidationBlock::KIND);
-                bytes.extend(b.data.pack_to_vec());
-            }
-        }
+        let bytes = match self {
+            Self::Basic(b) => b.block_hash(),
+            Self::Validation(b) => b.block_hash(),
+        };
         Blake2b256::digest(bytes).into()
```
Thoralf-M: This should be removed now, since the other `block_hash()` already call this

Thoralf-M:
```suggestion
        match self {
            Self::Basic(b) => b.block_hash(),
            Self::Validation(b) => b.block_hash(),
        }
```

Thoralf-M (dismissed):

Thoralf-M (dismissed):

abdulmth (dismissed):

Thoralf-M (dismissed):

abdulmth (dismissed):

Thoralf-M (dismissed):

Thoralf-M (dismissed):

abdulmth (dismissed):

Thoralf-M (dismissed):

Thoralf-M (dismissed):

Thoralf-M (dismissed):

Thoralf-M (dismissed):

sdk/src/types/block/core/basic.rs
```diff
@@ -95,6 +118,75 @@ impl BasicBlockBuilder {
     }
 }
 
+impl BlockBuilder<BasicBlockBuilder> {
```
thibault-martinez: Honestly I feel like this is a remnant from the previous version where the wrapper itself was generic. I don't think we need generics for the builder and I even think it makes it more complex and duplicates code. We already have a `BasicBlockBuilder`, we don't need more. We probably just need to add a very simple and straightforward BlockWrapperBuilder.

DaughterOfMars: In that case we would also need a `BlockBuilder` enum and everything that comes with that

thibault-martinez: Hmm not sure about that, if the wrapper builder takes an already built block I don't see why we'd need an enum. But I haven't tried it so maybe I'm missing something. I can have a look tomorrow.

DaughterOfMars: I went ahead and removed it

sdk/src/types/block/core/wrapper.rs
```diff
@@ -16,11 +16,136 @@ use crate::types::block::{
...
+        self.block = block.into();
+        self
+    }
+
+    pub(crate) fn header_hash(&self) -> [u8; 32] {
+        let mut bytes = [0u8; BlockHeader::LENGTH];
+        BlockHeader {
+            protocol_version: self.protocol_version,
+            network_id: self.network_id,
+            // TODO: what do here
```
Thoralf-M: Why not make it required in the new() method?

DaughterOfMars: felt like it was something we should allow to be auto generated, but I dunno

thibault-martinez: I feel like we should take an already built BlockHeader the same way we take an already built Block. This moves the problem of coming up with a value to somewhere else because I don't think it should be here

sdk/src/client/api/block_builder/mod.rs
```diff
@@ -8,22 +8,20 @@ pub use self::transaction::verify_semantic;
...
         payload::Payload,
-        signature::Ed25519Signature,
-        IssuerId,
+        Block, IssuerId,
     },
 };
 
 impl ClientInner {
-    pub async fn finish_basic_block_builder(
+    pub async fn unsigned_basic_block_builder(
```
thibault-martinez: I'd simplify the name tbh, it returns a builder, it's assumed it's unsigned

sdk/src/client/secret/mod.rs
```diff
@@ -602,3 +604,29 @@ where
 
     Ok(tx_payload)
 }
+
+#[async_trait]
+pub trait SignBlockExt {
```
thibault-martinez: Can we just call it SignBlock? I think Ext would fit if it was a super-trait of SecretManage but it's not

sdk/src/types/block/rand/block.rs
```diff
@@ -3,19 +3,22 @@
 
 use alloc::vec::Vec;
 
+use super::signature::rand_sign_ed25519;
```
thibault-martinez: Why move it out 😭 ?

sdk/src/client/api/block_builder/mod.rs
DaughterOfMars: good point

sdk/src/types/block/rand/block.rs
DaughterOfMars: probably a merge thing

sdk/src/types/block/core/wrapper.rs
```diff
@@ -16,11 +16,50 @@ use crate::types::block::{
...
+        self
+    }
+
+    /// Get the signing input that can be used to generate an
+    /// [`Ed25519Signature`](crate::types::block::signature::Ed25519Signature) for the resulting block.
+    pub fn signing_input(&self) -> Vec<u8> {
+        [self.header.hash(), self.block.hash()].concat()
+    }
+
+    pub fn finish(self, signature: Ed25519Signature) -> Result<BlockWrapper, Error> {
```
thibault-martinez: Take an `impl Into<Signature>` instead

thibault-martinez (changes requested):
Common feature set CI failing

thibault-martinez (dismissed):

Thoralf-M (dismissed):

Alex6323 (dismissed):

Thoralf-M (approved):

Alex6323 (approved):

thibault-martinez (approved):

PR #998, opened 2023-08-08
Fix block id computation, fix return type, reexport BlockHash, clippy

Thoralf-M:


sdk/tests/types/block_id.rs
```diff
@@ -49,3 +52,17 @@ fn pack_unpack_valid() {
         PackableExt::unpack_verified(packed_block_id.as_slice(), &()).unwrap()
     );
 }
+
+#[test]
+fn compute() {
+    let block_dto_json = r#"{"protocolVersion":3,"networkId":"0","issuingTime":"0","slotCommitment":"0x8633b2eb1845fdecf12ee6c5e789c3cf1f0d0bbb3cee65cb5fb2757e995b5cd70000000000000000","latestFinalizedSlot":"0","issuerId":"0x0000000000000000000000000000000000000000000000000000000000000000","block":{"type":1,"strongParents":["0x417c5700320912627b604d4c376a5a1663634b09703538570b1d52440b3e474639490b100a6f3608"],"weakParents":[],"shallowLikeParents":[],"highestSupportedVersion":3,"protocolParametersHash":"0x0000000000000000000000000000000000000000000000000000000000000000"},"signature":{"type":0,"publicKey":"0x714f5f07067012267c21426d382a52752f0b3208443e0e3c49183e0110494148","signature":"0x3e4a492924302b3b093f1e4266757a1d2041480a3861271d4c2e646d4e3d08360a3e765e1a385a784f6753276c233123475867370a184573195d530b41643a1d"}}"#;
+    let block_dto = serde_json::from_str::<BlockDto>(block_dto_json).unwrap();
+    let block = Block::try_from_dto(block_dto).unwrap();
+    let block_hash = block.hash();
```
DaughterOfMars, line 61: could just call the id fn, which would still call both of these

DaughterOfMars (approved):

sdk/tests/types/block_id.rs
Thoralf-M: But then I need to provide the full `ProtocolParameters`

PR #1000, opened 2023-08-09
Store protocol params on block

DaughterOfMars:
# Description of change

Now that calculating the ID requires protocol parameters, it makes sense to store them on the block. As blocks already required the protocol params to unpack, this is simple. However, if you call `try_from_dto` without providing the params, they will be defaulted. So, we should never do this without understanding that the ID generated from such a block will be wrong. For instance, in `Utils::GetBlockId` we circumvent this by manually constructing the ID from the hash and the necessary params.


sdk/src/types/block/basic.rs
```diff
@@ -219,12 +213,12 @@ impl Packable for BasicBlock {
...
         }
 
-        let data = BasicBlockData::unpack::<_, VERIFY>(unpacker, visitor)?;
+        let data = BasicBlockData::unpack::<_, VERIFY>(unpacker, protocol_params)?;
 
         let Signature::Ed25519(signature) = Signature::unpack::<_, VERIFY>(unpacker, &())?;
 
         let block = Self {
-            protocol_version,
+            protocol_params: protocol_params.clone(),
```
kwek20, line 226: doesnt look like this clone should be needed

sdk/src/types/block/slot/index.rs
```diff
@@ -17,6 +17,11 @@ impl SlotIndex {
     pub fn new(index: u64) -> Self {
         Self::from(index)
     }
+
+    /// Returns the inner value.
+    pub fn get(&self) -> u64 {
```
kwek20: impl deref ;) or make it `pub struct SlotIndex(pub u64);`

sdk/src/types/block/core.rs
```diff
@@ -145,11 +146,11 @@ impl Packable for Block {
 
         let block = match kind {
             BasicBlock::KIND => {
-                let data = BasicBlockData::unpack::<_, VERIFY>(unpacker, visitor)?;
+                let data = BasicBlockData::unpack::<_, VERIFY>(unpacker, protocol_params)?;
                 let Signature::Ed25519(signature) = Signature::unpack::<_, VERIFY>(unpacker, &())?;
 
                 Self::from(BlockWrapper {
-                    protocol_version,
+                    protocol_params: protocol_params.clone(),
```
kwek20, line 158: another clone, i guess theyre needed? please double check

sdk/src/types/block/basic.rs
DaughterOfMars: Protocol params provided to the fn are by reference!

sdk/src/types/block/slot/index.rs
DaughterOfMars: It already derives Deref

sdk/src/types/block/core.rs
```diff
@@ -292,6 +299,7 @@ impl Block {
     /// Creates a new [`BlockBuilder`] to construct an instance of a [`BasicBlock`].
     #[inline(always)]
     pub fn build_basic(
+        protocol_params: ProtocolParameters,
         network_id: u64,
```
Thoralf-M: Shouldn't this also be used from the ProtocolParameters now (with network_name_to_id())?

DaughterOfMars: I did not know that was possible, thanks 👍

Thoralf-M: There is even `ProtocolParameters::network_id()` which calls this

DaughterOfMars: done

Thoralf-M (approved):

kwek20 (approved):

abdulmth (approved):

PR #1003, opened 2023-08-11
Add  lint check for examples

### LINKED ISSUE #935: CI: verify python example syntax without running them (compile them?)

qrayven:
# Description of change

Adds the linter for Python examples. Linter 
 - a compilation of the examples
 - imports
 - whether methods are defined or not

fixes #935 

## Type of change

## Change checklist


bindings/python/examples/.pylintrc
```diff
@@ -0,0 +1,8 @@
+[MESSAGES CONTROL]
+disable=missing-module-docstring,
```
DaughterOfMars, line 2: is the spacing supposed to be like this?

bindings/python/examples/client/build_basic.py
```diff
@@ -43,7 +48,7 @@
     unlock_conditions=[
         address_unlock_condition,
         StorageDepositReturnUnlockCondition(
-            return_address=Ed25519Address(hex_address),
+            returnAddress=Ed25519Address(hex_address),
```
DaughterOfMars, line 51: uhhhh, is this casing right?

bindings/python/examples/client/mqtt.py
```diff
@@ -22,8 +24,10 @@
 
 
 def callback(event):
+    """Callback function for the MQTT listener"""
     event_dict = json.loads(event)
     print(event_dict)
+    # pylint: disable=global-statement
```
DaughterOfMars, line 30: ?

bindings/python/examples/how_tos/accounts_and_addresses/consolidate_outputs.py
```diff
@@ -32,6 +34,7 @@
 
 for i, output_data in enumerate(outputs):
     print(f'OUTPUT #{i}')
+    # pylint: disable=consider-using-f-string
     print(
```
DaughterOfMars: Maybe instead of disabling lints, fix the issue?

bindings/python/examples/how_tos/alias_wallet/transaction.py
```diff
@@ -31,14 +34,15 @@
 
 # Find first output unlockable by the alias address
 query_parameters = NodeIndexerAPI.QueryParameters(alias_address)
-input = wallet.get_client().basic_output_ids(query_parameters).items[0]
+transaction_input = wallet.get_client(
```
DaughterOfMars: Oof, can we perhaps configure the lint so this doesn't happen?

bindings/python/examples/wallet/logger.py
```diff
@@ -1,8 +1,12 @@
-from iota_sdk import Wallet, StrongholdSecretManager, init_logger, CoinType, ClientOptions
-from dotenv import load_dotenv
 import json
 import os
 
+from dotenv import load_dotenv
+
+# pylint: disable=no-name-in-module
```
DaughterOfMars, line 6: ?

.github/workflows/bindings-python.yml
```diff
@@ -40,7 +40,7 @@ concurrency:
 
 jobs:
   lint:
-    name: PEP8 style check
+    name: Python PEP8 format and lint check
```
kwek20: this one says lint, but we dont lint here (also on line 61)

bindings/python/examples/client/build_basic.py
kwek20: yea, the name is different here, but we should fix it instead

bindings/python/examples/client/custom_plugin.py
```diff
@@ -1,9 +1,11 @@
-from iota_sdk import Client, init_logger
-from dotenv import load_dotenv
-
 import json
 import os
 
+from dotenv import load_dotenv
+
+# pylint: disable=no-name-in-module
```
kwek20, line 6: yea these should all go, its aparently requiring a __init__.py file somewhere

qrayven: sorry I forgot to explain it in the PR. This must be disabled because `init_logger`  comes from external bindings and `pylint` doesn't work well with the case when we have a mixin of  C  and native ones important from one enrypoint (`iota_sdk`) So far I didn't find any solution that would fix that error (it's a known problem of `pylint`) I propose spend additional hours specifically on that in the separate story

bindings/python/examples/how_tos/accounts_and_addresses/consolidate_outputs.py
qrayven: It this case, IMO, the `format` fits better than `f-string`. But, ok, I can change that

bindings/python/examples/how_tos/alias_wallet/transaction.py
qrayven: The `input`'s override was actually triggered by `pylint`, so this is already detected by linter

bindings/python/examples/wallet/logger.py
qrayven: This is the same case as here:
https://github.com/iotaledger/iota-sdk/pull/1003/files#r1295416035

bindings/python/examples/client/build_basic.py
qrayven: Initially I wanted to fix it too, but isn't the breaking change? If you look, this applies to all unclock conditions.

.github/workflows/bindings-python.yml
qrayven: yeah, lefover, fixed

bindings/python/examples/.pylintrc
qrayven: no, it shouldn't. This is a weird behavior, because vscode shows it differently, Fixed

bindings/python/examples/client/mqtt.py
qrayven: I would allow to use `global` here.
 - makes the example simple
 - otherwise we need to build a stateful callback listener
 - python doesn't allow to pass by reference non-compound types

bindings/python/examples/client/custom_plugin.py
Thoralf-M: Can we define `init_logger` in the Python code which just calls the external one, so we don't need an extra comment here?

bindings/python/examples/client/mqtt.py
DaughterOfMars: Alright

bindings/python/examples/how_tos/alias_wallet/transaction.py
DaughterOfMars: I mean the spacing

bindings/python/examples/how_tos/accounts_and_addresses/consolidate_outputs.py
DaughterOfMars: You're right, my mistake. Perhaps we should disable this lint globally, it seems like a bad lint.

.github/workflows/bindings-python.yml
```diff
@@ -111,6 +112,12 @@ jobs:
           sudo apt-get update
           sudo apt-get install libudev-dev libusb-1.0-0-dev
 
-      - name: Run Tox
+      - name: Run linter for examples
+        if: ${{ startsWith(matrix.os, 'ubuntu-latest') }}
```
thibault-martinez, line 116: Any specific reason for this `if`?

qrayven: Just a matter of saving time. We don't need to run `pylint` on Windows because it will give the same results.

thibault-martinez: Then we also don't need to run it on macos?

thibault-martinez: Lol ignore my comment, brainfart

bindings/python/examples/client/custom_plugin.py
qrayven: Ok, unfortunately, it's not easy. When trying this I always ended up in an import loop. The 'right' way it should be done is with `pyi` file that contains the definitions for FFI functions. Apparently `pylint` version 2.5.17 doesn't work well those files, so we probably need to wait for version 3. So my proposition is to keep it as is

thibault-martinez (dismissed):

bindings/python/examples/how_tos/accounts_and_addresses/consolidate_outputs.py
qrayven: brought this back

bindings/python/examples/how_tos/alias_wallet/transaction.py
qrayven: I mean this is a PEP8  formatted line braking. I think there is nothing we can do as we want to keep pep8. Imo `black` formatter does much better in this aspect.

bindings/python/examples/how_tos/accounts_and_addresses/consolidate_outputs.py
DaughterOfMars: Can we disable the lint globally?

DaughterOfMars: Oh, you did, so remove it here :)

qrayven: done

bindings/python/examples/how_tos/alias_wallet/transaction.py
DaughterOfMars: We can change the allowed length of each line, right?

qrayven: yes this could be changed, I see PEP8 has default, 79, change it to 119?

qrayven: changed to 120

DaughterOfMars (approved):

thibault-martinez (approved):

kwek20 (approved):

Issue #646, opened 2023-06-21
Implement 2.0 routes


thibault-martinez:
# `node`

- [x] GET /health
- [x] GET /api/routes
- [x] GET /api/core/v3/info

# `blocks`

- [x] GET /api/core/v3/blocks/issuance
- [x] POST /api/core/v3/blocks
    - [x] JSON
    - [x] RAW
- [x] GET /api/core/v3/blocks/{blockId}
    - [x] JSON
    - [x] RAW
- [x] GET /api/core/v3/blocks/{blockId}/metadata

# `UTXO`

- [x] GET /api/core/v3/outputs/{outputId}
    - [x] JSON
    - [x] RAW
- [x] GET /api/core/v3/outputs/{outputId}/metadata
- [x] GET /api/core/v3/transactions/{transactionId}/included-block
    - [x] JSON
    - [x] RAW
- [x] GET /api/core/v3/transactions/{transactionId}/included-block/metadata

# `commitments`

- [x] GET /api/core/v3/commitments/{commitmentId}
    - [x] JSON
    - [x] RAW
- [x] GET /api/core/v3/commitments/{commitmentId}/utxo-changes
- [x] GET /api/core/v3/commitments/by-index/{index}
    - [x] JSON
    - [x] RAW
- [x] GET /api/core/v3/commitments/by-index/{index}/utxo-changes

thibault-martinez:
Closed by https://github.com/iotaledger/iota-sdk/pull/1012

PR #1012, opened 2023-08-15
Add Get Issuance Route

DaughterOfMars:
# Description of change

Adds the `get_issuance` node API route to replace `get_tips`.

#646

### DIFF
--- a/bindings/core/src/response.rs
+++ b/bindings/core/src/response.rs
@@ -102,9 +102,6 @@ pub enum Response {
     /// - [`GetPeers`](crate::method::ClientMethod::GetPeers)
     Peers(Vec<PeerResponse>),
     /// Response for:
-    /// - [`GetTips`](crate::method::ClientMethod::GetTips)
-    Tips(Vec<BlockId>),
-    /// Response for:
     /// - [`GetIssuance`](crate::method::ClientMethod::GetIssuance)
     Issuance(IssuanceBlockHeaderResponse),
     /// Response for:
--- a/bindings/core/src/response.rs
+++ b/bindings/core/src/response.rs
@@ -16,7 +16,8 @@ use iota_sdk::{
     types::{
         api::{
             core::response::{
-                BlockMetadataResponse, InfoResponse as NodeInfo, OutputWithMetadataResponse, PeerResponse,
+                BlockMetadataResponse, InfoResponse as NodeInfo, IssuanceBlockHeaderResponse,
+                OutputWithMetadataResponse, PeerResponse,
             },
             plugins::indexer::OutputIdsResponse,
         },
@@ -103,6 +104,9 @@ pub enum Response {
     /// Response for:
     /// - [`GetTips`](crate::method::ClientMethod::GetTips)
     Tips(Vec<BlockId>),
+    /// Response for:
+    /// - [`GetIssuance`](crate::method::ClientMethod::GetIssuance)
+    Issuance(IssuanceBlockHeaderResponse),
     /// Response for:
     /// - [`GetBlock`](crate::method::ClientMethod::GetBlock)
     /// - [`GetIncludedBlock`](crate::method::ClientMethod::GetIncludedBlock)
--- a/sdk/examples/client/node_api_core/03_get_tips.rs
+++ b/sdk/examples/client/node_api_core/03_get_tips.rs
@@ -24,7 +24,7 @@ async fn main() -> Result<()> {
     let client = Client::builder().with_node(&node_url)?.finish().await?;
 
     // Get tips.
-    let tips = client.get_tips().await?;
+    let tips = client.get_issuance().await?.strong_parents;
 
     println!("Tips:\n{tips:#?}");
 
--- a/sdk/src/client/node_api/core/routes.rs
+++ b/sdk/src/client/node_api/core/routes.rs
@@ -101,7 +101,7 @@ impl ClientInner {
         self.node_manager
             .read()
             .await
-            .get_request::<IssuanceBlockHeaderResponse>(PATH, None, self.get_timeout().await, false, false)
+            .get_request(PATH, None, self.get_timeout().await, false, false)
             .await
     }
 
--- a/sdk/src/client/node_api/core/routes.rs
+++ b/sdk/src/client/node_api/core/routes.rs
@@ -15,8 +15,8 @@ use crate::{
     },
     types::{
         api::core::response::{
-            BlockMetadataResponse, InfoResponse, PeerResponse, RoutesResponse, SubmitBlockResponse, TipsResponse,
-            UtxoChangesResponse,
+            BlockMetadataResponse, InfoResponse, IssuanceBlockHeaderResponse, PeerResponse, RoutesResponse,
+            SubmitBlockResponse, UtxoChangesResponse,
         },
         block::{
             output::{dto::OutputDto, Output, OutputId, OutputMetadata},
@@ -93,19 +93,16 @@ impl ClientInner {
 
     // Blocks routes.
 
-    /// Returns tips that are ideal for attaching a block.
-    /// GET /api/core/v3/tips
-    pub async fn get_tips(&self) -> Result<Vec<BlockId>> {
-        let path = "api/core/v3/tips";
+    /// Returns information that are ideal for attaching a block.
+    /// GET /api/core/v3/blocks/issuance
+    pub async fn get_issuance(&self) -> Result<IssuanceBlockHeaderResponse> {
+        const PATH: &str = "api/core/v3/blocks/issuance";
 
-        let response = self
-            .node_manager
+        self.node_manager
             .read()
             .await
-            .get_request::<TipsResponse>(path, None, self.get_timeout().await, false, false)
-            .await?;
-
-        Ok(response.tips)
+            .get_request::<IssuanceBlockHeaderResponse>(PATH, None, self.get_timeout().await, false, false)
+            .await
     }
 
     /// Returns the BlockId of the submitted block.
--- a/sdk/src/client/node_api/core/routes.rs
+++ b/sdk/src/client/node_api/core/routes.rs
@@ -47,10 +47,10 @@ impl ClientInner {
     /// Returns the health of the node.
     /// GET /health
     pub async fn get_health(&self, url: &str) -> Result<bool> {
-        let path = "health";
+        const PATH: &str = "health";
 
         let mut url = Url::parse(url)?;
-        url.set_path(path);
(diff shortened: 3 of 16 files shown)


bindings/core/src/response.rs
```diff
@@ -104,6 +105,9 @@ pub enum Response {
     /// - [`GetTips`](crate::method::ClientMethod::GetTips)
     Tips(Vec<BlockId>),
```
Thoralf-M: Remove this then?

sdk/src/types/api/core/response.rs
```diff
@@ -93,16 +94,24 @@ pub struct BaseTokenResponse {
...
-    pub tips: Vec<BlockId>,
+pub struct IssuanceBlockHeaderResponse {
+    /// Blocks that are strongly directly approved.
+    pub strong_parents: StrongParents,
+    /// Blocks that are weakly directly approved.
+    pub weak_parents: WeakParents,
+    /// Blocks that are directly referenced to adjust opinion.
+    pub shallow_like_parents: ShallowLikeParents,
+    /// The slot index of the latest finalized slot.
+    pub latest_finalized_slot: SlotIndex,
```
Thoralf-M, line 114: Missing `commitment` field

bindings/core/src/response.rs
DaughterOfMars: oop

sdk/src/types/api/core/response.rs
```diff
@@ -112,6 +112,8 @@ pub struct IssuanceBlockHeaderResponse {
     pub shallow_like_parents: ShallowLikeParents,
     /// The slot index of the latest finalized slot.
     pub latest_finalized_slot: SlotIndex,
+    /// TODO: help
```
Thoralf-M:
```suggestion
    /// A Slot Commitment contains a summary of a slot.
```
from https://github.com/iotaledger/tips-draft/blob/tip46/tips/TIP-0046/tip-0046.md#slot-commitment

DaughterOfMars: Oh, that's what I was looking for. So this raises a question. In the core API, the slot commitment has a protocol version field, but in this TIP it does not. What do?

Thoralf-M: I can't find a protocol version field in the core API

DaughterOfMars: It's in the `Commitment` definition

Thoralf-M: Ah I was looking for `protocolVersion`, but it's just called `version`
In iota.go it's also there https://github.com/iotaledger/iota.go/blob/8290f88bd360293b56595ed03cfe2253bd5c4af9/commitment.go#L21
So I guess it's just missing in the other TIP, added a comment https://github.com/iotaledger/tips-draft/pull/20/files#r1296134085

DaughterOfMars: Okay. Regarding the comment, I was hoping for something that describes the function of this field in this struct.

Thoralf-M: I guess it's required so the block issuer can use it to calculate the commitment id which is required in the block
But then maybe the id alone would be sufficient? Also not sure

Thoralf-M: Got added here https://github.com/iotaledger/tips-draft/pull/20/files/39b6de8c15eb8055afde141df98f9edc3ccdf0c9..422beb7dbca565bbe0bf2369d727daa6e6fa2be7

sdk/examples/client/node_api_core/03_get_tips.rs
thibault-martinez, line 1: Rename the file as well

sdk/examples/client/node_api_core/03_get_tips.rs
```diff
@@ -24,7 +24,7 @@ async fn main() -> Result<()> {
     let client = Client::builder().with_node(&node_url)?.finish().await?;
 
     // Get tips.
-    let tips = client.get_tips().await?;
+    let tips = client.get_issuance().await?.strong_parents;
```
thibault-martinez: The example is not about strong parents, just log the whole struct

sdk/src/types/block/slot/commitment.rs
```diff
@@ -18,6 +18,9 @@ use crate::{
     serde(rename_all = "camelCase")
 )]
 pub struct SlotCommitment {
+    // The version of the protocol running.
+    #[serde(rename = "version")]
+    protocol_version: u8,
```
thibault-martinez, line 23: You renamed to just `version` in other places, do it here as well?

sdk/src/types/block/slot/commitment.rs
```diff
@@ -18,6 +18,9 @@ use crate::{
     serde(rename_all = "camelCase")
 )]
 pub struct SlotCommitment {
```
thibault-martinez, line 20: A new field has been added since, RMC

sdk/src/types/api/core/response.rs
DaughterOfMars: Yeah, still no idea what the comment should be 😆

sdk/src/types/block/slot/commitment.rs
DaughterOfMars: I renamed it to version in the params because it's the protocol params, so the protocol part is implied. Here, I'm not sure that logic applies. If I had my way I would change it in the TIP.

sdk/src/types/block/slot/commitment.rs
DaughterOfMars: ffs

thibault-martinez: You don't have to do it here, just mentioning since you already added one

sdk/src/types/block/slot/commitment.rs
thibault-martinez: Fair enough, maybe suggest it in the TIP?

sdk/src/client/node_api/core/routes.rs
```diff
@@ -93,19 +93,16 @@ impl ClientInner {
...
-        let response = self
-            .node_manager
+        self.node_manager
             .read()
             .await
-            .get_request::<TipsResponse>(path, None, self.get_timeout().await, false, false)
-            .await?;
-
-        Ok(response.tips)
+            .get_request::<IssuanceBlockHeaderResponse>(PATH, None, self.get_timeout().await, false, false)
```
thibault-martinez: Just a nit but explicit type shouldn't be necessary

sdk/src/types/block/slot/commitment.rs
DaughterOfMars: Done

sdk/src/types/block/slot/commitment.rs
DaughterOfMars: Did it

sdk/src/client/node_api/core/routes.rs
```diff
@@ -93,19 +93,16 @@ impl ClientInner {
...
     // Blocks routes.
 
-    /// Returns tips that are ideal for attaching a block.
-    /// GET /api/core/v3/tips
-    pub async fn get_tips(&self) -> Result<Vec<BlockId>> {
-        let path = "api/core/v3/tips";
+    /// Returns information that are ideal for attaching a block.
+    /// GET /api/core/v3/blocks/issuance
+    pub async fn get_issuance(&self) -> Result<IssuanceBlockHeaderResponse> {
+        const PATH: &str = "api/core/v3/blocks/issuance";
```
thibault-martinez, line 99: In case you are interested, there are more places where you can made it const PATH

DaughterOfMars: Sure

sdk/src/types/block/slot/commitment.rs
```diff
@@ -31,21 +34,28 @@ pub struct SlotCommitment {
     /// the switching of a chain.
     #[serde(with = "string")]
     cumulative_weight: u64,
+    /// Reference Mana Cost (RMC) to be used in the slot with index at `index + Max Committable Age`.
+    #[serde(with = "string")]
+    reference_mana_cost: u64,
```
thibault-martinez, line 39: It's not yet been added to the HTTP API but knowing them, it'll probably be `RMC` as JSON. We can update later, just mentioning it.

DaughterOfMars: I added a comment about the naming of this field too, because I think it should be the full name.

sdk/src/client/node_api/core/routes.rs
```diff
@@ -93,33 +93,30 @@ impl ClientInner {
 
     // Blocks routes.
 
-    /// Returns tips that are ideal for attaching a block.
-    /// GET /api/core/v3/tips
-    pub async fn get_tips(&self) -> Result<Vec<BlockId>> {
-        let path = "api/core/v3/tips";
+    /// Returns information that are ideal for attaching a block.
```
kwek20: is ideal; or maybe better "Return information that is used to attach a block in the network"?

sdk/src/types/api/core/response.rs
```diff
@@ -93,16 +94,26 @@ pub struct BaseTokenResponse {
     pub use_metric_prefix: bool,
 }
 
-/// Response of GET /api/core/v3/tips.
-/// Returns non-lazy tips.
+/// Response of
+/// - GET /api/core/v3/blocks/issuance
+/// Returns information that are ideal for attaching a block.
```
kwek20: same comment here

sdk/src/client/node_api/core/routes.rs
thibault-martinez: FYI, this is the documentation from the TIP so feel free to also comment there 🙏🏻

sdk/src/types/api/core/response.rs
thibault-martinez: If it's missing in the TIP, we should ask them to add it

DaughterOfMars: There's a comment but it's not useful

DaughterOfMars: I just added something, hopefully it's true lol

thibault-martinez (approved):

Thoralf-M (approved):

sdk/src/client/node_api/core/routes.rs
kwek20: Commented on the tip

kwek20 (approved):

qrayven (approved):

PR #1023, opened 2023-08-16
Remove uses of unmaintained actions-rs/cargo

thibault-martinez:


.github/actions/setup-rust/action.yml
```diff
@@ -50,7 +50,5 @@ runs:
 
     - name: Install ${{ inputs.install }}
       if: ${{ inputs.install != '' }}
-      uses: actions-rs/cargo@v1
-      with:
-        command: install
-        args: ${{ inputs.install }}
+      shell: bash
```
DaughterOfMars, line 53: Why here? And is this needed in other places?

thibault-martinez: It was only complaining here, I think it's because it's an action and not a workflow

Thoralf-M (approved):

abdulmth (approved):

DaughterOfMars (approved):

PR #1044, opened 2023-08-21
Some mana nits

thibault-martinez:


sdk/src/types/block/rand/mana.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::types::block::{mana::ManaAllotment, rand::output::rand_account_id};
+
+/// Generates a random mana allotment.
+pub fn rand_mana_allotment(mana: u64) -> ManaAllotment {
+    ManaAllotment::new(rand_account_id(), mana).unwrap()
```
Thoralf-M: Why don't make the amount also random?

thibault-martinez: I'm assuming that in future tests we'd need to provide an actual value to test edge cases

Thoralf-M: But then just not use the random function? account id should probably also be tested in some cases

thibault-martinez: Yeah ok, I'll change it later

thibault-martinez: Changed

Thoralf-M (dismissed):

Thoralf-M (dismissed):

sdk/src/types/block/rand/mana.rs
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::types::block::{
+    mana::ManaAllotment,
+    rand::{number::rand_number, output::rand_account_id},
+};
+
+/// Generates a random mana allotment.
+pub fn rand_mana_allotment() -> ManaAllotment {
+    ManaAllotment::new(rand_account_id(), rand_number()).unwrap()
```
Thoralf-M: rand_number() should consider the theoretical mana max

thibault-martinez: Done

kwek20 (commented):
Theyre just called Allotment in the TIPs, did you request the rename there as well?

sdk/src/types/block/mana/allotment.rs
```diff
@@ -8,34 +8,34 @@ use packable::{
     Packable,
 };
 
-use super::MAX_THEORETICAL_MANA;
+use super::THEORETICAL_MANA_MAX;
```
DaughterOfMars, line 11: I prefer the old name

thibault-martinez: I don't really care if it's a prefix or suffix but I do want consistency and most of our constants have a suffix, there is even a `COUNT_MAX` in the same file. There is a mix of both in the repository, if we all agree on what we prefer I can do a PR to change them all.

DaughterOfMars: bleh, okay, it's not that important

DaughterOfMars (approved):

abdulmth (approved):

sdk/src/types/block/mana/allotment.rs
thibault-martinez: I opened an issue anyway, it's all a bit inconsistent atm

Thoralf-M (approved):

kwek20 (approved):

qrayven (approved):

PR #1047, opened 2023-08-21
Fix Account::prepareBurn() return type

Thoralf-M:
# Description of change

Fix Account::prepareBurn() return type

## Type of change

Running an example which calls it

## Change checklist

### DIFF
--- a/bindings/nodejs/lib/wallet/account.ts
+++ b/bindings/nodejs/lib/wallet/account.ts
@@ -121,12 +121,12 @@ export class Account {
      * @param burn The outputs to burn
      * @param transactionOptions The options to define a `RemainderValueStrategy`
      * or custom inputs.
-     * @returns The resulting transaction.
+     * @returns The prepared transaction.
      */
     async prepareBurn(
         burn: Burn,
         transactionOptions?: TransactionOptions,
-    ): Promise<Transaction> {
+    ): Promise<PreparedTransaction> {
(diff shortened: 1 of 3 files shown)


bindings/nodejs/lib/wallet/account.ts
```diff
@@ -121,12 +121,12 @@ export class Account {
...
      * @param transactionOptions The options to define a `RemainderValueStrategy`
      * or custom inputs.
-     * @returns The resulting transaction.
+     * @returns The prepared transaction.
      */
     async prepareBurn(
         burn: Burn,
         transactionOptions?: TransactionOptions,
-    ): Promise<Transaction> {
+    ): Promise<PreparedTransaction> {
```
thibault-martinez, line 129: Not sure we can do that, breaking change

Thoralf-M: How can it break anything, if nothing can work with it, like it is without this change?

thibault-martinez: What would happen if I call this function now? Does it crash ?

Thoralf-M: Apparently `plainToInstance()` doesn't crash even though it returns a wrong object
But you can't call `.send()` on the resulting object, because then the TS compiler complains

thibault-martinez (approved):

DaughterOfMars (approved):

PR #1056, opened 2023-08-21
Simplify nodejs client and wallet

### LINKED ISSUE #699: Investigate: Nodejs bindings wrapper type
Investigate if we can change the bindings wrappers into something more usable.
Currently its a type wrapped in a struct, that aint pretty

See #679

DaughterOfMars:
# Description of change

This PR cleans up the client and wallet nodejs rust bindings.

Closes #699


thibault-martinez (commented):
Do we want to do something similar with `SecretManagerMethodHandler`? It has internal Arc/RwLock. Or why is this one different?

DaughterOfMars:
They all have additional internal locking. The difference is that the secret manager doesn't have the outer Arc with the `Option` inside to control whether it's created or not

thibault-martinez:
Did you say you needed help to try it or is this completely ready to be reviewed/merged?

DaughterOfMars:
Yeah I need help ensuring the nodejs side is correct.

Thoralf-M:
Maybe enable the CI for pull requests to the 1.1 branch

thibault-martinez:
> Maybe enable the CI for pull requests to the 1.1 branch

They are already enabled

Thoralf-M:
> They are already enabled

Why did nothing run on this PR then?

DaughterOfMars:
Because I did a base change and github is bad

DaughterOfMars:
```Jest has detected the following 1 open handle potentially keeping Jest from exiting:```

I think this is because we are no longer returning a promise, so evidence that nodejs needs changes.

**Comment by kwek20** [comment]
> `Jest has detected the following 1 open handle potentially keeping Jest from exiting:`
> 
> I think this is because we are no longer returning a promise, so evidence that nodejs needs changes.

No i've tried to solve this before using the .finish() on client, but i dont think its actually related to our code

**Comment by Thoralf-M** [comment]
Tests need to be updated because the error message changed

File: bindings/wasm/src/wallet.rs [hunk 1]
```diff
@@ -38,38 +57,34 @@ pub fn create_wallet(options: String) -> Result<WalletMethodHandler, JsValue> {
         .map_err(|e| e.to_string())?;
 
     Ok(WalletMethodHandler {
-        wallet: Arc::new(Mutex::new(Some(wallet_method_handler))),
+        wallet: Arc::new(RwLock::new(Some(wallet_method_handler))),
```
DaughterOfMars: Why use the std version?

kwek20: we also do in nodejs. but the std one is not async which is the main reason

DaughterOfMars: Yeah but we are in async fns...

kwek20: .client() isnt, just like in rust

bindings/nodejs/src/client.rs
```diff
@@ -60,71 +58,69 @@ impl ClientMethodHandler {
...
+                        let cb = callback.into_inner(&mut cx);
+                        let this = cx.undefined();
+
+                        let args = [
+                            if is_error {
+                                cx.string(response.clone()).upcast::<JsValue>()
+                            } else {
+                                cx.undefined().upcast::<JsValue>()
+                            },
+                            cx.string(response).upcast::<JsValue>(),
```
Alex6323: just for my understanding ... as it's wasn't changed in this PR ... but why do we need line 101?

bindings/wasm/src/wallet.rs
DaughterOfMars:
```rust
pub async fn get_client(method_handler: &WalletMethodHandler) -> Result<ClientMethodHandler, JsValue> {
```

bindings/nodejs/src/client.rs
DaughterOfMars:
```
callClientMethod(method, handler, (error: Error, result: string)
```

The TS expects that tuple so this is building that.

PR #1057, opened 2023-08-21
Storage migration

Thoralf-M:
# Description of change

Add storage migration for chrysalis DBs and stronghold snapshots to migrate to stardust.
Only accounts and their addresses will be migrated to the new format, all other data will be merged into a HashMap, so historically data like transaction ids can still be retrieved from there.

Is the stronghold migration automatically like it's now fine or do we want a separate method as for the storage? If so, it should probably modify the snapshot file we want to use to restore

## Type of change

Added tests

## Change checklist


bindings/nodejs/lib/wallet/wallet.ts
```diff
@@ -148,6 +148,18 @@ export class Wallet {
         return this.methodHandler.getClient();
     }
 
+    /**
+     * Get chrysalis data.
+     */
+    // TODO: is this the correct type?
```
DaughterOfMars: is it?

sdk/src/types/block/output/feature/mod.rs
```diff
@@ -37,6 +37,7 @@ pub enum Feature {
 }
 
 impl PartialOrd for Feature {
+    #[allow(clippy::incorrect_partial_ord_impl_on_ord_type)]
```
DaughterOfMars: Or fix it?

sdk/src/wallet/storage/manager.rs
```diff
@@ -112,6 +112,20 @@ impl StorageManager {
         let key = format!("{ACCOUNT_INDEXATION_KEY}{account_index}-{ACCOUNT_SYNC_OPTIONS}");
         self.get(&key).await
     }
+
+    #[cfg(not(target_family = "wasm"))]
+    pub(crate) async fn set_chrysalis_data(
```
DaughterOfMars: These feel like they should not exist.

sdk/tests/wallet/mod.rs
```diff
@@ -9,6 +9,9 @@ mod backup_restore;
 mod balance;
 mod bech32_hrp_validation;
 mod burn_outputs;
+#[cfg(not(target_os = "windows"))]
```
DaughterOfMars, line 12: why no windows?

DaughterOfMars (commented):
Can you give a high-level explanation of what this is doing and why?

sdk/src/types/block/output/feature/mod.rs
Thoralf-M: But what it suggests would actually be wrong, no?

sdk/tests/wallet/mod.rs
Thoralf-M: Because rocksdb expect different files there

sdk/src/types/block/output/feature/mod.rs
DaughterOfMars: no?

sdk/src/wallet/storage/manager.rs
Thoralf-M: Removed

sdk/src/types/block/output/feature/mod.rs
Thoralf-M: It suggested `Some(self.cmp(other))`, but `cmp()` calls `partial_cmp()`

sdk/src/wallet/storage/manager.rs
```diff
@@ -113,14 +113,6 @@ impl StorageManager {
...
 
-    #[cfg(not(target_family = "wasm"))]
-    pub(crate) async fn set_chrysalis_data(
-        &self,
-        chrysalis_data: std::collections::HashMap<String, String>,
-    ) -> crate::wallet::Result<()> {
-        self.set(CHRYSALIS_STORAGE_KEY, &chrysalis_data).await
-    }
-
     pub(crate) async fn get_chrysalis_data(
```
DaughterOfMars: Same?

sdk/src/types/block/output/feature/mod.rs
DaughterOfMars: So make it not?

Thoralf-M: But that would change `cmp()` then

DaughterOfMars: Yes, you should flip the impls

Thoralf-M: But that's different behavior then and could break something, no?

bindings/nodejs/CHANGELOG.md
```diff
@@ -21,6 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## 1.0.6 - 2023-MM-DD
 
+### Added
+
+- `migrateDbFromChrysalisToStardust` function;
+- `Wallet::getChrysalisData`;
```
```suggestion
- `Wallet::getChrysalisData` method;
```

bindings/nodejs/lib/bindings.ts
```diff
@@ -27,6 +27,7 @@ const {
     getClientFromWallet,
     getSecretManagerFromWallet,
     migrateStrongholdSnapshotV2ToV3,
+    migrateDbFromChrysalisToStardust,
```
thibault-martinez: Ignore if you want but we haven't put the From in the stronghold migration

bindings/nodejs/lib/wallet/wallet.ts
```diff
@@ -148,6 +148,17 @@ export class Wallet {
         return this.methodHandler.getClient();
     }
 
+    /**
+     * Get chrysalis data.
+     */
+    async getChrysalisData(): Promise<Record<string, string>> {
```
thibault-martinez, line 154: Is Record a hashmap?

Thoralf-M: Not necessarily I think, but it's a type that can be used for this https://stackoverflow.com/a/62650503

bindings/nodejs/lib/bindings.ts
Thoralf-M: Removed it

sdk/src/types/block/output/feature/mod.rs
DaughterOfMars: no, that's why clippy is complaining 😆

Thoralf-M: Yes now I see it as well, somehow I thought `cmp()` would compare also the content and not only the `kind()`

Thoralf-M: Done

sdk/src/wallet/core/mod.rs
```diff
@@ -200,6 +203,11 @@ impl<S: SecretManage> WalletInner<S> {
         Ok(Client::generate_mnemonic()?)
     }
 
+    #[cfg(feature = "storage")]
+    pub async fn get_chrysalis_data(&self) -> crate::wallet::Result<Option<std::collections::HashMap<String, String>>> {
```
thibault-martinez: Is there nowhere we could put it in the storage module?

thibault-martinez (commented):
Entry in the main changelog?

sdk/src/wallet/core/mod.rs
Thoralf-M: Like `impl Wallet{}` in the storage module?

thibault-martinez: Or core/operations/storage.rs idk, somewhere we already are under the storage feature

sdk/src/wallet/storage/constants.rs
```diff
@@ -31,3 +31,5 @@ pub(crate) const DATABASE_SCHEMA_VERSION_KEY: &str = "database-schema-version";
 pub(crate) const PARTICIPATION_EVENTS: &str = "participation-events";
 #[cfg(feature = "participation")]
 pub(crate) const PARTICIPATION_CACHED_OUTPUTS: &str = "participation-cached-outputs";
+
+pub(crate) const CHRYSALIS_STORAGE_KEY: &str = "CHRYSALIS_DATA";
```
thibault-martinez: Curious as to why it's uppercase?

Thoralf-M: Maybe because consts are uppercase and then I just continued :trollface:

Thoralf-M:
```suggestion
pub(crate) const CHRYSALIS_STORAGE_KEY: &str = "chrysalis-data";
```

sdk/src/wallet/migration/mod.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2023 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+#[cfg(not(target_family = "wasm"))]
+pub(crate) mod chrysalis;
```
thibault-martinez, line 5: Nit but I'd make this private and export `key_to_chrysalis_key` if needed, feel free to ignore

sdk/src/wallet/core/mod.rs
Thoralf-M: So create a new file just for this?

thibault-martinez: `core/operations/storage.rs` already exists?

thibault-martinez: Or idk just add it to an existing file in the storage module. It just feels weird to have this random storage function here

sdk/src/wallet/migration/mod.rs
Thoralf-M: Would rather keep like it is, it's also `key_to_chrysalis_key` and reexporting them separated would be more code

sdk/src/wallet/core/mod.rs
Thoralf-M: Looked into another folder, it exists already, moved the function to `core/operations/storage.rs`

sdk/src/wallet/migration/mod.rs
thibault-martinez: Fair enough

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,311 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use core::str::FromStr;
```
thibault-martinez: import this from std

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,311 @@
...
+    outputs: HashMap<String, Value>,
+    locked_outputs: HashSet<String>,
+    unspent_outputs: HashMap<String, Value>,
+    transactions: HashMap<String, Value>,
+    pending_transactions: HashSet<String>,
+    incoming_transactions: HashMap<String, Value>,
+    native_token_foundries: HashMap<String, Value>,
+}
+
+pub async fn migrate_db_chrysalis_to_stardust(storage_path: String, password: Option<Password>) -> Result<()> {
```
thibault-martinez: Take an impl Into<Path> ?

sdk/tests/wallet/chrysalis_migration.rs
```diff
@@ -0,0 +1,221 @@
...
+use crate::wallet::common::{setup, tear_down};
+
+#[tokio::test]
+async fn migrate_chrysalis_db() -> Result<()> {
+    let storage_path = "migrate_chrysalis_db/db";
+    setup(storage_path)?;
+    // Copy db so the original doesn't get modified
+    copy_folder("./tests/wallet/fixtures/chrysalis-db/db", storage_path).unwrap();
+
+    migrate_db_chrysalis_to_stardust("migrate_chrysalis_db".into(), None).await?;
```
thibault-martinez: then can remove into here and following

sdk/tests/wallet/chrysalis_migration.rs
```diff
@@ -0,0 +1,221 @@
...
+        "rms1qql3h5vxh2sxa93yadh7f4rkr7f9g9e65wlytazeu688mpcvhvmd2xvfq8y"
+    );
+    assert_eq!(bob_acc_details.internal_addresses().len(), 1);
+    assert_eq!(
+        bob_acc_details.internal_addresses()[0].address().try_to_bech32("rms")?,
+        "rms1qq4c9kl7vz0yssjw02w7jda56lec4ss3anfq03gwzdxzl92hcfjz7daxdfg"
+    );
+
+    let chrysalis_data = wallet.get_chrysalis_data().await?;
+    assert!(!chrysalis_data.unwrap().is_empty());
```
thibault-martinez: We should be able to test at least one thing from this?

sdk/tests/wallet/chrysalis_migration.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    // // Tests if setting stronghold password still works, commented because age encryption is very slow in CI
```
thibault-martinez: Can't we set the factor to 0 like other tests ?

sdk/tests/wallet/chrysalis_migration.rs
```diff
@@ -0,0 +1,221 @@
...
+#[tokio::test]
+async fn migrate_chrysalis_db_encrypted() -> Result<()> {
+    let storage_path = "migrate_chrysalis_db_encrypted/db";
+    setup(storage_path)?;
+    // Copy db so the original doesn't get modified
+    copy_folder("./tests/wallet/fixtures/chrysalis-db-encrypted/db", storage_path).unwrap();
+
+    migrate_db_chrysalis_to_stardust(
+        "migrate_chrysalis_db_encrypted".into(),
+        Some("password".to_string().into()),
```
thibault-martinez, line 120: Can we add a small test with invalid password and expect an error pls?

sdk/tests/wallet/chrysalis_migration.rs
```diff
@@ -0,0 +1,221 @@
...
+    //     "rms1qqqu7qry22f6v7d2d9aesny9vjtf56unpevkfzfudddlcq5ja9clv44sef6"
+    // );
+
+    tear_down("migrate_chrysalis_db_encrypted")
+}
+
+#[tokio::test]
+async fn migrate_chrysalis_stronghold() -> Result<()> {
+    iota_stronghold::engine::snapshot::try_set_encrypt_work_factor(0).unwrap();
+    let storage_path = "migrate_chrysalis_stronghold";
```
thibault-martinez, line 274: Does this db contain anything ?

Thoralf-M: It gets newly created

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: Why? We import from core in many places, also in the other migrations

thibault-martinez: It makes sense where we are in a no_std context, but that's not the case at all for the wallet so it's useless and noisy

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: That's not `Send` and then it returns an error

thibault-martinez: You can just add a Send bound, that's what we usually do when we use an impl Into

thibault-martinez: Also fine to keep like this, was a minor suggestion

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: Done

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: Done

sdk/tests/wallet/chrysalis_migration.rs
Thoralf-M: Done

sdk/tests/wallet/chrysalis_migration.rs
Thoralf-M: Added one

sdk/tests/wallet/chrysalis_migration.rs
Thoralf-M: Done

sdk/tests/wallet/chrysalis_migration.rs
Thoralf-M: Added a check with wrong password above

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -58,8 +58,12 @@ pub(crate) struct AccountDetailsDto {
     native_token_foundries: HashMap<String, Value>,
 }
 
-pub async fn migrate_db_chrysalis_to_stardust(storage_path: String, password: Option<Password>) -> Result<()> {
-    let storage_path = Path::new(&storage_path);
+pub async fn migrate_db_chrysalis_to_stardust(
+    storage_path: impl Into<String> + Send,
```
DaughterOfMars: `impl Into<PathBuf>` could also work probably

DaughterOfMars (dismissed):

thibault-martinez (dismissed):

bindings/nodejs/lib/wallet/wallet.ts
thibault-martinez: Is this resolved ?

Thoralf-M: I think so, I changed the type and removed the comment

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    if let Some(secret_manager_dto) = secret_manager_dto {
+        // This is required for the secret manager to be loaded
+        stardust_db.put(WALLET_INDEXATION_KEY, format!("{{ \"coinType\": {IOTA_COIN_TYPE}}}"))?;
+        stardust_db.put(SECRET_MANAGER_KEY, secret_manager_dto)?;
+    }
+
+    // set db migration version
+    let migration_version = MigrationVersion {
```
thibault-martinez: Reuse the existing one from migrate_4

sdk/src/wallet/migration/chrysalis.rs
thibault-martinez: We doing this or no?

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: You mean use the consts from here
```
impl MigrationData for Migrate {
    const ID: usize = 4;
    const SDK_VERSION: &'static str = "1.0.0-rc.0";
    const DATE: time::Date = time::macros::date!(2023 - 07 - 19);
```
? Because there is no built MigrationVersion

Thoralf-M: Done

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: Yes, updated

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    if let Some(secret_manager_dto) = secret_manager_dto {
+        // This is required for the secret manager to be loaded
+        stardust_db.put(WALLET_INDEXATION_KEY, format!("{{ \"coinType\": {IOTA_COIN_TYPE}}}"))?;
+        stardust_db.put(SECRET_MANAGER_KEY, secret_manager_dto)?;
+    }
+
+    // set db migration version
+    let migration_version = crate::wallet::migration::migrate_4::Migrate::version();
```
thibault-martinez, line 128: Now that I think about it, this seems super error-prone. We'd definitely forget to update it if ever we have a new migration. Use `latest_backup_migration_version` instead ?

thibault-martinez: I mean, maybe add a similar method or change it

thibault-martinez: Ah it's pub, my bad. Maybe just add one then, but you understand the idea, don't use migrate4, use latest.

Thoralf-M: No it has to be migrate 4, because the types are based on this version
Otherwise we would always need to update the chrysalis migration to the latest types

thibault-martinez: Ahh I see, my bad!

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,311 @@
...
+    incoming_transactions: HashMap<String, Value>,
+    native_token_foundries: HashMap<String, Value>,
+}
+
+pub async fn migrate_db_chrysalis_to_stardust(
+    storage_path: impl Into<PathBuf> + Send,
+    password: Option<Password>,
+) -> Result<()> {
+    let storage_path_string = storage_path.into();
+    let storage_path = Path::new(&storage_path_string);
```
thibault-martinez: You don't really need this extra step

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,311 @@
...
+        }
+    }
+    // Accounts must be ordered by index
+    new_accounts.sort_unstable_by_key(|a| a.index);
+    Ok((new_accounts, secret_manager_dto))
+}
+
+fn get_chrysalis_data(chrysalis_storage_path: &Path, password: Option<Password>) -> Result<HashMap<Vec<u8>, String>> {
+    let encryption_key = password.map(storage_password_to_encryption_key);
+    let chrysalis_db = DB::open_default(chrysalis_storage_path).unwrap();
```
thibault-martinez:
```suggestion
    let chrysalis_db = DB::open_default(chrysalis_storage_path)?;
```

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,311 @@
...
+    // iterate over all rocksdb keys
```
thibault-martinez: I'd put this comment right before the for

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,311 @@
...
+    let mut chrysalis_data = HashMap::new();
+    for item in chrysalis_db.iterator(IteratorMode::Start) {
+        let (key, value) = item.unwrap();
```
thibault-martinez:
```suggestion
        let (key, value) = item?;
```

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,310 @@
...
+    let mut chrysalis_data = HashMap::new();
+    // iterate over all rocksdb keys
+    for item in chrysalis_db.iterator(IteratorMode::Start) {
+        let (key, value) = item?;
+
+        let key_utf8 = String::from_utf8(key.to_vec()).map_err(|_| Error::Migration("invalid utf8".into()))?;
+        let value = if let Some(encryption_key) = &encryption_key {
+            let value_utf8 = String::from_utf8(value.to_vec()).map_err(|_| Error::Migration("invalid utf8".into()))?;
+            // "iota-wallet-key-checksum_value" is never an encrypted value
+            if serde_json::from_str::<Vec<u8>>(&value_utf8).is_ok() && key_utf8 != "iota-wallet-key-checksum_value" {
```
thibault-martinez: nit but could maybe rework it with an if let so we don't have to parse it L276

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,310 @@
...
+    chrysalis_data: &HashMap<Vec<u8>, String>,
+    storage_path: &Path,
+    // in stronghold the keys are hashed first
+    stronghold: bool,
+) -> Result<(Vec<AccountDetailsDto>, Option<String>)> {
+    let mut new_accounts: Vec<AccountDetailsDto> = Vec::new();
+    let mut secret_manager_dto: Option<String> = None;
+
+    let account_indexation_key = if stronghold {
+        key_to_chrysalis_key(b"iota-wallet-account-indexation")
```
thibault-martinez: I would have this function to take the bool, less outside ifs

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,310 @@
...
+                        .to_vec()
+                };
+
+                if let Some(account_data) = chrysalis_data.get(&account_key) {
+                    let account_data = serde_json::from_str::<serde_json::Value>(account_data)?;
+                    if secret_manager_dto.is_none() {
+                        let dto = match &account_data["signerType"]["type"].as_str() {
+                            Some("Stronghold") => format!(
+                                r#"{{"Stronghold": {{"password": null, "timeout": null, "snapshotPath": "{}/wallet.stronghold"}} }}"#,
+                                storage_path.to_string_lossy()
```
thibault-martinez, line 163: Should we move it though? Wouldn't it end up inside the database folder itself otherwise?

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    log::debug!(
+        "Chrysalis data: {}",
+        serde_json::to_string_pretty(&chrysalis_data_with_string_keys)?
+    );
+
+    let mut opts = Options::default();
+    opts.create_if_missing(true);
+    opts.create_missing_column_families(true);
+    let stardust_db = DB::open(&opts, storage_path_string).unwrap();
```
thibault-martinez:
```suggestion
    let stardust_db = DB::open(&opts, storage_path_string)?;
```

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: Move it where? Like this it would be the same as in chrysalis

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    // store chrysalis data in a new key
+    stardust_db.put(
```
thibault-martinez: What about encryption? Should we also encrypt the stardust db if a password is given?

sdk/src/wallet/migration/chrysalis.rs
thibault-martinez: Well not exactly. There used to be a /db for the database and the snapshot would be outside. Now it's together with the rocksdb files.

Thoralf-M: The rocksdb folder moved one folder up, but stronghold stays at the exact same place

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: Done

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: Done

bindings/nodejs/lib/wallet/wallet.ts
thibault-martinez: I'm resolving this for you then.

thibault-martinez (changes requested):
Clippy

thibault-martinez (changes requested):
CI failing

bindings/nodejs/lib/wallet/index.ts
```diff
@@ -5,3 +5,5 @@ export * from './account';
 export * from './wallet';
 export * from './wallet-method-handler';
 export * from '../types/wallet';
+import { migrateDbChrysalisToStardust } from '../bindings';
+export { migrateDbChrysalisToStardust };
```
thibault-martinez: Just curious, can't this be a single export from ?

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,314 @@
...
+    transactions: HashMap<String, Value>,
+    pending_transactions: HashSet<String>,
+    incoming_transactions: HashMap<String, Value>,
+    native_token_foundries: HashMap<String, Value>,
+}
+
+pub async fn migrate_db_chrysalis_to_stardust(
+    storage_path: impl Into<PathBuf> + Send,
+    password: Option<Password>,
+    new_db_encryption_key: impl Into<Option<Zeroizing<[u8; 32]>>> + Send,
```
thibault-martinez, line 68: Shouldn't this be a password?

Thoralf-M: No, https://github.com/iotaledger/iota-sdk/blob/5cfc4aa4cb8ef14708cd42ab313198536f6ee4fa/sdk/src/wallet/storage/options.rs#L44

thibault-martinez: Well yeah sure, but `migrate_db_chrysalis_to_stardust` is a bit more user-facing, we could convert a password to an encryption key, don't we have a function for that?

bindings/nodejs/lib/wallet/index.ts
Thoralf-M: Ah yeah probably can, will try

thibault-martinez (dismissed):

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: `with_encryption_key()` is the only way users can set the storage encryption key, we had `storage_password_to_encryption_key()` in Chrysalis, but I think Vlad wasn't happy with it and in Stardust we don't have anything

thibault-martinez: Alright 👍🏻

sdk/src/wallet/core/operations/stronghold_backup/stronghold_snapshot.rs
```diff
@@ -89,5 +98,123 @@ pub(crate) async fn read_data_from_stronghold_snapshot<S: 'static + SecretManage
...
+    // check if snapshot contains chrysalis data
+    if !keys.iter().any(|k| k == &wallet_indexation_key) {
+        return Ok(None);
+    }
+
+    let mut chrysalis_data: HashMap<Vec<u8>, String> = HashMap::new();
+    for key in keys {
+        let value = stronghold_store
+            .get(&key)
+            .map_err(|e| crate::wallet::Error::Client(Box::new(crate::client::Error::Stronghold(e.into()))))?;
```
DaughterOfMars: I would help clean up this code if we used this above

```rust
use crate::client::Error as ClientError;
use crate::wallet::Error as WalletError;
```

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,314 @@
...
+            ))
+        })
+        .collect::<Result<HashMap<String, String>>>()?;
+
+    log::debug!(
+        "Chrysalis data: {}",
+        serde_json::to_string_pretty(&chrysalis_data_with_string_keys)?
+    );
+
+    let stardust_db = crate::wallet::storage::adapter::rocksdb::RocksdbStorageAdapter::new(storage_path_string)?;
```
DaughterOfMars, line 95: Does this need to be feature gated?

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,314 @@
...
+        &mut pt,
+        &ct,
+        tag.as_slice().try_into().unwrap(),
+    )
+    .map_err(|e| Error::Migration(format!("{:?}", e)))?;
+
+    Ok(String::from_utf8_lossy(&pt).to_string())
+}
+
+pub(crate) fn key_to_chrysalis_key(key: &[u8], stronghold: bool) -> Vec<u8> {
```
DaughterOfMars: Should be private probably. Also can we name it a little better? Maybe `stardust_key_to_chrysalis` or just `to_chrysalis_key`

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -0,0 +1,314 @@
...
+        encryption_key.try_into().unwrap(),
+        &nonce.try_into().unwrap(),
+        &[],
+        &mut pt,
+        &ct,
+        tag.as_slice().try_into().unwrap(),
+    )
+    .map_err(|e| Error::Migration(format!("{:?}", e)))?;
+
+    Ok(String::from_utf8_lossy(&pt).to_string())
```
DaughterOfMars: Perhaps instead of lossy we create an error here like above?

sdk/src/wallet/migration/mod.rs
```diff
@@ -1,16 +1,20 @@
...
 mod migrate_0;
 mod migrate_1;
 mod migrate_2;
 mod migrate_3;
-mod migrate_4;
+pub(crate) mod migrate_4;
```
DaughterOfMars, line 10: Maybe instead we should have a `latest_storage_migration_version` fn

Thoralf-M: But we only need the migration _4 version, because the types have the same layout as in this migration, later version could change again

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: It's also used in stronghold_snapshot.rs, but renamed it

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: Yes added `#[cfg(feature = "rocksdb")]` above the function

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: Done

sdk/src/wallet/migration/mod.rs
DaughterOfMars: fair enough

DaughterOfMars (approved):

thibault-martinez (approved):

Issue #892, opened 2023-07-24
Add StakingFeature - nodejs


thibault-martinez:
https://github.com/iotaledger/tips-draft/blob/tip42/tips/TIP-0042/tip-0042.md#staking-feature

thibault-martinez:
Closed by https://github.com/iotaledger/iota-sdk/pull/1058

PR #1058, opened 2023-08-22
Add `StakingFeature`  Node.js

abdulmth:
#fixes https://github.com/iotaledger/iota-sdk/issues/892

### DIFF
--- a/bindings/nodejs/lib/types/block/output/feature.ts
+++ b/bindings/nodejs/lib/types/block/output/feature.ts
@@ -14,6 +14,7 @@ enum FeatureType {
     Metadata = 2,
     Tag = 3,
     BlockIssuer = 4,
+    Staking = 5,
 }
 
 abstract class Feature {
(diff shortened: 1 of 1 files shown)


bindings/nodejs/lib/types/block/output/feature.ts
```diff
@@ -90,6 +91,41 @@ class BlockIssuerFeature extends Feature {
...
 }
 
+/**
+ * Block Issuer feature.
+ */
+class StakingFeature extends Feature {
+    /**
+     * The amount of coins that are locked and staked in the containing account.
+     **/
+    readonly stakedAmount: bigint;
```
abdulmth: `bigint` instead of the type alias `u64` is maybe better for the user to be directly aware of the type, we might want to not use `u64` anymore because it can be confusing to the user as @kwek20 hinted to if I'm not mistaken.

bindings/nodejs/lib/types/block/output/feature.ts
```diff
@@ -90,6 +91,41 @@ class BlockIssuerFeature extends Feature {
     }
 }
 
+/**
+ * Block Issuer feature.
```
Thoralf-M: Wrong comment

Thoralf-M (dismissed):

bindings/nodejs/lib/types/block/output/feature.ts
thibault-martinez: Not fully sure to understand, these values are internally (rust) u64, so we should probably use u64? Even though in JS u64 point to bigint, these are not bigints.

abdulmth: yes, but when people want to create the value, they have to use the `bigint` constructor and `typeof value` would be `bigint`

thibault-martinez: Oh alright. Is there no other way to define `u64` than a simple typedef? A more strongly typed alternative? Otherwise these typedef seems a bit useless to me...

abdulmth: The type alias is just an indication that you can't use negative values or values bigger than the max of u64, but it doesn't enforce anything. I don't think we have any good alternative to using just `bigint` as it has many advantages, for example you can use normal arithmetic operation on bigint values. but if you wrap it somehow, you can't do that anymore.

thibault-martinez (dismissed):

DaughterOfMars (dismissed):

qrayven (dismissed):

bindings/nodejs/lib/types/block/output/feature.ts
thibault-martinez: Are we changing these then?

abdulmth: I changed them, here it goes with the sweet sweet reviews 🥲

thibault-martinez (approved):

Thoralf-M (approved):

qrayven (approved):

kwek20 (approved):

PR #1061, opened 2023-08-22
Add PublicKey type

### LINKED ISSUE #947: Update BlockIssuerFeature  - nodejs
https://github.com/iotaledger/tips-draft/pull/11/commits/0dfe5788ed3b4a596bf94a9db932c47229119bf0?short_path=e59aa6b#diff-e59aa6be7b496f4aed7fb5358177213dc4e66dde764876c61559bdf4c242ecec

Thoralf-M:
# Description of change

Add PublicKey type

Closes #947 

## Type of change

## Change checklist


thibault-martinez (changes requested):
CI failing

bindings/nodejs/examples/how_tos/sign_and_verify_ed25519/sign-ed25519.ts
```diff
@@ -62,11 +62,11 @@ async function run() {
             bip44Chain,
         );
         console.log(
-            `Public key: ${ed25519Signature.publicKey}\nSignature: ${ed25519Signature.signature}`,
+            `Public key: ${ed25519Signature.publicKey.publicKey}\nSignature: ${ed25519Signature.signature}`,
```
DaughterOfMars: bummer

bindings/nodejs/examples/how_tos/sign_and_verify_ed25519/verify-ed25519-signature.ts
```diff
@@ -24,7 +29,10 @@ function run() {
     try {
         const message = utf8ToHex(JSON.stringify(FOUNDRY_METADATA));
         const validSignature = Utils.verifyEd25519Signature(
-            new Ed25519Signature(PUBLIC_KEY, ED25519_SIGNATURE),
+            new Ed25519Signature(
+                new Ed25519PublicKey(PUBLIC_KEY),
```
DaughterOfMars: Can we just pass the inner type and wrap it in the constructor?

DaughterOfMars (dismissed):

qrayven (dismissed):

bindings/nodejs/lib/types/block/public-key.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+// Temp solution for not double parsing JSON
```
thibault-martinez: Temp? Should there be a TODO?

bindings/nodejs/lib/types/block/public-key.ts
```diff
@@ -0,0 +1,40 @@
...
+     */
+    publicKey: HexEncodedString;
+
+    constructor(publicKey: HexEncodedString) {
+        super(PublicKeyType.Ed25519);
+        this.publicKey = publicKey;
+    }
+}
+
+export { Ed25519PublicKey, PublicKey, PublicKeyType };
```
thibault-martinez: Do we not need the `Discriminator` thing? It's not very clear to me what it does and when we need it

bindings/nodejs/lib/types/block/public-key.ts
Thoralf-M: I copied it from somewhere else, but it was also wrong there I think
@kwek20 please confirm

bindings/nodejs/lib/types/block/public-key.ts
Thoralf-M: Also not clear to me if it's needed, especially if there is only a single variant atm 
@kwek20 do we need it?

abdulmth: If I'm not mistaken,  the Discriminator let the class transformer decides what class should it be used to create the intance depending on some value, usually the `type` property for example: 

```Typescript
const FeatureDiscriminator = {
    property: 'type',
    subTypes: [
        { value: SenderFeature, name: FeatureType.Sender as any },
        { value: IssuerFeature, name: FeatureType.Issuer as any },
        { value: MetadataFeature, name: FeatureType.Metadata as any },
        { value: TagFeature, name: FeatureType.Tag as any },
        { value: BlockIssuerFeature, name: FeatureType.BlockIssuer as any },
        { value: StakingFeature, name: FeatureType.Staking as any },
    ],
};
``` 
if the property `type` == `FeatureType.Sender` (which is 0) it will create an instance of class `SenderFeature` and so on.

bindings/nodejs/lib/types/block/public-key.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/**
+ * Ed25519 public key.
+ */
+class Ed25519PublicKey extends PublicKey {
+    /**
+     * The public key.
+     */
+    publicKey: HexEncodedString;
```
abdulmth:
```suggestion
    readonly publicKey: HexEncodedString;
```

bindings/nodejs/examples/how_tos/sign_and_verify_ed25519/sign-ed25519.ts
abdulmth: Maybe we can add a method like `toHexString`  or something to avoid this duplication and give a hint about the type?

Thoralf-M: > and give a hint about the type

What do you  mean with this?

bindings/nodejs/lib/types/block/public-key.ts
Thoralf-M: Why is it `name:` and not `type:`?

bindings/nodejs/examples/how_tos/sign_and_verify_ed25519/sign-ed25519.ts
abdulmth: only that the retuned type is string pretty much.

bindings/nodejs/lib/types/block/public-key.ts
abdulmth: That I don't know 😆

bindings/nodejs/examples/how_tos/sign_and_verify_ed25519/sign-ed25519.ts
Thoralf-M: Sounds stupid to me to add a method just to have a different name
Anyone else have an opinion on this?

abdulmth: yeah, it's not the nicest thing, but `publicKey.publicKey` is also not great :D

abdulmth: I don't mind it much though.

abdulmth (dismissed):

kwek20 (dismissed):

bindings/nodejs/lib/types/block/public-key.ts
kwek20: huh yea this makes no sense, should be gone. (so in other places)

bindings/nodejs/lib/types/block/public-key.ts
kwek20: it has a field: `property: 'type',` which sets based on which json property it runs. `value: and name:` are just parts of this class-transformer, its its values.

and yes its used for types with subclasses, youll see the `@type()` annotation usually with a discriminator

thibault-martinez (dismissed):

bindings/nodejs/examples/how_tos/sign_and_verify_ed25519/sign-ed25519.ts
DaughterOfMars: Accessor method?

Thoralf-M: We removed them for the other types since the field is public

DaughterOfMars: Sure, but in this case I think it would make more sense than accessing it like this

Thoralf-M: So want me to add this?
```TS
    /**
     * The hex encoded public key.
     */
    toHexString(): HexEncodedString {
        return this.publicKey;
    }
```

DaughterOfMars: That's not what I meant. I think it would be better to call it like this ```ed25519Signature.publicKey()```

Thoralf-M: `publicKey()` is not allowed, because the field is already named like this, added `ed25519PublicKey()` instead

abdulmth (dismissed):

abdulmth:
CI is failing though, probably something is wrong now after fixing the signature dto.

bindings/nodejs/examples/how_tos/sign_and_verify_ed25519/sign-ed25519.ts
```diff
@@ -62,11 +62,13 @@ async function run() {
             bip44Chain,
         );
         console.log(
-            `Public key: ${ed25519Signature.publicKey.publicKey}\nSignature: ${ed25519Signature.signature}`,
+            `Public key: ${ed25519Signature.ed25519PublicKey()}\nSignature: ${
```
DaughterOfMars: I feel like just `publicKey` is fine, since this is a `ed25519Signature`

abdulmth: I think Thoralf already commented somewhere that `publicKey` is already taken by the property.

Thoralf-M: Yes unfortunately `publicKey` can't be used

DaughterOfMars: Oh, double bummer. Maybe make it non public then?

DaughterOfMars: You know what, nvm

abdulmth: I don't think making it private would work, it's a JS issue not a TS.

DaughterOfMars (dismissed):

bindings/nodejs/lib/types/block/public-key.ts
```diff
@@ -0,0 +1,45 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+import { plainToInstance } from 'class-transformer';
```
thibault-martinez: warning

bindings/nodejs/examples/how_tos/sign_and_verify_ed25519/sign-ed25519.ts
```diff
@@ -48,7 +48,7 @@ async function run() {
         // A mnemonic can be generated with `Utils.generateMnemonic()`.
         // Store the mnemonic in the Stronghold snapshot, this needs to be done only the first time.
         // The mnemonic can't be retrieved from the Stronghold file, so make a backup in a secure place!
-        await secretManager.storeMnemonic(process.env.MNEMONIC);
+        // await secretManager.storeMnemonic(process.env.MNEMONIC);
```
thibault-martinez: Why?

bindings/nodejs/lib/types/block/signature.ts
```diff
@@ -1,8 +1,13 @@
 // Copyright 2023 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-// Temp solution for not double parsing JSON
+import { plainToInstance, Type } from 'class-transformer';
```
thibault-martinez: warnings

bindings/nodejs/examples/how_tos/sign_and_verify_ed25519/sign-ed25519.ts
Thoralf-M: Disabled for testing, removed again

thibault-martinez (dismissed):

abdulmth (dismissed):

bindings/nodejs/lib/types/block/signature.ts
```diff
@@ -33,17 +33,24 @@ class Ed25519Signature extends Signature {
     /**
      * The public key.
      */
-    publicKey: HexEncodedString;
+    publicKey: Ed25519PublicKey;
```
kwek20: change to readonly and remove the getter? weve been switching to that style

Thoralf-M: That's out of scope I would say, we have an own issue for the signature https://github.com/iotaledger/iota-sdk/issues/880

Thoralf-M: Actually this change is wrong I think

thibault-martinez:
Why closed ?

Thoralf-M:
I didn't close it, just force pushed so there are no changes anymore :joy: Will update soon with the updated code

bindings/nodejs/lib/types/block/signature.ts
Thoralf-M: Undid the change

sdk/src/types/block/macro.rs
DaughterOfMars, line 1: What happened here? Feels like a bad merge

sdk/src/types/block/public_key/ed25519.rs
```diff
@@ -66,15 +66,12 @@ pub(crate) mod dto {
     #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
     #[serde(rename_all = "camelCase")]
     pub struct Ed25519PublicKeyDto {
```
DaughterOfMars: DTO isn't needed anymore

sdk/src/types/block/macro.rs
DaughterOfMars: Or was this just not merged up yet?

sdk/src/types/block/public_key/mod.rs
```diff
@@ -82,6 +82,8 @@ pub(crate) mod dto {
             }
         }
     }
+
+    impl_serde_typed_enum_dto!(PublicKey, PublicKeyDto, "public key");
```
DaughterOfMars: I think this was intended to be used for the variants

sdk/tests/types/public_key.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use iota_sdk::types::block::public_key::PublicKey;
+
+#[test]
+#[cfg(feature = "serde")]
+fn public_key_de_serialization() {
```
DaughterOfMars:
```suggestion
fn public_key_serde_roundtrip() {
```

sdk/src/types/block/public_key/mod.rs
Thoralf-M: I added a new one

DaughterOfMars: I see, but it isn't correct to do it this way because if the type field is wrong this just ignores it.

Thoralf-M: Yes not sure if there is a good way to do this with a macro then
Should I implement the conversion manually only for this type and remove the macro?

DaughterOfMars: I just used the macro on the individual type DTOs

Thoralf-M: But this doesn't work with this change? `Ed25519PublicKey` should be serialized just as `String` in all cases apart from the public keys field in the issuer keys, there the type field should be added

DaughterOfMars: Hm, that is a problem indeed. In that case I think you have to just define the manual deserialization which checks the type field

sdk/src/types/block/public_key/ed25519.rs
Thoralf-M: Removed

DaughterOfMars (approved):

kwek20 (approved):

DaughterOfMars (approved):

PR #1064, opened 2023-08-22
Add `ManaAllotment` in Node.js

abdulmth:
fixes https://github.com/iotaledger/iota-sdk/issues/914

I think that's all what we need in Node.js since we don't need the `ManaAllotments` as in Rust because it can be just represented as `ManaAllotment[]`.

### DIFF
--- a/bindings/nodejs/lib/types/block/mana-allotment.ts
+++ b/bindings/nodejs/lib/types/block/mana-allotment.ts
@@ -25,5 +25,5 @@ class ManaAllotment {
 
 export {
   ManaAllotment
-}
+};
 
--- a/bindings/nodejs/lib/types/block/mana-allotment.ts
+++ b/bindings/nodejs/lib/types/block/mana-allotment.ts
@@ -1,29 +1,26 @@
 // Copyright 2023 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-import { AccountId } from "..";
+import { AccountId } from '..';
 
 /**
  * An allotment of Mana which will be added upon commitment of the slot in which the containing transaction was issued,
  * in the form of Block Issuance Credits to the account.
  */
 class ManaAllotment {
-  /**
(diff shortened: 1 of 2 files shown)


bindings/nodejs/lib/types/block/mana-allotment.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+class ManaAllotment {
+    /**
+     * The Account to allot Mana to.
+     */
+    readonly accountId: AccountId;
+    /**
+     * The Amount of Mana to allot.
+     */
+    readonly mana: bigint;
```
thibault-martinez: Same discussion as before but I actually still don't understand. Why not u64? What is different between this mana and the one on the Output that is a u64?

abdulmth: I thought we wanted to change all of them to `bigint` 😅

thibault-martinez: But we just introduced u64 a couple of days ago?

abdulmth: Changed it

thibault-martinez (approved):

qrayven (approved):

Thoralf-M (approved):

kwek20 (commented):
We could add a type alias Mana Allotments  anyway to mimic rust code more.

thibault-martinez:
> We could add a type alias Mana Allotments anyway to mimic rust code more.

Have we done that with the other list/set types? Like Inputs, Outputs, NativeTokens, ... ?

DaughterOfMars (approved):

abdulmth:
> We could add a type alias Mana Allotments anyway to mimic rust code more.

In Rust it's necessary, but I'm not sure how the type alias would bring any benefit to TS. I would just use a normal array.

PR #1065, opened 2023-08-22
Refactor rent cost traits

### LINKED ISSUE #1124: Add `vByteFactorDelegation` to RentStructure
https://github.com/iotaledger/tips-draft/pull/24/files/51c445ef58ae9610de2d8c38c7f68380405ff5e1..93227c50259fdc413b1beaec56c35360a14a3efb

DaughterOfMars:
# Description of change

This PR refactors rent cost so that it can be deeply derived and adds some helpful builder functionality. It also allows builder types to calculate rent cost so that you no longer need to convert back and forth between builder and output.

## Related Issues

Closes #1124


sdk/src/types/block/output/basic.rs
```diff
@@ -152,8 +156,34 @@ impl BasicOutputBuilder {
         self
     }
 
+    /// Adds a storage deposit if one is needed to cover the current amount.
+    pub fn with_sufficient_storage_deposit(
```
Thoralf-M: From the comment and function name it's a bit unclear to me why it adding a StorageDepositReturnUnlockCondition if amount < rent_cost

Not clear to me at all why we want this function

sdk/src/types/block/output/basic.rs
```diff
@@ -152,8 +156,34 @@ impl BasicOutputBuilder {
...
+        rent_structure: RentStructure,
+        token_supply: u64,
+    ) -> Result<Self, Error> {
+        Ok(match self.amount {
+            OutputBuilderAmount::Amount(amount) => {
+                let rent_cost = self.rent_cost(rent_structure);
+                if amount < rent_cost {
+                    self.with_amount(amount + rent_cost).replace_unlock_condition(
+                        StorageDepositReturnUnlockCondition::new(return_address, rent_cost, token_supply)?,
+                    )
```
Thoralf-M: This could be wrong if there was no StorageDepositReturnUnlockCondition before, no? Because adding it increases the required storage deposit and if `amount` is not enough to cover it, it wouldn't be enough amount at the end

sdk/src/types/block/output/basic.rs
DaughterOfMars: It does exactly the same thing as `MinimumRequiredStorageDepositBasicOutput` used to.

sdk/src/types/block/output/basic.rs
DaughterOfMars: Well this does a replace, not an add

Thoralf-M: But it's adding if there is none yet

DaughterOfMars: Hmm, yes, I think you are right. I will change it

sdk/src/types/block/output/basic.rs
Thoralf-M: Yes, but just reading these two lines I would assume it just increases the amount to the minimum required storage deposit and that's it, but it also adds a storage deposit return unlock condition if not enough amount is already there

sdk/src/types/block/output/basic.rs
Thoralf-M: If the amount is not 0 it would add more than actually required
`amount + rent_cost`, because to return something we just need the minimum required storage deposit of an output, so the difference between that and the rent_cost could be subtracted from the `amount` part of `amount + rent_cost` (if `amount` would then be negative, this number would need to be added to the return amount)

So for example provided amount: 10
`rent_cost_basic` =  min required storage deposit basic output: 42600
`rent_cost_basic_sdr` = min required storage deposit basic output with storage deposit return: 42800

Then `amount + rent_cost_basic_sdr` == 10+42800 => 42810
But 42800 would be sufficient, because we can just have have 42790 in the storage deposit return unlock condition

sdk/src/types/block/output/basic.rs
DaughterOfMars: Well increasing the amount would be invalid, because that would change the ultimate transferred amount. I get what you're saying though, I suppose. Do you have a suggestion?

sdk/src/types/block/output/basic.rs
DaughterOfMars: Yeah but you're trying to send a specific amount to someone and this would change that

Thoralf-M: How does this change? The recipient would still get exactly the same amount

sdk/src/types/block/output/basic.rs
Thoralf-M: `Adds a storage deposit`
Did you mean a storage deposit return unlock condition?
I think something like this would make it more clear
`If amount < rent_cost, adds a storage deposit return unlock condition for rent_cost and increases the amount to amount + rent_cost`

sdk/src/types/block/output/basic.rs
DaughterOfMars: Oh, I misread. You are correct.

sdk/src/types/block/output/basic.rs
DaughterOfMars: Hopefully this is better now

sdk/src/types/block/address/ed25519.rs
```diff
@@ -24,6 +24,11 @@ impl Ed25519Address {
     pub fn new(address: [u8; Self::LENGTH]) -> Self {
         Self::from(address)
     }
+
+    #[allow(dead_code)]
```
thibault-martinez: Is this really needed since we actually use it?

DaughterOfMars: features

sdk/src/types/block/output/basic.rs
```diff
@@ -172,18 +172,32 @@ impl BasicOutputBuilder {
...
-                    let rent_cost = self.rent_cost(rent_structure);
+                    // Get the rent cost of the output with the added storage deposit return unlock condition
+                    let rent_cost_with_sdruc = self.rent_cost(rent_structure);
+                    // If the return rent cost and amount are less than the required min
+                    let (amount, sdruc_amount) = if rent_cost_with_sdruc >= return_rent_cost + amount {
+                        // Then sending rent_cost_with_sdruc covers both minimum requirements
+                        (rent_cost_with_sdruc, rent_cost_with_sdruc - amount)
+                    } else {
+                        // Otherwise we must use the total of the return minimum and the original amount
+                        // which is unfortunately more than the rent_cost_with_sdruc
```
Thoralf-M: Why `unfortunately`? :D Otherwise it's looking good now :+1:

sdk/src/types/block/output/basic.rs
Thoralf-M: Looks correct to me now

sdk/src/types/block/output/basic.rs
DaughterOfMars: just because it has to add more that way

sdk/src/types/block/output/feature/block_issuer.rs
```diff
@@ -48,6 +49,18 @@ impl BlockIssuerFeature {
...
 
+impl Rent for BlockIssuerFeature {
+    fn build_weighted_bytes(&self, builder: &mut RentBuilder) {
+        builder
+            // Feature Type
+            .data_field::<u8>()
+            // Expiry Slot
+            .data_field::<SlotIndex>()
+            // Public Keys
+            .packable_data_field(&self.public_keys);
```
Thoralf-M: The keys here should be weighted with `block_issuer_key`

DaughterOfMars: God damn it

Thoralf-M (dismissed):

thibault-martinez (changes requested):
Conflicts

sdk/src/wallet/account/operations/output_claiming.rs
```diff
@@ -294,15 +294,10 @@ where
...
-        } else {
-            required_amount_for_nfts
-                + BasicOutputBuilder::new_with_amount(Output::AMOUNT_MIN)
-                    .with_native_tokens(option_native_token.into_iter().flatten())
-                    .rent_cost(rent_structure)
-        };
+        // Get the required amount using a sufficient storage deposit
+        let mut required_amount = BasicOutputBuilder::new_with_amount(required_amount_for_nfts)
+            .with_native_tokens(option_native_token.into_iter().flatten())
+            .min_storage_deposit_amount(rent_structure, token_supply)?;
```
Thoralf-M: This change doesn't seem right, before depending on `if possible_additional_inputs.is_empty() {` it did something different

sdk/src/wallet/account/operations/output_claiming.rs
```diff
@@ -318,10 +313,9 @@ where
                 };
                 // Recalculate every time, because new inputs can also add more native tokens, which would increase
                 // the required storage deposit
-                required_amount = required_amount_for_nfts
-                    + BasicOutputBuilder::new_with_amount(Output::AMOUNT_MIN)
-                        .with_native_tokens(option_native_token.into_iter().flatten())
-                        .rent_cost(rent_structure);
+                required_amount = BasicOutputBuilder::new_with_amount(required_amount_for_nfts)
+                    .with_native_tokens(option_native_token.into_iter().flatten())
+                    .min_storage_deposit_amount(rent_structure, token_supply)?;
```
Thoralf-M: min_storage_deposit_amount() just overwrites `required_amount_for_nfts` and doesn't add to it?

DaughterOfMars: No, it adds to it using the logic from the sufficient fn

sdk/src/wallet/account/operations/output_claiming.rs
DaughterOfMars: Yeah, but I'm pretty sure that was wrong

Thoralf-M: I don't think it was wrong, it was for the case when there are only NFTs to claim and no other inputs are available
If you add the min_storage_deposit_amount() for a basic output there, you make this amount additional required even though it's not
https://github.com/iotaledger/iota-sdk/commit/33069b6ef1b1709eac2dc65a59a92b9414d38353

sdk/src/types/block/output/basic.rs
```diff
@@ -151,8 +160,73 @@ impl BasicOutputBuilder {
...
+                        )?)
+                } else {
+                    self
+                }
+            }
+            OutputBuilderAmount::MinimumStorageDeposit(_) => self,
+        })
+    }
+
+    pub fn min_storage_deposit_amount(&self, rent_structure: RentStructure, token_supply: u64) -> Result<u64, Error> {
```
Thoralf-M: From the name it's very unclear to me why it calls `with_sufficient_storage_deposit()` and not just returns the minimum required storage deposit amount without potentially adding a SDRUC
`with_sufficient_storage_deposit()` just returns the amount if it's enough to cover the rent_cost, but this could also be much higher than the minimum

DaughterOfMars: It clones the value, so it does not add a SDRUC. The reason I call the other fn is because most of the functionality would have be re-implemented either way.

sdk/src/wallet/account/operations/output_claiming.rs
Thoralf-M: But this can be wrong if the `required_amount_for_nfts` is less than the min required storage deposit for such a basic output (since the function then adds the required amount for a storage deposit return unlock condition)

sdk/src/types/block/output/basic.rs
Thoralf-M: It doesn't add the SDRUC to the initial builder, but it adds it before calculating the amount if not enough amount is there
This function doesn't return the min_storage_deposit_amount as the name suggests, but just the amount the output has if higher than the rent_cost or the rent cost it would need with a SDRUC

DaughterOfMars: Exactly, which is the min storage deposit amount?

Thoralf-M: If min rent cost for an output is 42600, but you call this on an output with 100000000 as amount, it would return 100000000

DaughterOfMars: Yes, that is correct. Because that is the minimum amount needed.

Thoralf-M: I don't get how you can call this a `minimum`. To me its just the amount of the output with potentially a SDRUC if needed (only in this case `minimum` makes sense to me)

DaughterOfMars: Think of it like this
```rust
current_amount.min(needed_amount)
```
It's not exactly that of course but that's what I mean by minimum

sdk/src/types/block/output/account.rs
```diff
@@ -87,9 +87,9 @@ impl AccountOutputBuilder {
     }
 
     /// Creates an [`AccountOutputBuilder`] with a provided rent structure.
-    /// The amount will be set to the minimum storage deposit.
-    pub fn new_with_minimum_storage_deposit(rent_structure: RentStructure, account_id: AccountId) -> Self {
-        Self::new(OutputBuilderAmount::MinimumStorageDeposit(rent_structure), account_id)
+    /// The amount will be set to the rent cost of the resulting output.
+    pub fn new_with_minimum_amount(rent_structure: RentStructure, account_id: AccountId) -> Self {
```
thibault-martinez: I think I would rename these to either `new_with_rent_cost` or `new_with_rent_cost_amount` to better match the variant name

sdk/src/types/block/output/basic.rs
```diff
@@ -198,6 +273,27 @@ impl BasicOutputBuilder {
     }
 }
 
+impl Rent for BasicOutputBuilder {
```
thibault-martinez: This is so tedious and error prone 😭

DaughterOfMars: yeah, you're not wrong. But what choice do we have?

sdk/src/types/block/output/account.rs
thibault-martinez: ?

DaughterOfMars: I don't like those names really :/

sdk/src/types/block/output/basic.rs
```diff
@@ -198,6 +279,27 @@ impl BasicOutputBuilder {
...
+        Output::byte_offset(builder)
+            // Kind
+            .data_field::<u8>()
+            // Amount
+            .data_field::<u64>()
+            // Mana
+            .data_field::<u64>()
+            // Native Tokens
+            .data_field::<u8>()
+            .iter_field(&self.native_tokens)
```
Thoralf-M: This is now a Feature with only a single token

sdk/src/types/block/output/basic.rs
```diff
@@ -323,6 +425,24 @@ impl BasicOutput {
...
+    fn build_weighted_bytes(&self, builder: RentBuilder) -> RentBuilder {
+        Output::byte_offset(builder)
+            // Kind
+            .data_field::<u8>()
+            // Amount
+            .data_field::<u64>()
+            // Mana
+            .data_field::<u64>()
+            // Native Tokens
+            .packable_data_field(&self.native_tokens)
```
Thoralf-M: Remove, feature now

sdk/src/types/block/output/feature/block_issuer.rs
```diff
@@ -73,6 +85,18 @@ impl Ed25519BlockIssuerKey {
     pub fn try_from_bytes(bytes: [u8; Self::LENGTH]) -> Result<Self, Error> {
         Ok(Self(ed25519::PublicKey::try_from_bytes(bytes)?))
     }
+
+    /// Creates a dummy [`Ed25519BlockIssuerKey`] used to calculate a storage score for implicit account addresses.
```
Alex6323: Wouldn't probably use the term `dummy` anymore. I forgot to remove it.

thibault-martinez: Agree

sdk/src/types/block/output/rent.rs
```diff
@@ -1,224 +1,174 @@
...
+/// A trait to facilitate the computation of the byte cost of block outputs, which is central to dust protection.
+pub trait StorageScore {
+    fn storage_score(&self, _params: RentParameters) -> u64 {
+        0
     }
 
-    pub fn finish(self) -> Result<u64, Error> {
-        Ok(self.builder.finish_output(self.token_supply)?.rent_cost(self.config))
+    /// Computes the minimum deposit of this instance given [`RentParameters`].
+    fn min_deposit(&self, params: RentParameters) -> u64 {
```
thibault-martinez: Thinking out loud, this is kinda only relevant to outputs, maybe it shouldn't be part of the trait

sdk/src/types/block/address/mod.rs
```diff
@@ -21,6 +21,7 @@ pub use self::{
     nft::NftAddress,
     restricted::{AddressCapabilities, AddressCapabilityFlag, RestrictedAddress},
 };
+use super::output::StorageScore;
```
thibault-martinez: Module already imported

sdk/src/types/block/output/unlock_condition/mod.rs
```diff
@@ -30,6 +30,7 @@ pub use self::{
     state_controller_address::StateControllerAddressUnlockCondition,
     storage_deposit_return::StorageDepositReturnUnlockCondition, timelock::TimelockUnlockCondition,
 };
+use super::{RentParameters, StorageScore};
```
thibault-martinez: merge

sdk/src/types/block/output/feature/issuer.rs
```diff
@@ -26,8 +26,10 @@ impl IssuerFeature {
     }
 }
 
+impl StorageScore for IssuerFeature {}
```
thibault-martinez: Need to forward to the address

sdk/src/types/block/output/feature/block_issuer.rs
thibault-martinez: It's pub(crate) anyway, let's just remove the doc

sdk/src/types/block/output/rent.rs
DaughterOfMars: I renamed the fn because it's really not the min deposit after all, it's just the rent cost. For instance, we use it to get the cost of just the SDRUC.

sdk/src/wallet/account/operations/balance.rs
```diff
@@ -220,22 +220,23 @@ where
...
                                         if output.is_basic() {
-                                            balance.required_storage_deposit.basic += rent;
-                                            // Amount for basic outputs isn't added to total_rent_amount if there aren't
-                                            // native tokens, since we can
-                                            // spend it without burning.
+                                            balance.required_storage_deposit.basic += storage_cost;
+                                            // Amount for basic outputs isn't added to total_storage_cost if there
+                                            // aren't native tokens, since
+                                            // we can spend it without
+                                            // burning.
```
Alex6323: maybe just 2 lines?

DaughterOfMars: weird

thibault-martinez (changes requested):
Can you do the changes to bindings as well please

sdk/src/types/block/address/restricted.rs
```diff
@@ -54,6 +55,8 @@ impl RestrictedAddress {
     }
 }
 
+impl StorageScore for RestrictedAddress {}
```
thibault-martinez: I know iota.go returns 0 and that it's currently correct but I'd personally forward to the address, less error-prone. Just resolve if you think it's not worth it.

sdk/src/types/block/address/mod.rs
```diff
@@ -156,6 +156,18 @@ impl Address {
     }
 }
 
+impl StorageScore for Address {
+    fn storage_score(&self, params: super::output::StorageScoreParameters) -> u64 {
```
thibault-martinez: Can you just import StorageScoreParameters with the rest pls

sdk/src/types/block/output/storage_score.rs
```diff
@@ -0,0 +1,182 @@
+// Copyright 2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use packable::Packable;
+
+use super::OutputId;
```
thibault-martinez: merge import

bindings/core/src/method/utils.rs
```diff
@@ -131,7 +131,10 @@ pub enum UtilsMethod {
     /// Computes the input commitment from the output objects that are used as inputs to fund the transaction.
     ComputeInputsCommitment { inputs: Vec<OutputDto> },
     /// Computes the required storage deposit of an output.
-    ComputeStorageDeposit { output: OutputDto, rent: RentStructure },
+    ComputeStorageDeposit {
+        output: OutputDto,
+        storage_params: StorageScoreParameters,
```
thibault-martinez: Let's have the full name, this is too ambiguous

sdk/src/types/block/output/storage_score.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+/// A trait to facilitate the computation of the byte cost of block outputs, which is central to dust protection.
+pub trait StorageScore {
+    fn storage_score(&self, _params: StorageScoreParameters) -> u64 {
+        0
+    }
+
+    /// Computes the storage cost of this instance given [`StorageScoreParameters`].
+    fn storage_cost(&self, params: StorageScoreParameters) -> u64 {
```
thibault-martinez: I don't think we should call this `storage_cost` as the name is already reserved for one of the components of what this is actually returning. TIP calls this `Minimum Storage Deposit` so we should call it exactly the same.

DaughterOfMars: but that's not what it is...

thibault-martinez: How so ?

DaughterOfMars: It's not the min storage deposit. it can be calculated for an individual unlock condition, for instance

thibault-martinez: Ok so there is a single place where this is being used as non min_storage_deposit, in a test. I think I need to better understand what you are trying to achieve because it doesn't feel very correct to me. Storage score of a SDRUC is most of the time 0, only case would be if it contains an implicit account creation one but I'm not even sure this is possible/legal, I need to ask iota-core.

thibault-martinez: That being said, I don't see why it couldn't be called `Minimum Storage Deposit`, it's still really is a share of it, even if not an output.

sdk/src/types/block/output/basic.rs
```diff
@@ -323,6 +396,14 @@ impl BasicOutput {
     }
 }
 
+impl StorageScore for BasicOutput {
+    fn storage_score(&self, params: StorageScoreParameters) -> u64 {
+        params.output_offset()
+            + self.packed_len() as u64 * params.data_factor() as u64
+            + self.unlock_conditions.storage_score(params)
```
thibault-martinez, line 386: What about features ?

sdk/src/types/block/output/storage_score.rs
DaughterOfMars: No, we use it to calculate the minimum required storage deposit, where we call it on a sdruc

thibault-martinez (changes requested):
Conflicts (I'll review this tomorrow)

thibault-martinez (changes requested):
Conflicts

sdk/src/types/block/output/account.rs
```diff
@@ -504,6 +542,33 @@ impl AccountOutput {
...
+            + size_of::<u64>()
+            // Mana
+            + size_of::<u64>()
+            // Native Tokens
+            + size_of::<u8>()
+            + self.native_tokens.iter().map(|nt| nt.packed_len()).sum::<usize>()
+            // AccountId
+            + AccountId::LENGTH
+            // State Index
+            + size_of::<u32>()
```
Thoralf-M: Doesn't exist anymore

thibault-martinez: I think we should remove these methods anyway, it's just packed_len but error prone

Thoralf-M: So you would use `self.{field}.packed_len()` everywhere?

thibault-martinez: I'm just wondering what's the point of `stored_len` to start with

sdk/src/types/block/output/mod.rs
```diff
@@ -73,7 +73,7 @@ pub const OUTPUT_INDEX_RANGE: RangeInclusive<u16> = 0..=OUTPUT_INDEX_MAX; // [0.
 #[derive(Copy, Clone)]
 pub enum OutputBuilderAmount {
     Amount(u64),
-    MinimumStorageDeposit(RentStructure),
+    StorageCost(StorageScoreParameters),
```
thibault-martinez: I would very much appreciate if the variants match the method they are associated to.
`OutputBuilderAmount::Amount` => `new_with_amount`
So I would either do
`OutputBuilderAmount::StorageCost` => `new_with_storage_cost`
or 
`OutputBuilderAmount::MinimumAmount` => `new_with_minimum_amount`
You choose

sdk/src/types/block/output/foundry.rs
```diff
@@ -539,6 +571,39 @@ impl FoundryOutput {
 
         Ok(())
     }
+
+    fn stored_len(&self) -> usize {
```
thibault-martinez: Let's remove these, it's just packed_len + 1, or am I missing something?

sdk/src/types/block/output/storage_score.rs
Alex6323: I think the term `Minimum Storage Deposit` is really only defined for the outputs (could be made clearer in the TIP tho), and rather not for its constituents, so I think we're good calling it however we like. However, in the protocol parameters a slightly more accurate term would have been: `StorageCostPerScore` to determine the individual "Storage Cost" of a constituent depending on its score, which then is used to determine the "minimum storage deposit" of an output. So IMO what we have currently is a good name.

thibault-martinez: My point is that currently `storage_cost` is used for only ONE thing that is not minimum storage deposit and I'm not convinced it's needed in the first place. But we can discuss that in the huddle.

sdk/src/types/block/output/storage_score.rs
```diff
@@ -0,0 +1,190 @@
...
+        let basic_output_score = BasicOutputBuilder::new_with_amount(0)
+            .add_unlock_condition(AddressUnlockCondition::new(null_address))
+            .finish()
+            .unwrap()
+            .storage_score(*self);
+        let ed25519_address_score = null_address.storage_score(*self);
+        let basic_score_without_address = basic_output_score - ed25519_address_score;
+        let account_output_score = AccountOutputBuilder::new_with_amount(0, AccountId::null())
+            .add_unlock_condition(GovernorAddressUnlockCondition::new(null_address))
+            .add_unlock_condition(StateControllerAddressUnlockCondition::new(null_address))
```
thibault-martinez: This should be a single address now, not gov + state

sdk/src/types/block/address/implicit_account_creation.rs
```diff
@@ -26,6 +29,12 @@ impl ImplicitAccountCreationAddress {
     }
 }
 
+impl StorageScore for ImplicitAccountCreationAddress {
+    fn storage_score(&self, params: StorageScoreParameters) -> u64 {
+        params.implicit_account_creation_address_offset() + self.0.storage_score(params)
```
thibault-martinez: Can we move `self.0.storage_score(params)` to the offset definition? Closer to the TIP

sdk/src/types/block/output/nft.rs
```diff
@@ -424,6 +420,17 @@ impl NftOutput {
     }
 }
 
+impl StorageScore for NftOutput {
+    fn storage_score(&self, params: StorageScoreParameters) -> u64 {
+        params.output_offset()
+            + 1 // Type byte
```
thibault-martinez: Thinking out loud here, do you think the storage score of outputs should be without the 1 and we also have storage score with the 1 for the enum? It feels a bit weird to consider the type byte here when we usually don't

thibault-martinez: Thinking out loud here, do you think the storage score of outputs should be without the 1 and we also have storage score with the 1 for the enum? It feels a bit weird to consider the type byte here when we usually don't

sdk/src/types/block/output/anchor.rs
```diff
@@ -557,6 +549,16 @@ impl AnchorOutput {
     }
 }
 
+impl StorageScore for AnchorOutput {
+    fn storage_score(&self, params: StorageScoreParameters) -> u64 {
+        params.output_offset()
+            + 1 // Type byte
```
thibault-martinez: The 1 should be in parenthesis together with ` self.packed_len()`

sdk/src/types/block/output/nft.rs
DaughterOfMars: I don't think the storage score makes sense for an output variant if the type byte isn't included so I think this is better

sdk/src/types/block/address/implicit_account_creation.rs
DaughterOfMars: I can't even find this TIP

thibault-martinez: https://github.com/iotaledger/tips/blob/tip42/tips/TIP-0042/tip-0042.md#storage-score-2

sdk/src/types/block/output/nft.rs
thibault-martinez: Alright

bindings/core/src/method/utils.rs
```diff
@@ -129,7 +129,10 @@ pub enum UtilsMethod {
         transaction: TransactionDto,
     },
     /// Computes the required storage deposit of an output.
-    ComputeStorageDeposit { output: OutputDto, rent: RentStructure },
+    ComputeStorageDeposit {
```
Thoralf-M:
```suggestion
    #[serde(rename_all = "camelCase")]
    ComputeStorageDeposit {
```

sdk/src/types/block/output/feature/block_issuer.rs
```diff
@@ -63,6 +75,17 @@ impl Ed25519BlockIssuerKey {
     pub fn try_from_bytes(bytes: [u8; Self::LENGTH]) -> Result<Self, Error> {
         Ok(Self(ed25519::PublicKey::try_from_bytes(bytes)?))
     }
+
+    pub(crate) fn null() -> Self {
+        // Unwrap: we provide a valid byte array
```
thibault-martinez, line 80: Are we sure full 0 is actually a valid public key ?

sdk/src/types/block/address/anchor.rs
```diff
@@ -40,6 +40,8 @@ impl AnchorAddress {
     }
 }
 
+impl StorageScore for AnchorAddress {}
```
Thoralf-M, line 43: So this is just makes it 0?

thibault-martinez: Yes

sdk/src/wallet/operations/balance.rs
```diff
@@ -34,7 +34,7 @@ where
         wallet_data: &WalletData,
     ) -> Result<Balance> {
         let network_id = self.client().get_network_id().await?;
-        let rent_structure = self.client().get_rent_structure().await?;
+        let storage_score_params = self.client().get_storage_score_parameters().await?;
         let mut balance = Balance::default();
         let mut total_rent_amount = 0;
```
thibault-martinez: Maybe we want to rename the variable as well

sdk/src/wallet/operations/output_claiming.rs
```diff
@@ -244,7 +243,10 @@ where
             .iter()
             .map(|i| i.output.amount())
             .sum::<u64>()
-            >= MinimumStorageDepositBasicOutput::new(rent_structure, token_supply).finish()?;
+            >= BasicOutputBuilder::new_with_minimum_amount(storage_score_params)
+                .add_unlock_condition(AddressUnlockCondition::new(Ed25519Address::null()))
+                .finish()?
+                .amount();
```
thibault-martinez: Maybe would be nice to have a shorter way for this

sdk/src/wallet/operations/output_claiming.rs
```diff
@@ -274,14 +276,14 @@ where
 
                 let nft_output = if !enough_amount_for_basic_output {
                     // Only update address and nft id if we have no additional inputs which can provide the storage
-                    // deposit for the remaining amount and possible NTs
+                    // deposit for the remaining amount and possible NFTs
```
thibault-martinez: I think this meant Native Tokens?

sdk/src/types/block/address/anchor.rs
Thoralf-M: Bit weird to even have it then but okay, guess we have it for consistency then

sdk/src/types/block/output/account.rs
```diff
@@ -77,10 +77,10 @@ impl AccountOutputBuilder {
         Self::new(OutputBuilderAmount::Amount(amount), account_id)
     }
 
-    /// Creates an [`AccountOutputBuilder`] with a provided rent structure.
-    /// The amount will be set to the minimum storage deposit.
-    pub fn new_with_minimum_storage_deposit(rent_structure: RentStructure, account_id: AccountId) -> Self {
-        Self::new(OutputBuilderAmount::MinimumStorageDeposit(rent_structure), account_id)
+    /// Creates an [`AccountOutputBuilder`] with a provided storage score structure.
```
Thoralf-M:
```suggestion
    /// Creates an [`AccountOutputBuilder`] with provided storage score parameters.
```

Thoralf-M: Same for the other outputs

sdk/src/types/block/output/storage_score.rs
Alex6323: I leave it up to you guys to decide. My preference tho would be to just ask the TIP writers to rename `storage_cost` to `storage_cost_factor` on the `StorageScoreParameters` to remove the name collision and then leave it as it is. I kinda dislike renaming this to `minimum_storage_deposit` because 
1) if we ever wanted to use "storage cost" for something else we'ld have to rename this trait fn, while "storage_cost" is probably fairly future-proof.
2) it's uglier and who likes that? :trollface: 
3) If that's not acceptable I'ld rather remove this fn completely from the trait and make the final calculation in the caller.

thibault-martinez: 1. I don't think we should ever account for future proofing names, this logic could apply to way more than this.
2. It's not uglier and everybody wants that, it's always been named like that, renaming it would create confusion.
3. I would do that.

sdk/src/types/block/output/account.rs
```diff
@@ -77,10 +77,10 @@ impl AccountOutputBuilder {
...
+    /// The amount will be set to the storage cost of the resulting output.
```
thibault-martinez:
```suggestion
    /// The amount will be set to the minimum storage deposit of the resulting output.
```
Same for the other outputs

sdk/src/types/block/output/account.rs
```diff
@@ -103,10 +103,10 @@ impl AccountOutputBuilder {
         self
     }
 
-    /// Sets the amount to the minimum storage deposit.
+    /// Sets the amount to the storage cost.
```
thibault-martinez:
```suggestion
    /// Sets the amount to the minimum storage deposit.
```
Same for the other outputs

sdk/src/types/block/output/account.rs
```diff
@@ -258,15 +258,15 @@ impl AccountOutputBuilder {
             immutable_features,
         };
 
+        if let OutputBuilderAmount::MinimumAmount(params) = self.amount {
+            output.amount = output.storage_cost(params);
+        }
```
thibault-martinez: This is already in the match just after ?

sdk/src/types/block/output/anchor.rs
```diff
@@ -557,6 +549,16 @@ impl AnchorOutput {
     }
 }
 
+impl StorageScore for AnchorOutput {
+    fn storage_score(&self, params: StorageScoreParameters) -> u64 {
+        params.output_offset()
+            // Type byte
+            + (1 + self.packed_len() as u64) * params.data_factor() as u64
+            + self.unlock_conditions.storage_score(params)
+            + self.features.storage_score(params)
```
thibault-martinez, line 554: missing immutable features

sdk/src/types/block/output/mod.rs
```diff
@@ -436,17 +433,16 @@ impl Packable for Output {
...
-        Ok(())
+impl StorageScore for Output {
+    fn storage_score(&self, params: StorageScoreParameters) -> u64 {
+        match self {
+            Self::Basic(o) => o.storage_score(params),
+            Self::Account(o) => o.storage_score(params),
+            Self::Foundry(o) => o.storage_score(params),
+            Self::Nft(o) => o.storage_score(params),
+            Self::Delegation(o) => o.storage_score(params),
+            Self::Anchor(o) => o.storage_score(params),
```
thibault-martinez: Sorry super nitpicky but can you put it in its place, after Account pls

sdk/src/wallet/operations/output_claiming.rs
DaughterOfMars: Not possible anymore

sdk/src/types/block/output/feature/block_issuer.rs
Thoralf-M: Works
```Rust
    let public_key = prefix_hex::decode("0x0000000000000000000000000000000000000000000000000000000000000000")?;
    let public_key = PublicKey::try_from_bytes(public_key)?;
    println!("{public_key:?}");
```
```
PublicKey(VerificationKey { A_bytes: VerificationKeyBytes("0000000000000000000000000000000000000000000000000000000000000000"), minus_A: EdwardsPoint{
        X: FieldElement51([533094393274173, 2016890930128738, 18285341111199, 134597186663265, 1486323764102114]),
        Y: FieldElement51([0, 0, 0, 0, 0]),
        Z: FieldElement51([1, 0, 0, 0, 0]),
        T: FieldElement51([2251799813685229, 2251799813685247, 2251799813685247, 2251799813685247, 2251799813685247])
} })
```

bindings/core/src/method_handler/client.rs
```diff
@@ -294,13 +295,13 @@ pub(crate) async fn call_client_method_internal(client: &Client, method: ClientM
         ClientMethod::HexPublicKeyToBech32Address { hex, bech32_hrp } => {
             Response::Bech32Address(client.hex_public_key_to_bech32_address(&hex, bech32_hrp).await?)
         }
-        ClientMethod::MinimumRequiredStorageDeposit { output } => {
+        ClientMethod::MinimumRequiredOutputAmount { output } => {
```
thibault-martinez: I know it was like that already but it bothers me that `ClientMethod::MinimumRequiredOutputAmount` has a different name than `UtilsMethod::ComputeMinimumOutputAmount` when they are doing the same thing but with or without a client. I'd use `ComputeMinimumOutputAmount` for both and drop the required from everywhere

sdk/src/types/block/output/feature/block_issuer.rs
thibault-martinez: Thanks

sdk/src/types/block/output/unlock_condition/expiration.rs
```diff
@@ -55,6 +60,12 @@ impl ExpirationUnlockCondition {
     }
 }
 
+impl StorageScore for ExpirationUnlockCondition {
+    fn storage_score(&self, params: StorageScoreParameters) -> u64 {
+        self.return_address().storage_score(params)
```
Thoralf-M, line 65: What about the expiration slot index?
Guess it doesn't really matter because it's just 0, but same would be true for the address?

Thoralf-M: Same about the amount field in the SDRUC

thibault-martinez: I don't know if TIPs disallow that but you could have an implicit account creation address in the address and it then comes with an offset. Slot index has no chance to incur any additional cost.

sdk/tests/client/input_selection/basic_outputs.rs
```diff
@@ -1123,44 +1123,45 @@ fn two_inputs_remainder_3() {
...
-        1_000_000,
-        BECH32_ADDRESS_ED25519_0,
-        None,
-        None,
-        None,
-        None,
-        None,
-        None,
-    )]);
+// TODO: re-enabled when rent is figured out
```
Thoralf-M, line 1126: What's still missing?

sdk/src/types/block/output/unlock_condition/expiration.rs
Thoralf-M: image
Implicit account creation address is not listed here

thibault-martinez: Fair enough. But tbh I think I would keep it like that, a bit less error prone if we ever add an address type that comes with an offset.

sdk/src/types/block/output/basic.rs
```diff
@@ -146,6 +149,59 @@ impl BasicOutputBuilder {
         self
     }
 
+    /// Adds a storage deposit return unlock condition if one is needed to cover the current amount
+    /// (i.e. `amount < storage_cost`). This will increase the total amount to equal the `storage_cost` with
```
thibault-martinez: comment still mentions storage_cost, same for the NFT file

sdk/src/types/block/output/storage_score.rs
```diff
@@ -0,0 +1,190 @@
...
+/// A trait to facilitate the computation of the byte cost of block outputs, which is central to dust protection.
+pub trait StorageScore {
+    fn storage_score(&self, _params: StorageScoreParameters) -> u64 {
+        0
+    }
+}
+
+/// A trait that is shared by all output types, which is used to calculate the minimum amount the output
+/// must contain to satisfy its storage cost.
+pub trait MinimumOutputAmount: StorageScore {
```
thibault-martinez: I'd move that to the output mod.rs

sdk/src/types/block/protocol.rs
```diff
@@ -40,8 +42,8 @@ pub struct ProtocolParameters {
     pub(crate) network_name: StringPrefix<u8>,
     /// The HRP prefix used for Bech32 addresses in the network.
     pub(crate) bech32_hrp: Hrp,
-    /// The rent structure used by given node/network.
-    pub(crate) rent_structure: RentStructure,
+    /// The storage store parameters used by given node/network.
```
thibault-martinez:
```suggestion
    /// The storage score parameters used by given node/network.
```

sdk/src/wallet/operations/balance.rs
```diff
@@ -34,9 +36,9 @@ where
         wallet_data: &WalletData,
     ) -> Result<Balance> {
         let network_id = self.client().get_network_id().await?;
-        let rent_structure = self.client().get_rent_structure().await?;
+        let storage_score_params = self.client().get_storage_score_parameters().await?;
         let mut balance = Balance::default();
-        let mut total_rent_amount = 0;
+        let mut total_storage_cost = 0;
```
thibault-martinez, line 41: this variable name is still a bit off

DaughterOfMars: what do you think it should be?

sdk/tests/types/output/account.rs
```diff
@@ -63,7 +63,7 @@ fn builder() {
 
     assert_eq!(
         output.amount(),
-        Output::Account(output.clone()).rent_cost(protocol_parameters.rent_structure())
+        Output::Account(output.clone()).minimum_amount(protocol_parameters.storage_score_parameters())
```
thibault-martinez: I know it's only a test but we don't need to wrap/clone, right? Trying to prevent this pattern for proliferating if it's not necessary.

sdk/tests/types/output/basic.rs
```diff
@@ -48,15 +48,15 @@ fn builder() {
...
+        .with_minimum_amount(protocol_parameters.storage_score_parameters())
         .add_unlock_condition(rand_address_unlock_condition())
         .with_features([Feature::from(metadata.clone()), sender_1.clone().into()])
         .finish()
         .unwrap();
 
     assert_eq!(
         output.amount(),
-        Output::Basic(output.clone()).rent_cost(protocol_parameters.rent_structure())
+        Output::Basic(output.clone()).minimum_amount(protocol_parameters.storage_score_parameters())
```
thibault-martinez: same

thibault-martinez: and other similar files

sdk/tests/wallet/output_preparation.rs
```diff
@@ -404,8 +407,8 @@ async fn output_preparation() -> Result<()> {
             None,
         )
         .await?;
-    let rent_structure = wallet.client().get_rent_structure().await?;
-    let minimum_storage_deposit = output.rent_cost(rent_structure);
+    let storage_score_params = wallet.client().get_storage_score_parameters().await?;
+    let minimum_storage_deposit = output.minimum_amount(storage_score_params);
```
thibault-martinez: do you want to rename this variable?

sdk/tests/client/input_selection/basic_outputs.rs
DaughterOfMars: Just don't really know how to write this test without a point of reference, I guess

bindings/core/src/method/client.rs
```diff
@@ -351,10 +351,10 @@ pub enum ClientMethod {
         /// Human readable part
         bech32_hrp: Option<Hrp>,
     },
-    /// Calculate the minimum required storage deposit for an output.
+    /// Calculate the minimum required amount for an output.
     /// Expected response:
-    /// [`MinimumRequiredStorageDeposit`](crate::Response::MinimumRequiredStorageDeposit)
-    MinimumRequiredStorageDeposit { output: OutputDto },
+    /// [`MinimumRequiredOutputAmount`](crate::Response::MinimumRequiredOutputAmount)
```
thibault-martinez: I believe you changed that

bindings/nodejs/lib/types/client/bridge/index.ts
```diff
@@ -49,7 +49,7 @@ import type {
     __BuildNftOutputMethod__,
     __ClearListenersMethod__,
     __SignatureUnlockMethod__,
-    __MinimumRequiredStorageDepositMethod__,
+    __MinimumRequiredOutputAmountMethod__,
```
thibault-martinez: This needs to be changed to `__ComputeMinimumOutputAmountMethod__`
Same for the second one.

Thoralf-M (dismissed):

sdk/tests/wallet/output_preparation.rs
```diff
@@ -608,12 +611,13 @@ async fn prepare_output_remainder_dust() -> Result<()> {
...
-    let token_supply = wallet_0.client().get_token_supply().await?;
+    let storage_score_params = wallet_0.client().get_storage_score_parameters().await?;
 
     let balance = wallet_0.sync(None).await?;
-    let minimum_required_storage_deposit =
-        MinimumStorageDepositBasicOutput::new(rent_structure, token_supply).finish()?;
+    let minimum_required_storage_deposit = BasicOutputBuilder::new_with_amount(0)
+        .add_unlock_condition(AddressUnlockCondition::new(wallet_1.address().await))
+        .finish()?
+        .minimum_amount(storage_score_params);
```
thibault-martinez: There is a `minimum_storage_deposit` function in the output mod.rs, I'm wondering if we should move it as a static method to BasicOutput or something. This just seems like a pattern we may be doing quite a lot

sdk/src/wallet/operations/transaction/prepare_output.rs
```diff
@@ -114,10 +114,13 @@ where
             OutputBuilder::Basic(BasicOutputBuilder::from(first_output.as_basic()))
         };
 
-        let min_storage_deposit_basic_output =
-            MinimumStorageDepositBasicOutput::new(rent_structure, token_supply).finish()?;
+        // TODO: Probably not good to use ed25519 always here, even if technically it's the same for now..
+        let min_amount_basic_output = BasicOutputBuilder::new_with_minimum_amount(storage_score_params)
+            .add_unlock_condition(AddressUnlockCondition::new(Ed25519Address::null()))
+            .finish()?
+            .amount();
```
thibault-martinez: This is another example

sdk/src/wallet/operations/output_claiming.rs
thibault-martinez: And another

sdk/src/wallet/operations/balance.rs
thibault-martinez: Idk, let's keep it a bit more, I need to refactor this file eventually anyway

thibault-martinez (dismissed):

thibault-martinez (approved):

bindings/python/iota_sdk/client/_utils.py
```diff
@@ -49,11 +49,11 @@ def hex_public_key_to_bech32_address(
...
         })
 
-    def minimum_required_storage_deposit(self, output: Output) -> int:
-        """Minimum required storage deposit.
+    def computer_minimum_output_amount(self, output: Output) -> int:
+        """Minimum required output amount.
         """
         return int(self._call_method(
-            'minimumRequiredStorageDeposit', {
+            'computeMinimumOutputAmount', {
```
kwek20, line 56: missing storage_score_parameter?

thibault-martinez: No, this is a client method, parameters will be provided internally

Alex6323 (approved):

kwek20 (approved):
YES!

Issue #918, opened 2023-07-25
Adapt fields of RegularTransactionEssence - nodejs


thibault-martinez:
Blocked by https://github.com/iotaledger/iota-sdk/issues/912 and https://github.com/iotaledger/iota-sdk/issues/914

https://github.com/iotaledger/tips-draft/blob/tip45/tips/TIP-0045/tip-0045.md#serialized-layout

thibault-martinez:
Closed by https://github.com/iotaledger/iota-sdk/pull/1075

PR #1075, opened 2023-08-23
Adapt fields of `RegularTransactionEssence` in Node.js

abdulmth:
#fixes https://github.com/iotaledger/iota-sdk/issues/918

### DIFF
--- a/bindings/nodejs/lib/types/block/payload/transaction/essence.ts
+++ b/bindings/nodejs/lib/types/block/payload/transaction/essence.ts
@@ -73,7 +73,7 @@ class RegularTransactionEssence extends TransactionEssence {
     ) {
         super(TransactionEssenceType.Regular);
         this.networkId = networkId;
-        this.contextInput = contextInput
+        this.contextInput = contextInput;
         this.inputs = inputs;
         this.inputsCommitment = inputsCommitment;
         this.outputs = outputs;
--- a/bindings/nodejs/lib/types/block/payload/transaction/essence.ts
+++ b/bindings/nodejs/lib/types/block/payload/transaction/essence.ts
@@ -3,9 +3,12 @@
 
 import { Type } from 'class-transformer';
 import { PayloadDiscriminator } from '..';
-import { HexEncodedString } from '../../../utils';
+import { HexEncodedString, u64 } from '../../../utils';
+import { ContextInput, ContextInputDiscriminator } from '../../context_input';
 import { Input, InputDiscriminator } from '../../input';
+import { ManaAllotment } from '../../mana-allotment';
 import { Output, OutputDiscriminator } from '../../output';
+import { SlotIndex } from '../../slot';
 import { Payload } from '../payload';
 
 /**
@@ -16,17 +19,10 @@ enum TransactionEssenceType {
 }
 
 abstract class TransactionEssence {
-    private type: TransactionEssenceType;
+    readonly type: TransactionEssenceType;
 
     constructor(type: TransactionEssenceType) {
         this.type = type;
-    }
-
-    /**
-     * The type of essence.
-     */
-    getType(): TransactionEssenceType {
-        return this.type;
     }
 }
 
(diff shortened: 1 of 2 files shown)


bindings/nodejs/lib/types/block/payload/transaction/essence.ts
```diff
@@ -16,56 +19,67 @@ enum TransactionEssenceType {
...
-    inputsCommitment: HexEncodedString;
+    /**
+     * The unique value denoting whether the block was meant for mainnet, testnet, or a private network.
+     */
+    networkId: u64;
+
+    @Type(() => Input, {
+        discriminator: ContextInputDiscriminator,
+    })
+    contextInput: ContextInput[];
```
abdulmth: > Context Inputs **optAnyOf**

In the builder in Rust it has the type `Vec<ContextInput>` so it's not under `Option`. In case it's not provided it will be an empty array? I kept it the same here, but we could possibly make it an optional property `contextInput?: contextInput[]` Not sure how serialization would work though.

bindings/nodejs/lib/types/block/payload/transaction/essence.ts
```diff
@@ -16,56 +19,67 @@ enum TransactionEssenceType {
...
 
     @Type(() => Input, {
         discriminator: InputDiscriminator,
     })
-    inputs: [Input];
+    inputs: Input[];
```
abdulmth, line 48: This is a bug which was fixed in a patch for 1.0 I think.

bindings/nodejs/lib/types/block/payload/transaction/essence.ts
```diff
@@ -16,56 +19,67 @@ enum TransactionEssenceType {
...
-        networkId: number,
+        networkId: u64,
+        contextInput: ContextInput[],
+        inputs: Input[],
         inputsCommitment: HexEncodedString,
-        inputs: [Input],
-        outputs: [Output],
-        payload: Payload | undefined,
+        outputs: Output[],
+        allotments: ManaAllotment[],
```
abdulmth, line 71: This should be an ordered set. Sets in JS are not ordered, so we have to use arrays, but there's no validation for duplicates, we can validate that in the constructor. If we don't and the array has duplicates, an error would be returned from somewhere in Rust at a later point, not when an instance is created ofc.

DaughterOfMars: Since that validation is done on the rust side, it is not needed here.

bindings/nodejs/lib/types/block/payload/transaction/essence.ts
```diff
@@ -16,56 +19,67 @@ enum TransactionEssenceType {
...
+        payload?: Payload,
+        creationSlot?: SlotIndex,
```
Thoralf-M: Not optional and can you move it right after networkId?

bindings/nodejs/lib/types/block/payload/transaction/essence.ts
Thoralf-M:
```suggestion
    contextInputs: ContextInput[];
```

bindings/nodejs/lib/types/block/payload/transaction/essence.ts
abdulmth: done

bindings/nodejs/lib/types/block/payload/transaction/essence.ts
abdulmth: done

Thoralf-M (approved):

abdulmth:
@Thoralf-M wait, `creationSlot` is actually optional in the `RegularTransactionEssenceBuilder`. shouldn't we keep it optional in TS so it uses the current time?

Thoralf-M:
It's only optional in the builder, but not in the transaction essence itself

bindings/nodejs/lib/types/block/payload/transaction/essence.ts
kwek20: Keep in mind that after compilation none of these types exist anymore, itll be `undefined` probably if left completely out of the json. Thats fine though

kwek20 (approved):

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1082, opened 2023-08-24
Fix serde feature gating in types

DaughterOfMars:
# Description of change

This PR fixes compilation without the `serde` feature. Part of the change is to feature gate all API types with `serde` because I don't think they are useful without it anyway. This isn't breaking because everything that was available with the feature enabled is untouched, and without the feature it would not compile.

## Type of change

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -4,6 +4,7 @@ ci-check-nodejs = "check --manifest-path bindings/nodejs/Cargo.toml --all-target
 ci-check-python = "check --manifest-path bindings/python/Cargo.toml --all-targets --lib"
 ci-check-wasm = "check --manifest-path bindings/wasm/Cargo.toml --target wasm32-unknown-unknown --lib"
 ci-check-nostd = "check --no-default-features -F serde -p iota-sdk --target riscv64gc-unknown-none-elf"
+ci-check-types = "check --no-default-features -F std -p iota-sdk"
 
 ci-test = "nextest run --all-features --profile ci --cargo-profile ci -p iota-sdk -p iota-sdk-bindings-core"
 
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -21,4 +21,4 @@ ci-license = "license-template --template .license_template"
 #
 # - RUSTSEC-2021-0065: https://rustsec.org/advisories/RUSTSEC-2021-0065
 #   - anymap is unmaintained 🤷‍♀️
-ci-audit = "audit --file Cargo.lock --deny warnings --ignore RUSTSEC-2021-0145 --ignore RUSTSEC-2021-0065"
+ci-audit = "audit --file Cargo.lock --deny warnings --ignore RUSTSEC-2021-0065"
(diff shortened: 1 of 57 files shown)


sdk/src/types/block/output/native_token.rs
```diff
@@ -20,7 +20,7 @@ use crate::types::block::{output::TokenId, Error};
 #[packable(unpack_error = Error)]
 pub struct NativeToken {
     // Identifier of the native token.
-    #[serde(rename = "id")]
+    #[cfg_attr(feature = "serde", serde(rename = "id"))]
```
thibault-martinez, line 23: remove cfg_attr ?

.github/workflows/lint.yml
```diff
@@ -153,3 +153,17 @@ jobs:
...
+
+    steps:
+      - name: Checkout the Source Code
+        uses: actions/checkout@v3
+
+      - name: Set up Rust
+        uses: ./.github/actions/setup-rust
+
+      - name: Check
+        run: cargo check --no-default-features -p iota-sdk
```
thibault-martinez: Not using the alias?

.cargo/config.toml
```diff
@@ -14,3 +15,10 @@ ci-doc = "doc --all-features --no-deps --document-private-items"
 ci-fmt = "fmt --all -- --check"
 ci-toml = "sort --grouped --check"
 ci-udeps = "udeps --all-targets --all-features --backend=depinfo"
+ci-license = "license-template --template .license_template"
+
+# The ignored security advisories:
+#
+# - RUSTSEC-2021-0065: https://rustsec.org/advisories/RUSTSEC-2021-0065
+#   - anymap is unmaintained 🤷‍♀️
+ci-audit = "audit --file Cargo.lock --deny warnings --ignore RUSTSEC-2021-0145 --ignore RUSTSEC-2021-0065"
```
thibault-martinez: Our `atty` (RUSTSEC-2021-0145) vulnerability is fone, you can remove it

.github/workflows/lint.yml
```diff
@@ -152,4 +148,18 @@ jobs:
           targets: riscv64gc-unknown-none-elf
 
       - name: Check
-        run: cargo check --no-default-features -F serde -p iota-sdk --target=riscv64gc-unknown-none-elf
+        run: cargo ci-check-wasm
```
thibault-martinez:
```suggestion
        run: cargo ci-check-nostd
```
?

thibault-martinez (approved):

thibault-martinez (commented):
Anything worth mentioning in the CHANGELOG btw?

thibault-martinez (changes requested):
CI failing

sdk/src/types/block/output/native_token.rs
DaughterOfMars: no 😅

DaughterOfMars:
> Anything worth mentioning in the CHANGELOG btw?

I don't think so

sdk/src/types/block/output/native_token.rs
thibault-martinez: I thought it wasn't necessary when we already had it on the struct?

DaughterOfMars: Yeah I was wrong about that as I explained yesterday. I just didn't see it break because we always compile with serde

thibault-martinez: Oh alright, yeah I see

thibault-martinez (approved):

Thoralf-M (approved):

PR #1085, opened 2023-08-25
Balance refactoring part 1

thibault-martinez:


sdk/src/wallet/account/operations/balance.rs
```diff
@@ -89,178 +92,112 @@ where
...
-                                // add nft_id for nft outputs
-                                if let Output::Nft(output) = &output {
-                                    let nft_id = output.nft_id_non_null(output_id);
-                                    balance.nfts.push(nft_id);
-                                }
+                    }
+
+                    if let Some(native_tokens) = output.native_tokens() {
+                        total_native_tokens.add_native_tokens(native_tokens.clone())?;
+                    }
```
Thoralf-M, line 136: So where do you check if we can actually unlock these?

thibault-martinez: Hmm yeah you're right. I wanted to split more changes to later but it won't work without it.

thibault-martinez:
Will redo that later, differently.

PR #1098, opened 2023-08-29
Rename AccountAddress to Bip44Address

thibault-martinez:


sdk/src/wallet/account/operations/transaction/options.rs
```diff
@@ -39,7 +39,7 @@ pub enum RemainderValueStrategy {
     /// Move the remainder value to a change address.
     ChangeAddress,
     /// Move the remainder value to any specified address.
-    CustomAddress(AccountAddress),
+    CustomAddress(Bip44Address),
```
thibault-martinez: This doesn't seem to actually need a Bip44 one, I could replace with simply Address

DaughterOfMars: do it 🗡️

Thoralf-M: I think it was initially required to provide it to the ledger nano here https://github.com/iotaledger/iota-sdk/blob/4bf72ecb13c3540895d03cc62eb527241e41675a/sdk/src/client/secret/ledger_nano.rs#L297-L308
But looks like we don't pass this information down anymore
In https://github.com/iotaledger/iota-sdk/blob/4bf72ecb13c3540895d03cc62eb527241e41675a/sdk/src/client/api/block_builder/input_selection/core/remainder.rs#L26 it just gets set to None

DaughterOfMars (dismissed):

sdk/src/wallet/account/operations/transaction/options.rs
thibault-martinez: Changed

DaughterOfMars (approved):

Thoralf-M (approved):

kwek20 (approved):

abdulmth (approved):

PR #1107, opened 2023-08-30
Add a worker pool to the client to limit parallel requests

### LINKED ISSUE #1100: Provide option to limit requests per second
## Description

Currently there can be > 1000 API requests per second depending on how the accounts look and what options are used, this can lead to some errors because it exceeds rate limits from the server side.
Maybe we can add a global rate limit using a crate like https://crates.io/crates/governor? Or are there other options?
We also limit our parallel requests in some places with https://github.com/iotaledger/iota-sdk/blob/4bf72ecb13c3540895d03cc62eb527241e41675a/sdk/src/client/constants.rs#L24, could this then be removed, or should we also make it modifiable by the user?

DaughterOfMars:
# Description of change

This PR adds a pool of workers (tasks) that accept futures, poll them, and return the result. The worker pool forms a least-recently-used cache and each worker holds a priority queue which is processed synchronously in order.  

## Outstanding Questions

- Is this overkill? 
    - Yes...probably.
- Does it even solve the original problem? 
    - 🤷‍♀️ hard to say

#1100 

## Type of change


sdk/src/client/worker.rs
```diff
@@ -0,0 +1,239 @@
...
+            // Wait to be awakened by the channel
+            while let Some(task) = recv.recv().await {
+                match task {
+                    WorkerEvent::Task(task) => queue.push(task),
+                    WorkerEvent::Exit => exiting = true,
+                }
+                loop {
+                    if !exiting {
+                        // Get up to 10 messages at a time
+                        for _ in 0..10 {
```
Thoralf-M, line 122: Why 10?

sdk/src/client/worker.rs
```diff
@@ -0,0 +1,239 @@
...
+                                },
+                                Err(e) => match e {
+                                    TryRecvError::Empty => break,
+                                    TryRecvError::Disconnected => return,
+                                },
+                            }
+                        }
+                    }
+                    if let Some(next) = queue.pop() {
+                        next.await;
```
Thoralf-M, line 136: Does it send any requests in parallel then?

sdk/src/client/worker.rs
DaughterOfMars: Just seemed like a good number 🤷‍♀️

sdk/src/client/worker.rs
DaughterOfMars: Yes, each worker is a parallel task, but the worker itself only processes in order.

sdk/src/client/worker.rs
Thoralf-M: But this will send 10 requests sequentially, only if more than 10 requests are waiting, it will be done in parallel by using another worker, right? Just having a single message seems to be faster

DaughterOfMars: No, it will receive 10 messages from the incoming requests, and add those to the priority queue. Then it will process them sequentially by priority.

DaughterOfMars: Also, the workers do not pull from the same messages, they cannot steal messages from another worker.

Thoralf-M: Well that's what I mean, if one worker has 10 messages, and no other worker anything, then it will send these 10 requests sequentially
If every worker would only have a single message, then the other 9 messages would be split over other workers and could all be sent at the same time (as long as they don't exceed the limit)?

DaughterOfMars: yes, but that cannot happen because the work is distributed across the workers evenly

DaughterOfMars: It is still possible for a worker to be assigned high cost work and thereby end up with a backlog though. If they shared an input source then this could not happen, but then you have to deal with lock contention and I'm not sure it's worth that.

Thoralf-M: Yeah you're right probably not worth it

Thoralf-M (approved):

sdk/src/client/core.rs
```diff
@@ -95,9 +104,33 @@ impl Client {
...
+    where
+        F: 'static + Send + Sync + FnOnce(Self) -> Fut,
+        Fut: futures::Future + Send,
+        Fut::Output: Send,
+    {
+        let client = self.clone();
+        self.worker_pool
+            .process_task(priority, async move { f(client).await })
+            .await
+            .unwrap() // TODO
```
kwek20, line 129: todo 👀

sdk/src/client/api/block_builder/pow.rs
```diff
@@ -8,11 +8,11 @@ use crate::pow::miner::{Miner, MinerBuilder, MinerCancel};
 #[cfg(target_family = "wasm")]
 use crate::pow::wasm_miner::{SingleThreadedMiner, SingleThreadedMinerBuilder};
 use crate::{
-    client::{ClientInner, Error, Result},
+    client::{Client, Error, Result},
     types::block::{parent::Parents, payload::Payload, Block, BlockBuilder, Error as BlockError},
 };
 
-impl ClientInner {
+impl Client {
```
thibault-martinez, line 15: Just making sure, this is not breaking in any way since Client derefs to ClientInner, right?

thibault-martinez: But wait, `ClientInner` is public, so this is breaking change?

DaughterOfMars: Mmmm, yes I suppose so. Technically. But if you used the client this way you probably deserve to have your code break 😆

thibault-martinez: It this change (Client/ClientInner) really required for this PR? Otherwise we just do it in 2.0

DaughterOfMars: Unfortunately it is most definitely required

DaughterOfMars: I will think about alternative impls though. I'm not 100% with this one.

Issue #932, opened 2023-07-26
Remove pow/nonce - python


thibault-martinez:

thibault-martinez:
Closed by https://github.com/iotaledger/iota-sdk/pull/1113

PR #1113, opened 2023-09-01
Remove pow/nonce - python

abdulmth:
fixes https://github.com/iotaledger/iota-sdk/issues/932


abdulmth:
Not sure about removing the nonce from the test values.

thibault-martinez:
> Not sure about removing the nonce from the test values.

Why? Would fail if we leave them.

thibault-martinez (approved):

Thoralf-M (approved):

qrayven (approved):

Alex6323 (approved):

PR #1121, opened 2023-09-04
Fix UNLOCK_INDEX_MAX

thibault-martinez:


sdk/src/types/block/unlock/mod.rs
```diff
@@ -24,9 +24,9 @@ pub const UNLOCK_COUNT_MAX: u16 = INPUT_COUNT_MAX; // 128
 /// The range of valid numbers of unlocks of a transaction.
 pub const UNLOCK_COUNT_RANGE: RangeInclusive<u16> = INPUT_COUNT_RANGE; // [1..128]
 /// The maximum index of unlocks of a transaction.
-pub const UNLOCK_INDEX_MAX: u16 = INPUT_INDEX_MAX; // 127
+pub const UNLOCK_INDEX_MAX: u16 = INPUT_INDEX_MAX - 1; // 126
 /// The range of valid indices of unlocks of a transaction.
```
Thoralf-M: Should we update the comment like this? Since an index can be 127, it just can't be used in other unlocks
```suggestion
/// The range of valid indices of unlocks of a transaction that can be referenced in Reference, Alias or NFT unlocks.
```

abdulmth (dismissed):

sdk/src/types/block/unlock/mod.rs
thibault-martinez: Hmm not sure that improves anything, we're just speaking of the possible values of unlock indices, adding what types of unlock doesn't change much

Thoralf-M: But the ranges only are correct in the context of these unlocks, without the unlocks it's one more

thibault-martinez: But the comment is on an Unlock constant so what's the matter ?

Thoralf-M: `The range of valid indices of unlocks of a transaction.` to me sounds like transaction.unlocks.get(UNLOCK_INDEX_MAX+1) would always fail
But if you think it's fine then we can leave it as it is

Thoralf-M (dismissed):

sdk/src/types/block/unlock/mod.rs
thibault-martinez: Doesn't harm to change it anyway

Thoralf-M (approved):

abdulmth:
CI failing  🤔

abdulmth (approved):

Alex6323 (approved):

kwek20 (approved):

PR #1146, opened 2023-09-06
Simplify nodejs and wasm bindings p. 2

kwek20:
====
This merges into the original to show the difference
Please discuss below ;)
====

Changes:
- Wrapped all errors returned to js in `èrrorHandler`, making parsing on 1 spot
- Added a lot of try/catch to otherwise methods (and make them call errorHandle. This used to return just the str error, not a serialized object)
- Made most methods that didnt need async sync. neon does NOT like async.
- Made wasm side methods sync when possible
- Removed Unneeded error logging

Questions:
- Do we just want to return a string again, or return the whole error object (I vote string for now)
- Alternatively, we can make all methods async, this needs more work on neon side, but makes JS code more maintainable
- Anyone knows what the heck is up with SecretManager on wasm?? (https://github.com/iotaledger/iota-sdk/blob/25d9aa466c4ef40523e8710154664015634ef820/bindings/wasm/lib/bindings.ts#L27)

Todo:
- [x] I noticed i missed wallet.listen cannot be sync, need to revert

Heres my beautifull design incase youre wondering how it all stitches together
Screenshot_841


bindings/wasm/lib/bindings.ts
```diff
@@ -20,6 +21,15 @@ const callUtilsMethod = (method: __UtilsMethods__): any => {
     }
 };
 
+const getSecretManagerFromWallet = (
+    method: WalletMethodHandler,
+): SecretManagerMethodHandler => {
+    // TODO figure out why this one is extensible but client isnt
```
DaughterOfMars, line 27: TODO

DaughterOfMars (approved):

bindings/wasm/lib/bindings.ts
kwek20: yea i cannot figure this one out unfortunately, should i make a ticket or do we accept this solution? :)

DaughterOfMars: what is the problem exactly?

kwek20: SecretManagerMethodHandler constructor checks the type it receives, if its extensible, its a config/serialized  info. if not, its a rust object, and its ready for use. However, for some reason the SecretManagerMethodHandler we return in wasm is diffferent from the nodejs one, and is extensible, causing it to crash. (trying to deserialize a rust pointer)
Thats why here i intercept the wasm method and just force "not extensible". it works :D

kwek20: (probably because nodejs returns Arc<>, but we cant do that in wasm as its not implementing some trait (i forgot which, some wasm stuff)

PR #1152, opened 2023-09-06
Python: add burned_mana field

### LINKED ISSUE #885: Add `burned_mana` field to Block - python
https://github.com/iotaledger/tips-draft/blob/tip46/tips/TIP-0046/tip-0046.md#basic-block

Thoralf-M:
# Description of change

Python: add burned_mana field

Closes #885 

## Type of change


thibault-martinez (approved):

Alex6323 (approved):

kwek20 (approved):

bindings/python/iota_sdk/types/block.py
```diff
@@ -18,11 +18,13 @@ class Block:
     Attributes:
         protocol_version: The protocol version with which this block was issued.
         parents: The parents of this block.
+        burned_mana: The amount of Mana the Account identified by the IssuerId is at most willing to burn for this block.
         payload: The optional payload of this block.
     """
 
     protocol_version: int
     parents: List[HexStr]
+    burned_mana: str
```
DaughterOfMars, line 27: Not a number?

Thoralf-M: No, it gets serialized as string

DaughterOfMars: Yeah, but does that mean the field has to be a string in the type?

Thoralf-M: Hm yeah if we add some custom dict/json methods then we should be able to do it, will have to investigate

DaughterOfMars: Do we really use strings for all u64 in python?

Thoralf-M: Currently we do

DaughterOfMars: Dang, alright

DaughterOfMars (approved):

PR #1156, opened 2023-09-06
Fix Blocking behavior in Node.js bindings

### LINKED ISSUE #1097: Improve perfomance of @iota/sdk
## Description

The performance of our app degrades a lot when switching over from @iota/wallet to @iota/sdk.

We measured the blocking time, which is a metric that makes the UI non-responsive due to processes running in the main process. On figure 1. you can see the performance profiling for @iota/wallet (blocking time 935 ms). On figure 2. you can see the performance profileing of @iota/sdk (blocking time 4583 ms). 

Additionally, we measured the time it takes to perform the API calls during login (I'll add some time measurements below):

The following functions take consistently longer:

    getNodeInfo()
    getAccounts()

The following functions take a long time generally:

    registerProposals
    getParticipationEventStatus
    
Time Measurements:
[Uploading performance-measure-sdk.md…]()

You can see that on the top, there is a full yellow bar for figure 2, which reflects that the main process is fully blocked. On figure one, the resources are much more balanced out. 

Figure 1 - Profiling of the login process with @iota/wallet 

image

Figure 2 - Profiling of the login process with @iota/sdk

image

After doing some more profiling, we discovered that `listen_wallet` might be the culprit:

Figure 3 - Profiling indicates that listen_wallet blocks the main process
image
image

## Motivation

Without performance optimizations the UI becomes non-responsive resulting in a terrible UX.

No.

abdulmth:
fixes https://github.com/iotaledger/iota-sdk/issues/1097

Tokio's mpsc channel doesn't work since receiving is async. So I think panic is good enough for now, and it's the same behavior in the old bindings.

### DIFF
--- a/bindings/nodejs/src/client.rs
+++ b/bindings/nodejs/src/client.rs
@@ -86,7 +86,6 @@ pub fn call_client_method(mut cx: FunctionContext) -> JsResult<JsUndefined> {
     let method_handler = Arc::clone(&cx.argument::<JsBox<ClientMethodHandlerWrapper>>(1)?.0);
     let callback = cx.argument::<JsFunction>(2)?.root(&mut cx);
 
-    let (sender, receiver) = std::sync::mpsc::channel();
     crate::RUNTIME.spawn(async move {
         if let Some(method_handler) = &*method_handler.read().await {
             let (response, is_error) = method_handler.call_method(method).await;
@@ -108,17 +107,9 @@ pub fn call_client_method(mut cx: FunctionContext) -> JsResult<JsUndefined> {
                 Ok(())
             });
         } else {
-            // Notify that the client got destroyed
-            // Safe to unwrap because the receiver is waiting on it
-            sender.send(()).unwrap();
+            panic!("Client handler got destroyed")
         }
     });
-
-    if receiver.recv().is_ok() {
-        return cx.throw_error(
-            serde_json::to_string(&Response::Panic("Client got destroyed".to_string())).expect("json to string error"),
-        );
-    }
 
     Ok(cx.undefined())
 }
--- a/bindings/nodejs/src/client.rs
+++ b/bindings/nodejs/src/client.rs
@@ -107,7 +107,7 @@ pub fn call_client_method(mut cx: FunctionContext) -> JsResult<JsUndefined> {
                 Ok(())
             });
         } else {
-            panic!("Client handler got destroyed")
+            panic!("Client got destroyed")
         }
     });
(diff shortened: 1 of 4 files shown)


DaughterOfMars (dismissed):

thibault-martinez (changes requested):
Changelog

DaughterOfMars (dismissed):

bindings/nodejs/src/client.rs
```diff
@@ -108,18 +107,10 @@ pub fn call_client_method(mut cx: FunctionContext) -> JsResult<JsUndefined> {
                 Ok(())
             });
         } else {
-            // Notify that the client got destroyed
-            // Safe to unwrap because the receiver is waiting on it
-            sender.send(()).unwrap();
+            panic!("Client handler got destroyed")
```
thibault-martinez:
```suggestion
            panic!("Client got destroyed")
```

thibault-martinez (dismissed):

Thoralf-M (dismissed):

Alex6323 (dismissed):

bindings/nodejs/tests/wallet/wallet.spec.ts
```diff
@@ -132,53 +132,6 @@ describe('Wallet', () => {
...
-            throw 'Should return an error because the wallet got destroyed';
-        } catch (err: any) {
-            expect(err).toContain('Wallet got destroyed');
-        }
-
-        try {
-            const client = await wallet.getClient();
-            throw 'Should return an error because the wallet got destroyed';
-        } catch (err: any) {
-            expect(err).toContain('Wallet got destroyed');
```
thibault-martinez, line 178: Just wondering, no way to test a panic like in rust?

abdulmth: I'm not aware of something that does it, it just exits :|

abdulmth: Also, this is the worst case, so testing it isn't very useful IMO

thibault-martinez (approved):

bindings/nodejs/tests/wallet/wallet.spec.ts
kwek20: did a quick check, yes neon with Throw an `Error("internal error in Neon module: Wallet got destroyed"` which can be caught with try/catch

Thoralf-M (approved):

PR #1160, opened 2023-09-06
Add IRC27 and IRC30 metadata utilities

### LINKED ISSUE #504: Provide functions to create TIP27 and TIP30 compatible assets
## Description

We should provide a function/class to provide TIP27 or TIP30 compatible data. But ideally we probably create helper functions like `mint_IRC27_nft` or `mint_IRC30_token`. (Could also just be `mint` from different namespaces)

No.

DaughterOfMars:
# Description of change

This PR adds two kinds of NFT Immutable Metadata schemas. These can be created and used directly as a `MetadataFeature` in the immutable features of an NFT.

Closes #504 

## Type of change

I added tests for serialization, and used the metadata in a separate NFT test. However, there is an issue with that test. The input selection returns an `Err(InsufficientAmount { found: 302000, required: 490500 })` when it should return `UnfulfillableRequirement`.

### DIFF
--- a/sdk/src/types/block/output/feature/metadata.rs
+++ b/sdk/src/types/block/output/feature/metadata.rs
@@ -75,6 +75,333 @@ impl core::fmt::Debug for MetadataFeature {
 }
 
 #[cfg(feature = "serde")]
+pub(crate) mod irc_27 {
+    use alloc::collections::{BTreeMap, BTreeSet};
+
+    use getset::Getters;
+    use serde::{Deserialize, Serialize};
+    use url::Url;
+
+    use super::*;
+    use crate::types::block::address::Bech32Address;
+
+    /// The IRC27 NFT standard schema.
+    #[derive(Clone, Debug, Serialize, Deserialize, Getters, PartialEq)]
+    #[serde(rename_all = "camelCase")]
+    #[serde(tag = "standard", rename = "IRC27")]
+    #[getset(get = "pub")]
+    pub struct Irc27Metadata {
+        version: String,
+        /// The media type (MIME) of the asset.
+        ///
+        /// ## Examples
+        /// - Image files: `image/jpeg`, `image/png`, `image/gif`, etc.
+        /// - Video files: `video/x-msvideo` (avi), `video/mp4`, `video/mpeg`, etc.
+        /// - Audio files: `audio/mpeg`, `audio/wav`, etc.
+        /// - 3D Assets: `model/obj`, `model/u3d`, etc.
+        /// - Documents: `application/pdf`, `text/plain`, etc.
+        #[serde(rename = "type")]
+        media_type: String,
+        /// URL pointing to the NFT file location.
+        uri: Url,
+        /// The human-readable name of the native token.
+        name: String,
+        /// The human-readable collection name of the native token.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        collection_name: Option<String>,
+        /// Royalty payment addresses mapped to the payout percentage.
+        #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+        royalties: BTreeMap<Bech32Address, f64>,
+        /// The human-readable name of the native token creator.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        issuer_name: Option<String>,
+        /// The human-readable description of the token.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        description: Option<String>,
+        /// Additional attributes which follow [OpenSea Metadata standards](https://docs.opensea.io/docs/metadata-standards).
+        #[serde(default, skip_serializing_if = "BTreeSet::is_empty")]
+        attributes: BTreeSet<Attribute>,
+    }
+
+    impl Irc27Metadata {
+        pub fn new(media_type: impl Into<String>, uri: Url, name: impl Into<String>) -> Self {
+            Self {
+                version: "v1.0".to_owned(),
+                media_type: media_type.into(),
+                uri,
+                name: name.into(),
+                collection_name: Default::default(),
+                royalties: Default::default(),
+                issuer_name: Default::default(),
+                description: Default::default(),
+                attributes: Default::default(),
+            }
+        }
+
+        pub fn with_collection_name(mut self, collection_name: impl Into<String>) -> Self {
+            self.collection_name.replace(collection_name.into());
+            self
+        }
+
+        pub fn add_royalty(mut self, address: Bech32Address, percentage: f64) -> Self {
+            self.royalties.insert(address, percentage);
+            self
+        }
+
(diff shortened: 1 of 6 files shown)


sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -74,6 +74,333 @@ impl core::fmt::Debug for MetadataFeature {
...
+    impl TryFrom<Irc27Metadata> for MetadataFeature {
+        type Error = Error;
+        fn try_from(value: Irc27Metadata) -> Result<Self, Error> {
+            Self::new(value.to_bytes())
+        }
+    }
+
+    #[derive(Clone, Debug, Serialize, Deserialize, Getters, PartialEq, Eq)]
+    #[getset(get = "pub")]
+
```
kwek20: sad format noises.
also, missing rename camelCase (not that its likely used outside of Irc27Metadata but still)

Thoralf-M:
The test probably fails because the new metadata is smaller and then there is not enough amount for a remainder output and the amount requirement is just checked before. I would just revert the changes in this test.

Would be nice to have the examples make use of the new types, for example in  https://github.com/iotaledger/iota-sdk/blob/develop/sdk/examples/how_tos/native_tokens/create.rs#L68 and https://github.com/iotaledger/iota-sdk/blob/a3260a4ed7073167b20376077b2b90ecefa5b4e9/sdk/examples/how_tos/nfts/mint_nft.rs#L66

Also bindings are not updated

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -74,6 +74,333 @@ impl core::fmt::Debug for MetadataFeature {
...
+        description: Option<String>,
+        /// Additional attributes which follow [OpenSea Metadata standards](https://docs.opensea.io/docs/metadata-standards).
+        #[serde(default, skip_serializing_if = "BTreeSet::is_empty")]
+        attributes: BTreeSet<Attribute>,
+    }
+
+    impl Irc27Metadata {
+        pub fn new(media_type: impl Into<String>, uri: Url, name: impl Into<String>) -> Self {
+            Self {
+                version: "v1.0".to_owned(),
```
Thoralf-M, line 156: Also add `with_version()` so users can have a higher version without requiring a library update? (if no the other fields don't require a breaking change then)

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Not camelCase here https://docs.opensea.io/docs/metadata-standards#attributes

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -74,6 +74,333 @@ impl core::fmt::Debug for MetadataFeature {
...
+    /// The IRC30 NFT standard schema.
+    #[derive(Clone, Debug, Serialize, Deserialize, Getters, PartialEq, Eq)]
+    #[serde(rename_all = "camelCase")]
+    #[serde(tag = "standard", rename = "IRC30")]
+    #[getset(get = "pub")]
+    pub struct Irc30Metadata {
+        /// The human-readable name of the native token.
+        name: String,
+        /// The human-readable description of the token.
+        description: String,
```
Thoralf-M: This field is also optional

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -74,6 +74,333 @@ impl core::fmt::Debug for MetadataFeature {
...
+        /// The symbol/ticker of the token.
+        symbol: String,
+        /// Number of decimals the token uses (divide the token amount by 10^decimals to get its user representation).
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        decimals: Option<u32>,
```
Thoralf-M: Not optional

sdk/src/types/block/output/feature/metadata.rs
DaughterOfMars: But if the version changes, wouldn't the entire struct be different?

sdk/src/types/block/output/feature/metadata.rs
DaughterOfMars: thanks

DaughterOfMars:
Created #1169 and #1170 to add bindings

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Don't know, but maybe there can be a minor version change which keeps the structure the same, but adds some special handling with the data in some of the fields? Just sounds better to not have it hardcoded to this specific version

DaughterOfMars: Okay, but in that case why would a user change the version?

Thoralf-M: Can't tell you since I have no idea how it would look, I think it's just generally not a good idea to hardcode values and make them not adjustable
Such a change could at least easily be added later without breaking something, so resolving this conversation

Thoralf-M (changes requested):
Clippy

sdk/tests/client/input_selection/nft_outputs.rs
```diff
@@ -1188,9 +1188,16 @@ fn changed_immutable_metadata() {
     let protocol_parameters = protocol_parameters();
     let nft_id_1 = NftId::from_str(NFT_ID_1).unwrap();
 
+    let metadata = iota_sdk::types::block::output::feature::Irc27Metadata::new(
```
Thoralf-M, line 1192:
```
failed to resolve: could not find `Irc27Metadata` in `feature`
error[E0433]: failed to resolve: could not find `Irc27Metadata` in `feature`
    --> sdk/tests/client/input_selection/nft_outputs.rs:1191:61
     |
1191 |     let metadata = iota_sdk::types::block::output::feature::Irc27Metadata::new(
     |                                                             ^^^^^^^^^^^^^ could not find `Irc27Metadata` in `feature`
```

DaughterOfMars: so annoying

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -19,6 +19,29 @@ pub struct MetadataFeature(
...
+            type Error = Error;
+
+            fn try_from(value: $type) -> Result<Self, Self::Error> {
+                Vec::<u8>::from(value).try_into()
+            }
+        }
+    };
+}
+impl_from_vec!(&str);
+impl_from_vec!(String);
```
Thoralf-M, line 34: no_std complains now

DaughterOfMars: ffs kill me

Thoralf-M (approved):

kwek20 (approved):

PR #1171, opened 2023-09-07
Fix type of `value` property in `CustomAddress` (Node.js)

### LINKED ISSUE #1060: [Task]: Modify type of CustomAddress value in nodejs
## Description

The type for the value property for the `CustomAddress` remainder strategy in [nodejs](https://github.com/iotaledger/iota-sdk/blob/develop/bindings/nodejs/lib/types/wallet/transaction-options.ts#L59) is `string`, but it should be `AccountAddress` as it appears in [sdk](https://github.com/iotaledger/iota-sdk/blob/c0094b2a7a0c56bedf249e8ea0a7d3787cd6400f/sdk/src/wallet/account/operations/transaction/options.rs#L75)

## Motivation

Return correct type for `CustomAddress` value in remainder strategy for `nodejs`

## Requirements

1. Return correct type for `CustomAddress` value

No.

abdulmth:
fixes https://github.com/iotaledger/iota-sdk/issues/1060

After this change `CustomAddress` matches the rust side which is also used in `TransactionOptionsDto`

### DIFF
--- a/bindings/nodejs/CHANGELOG.md
+++ b/bindings/nodejs/CHANGELOG.md
@@ -19,6 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 1.0.10 - 2023-mm-dd
+
+### Fixed
+
+- Type of `value` property in `CustomAddress`;
+
 ## 1.0.9 - 2023-09-07
 
 ### Added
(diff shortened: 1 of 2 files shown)


DaughterOfMars (dismissed):

thibault-martinez (changes requested):
Changelog

DaughterOfMars (approved):

thibault-martinez (approved):

Thoralf-M (approved):

PR #1179, opened 2023-09-07
feat(cli): lists NTs/NFTs/Aliases/Foundries in addresses overview

### LINKED ISSUE #261: CLI: Add NTs/NFTs/Aliases/Foundries to addresses overview
## Description

Add NTs/NFTs/Aliases/Foundries to addresses overview
https://github.com/iotaledger/iota-sdk/pull/259#discussion_r1166529957
Probably should reuse something from the account.balance() method

## Motivation

Better visibility of what belongs to which address

## Open questions (optional)

Do we want to have the full balance object for every address or is that too large?

No.

Alex6323:
# Description of change

When calling the `addressess` command in an account using the CLI wallet assets like Native Tokens, NFTs, ... associated with an address will be listed.

Fixes #261 

## Type of change

Checked output in the CLI wallet.


cli/src/command/account.rs
```diff
@@ -919,40 +919,68 @@ async fn print_address(account: &Account, address: &AccountAddress) -> Result<()
...
+                    }
+                    if output_data.output.is_nft() {
+                        address_nfts.push(output_data.output.as_nft().nft_id_non_null(&output_id));
+                    }
+                    if output_data.output.is_alias() {
+                        address_aliases.push(output_data.output.as_alias().alias_id_non_null(&output_id));
+                    }
+                    if output_data.output.is_foundry() {
+                        address_foundries.push(output_data.output.as_foundry().id());
+                    }
```
thibault-martinez: if is_ + as_ sounds like a match...

DaughterOfMars: or if let

cli/src/command/account.rs
```diff
@@ -919,40 +919,68 @@ async fn print_address(account: &Account, address: &AccountAddress) -> Result<()
     let addresses = account.addresses_with_unspent_outputs().await?;
     let current_time = iota_sdk::utils::unix_timestamp_now().as_secs() as u32;
 
+    let mut output_ids = &Vec::new();
```
thibault-martinez: Bit weird. You're putting new lines in the log so you could just print in the if and not do this

cli/src/command/account.rs
Alex6323: normally it does, yes, but the `as_` methods panic if you call them on the wrong output type.

thibault-martinez: Yeah I know that, but why not match ?

cli/src/command/account.rs
Alex6323: not exactly sure what you mean tbh. Can you elaborate?

thibault-martinez: It looks like you're doing this weird thing for the log at the end but you could log it earlier and have the exact same result

cli/src/command/account.rs
Alex6323: dunno ... I found it readable enough I guess.

Alex6323: changing it ...

thibault-martinez: Yeah but it doesn't scale, we're 100% going to forget about adding delegation output in 2.0

cli/src/command/account.rs
Alex6323: still not sure what the weird thing is you're talking about :sweat_smile: , I basically(!) kept the log formatting as it was .. I just added the assets to it, no?

thibault-martinez: You're taking a ref of a new Vec but it's not necessary at all

Alex6323: see .. the way it was before ... you would have to edit 2 formatted strings (and keep them in sync). I consolidated that into just 1. That's why it has moved out of the if block.

thibault-martinez: Idk, looks like it could be reworked and logged individually since we new line anyway

Alex6323: I can change it so they are individually logged if you want. As I said ... this is how it was done before and I just kept it :shrug:

Alex6323: Lol...Gh didnt show me the text you wrote about the &Vec::new 🙈 when I replied. Yeah that's weird, but a relatively small price to pay and compiler checked contrary to maintaining 2 formatted strings.

cli/src/command/account.rs
Alex6323: replaced it as suggested with a match

cli/src/command/account.rs
Alex6323: not sure you're okay with this now ... feels like there was something else you didn't like with the logging (but as I said nothing I fundamentally changed with this PR). But I prefer this solution over having another format string with hard-coded `[]`s and `0` in case the if-block isn't executed. The downside of course are the allocations but no reason to optimize this part of the code IMO.

cli/src/command/account.rs
```diff
@@ -926,40 +926,58 @@ async fn print_address(account: &Account, address: &AccountAddress) -> Result<()
...
-        let mut address_amount = 0;
-        for output_id in addresses[index].output_ids() {
+        output_ids.extend(addresses[index].output_ids());
+
+        for output_id in &output_ids {
             if let Some(output_data) = account.get_output(output_id).await {
-                // Output might be associated with the address, but can't unlocked by it, so we check that here
+                // Output might be associated with the address, but can't be unlocked by it, so we check that here
+                // TODO: Should we just skip outputs for which this call fails? (that's only possible for
+                // TreasuryOutputs, right?)
```
Thoralf-M: TreasuryOutput will never be in an account, so no need to add extra logic for it here

cli/src/command/account.rs
```diff
@@ -926,40 +926,58 @@ async fn print_address(account: &Account, address: &AccountAddress) -> Result<()
...
                 let (required_address, _) =
                     output_data
                         .output
                         .required_and_unlocked_address(current_time, output_id, None)?;
-                if *address.address().as_ref() == required_address {
+
+                if address.address().as_ref() == &required_address {
+                    if let Some(nts) = output_data.output.native_tokens() {
+                        address_nts.extend(nts.iter().map(|nt| nt.token_id()));
```
Thoralf-M: If two outputs have the same native token, it would create two entries this way, so should probably be a HashSet where you insert or maybe NativeTokensBuilder would be even more interesting, so we can also have the amount?

Alex6323: yeah, good point

cli/src/command/account.rs
Alex6323: yeah, that basically means we can just unwrap, currently we early return and abort which feels off in this case because there's no other reason for the call to fail then this TreasuryOutput thing.

cli/src/command/account.rs
Alex6323: changed it to use the `NativeTokensBuilder`.

cli/src/command/account.rs
```diff
@@ -926,40 +926,62 @@ async fn print_address(account: &Account, address: &AccountAddress) -> Result<()
     let addresses = account.addresses_with_unspent_outputs().await?;
     let current_time = iota_sdk::utils::unix_timestamp_now().as_secs() as u32;
 
+    let mut output_ids = Vec::new();
+    let mut address_amount = 0;
+    let mut address_nts = NativeTokensBuilder::new();
+    let mut address_nfts = Vec::new();
+    let mut address_aliases = Vec::new();
+    let mut address_foundries = Vec::new();
```
Thoralf-M: Could also remove the `address_` prefix, but I don't really care :trollface:

Thoralf-M (dismissed):

cli/src/command/account.rs
Alex6323: yeah, it annoys me too :smile:

thibault-martinez: Well let's remove if it annoys the both of you? Also annoys me.

cli/src/command/account.rs
```diff
@@ -919,40 +919,68 @@ async fn print_address(account: &Account, address: &AccountAddress) -> Result<()
...
+
     if let Ok(index) = addresses.binary_search_by_key(&(address.key_index(), address.internal()), |a| {
         (a.key_index(), a.internal())
     }) {
-        let mut address_amount = 0;
-        for output_id in addresses[index].output_ids() {
+        output_ids = addresses[index].output_ids();
```
DaughterOfMars: would this work instead?

```rust
let mut output_ids = &[];
...
output_ids = addresses[index].output_ids().as_slice();
```

Alex6323: yes, that'll work, but you have to tell Rust that you want an empty slice (otherwise it'll infer an array of length 0)

Alex6323: I'll change it.

thibault-martinez (changes requested):
Changelog

DaughterOfMars (dismissed):

Thoralf-M (dismissed):

Thoralf-M (dismissed):

DaughterOfMars (dismissed):

cli/src/command/account.rs
```diff
@@ -926,40 +926,62 @@ async fn print_address(account: &Account, address: &AccountAddress) -> Result<()
...
-            "{log}\nOutputs: {:#?}\nBase coin amount: {}\n",
-            addresses[index].output_ids(),
-            address_amount
-        );
-    } else {
-        log = format!("{log}\nOutputs: []\nBase coin amount: 0\n");
     }
 
+    log = format!(
+        "{log}\n Outputs: {:#?}\n Base coin amount: {}\n NTs: {:?}\n NFTs: {:?}\n Aliases: {:?}\n Foundries: {:?}\n",
```
thibault-martinez:
```suggestion
        "{log}\n Outputs: {:#?}\n Base coin amount: {}\n Native Tokens: {:?}\n NFTs: {:?}\n Aliases: {:?}\n Foundries: {:?}\n",
```

cli/src/command/account.rs
```diff
@@ -926,40 +926,62 @@ async fn print_address(account: &Account, address: &AccountAddress) -> Result<()
...
+
+        for output_id in output_ids {
             if let Some(output_data) = account.get_output(output_id).await {
-                // Output might be associated with the address, but can't unlocked by it, so we check that here
-                let (required_address, _) =
-                    output_data
-                        .output
-                        .required_and_unlocked_address(current_time, output_id, None)?;
-                if *address.address().as_ref() == required_address {
+                // Output might be associated with the address, but can't be unlocked by it, so we check that here
```
thibault-martinez:
```suggestion
                // Output might be associated with the address, but can't be unlocked by it, so we check that here.
```

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1195, opened 2023-09-11
Add a protocol parameters hash test

### LINKED ISSUE #1132: Add ProtocolParameter hash test against TIP
https://github.com/iotaledger/tips-draft/pull/25/commits/63d68979c153c6856413c7c3d8b445da4af05ea5

DaughterOfMars:
# Description of change

Adds a test for the protocol parameters using validated data from the TIP.

Closes #1132
Closes https://github.com/iotaledger/iota-sdk/issues/1124


sdk/src/types/block/protocol.rs
```diff
@@ -320,3 +324,82 @@ impl_id!(
 
 #[cfg(feature = "serde")]
 string_serde_impl!(ProtocolParametersHash);
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn params_serde_hash() {
```
thibault-martinez: Can you put a link to the TIP example pls

sdk/src/types/block/protocol.rs
```diff
@@ -25,6 +25,9 @@ use crate::types::block::{helper::network_name_to_id, output::RentStructure, Con
 )]
 #[getset(get_copy = "pub")]
 pub struct ProtocolParameters {
+    /// The layout type.
+    #[serde(rename = "type")]
+    pub(crate) kind: u8,
```
thibault-martinez, line 30: So we're adding it in the end ?

sdk/src/types/block/address/bech32.rs
```diff
@@ -18,12 +18,21 @@ use packable::{
...
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct Hrp {
     inner: [u8; 83],
     len: u8,
 }
 
+impl core::fmt::Debug for Hrp {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("Hrp")
+            .field("inner", &prefix_hex::encode(&self.inner[..self.len as usize]))
```
thibault-martinez, line 31: I know it's debug but this is not very helpful...

thibault-martinez: For the record, it was previously also not helpful

sdk/src/types/block/protocol.rs
DaughterOfMars: I think it's the simplest solution.

sdk/src/types/block/address/bech32.rs
DaughterOfMars: How is it not helpful?

thibault-martinez: What's the point of printing the inner as hex? I don't see why we wouldn't print this as a String

DaughterOfMars: That's what the display impl is for. Debug is for debugging

thibault-martinez: Yeah sure, but there are no practical use of HRP as raw bytes. If anything, I'd say it's preventing us from debugging correctly as we'd have to manually convert it to see what's up

DaughterOfMars: You could also argue that having both impls print the string could cause debugging issues because you need to convert back to bytes. I like that if you want the string, you can just use display.

thibault-martinez: Sure but you can't do anything when it's nested in something and you are using debug 🤷🏻‍♂️

thibault-martinez:
Is this still a draft?

DaughterOfMars:
Until feat/rent_cost is merged, ye

sdk/tests/types/rent.rs
```diff
@@ -8,11 +8,21 @@ use iota_sdk::types::block::{
...
 
 const BYTE_COST: u32 = 1;
-const FACTOR_KEY: u8 = 10;
 const FACTOR_DATA: u8 = 1;
+const FACTOR_KEY: u8 = 10;
+const FACTOR_DELEGATION: u8 = 1;
+const FACTOR_STAKING_FEATURE: u8 = 1;
+const FACTOR_ISSUER_KEYS: u8 = 1;
 
 fn config() -> RentStructure {
```
thibault-martinez: Any reason to not use default here ?

thibault-martinez (changes requested):
Block ID test is failing

sdk/tests/types/rent.rs
DaughterOfMars: Nope

thibault-martinez (changes requested):
CI

DaughterOfMars:
Not sure what to do about CI, should I just disable the failing tests?

sdk/tests/types/rent.rs
```diff
@@ -2,21 +2,13 @@
...
     rand::output::{rand_account_output, rand_basic_output, rand_foundry_output, rand_nft_output},
 };
 
-const BYTE_COST: u32 = 1;
-const FACTOR_KEY: u8 = 10;
-const FACTOR_DATA: u8 = 1;
-
-fn config() -> RentStructure {
-    RentStructure::new(BYTE_COST, FACTOR_KEY, FACTOR_DATA)
-}
```
thibault-martinez, line 16: Seems like the CI fails are due to this? Values are actually different. Maybe just bring it back with a TODO to use default and change tests ?

DaughterOfMars: I think it fails because the structs cost more rent now

thibault-martinez: Ok well TODO and/or issue then

Thoralf-M (dismissed):

thibault-martinez (approved):

Thoralf-M (approved):

Alex6323 (approved):

kwek20 (approved):

PR #1203, opened 2023-09-12
Add alias transition examples

### LINKED ISSUE #1084: Create missing Alias Output Examples with related docs
## Description

Create examples for the Alias Output section. It should contain:

- [ ]  Governance Transitions
- [ ]  State Transitions
- [x]  Unlock Alias Funds https://github.com/iotaledger/iota-sdk/blob/develop/sdk/examples/how_tos/alias_wallet/transaction.rs

No

Thoralf-M:
# Description of change

Add alias transition examples

Closes #1084 

## Type of change

- Documentation enhancement

Running the examples


bindings/nodejs/examples/how_tos/alias/governance-transition.ts
```diff
@@ -0,0 +1,109 @@
...
+        const aliasOutputData = (
+            await account.unspentOutputs({ aliasIds: [aliasId] })
+        )[0];
+        console.log(
+            `Alias ${aliasId} found in unspent output: ${aliasOutputData.outputId}`,
+        );
+
+        await wallet.setStrongholdPassword(process.env.STRONGHOLD_PASSWORD);
+
+        const newStateController = Utils.parseBech32Address(
```
DaughterOfMars, line 61: Do we really need to call parse here?

bindings/nodejs/examples/how_tos/alias/governance-transition.ts
```diff
@@ -0,0 +1,109 @@
...
+                } else {
+                    return unlock;
+                }
+            },
+        );
+
+        const updatedAliasOutput = await (
+            await wallet.getClient()
+        ).buildAliasOutput({
+            aliasId,
```
DaughterOfMars, line 81: Do we need to check for null here like in rust?

bindings/nodejs/examples/how_tos/alias/governance-transition.ts
Thoralf-M: Yes because it's a bech32 address string, but we need to provide is as address with type

bindings/nodejs/examples/how_tos/alias/governance-transition.ts
Thoralf-M: No, since we get the id from the balance it's already not null
In the Rust example it's actually also not needed, but thought it could still be interesting to have in an example in case the id is not returned from the balance, but another output

bindings/nodejs/examples/how_tos/alias/governance-transition.ts
DaughterOfMars: Why do we not already deserialize the address as the proper type from the generate address call?

bindings/nodejs/examples/how_tos/alias/governance-transition.ts
DaughterOfMars: Hmm, well imo they should match so maybe just add a comment about it instead?

bindings/nodejs/examples/how_tos/alias/governance-transition.ts
Thoralf-M: Because in most cases users want the bech32 address

DaughterOfMars: I think I misunderstood 🤷‍♀️

bindings/nodejs/examples/how_tos/alias/governance-transition.ts
Thoralf-M: Just removed it, because I don't know where and how to put it without it being more confusing for users, since in this example it's not needed

DaughterOfMars (approved):

Alex6323 (approved):

DaughterOfMars (commented):
Do we want a changelog entry for this?

Thoralf-M:
Why would we? It's only examples

PR #1209, opened 2023-09-12
Replace `|` with `Union` type for Python 3.9 compatibility

### LINKED ISSUE #1202: Verify min Python version requirement
## Description

Currently we say `Python 3.10+` is required which is probably correct, but then we shouldn't build any binaries for 3.9 anymore 
https://github.com/iotaledger/iota-sdk/blob/75e919b896a9431221de6cb04fd53fc517a1a61b/.github/workflows/bindings-python-publish.yml#L14
https://github.com/iotaledger/iota-sdk/blob/75e919b896a9431221de6cb04fd53fc517a1a61b/.github/workflows/bindings-python-publish.yml#L109

```
    address: Ed25519Address | AliasAddress | NFTAddress
E   TypeError: unsupported operand type(s) for |: 'type' and 'type'
```
`|` for type unions seems to be introduced in 3.10, but if it's only in the typings, we can maybe change to still support 3.9 without breaking change?

Thoralf-M:
# Description of change

Replace `|` with `Union` type for Python 3.9 compatibility

Closes #1202 

## Type of change

Running an example with 3.9

### DIFF
--- a/bindings/python/CHANGELOG.md
+++ b/bindings/python/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
-## 1.0.2 - 2023-MM-DD
+## 1.0.2 - 2023-09-12
 
 ### Added
 
--- a/bindings/python/CHANGELOG.md
+++ b/bindings/python/CHANGELOG.md
@@ -25,6 +25,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `ClientOptions::maxParallelApiRequests`;
 
+### Changed
+
+- Replaced `|` with `Union` type for Python 3.9 compatibility;
+
 ### Fixed
 
 - `Utils::parse_bech32_address()`;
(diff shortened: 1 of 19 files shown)


kwek20 (dismissed):

bindings/python/CHANGELOG.md
```diff
@@ -25,6 +25,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `ClientOptions::maxParallelApiRequests`;
 
+### Changed
+
+- Replaced `|` with `Union` type for Python 3.9 compatibility;
```
thibault-martinez, line 30: Do you want to release this today?

Thoralf-M: Yeah let's try, but also with https://github.com/iotaledger/iota-sdk/pull/1203 if possible since this will fix another thing

Thoralf-M: Actually will split that one

Thoralf-M: https://github.com/iotaledger/iota-sdk/pull/1210, so it's easier to review

thibault-martinez (approved):

DaughterOfMars (dismissed):
I actually much prefer this syntax. I wonder if we can define types for them?

thibault-martinez (changes requested):
Conflicts

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1215, opened 2023-09-12
Error in migrate_db_chrysalis_to_stardust() if no chrysalis data exists

Thoralf-M:
# Description of change

Error in migrate_db_chrysalis_to_stardust() if no chrysalis data exists

## Type of change

Added test

## Change checklist

### DIFF
--- a/sdk/CHANGELOG.md
+++ b/sdk/CHANGELOG.md
@@ -19,6 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 1.0.3 - 2023-MM-DD
+
+### Changed
+
+- `migrate_db_chrysalis_to_stardust()` returns an error if no chrysalis data was found;
+
 ## 1.0.3 - 2023-09-07
 
 ### Added
--- a/sdk/CHANGELOG.md
+++ b/sdk/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
-## 1.0.3 - 2023-MM-DD
+## 1.0.4 - 2023-MM-DD
 
 ### Changed
 
--- a/sdk/src/wallet/migration/chrysalis.rs
+++ b/sdk/src/wallet/migration/chrysalis.rs
@@ -77,6 +77,11 @@ pub async fn migrate_db_chrysalis_to_stardust(
         ));
     }
     let chrysalis_data = get_chrysalis_data(chrysalis_storage_path, password)?;
+    if chrysalis_data.is_empty() {
+        return Err(crate::wallet::Error::Migration(
+            "no chrysalis data to migrate".to_string(),
+        ));
+    }
 
     // create new accounts base on previous data
     let (new_accounts, secret_manager_dto) = migrate_from_chrysalis_data(&chrysalis_data, &storage_path_string, false)?;
--- a/sdk/src/wallet/migration/chrysalis.rs
+++ b/sdk/src/wallet/migration/chrysalis.rs
@@ -71,12 +71,12 @@ pub async fn migrate_db_chrysalis_to_stardust(
     // `/db` will be appended to the chrysalis storage path, because that's how it was done in the chrysalis wallet
     let chrysalis_storage_path = &(*storage_path_string).join("db");
 
-    let chrysalis_data = get_chrysalis_data(chrysalis_storage_path, password)?;
-    if chrysalis_data.is_empty() {
+    if !chrysalis_storage_path.is_dir() {
         return Err(crate::wallet::Error::Migration(
             "no chrysalis data to migrate".to_string(),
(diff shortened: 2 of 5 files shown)


DaughterOfMars (dismissed):

sdk/CHANGELOG.md
```diff
@@ -19,6 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 1.0.3 - 2023-MM-DD
```
thibault-martinez:
```suggestion
## 1.0.4 - 2023-MM-DD
```

DaughterOfMars (dismissed):

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -71,6 +71,11 @@ pub async fn migrate_db_chrysalis_to_stardust(
     // `/db` will be appended to the chrysalis storage path, because that's how it was done in the chrysalis wallet
     let chrysalis_storage_path = &(*storage_path_string).join("db");
 
+    if !chrysalis_storage_path.is_dir() {
```
thibault-martinez, line 74: Hmm idk, this doesn't seem super robust. We're just relying on the existence of the `/db` part to determine if it's chrysalis or not?

sdk/tests/wallet/chrysalis_migration.rs
```diff
@@ -348,6 +348,20 @@ async fn migrate_chrysalis_stronghold() -> Result<()> {
     tear_down(storage_path)
 }
 
+#[tokio::test]
+async fn migrate_empty_chrysalis_db() -> Result<()> {
+    iota_stronghold::engine::snapshot::try_set_encrypt_work_factor(0).unwrap();
+    let storage_path = "migrate_empty_chrysalis_db";
```
thibault-martinez, line 354: I feel like the test would be more relevant if we were actually trying to migrate a stardust db, surely we have some in fixtures we could use.

sdk/tests/wallet/chrysalis_migration.rs
```diff
@@ -348,6 +348,27 @@ async fn migrate_chrysalis_stronghold() -> Result<()> {
...
+    setup(storage_path)?;
+
+    // Copy db so the original doesn't get modified
+    copy_folder("./tests/wallet/fixtures/check_existing_4_db_test", storage_path).unwrap();
+
+    assert!(matches!(
+        migrate_db_chrysalis_to_stardust("migrate_empty_chrysalis_db", None, None).await,
+        Err(iota_sdk::wallet::error::Error::Migration(msg)) if msg == "no chrysalis data to migrate"
+    ));
```
thibault-martinez, line 363: Maybe also do this assert before the copy so you check both ifs

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: Yes because it was always like that and in stardust it's not
Do you also want it to look for the account indexation key or something like that?

sdk/tests/wallet/chrysalis_migration.rs
Thoralf-M: The second isn't reached, since there is no /db folder

sdk/src/wallet/migration/chrysalis.rs
thibault-martinez: I was thinking something like that yes. But your new check of is_empty should cover that now?

sdk/tests/wallet/chrysalis_migration.rs
thibault-martinez: Right

thibault-martinez: Well maybe do another one that actually has a /db ? For completeness

sdk/src/wallet/migration/chrysalis.rs
Thoralf-M: It just checks if it's empty, if there is a db with random other data it would still overwrite

thibault-martinez: What do you mean random? get_chrysalis_data would only return stuff relevant to chrysalis, right ?

Thoralf-M: Updated, but no, get_chrysalis_data() returns just all keys, since we don't know all keys before we can't really check it (without a lot of complex logic to interpret the data)

sdk/tests/wallet/chrysalis_migration.rs
Thoralf-M: Updated it to add a db folder

sdk/tests/wallet/chrysalis_migration.rs
Thoralf-M: Done

thibault-martinez (approved):

DaughterOfMars (approved):

Issue #701, opened 2023-06-29
(non-)Slip-10 interface in `SecretManage`


DeppLearning:
## Description

The `SecretManage` trait assumes a Slip-10 compatible implementer, i.e. a hierarchical and deterministic wallet.

## Motivation

Hardware security modules (HSM) used by larger corporations to store keys usually do not support Slip-10 (or even Bip-32). Currently, in order to support HSMs, we are forced to work around the problem, i.e. by caching the `path` and other Slip-10 parameters. 

## Requirements /  Open questions (optional)

Would it make sense to provide a non-Slip-10 interface, that assumes just "normal" Ed25519 keys?

## Are you planning to do it yourself in a pull request?

Maybe

Aconitin:
This is blocking us quite heavily on using identity-rs properly. To make the issue a bit more clear:

Since everybody in our industry is using HSMs for key storage (as you should), but the `SecretManage` trait assumes "basically Stronghold" through Slip-10, you can't really build a secure application that passes the signing part off to a HSM. The result is a really dirty hack to "make it work somehow" by creating an intermediary layer that acts as if it implemented Slip-10 towards the iota-sdk.

It would help a lot of the `SecretManage` trait was unassuming and just dealt with normal Ed25519 keys. Would it be possible to bump this up a little? Otherwise, is there maybe some (more "canonical") preexisting implementation that could help us?

(For the identity people: Every identity will need to have its state controller / governor keypair in a Stronghold, even if you store document-level keys in an HSM, degrading the security of the whole identity to Stronghold level)

Thoralf-M:
Can't you just ignore the Slip-10 stuff as in https://github.com/iotaledger/iota-sdk/pull/937/files?

DeppLearning:
> Can't you just ignore the Slip-10 stuff as in https://github.com/iotaledger/iota-sdk/pull/937/files?

Thx a lot for the hint, while in fact I forgot the `SecretManage` is implemented for private keys, that doesn't really help I guess. The point of the HSM is that the raw keys never exit the HSM. We simply don't have access to the raw private key, we can only access it via a reference and sign arbitrary data with it using PKCS#11.

Thoralf-M:
My point was not about using this `PrivateKeySecretManager` for your case, but why can't you implement another secret manager in a similar way, where you just ignore the Slip-10 things?

DeppLearning:
[...]

PR #1219, opened 2023-09-12
Dynamic secret manager

DaughterOfMars:
# Description of change

This PR replaces the generic secret manager bounds with trait objects. This helps to clean up a lot of the code, but it also required pretty significant changes to some functionality.

In particular, the wallet builder `finish` method no longer loads data from the storage. Instead, there is a `load_storage` fn that will do this part optionally. The reason for this is that otherwise the secret manager type would need to be specified in `finish` even when no loading is needed or wanted.


sdk/src/client/secret/mod.rs
```diff
@@ -115,6 +116,160 @@ pub trait SecretManage: Send + Sync {
...
+pub trait DynSecretManage: std::fmt::Debug + Send + Sync {
+    async fn dyn_generate_ed25519_addresses(
+        &self,
+        coin_type: u32,
+        account_index: u32,
+        address_indexes: Range<u32>,
+        options: Option<GenerateAddressOptions>,
+    ) -> Result<Vec<Ed25519Address>, Error>;
+
+    async fn dyn_generate_evm_addresses(
```
Thoralf-M: Not sure if it would be more annoying, but should we maybe split the evm things into an own trait? Many people are probably not interested at that and would then still need to implement this with an error or something

sdk/src/client/secret/mod.rs
```diff
@@ -115,6 +116,160 @@ pub trait SecretManage: Send + Sync {
...
+    ) -> Result<Unlocks, Self::Error> {
+        self.dyn_sign_transaction_essence(prepared_transaction_data).await
+    }
+
+    async fn sign_transaction(
+        &self,
+        prepared_transaction_data: PreparedTransactionData,
+    ) -> Result<TransactionPayload, Self::Error> {
+        self.dyn_sign_transaction(prepared_transaction_data).await
+    }
```
Thoralf-M: Considering https://github.com/iotaledger/iota-sdk/issues/701, this also looks annoying to implement for other people
Probably worth to check if we can get rid of them and only rely on `sign_ed25519()` if that doesn't make it too complicated and maybe even with `chain: Option<Bip44>`

sdk/src/client/secret/mod.rs
DaughterOfMars: The main issue with that is that it means more bounds and in the end we'll probably have to require `T: Ed25519SecretManage + EvmSecretManage` most places anyway so they'd still have to.

sdk/src/client/secret/mod.rs
DaughterOfMars: They don't have to implement anything except for `SecretManage`

sdk/src/client/secret/mod.rs
Thoralf-M: Hm but where do we need `EvmSecretManage`? Not in the client and wallet, only on the secret managers (at least currently)

sdk/src/client/secret/mod.rs
Thoralf-M: Ok commented on the wrong section then, but there we have the same methods the need to implement, no?

sdk/src/client/secret/mod.rs
DaughterOfMars: Well, the problem is we can only have one dynamic secret manager, so it's either only an `dyn Ed25519SecretManage` or `dyn Ed25519SecretManage + EvmSecretManage`, both have their downsides. For secret manager methods, I doubt we even need to use the dynamic stuff really, we can just use a generic. If there's never a use for the evm fns outside of the secret manager methods, then splitting it might make sense.

sdk/src/client/secret/mod.rs
DaughterOfMars: Yeah, same as always

Thoralf-M: Yes we have them already, I'm just wondering if it now would make sense to get rid of something
Had a quick look and `sign_transaction()` for example seems to just forward to `default_sign_transaction()` which internally uses `sign_transaction_essence()`, so if I didn't miss something we can just remove that one?
And `signature_unlock()` could also be just a function that accepts a secret manager and internally calls `sign_ed25519()` on it

At least to me it looks like we would have then 2 functions less in the trait and can still do the same things

DaughterOfMars: I don't think we can remove either `sign_transaction` or `sign_transaction_essence` just because there are defaults for them. The defaults are internal anyway, not part of the trait. And I don't really understand the point about `signature_unlock` since users don't need to impl that fn anyway.

bindings/core/src/method_handler/wallet.rs
```diff
@@ -20,37 +21,44 @@ use iota_sdk::{
...
+        // #[cfg(feature = "stronghold")]
+        // WalletMethod::ClearStrongholdPassword => {
+        //     wallet.clear_stronghold_password().await?;
+        //     Response::Ok
+        // }
+        // #[cfg(feature = "stronghold")]
+        // WalletMethod::IsStrongholdPasswordAvailable => {
+        //     let is_available = wallet.is_stronghold_password_available().await?;
+        //     Response::Bool(is_available)
+        // }
```
Alex6323, line 61: Is that still a todo for this PR?

DaughterOfMars: yeah

PR #1238, opened 2023-09-15
feat(cli): `rustyline` instead of `dialoguer`

### LINKED ISSUE #260: CLI: handle CTRL-L to clear the terminal page

### LINKED ISSUE #437: Investigate alternative to `dialoguer`
## Description

I think we should look if there's something better out there. I feel like `RustyLine` could be worth a look.

## Motivation

There are few things that aren't so great about it:
* CTRL-C during a multiple-choice select kills the cursor
* weird behavior when pasting a mnemonic
* no general CTRL-L
* limited cursor movement on text inputs

## Open questions (optional)

What disadvantage do the alternative have over `dialoguer`?

Sure.

Alex6323:
# Description of change

Replace `dialoguer` with `rustyline` for the interactive account management in the CLI wallet. This makes the wallet overall more convenient to use.

With this PR the following is now supported:
* Improved TAB completion for commands (prints a list of matching commands on double TAB)
* CTRL-L to clear the terminal (no need to type `c` or `clear` anymore)
* HOME/END/... navigation
* Auto-complete for commands from the history
* Proper handling of CTRL-C (or any other interrupt)

Closes #437 
Closes #260 

## Type of change

Running the CLI wallet.


cli/Cargo.toml
```diff
@@ -34,15 +34,16 @@ clap = { version = "4.4.2", default-features = false, features = [
 ] }
 colored = { version = "2.0.4", default-features = false }
 dialoguer = { version = "0.10.4", default-features = false, features = [
-    "history",
-    "password",
-    "completion",
+   #"history",
+   "password",
+   #"completion",
```
Thoralf-M: Remove if not needed

Alex6323: Done

cli/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `addresses` now additionally prints the hex version of the address;
 - `outputs`, `unspent-outputs` print a list that includes number and type of the output;
 - `Account::switch` command to allow changing accounts quickly;
+- UX improvements (Ctrl+l, TAB completion/suggestions and more) during interactive account management;
```
thibault-martinez, line 30: We already had TAB completion?

cli/Cargo.toml
```diff
@@ -33,16 +33,13 @@ clap = { version = "4.4.2", default-features = false, features = [
     "env",
 ] }
 colored = { version = "2.0.4", default-features = false }
-dialoguer = { version = "0.10.4", default-features = false, features = [
-    "history",
-    "password",
-    "completion",
-] }
+dialoguer = { version = "0.10.4", default-features = false, features = [ "password" ] }
```
thibault-martinez: Doesn't look formatted

cli/src/account_completion.rs
```diff
@@ -47,17 +47,15 @@ pub(crate) const ACCOUNT_COMPLETION: &[&str] = &[
...
-            .iter()
-            .filter(|option| option.starts_with(input))
-            .collect::<Vec<_>>();
-
-        if matches.len() == 1 {
-            Some(matches[0].to_string())
-        } else {
-            None
+impl Completer for AccountCompleter {
+    type Candidate = String;
```
thibault-martinez:
```suggestion
    type Candidate = String;

```

cli/src/account_history.rs
thibault-martinez, line 1: History is all handled by rustyline?

thibault-martinez (changes requested):

cli/src/account_history.rs
Alex6323: yeah

cli/CHANGELOG.md
Alex6323: yeah, but not exactly like this. Now ... if you TAB and there are multiple matches it will show a list of candidates. If you have no idea whatsoever and TAB on an empty prompt it will show all available commands. Furthermore it will autocomplete commands from the history.

cli/Cargo.toml
Alex6323: hmm, what would you expect?

thibault-martinez: No spaces around the feature, like for rustyline. Cargo fmt is supposed to fix that.

Thoralf-M (dismissed):

cli/Cargo.toml
DaughterOfMars: I don't think my cargo fmt does that

cli/src/account.rs
```diff
@@ -24,11 +28,50 @@ use crate::{
...
+pub struct PromptHelper {
+    #[rustyline(Completer)]
+    completer: AccountCompleter,
+    #[rustyline(Hinter)]
+    hinter: HistoryHinter,
+    colored_prompt: String,
+}
+
+impl Highlighter for PromptHelper {
+    fn highlight_prompt<'b, 's: 'b, 'p: 'b>(&'s self, prompt: &'p str, default: bool) -> Cow<'b, str> {
```
DaughterOfMars: wow the author of this lib went a little overkill with the manual lifetimes

cli/src/account.rs
```diff
@@ -24,11 +28,50 @@ use crate::{
...
+        .completion_type(rustyline::CompletionType::List)
+        .edit_mode(rustyline::EditMode::Emacs)
+        .build();
+
+    // Panic: should never happen
+    let mut rl = Editor::with_history(config, MemHistory::with_config(config)).unwrap();
+
+    let helper = PromptHelper {
+        completer: AccountCompleter,
+        hinter: HistoryHinter {},
```
DaughterOfMars: No new fn?

DaughterOfMars: or default?

DaughterOfMars: Actually couldn't you impl/derive `Default` on PromptHelper too?

cli/src/account.rs
```diff
@@ -55,154 +98,173 @@ pub enum AccountPromptResponse {
...
-                AccountCommand::Faucet { address, url } => faucet_command(account, address, url).await,
-                AccountCommand::MeltNativeToken { token_id, amount } => {
-                    melt_native_token_command(account, token_id, amount).await
-                }
-                AccountCommand::MintNativeToken { token_id, amount } => {
-                    mint_native_token(account, token_id, amount).await
+
+    let prompt = format!("Account \"{}\": ", alias);
+    // Panic: should never happen since there's always a helper
+    rl.helper_mut().unwrap().colored_prompt = prompt.green().to_string();
```
DaughterOfMars: Maybe just do an `if let Some()` here since this isn't important

cli/src/account.rs
```diff
@@ -55,154 +98,173 @@ pub enum AccountPromptResponse {
 pub async fn account_prompt_internal(
     wallet: &Wallet,
     account: &Account,
-    history: &mut AccountHistory,
+    rl: &mut Editor<PromptHelper, MemHistory>,
 ) -> Result<AccountPromptResponse, Error> {
     let alias = {
         let account = account.details().await;
```
DaughterOfMars: Make this one line?

cli/src/account.rs
```diff
@@ -55,154 +98,173 @@ pub enum AccountPromptResponse {
...
-                    send_command(
-                        account,
-                        address,
-                        amount,
-                        return_address,
-                        expiration.map(|e| e.as_secs() as u32),
-                        allow_micro_amount,
-                    )
-                    .await
+                    if let Err(err) = match account_cli.command {
```
DaughterOfMars: I think we should move the error check to below and replace this with `let res = `

cli/src/account.rs
```diff
@@ -55,154 +98,173 @@ pub enum AccountPromptResponse {
...
-                AccountCommand::DecreaseVotingPower { amount } => decrease_voting_power_command(account, amount).await,
-                AccountCommand::VotingOutput => voting_output_command(account).await,
-            } {
-                println_log_error!("{err}");
             }
         }
+        Err(ReadlineError::Interrupted) => {
+            return Ok(AccountPromptResponse::Done);
+        }
+        Err(_) => {}
```
DaughterOfMars: No message here?

cli/src/account_completion.rs
```diff
@@ -47,17 +47,15 @@ pub(crate) const ACCOUNT_COMPLETION: &[&str] = &[
...
+    fn complete(&self, input: &str, _pos: usize, _ctx: &Context<'_>) -> rustyline::Result<(usize, Vec<String>)> {
+        let mut completions = vec![];
+        for command in ACCOUNT_COMMANDS {
```
DaughterOfMars: filter_map?

cli/src/account.rs
Alex6323: probably :+1:

cli/src/account_completion.rs
Alex6323: Done

cli/Cargo.toml
Alex6323: Fixed

cli/src/account_completion.rs
Alex6323: Done

cli/src/account.rs
Alex6323: impl'ed Default for `PromptHelper`, couldn't derive it bc `HistoryHinter` doesn't impl it.

DaughterOfMars: Errrrr what the hell rustyline, why would you define a struct like this 😆 
```rust
pub struct HistoryHinter {}
```

cli/src/account.rs
Alex6323: Done

cli/src/account.rs
Alex6323: Done

cli/src/account.rs
Alex6323: now printing and logging the error

cli/src/account_completion.rs
```diff
@@ -3,7 +3,7 @@
 
 use rustyline::{completion::Completer, Context};
 
-pub struct AccountCompleter;
+pub struct AccountCompleter {}
```
DaughterOfMars: God no, just derive Default for this

cli/src/account.rs
```diff
@@ -60,15 +70,9 @@ pub async fn account_prompt(wallet: &Wallet, mut account: Account) -> Result<(),
         .edit_mode(rustyline::EditMode::Emacs)
         .build();
 
-    // Panic: should never happen
+    // Panic: should never happen since it's all in-memory
     let mut rl = Editor::with_history(config, MemHistory::with_config(config)).unwrap();
```
DaughterOfMars: This failure can be from an error accessing the terminal, so the comment isn't quite accurate. Maybe we should create an error variant for this instead

cli/src/account_completion.rs
Alex6323: damn, this was only temporary and I forgot to undo it. I wanted to know what's the point of defining it with { } :joy: (obviously there's none)

cli/src/account.rs
Alex6323: oh damn, you're right ...

Alex6323: should we just map it to an IO error?

DaughterOfMars: If that's possible, then sure

cli/src/account.rs
Alex6323: I changed it, but didn't add the extra `res` binding ... just did an `unwrap_or_else` after the match block. Fine with you?

cli/src/account.rs
Alex6323: changed my mind and added an error variant as you proposed

cli/src/account.rs
```diff
@@ -1,34 +1,82 @@
...
+            voting_power_command, AccountCli, AccountCommand,
+        },
+        account_completion::AccountCompleter,
     },
     error::Error,
     helper::bytes_from_hex_or_file,
     println_log_error,
 };
 
+#[derive(Helper, Completer, Hinter, Validator)]
```
DaughterOfMars: This should probably move to the `account_completion` module

cli/src/account.rs
DaughterOfMars: sure

cli/src/account.rs
Alex6323: Done

DaughterOfMars (dismissed):

Thoralf-M (dismissed):

cli/src/command/account_completion.rs
```diff
@@ -0,0 +1,108 @@
...
+    "participation-overview",
+    "voting-power",
+    "increase-voting-power",
+    "decrease-voting-power",
+    "voting-output",
+    "help",
+];
+
+impl Completer for AccountCompleter {
+    type Candidate = String;
```
thibault-martinez: Nit but it doesn't look like this needs to be a String

thibault-martinez: I mean, it can just be static &str

Alex6323: oh yeah, good catch

Alex6323: Done

DaughterOfMars (approved):

Thoralf-M (approved):

thibault-martinez (approved):

PR #1240, opened 2023-09-15
Use packable BTreeSet impls

DaughterOfMars:
# Description of change

This PR uses BTreeSets rather than boxed slices where possible when unique, ordered data is required.

# Related Issues

Blocked by https://github.com/iotaledger/common-rs/pull/70


thibault-martinez (changes requested):
Conflicts

sdk/src/types/block/error.rs
```diff
@@ -364,9 +365,10 @@ impl fmt::Display for Error {
             Self::UnallowedFeature { index, kind } => {
                 write!(f, "unallowed feature at index {index} with kind {kind}")
             }
-            Self::UnallowedUnlockCondition { index, kind } => {
+            Self::DisallowedUnlockCondition { index, kind } => {
                 write!(f, "unallowed unlock condition at index {index} with kind {kind}")
```
Thoralf-M: Should also be renamed here then?
```suggestion
                write!(f, "disallowed unlock condition at index {index} with kind {kind}")
```

Thoralf-M (dismissed):

thibault-martinez:
Conflicts

Thoralf-M (dismissed):

thibault-martinez:
Solve conflicts and I'll review/merge today

Thoralf-M (dismissed):

sdk/src/types/block/rand/parents.rs
```diff
@@ -7,6 +7,9 @@ use crate::types::block::{
 };
 
 /// Generates random strong parents.
-pub fn rand_strong_parents<const MAX: u8>() -> Parents<1, MAX> {
-    Parents::from_vec(rand_block_ids(rand_number_range(Parents::<1, MAX>::COUNT_RANGE).into())).unwrap()
+pub fn rand_parents<const MIN: u8, const MAX: u8>() -> Parents<MIN, MAX> {
```
thibault-martinez, line 10: Nice, I almost commented in the previous PR that I didn't like it and ended up not doing it

DaughterOfMars: this comment made me notice the comment above is outdated, thanks lol

sdk/src/types/block/output/unlock_condition/mod.rs
```diff
@@ -337,83 +363,86 @@ impl UnlockConditions {
 
     /// Creates a new [`UnlockConditions`] from a vec.
     pub fn from_vec(unlock_conditions: Vec<UnlockCondition>) -> Result<Self, Error> {
```
thibault-martinez, line 365: So this wouldn't error anymore on duplicates ?

DaughterOfMars: yeah currently it won't

thibault-martinez: Can we somehow bring it back?

DaughterOfMars: done

Thoralf-M (dismissed):

Thoralf-M (dismissed):

Thoralf-M (dismissed):

Thoralf-M (approved):

sdk/src/types/block/error.rs
```diff
@@ -371,9 +372,10 @@ impl fmt::Display for Error {
             Self::UnallowedFeature { index, kind } => {
                 write!(f, "unallowed feature at index {index} with kind {kind}")
             }
-            Self::UnallowedUnlockCondition { index, kind } => {
-                write!(f, "unallowed unlock condition at index {index} with kind {kind}")
+            Self::DisallowedUnlockCondition { index, kind } => {
+                write!(f, "disallowed unlock condition at index {index} with kind {kind}")
             }
+            Self::TooManyUnlockConditions => write!(f, "too many unlock conditions"),
```
kwek20, line 378: maybe we want max/current added?

sdk/src/types/block/output/delegation.rs
```diff
@@ -359,7 +359,7 @@ impl DelegationOutput {
 
     // Transition, just without full ValidationContext.
     pub(crate) fn transition_inner(current_state: &Self, next_state: &Self) -> Result<(), StateTransitionError> {
-        if !(current_state.delegation_id.is_null() && !next_state.delegation_id().is_null()) {
+        if !current_state.delegation_id.is_null() | next_state.delegation_id().is_null() {
```
kwek20, line 362: nice bit stuff :D lets add a `|`

sdk/src/types/block/output/feature/mod.rs
```diff
@@ -67,6 +71,17 @@ impl Ord for Feature {
         self.kind().cmp(&other.kind())
     }
 }
+impl core::hash::Hash for Feature {
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        self.kind().hash(state);
```
kwek20, line 76: Are we sure we want the feature hash to just be the kind? that would make 2 different SenderFeatures have the same hash

kwek20: just noticed that the Eq is also just kind. for our code it works, but what if someone gets outputs and compares it with their feature, youd just expect == to work.

thibault-martinez: We started speaking about it on Slack, we should definitely continue the discussion. I really dislike this.

kwek20 (commented):
well im not sure about all these hash/eq  only taking the kind into consideration for external people. may be confusing. idk if we actually care about that tho.

DaughterOfMars:
I'm going to close this, as the concerns are valid and additionally I believe there is a better way to do this.

PR #1247, opened 2023-09-18
Merge 1.1 into 2.0

thibault-martinez:


bindings/core/src/method_handler/client.rs
```diff
@@ -168,7 +168,7 @@ pub(crate) async fn call_client_method_internal(client: &Client, method: ClientM
         }
         ClientMethod::GetNode => Response::Node(client.get_node().await?),
         ClientMethod::GetNetworkInfo => Response::NetworkInfo(client.get_network_info().await?),
-        ClientMethod::GetNetworkId => Response::NetworkId(client.get_network_id().await?),
+        ClientMethod::GetNetworkId => Response::NetworkId(client.get_network_id().await?.to_string()),
```
DaughterOfMars, line 171: Is this needed?

Thoralf-M (approved):

bindings/core/src/method_handler/client.rs
thibault-martinez: Yes

bindings/core/src/response.rs
```diff
@@ -68,7 +68,7 @@ pub enum Response {
     NetworkInfo(NetworkInfo),
     /// Response for:
     /// - [`GetNetworkId`](crate::method::ClientMethod::GetNetworkId)
-    NetworkId(u64),
+    NetworkId(String),
```
DaughterOfMars, line 71: pour quoi non `#[serde(with = "string")]`

thibault-martinez: @Thoralf-M Would that also work? But anyway, this would need to be changed in develop, not here.

Thoralf-M: Yeah don't see why not

bindings/python/examples/wallet/offline_signing/1_prepare_transaction.py
```diff
@@ -27,7 +27,8 @@
 
 # Recovers addresses from example `0_address_generation`.
 addresses_data = json.load(open(ADDRESSES_FILE_PATH, "r", encoding="utf-8"))
-addresses = list(map(lambda x: from_dict(data_class=AccountAddress, data=x), addresses_data))
+addresses = list(map(lambda x: from_dict(
```
DaughterOfMars, line 30: :c

DaughterOfMars (approved):

bindings/core/src/response.rs
DaughterOfMars: 👍

kwek20 (approved):

PR #1253, opened 2023-09-18
Node: Cyclic dep. in transaction payload

### LINKED ISSUE #1249: Node: Nested payload in TransactionDataPayload is not properly typed
### Issue description

parsing nested TransactionDataPayload causes class-transformer to stop due to cyclic parsing.

resulting in 
```
Block data: Block {
  protocolVersion: 2,
  parents: [
    '0x3689742bb257ffec5b9789022bb0af90c23574b605d208b84d2abd083d54085d'
  ],
  payload: {
    essence: {
      type: 1,
      networkId: '1856588631910923207',
      inputs: [Array],
      inputsCommitment: '0xff6bd7f6aeb9c6926da84267d859cf08ce44c228c13fb67b8893d642030a2f51',
      outputs: [Array],
      payload: [Object]
    },
    unlocks: [ [Object] ]
  },
  nonce: '233605'
} 
```

instead of 
```
Block data: Block {
payload: TransactionPayload {
    type: 6,
    essence: RegularTransactionEssence {
      type: 1,
      networkId: '1856588631910923207',
      inputsCommitment: '0xff6bd7f6aeb9c6926da84267d859cf08ce44c228c13fb67b8893d642030a2f51',
      inputs: [Array],
      outputs: [Array],
      payload: [Object]
    },
    unlocks: [ [SignatureUnlock] ]
  },
  nonce: '233605'
} 
```

possible fix: Custom @transform with a new `plainToInstance` call like here:
https://github.com/typestack/class-transformer/issues/212#issuecomment-451066261

### Version

1.1

### Expected behaviour

nested payload has a type

### Actual behaviour

no type

Yes

1. Change client 07 example to look up a transaction with payload type 6
2. yarn run-example ./client/07-get-block-data.ts
3. see that block data isnt typed

### Errors

_No response_

### Duplicate declaration

kwek20:
# Description of change

Transaction payload wasnt parsed due to cyclic dependency in class-transformer.
Setting it to Always being TaggedDataPayload solves this as the discriminator is no longer needed.

Fixes https://github.com/iotaledger/iota-sdk/issues/1249

### DIFF
--- a/bindings/nodejs/lib/types/block/payload/transaction/essence.ts
+++ b/bindings/nodejs/lib/types/block/payload/transaction/essence.ts
@@ -2,11 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { Type } from 'class-transformer';
-import { PayloadDiscriminator } from '..';
 import { HexEncodedString } from '../../../utils';
 import { Input, InputDiscriminator } from '../../input';
 import { Output, OutputDiscriminator } from '../../output';
-import { Payload } from '../payload';
+import { TaggedDataPayload } from '../tagged';
 
 /**
  * All of the essence types.
@@ -57,10 +56,8 @@ class RegularTransactionEssence extends TransactionEssence {
     })
     outputs: Output[];
 
-    @Type(() => Payload, {
-        discriminator: PayloadDiscriminator,
-    })
-    payload: Payload | undefined;
+    @Type(() => TaggedDataPayload)
+    payload?: TaggedDataPayload;
 
     /**
      * @param networkId The ID of the network the transaction was issued to.
@@ -75,7 +72,7 @@ class RegularTransactionEssence extends TransactionEssence {
         inputsCommitment: HexEncodedString,
         inputs: Input[],
         outputs: Output[],
-        payload: Payload | undefined,
+        payload: TaggedDataPayload | undefined,
     ) {
         super(TransactionEssenceType.Regular);
         this.networkId = networkId;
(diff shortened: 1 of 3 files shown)


Thoralf-M:
Changelog entry?

kwek20:
Erm well technically it always was a TaggedDataPayload, and it never worked before :P but alright

DaughterOfMars (approved):

Thoralf-M (approved):

bindings/nodejs/lib/types/block/payload/transaction/essence.ts
```diff
@@ -75,7 +72,7 @@ class RegularTransactionEssence extends TransactionEssence {
         inputsCommitment: HexEncodedString,
         inputs: Input[],
         outputs: Output[],
-        payload: Payload | undefined,
+        payload: TaggedDataPayload | undefined,
```
thibault-martinez, line 75: Will this serialize the type?

kwek20: Yes, I tested it

msarcev:
Maybe this can be a temporary fix while we find a better one. At the moment it would be nice to have a fix to unblock the integration in Explorer.

thibault-martinez:
> Maybe this can be a temporary fix while we find a better one. At the moment it would be nice to have a fix to unblock the integration in Explorer.

Abdul is having another look now and we'll merge it if it's not conclusive.

thibault-martinez (changes requested):
.

PR #1254, opened 2023-09-18
Add feature sets CI

### LINKED ISSUE #919: Add CI step for most common combinations of features
Find out about the most common combinations of features used by other projects (e.g. identiity) and add a CI step to check it compiles.

https://github.com/iotaledger/iota-sdk/issues/897

DaughterOfMars:
# Description of change

Adds a new CI workflow to test common feature sets. I had to make some code changes for some of the combinations.

Closes #919

## Type of change


sdk/src/wallet/core/operations/stronghold_backup/mod.rs
```diff
@@ -103,7 +105,10 @@ impl Wallet {
             .password(stronghold_password.clone())
             .build(backup_path.clone())?;
 
-        let (read_client_options, read_coin_type, read_secret_manager, read_accounts, chrysalis_data) =
+        #[cfg(feature = "storage")]
```
Thoralf-M: Why does this need the storage feature? Stronghold alone should be sufficient, no?

DaughterOfMars: No, see the feature gate below

Thoralf-M: You added the feature there, but the function also doesn't need it?

DaughterOfMars: both have it?

Thoralf-M: Why does `migrate_snapshot_from_chrysalis_to_stardust()` need it?

DaughterOfMars: Because otherwise it's unused

Thoralf-M: But why not have the stronghold feature instead of storage?

DaughterOfMars: because we use storage below this...

Thoralf-M: You mean for `.set(CHRYSALIS_STORAGE_KEY, &chrysalis_data)`? `migrate_snapshot_from_chrysalis_to_stardust()` does more than only returning this data

DaughterOfMars: I see, so we should call it regardless of whether we will use the result?

Thoralf-M: I would say so

DaughterOfMars: Hopefully it's fixed now

Thoralf-M (dismissed):

.github/workflows/bindings-nodejs.yml
```diff
@@ -2,7 +2,7 @@ name: Nodejs bindings checks
 
 on:
   push:
-    branches: [develop, production, 1.1]
+    branches: [develop, production, "1.1"]
```
thibault-martinez, line 5: For my curiosity, why? This has never been an issue ?

sdk/src/lib.rs
```diff
@@ -30,8 +30,6 @@ pub mod utils;
 pub use crypto;
 pub use packable;
 pub use primitive_types::U256;
-#[cfg(feature = "url")]
-pub use url::Url;
```
thibault-martinez: Why can it not be that way? And why specifically Url?
I think it was done this way because it's part of our public API, people should use it from us, otherwise if we update it on our side it would be breaking change for them. Or something like this.

.github/workflows/bindings-nodejs.yml
DaughterOfMars: It complains in my workspace 🤷‍♀️

Thoralf-M (dismissed):

sdk/src/lib.rs
DaughterOfMars: If there is a breaking change in Url while we are re-exporting it, we must release a breaking update as well. De-coupling doesn't change this since we expose the types anyway in our public API.

thibault-martinez: What motivated the change in the context of this PR?

DaughterOfMars: Just reducing feature complexity.

.github/workflows/bindings-nodejs.yml
thibault-martinez: Weird but alright

sdk/examples/wallet/participation.rs
```diff
@@ -19,8 +19,9 @@
 use iota_sdk::{
     client::node_manager::node::Node,
     wallet::{account::types::participation::ParticipationEventRegistrationOptions, Result},
-    Url, Wallet,
+    Wallet,
 };
+use url::Url;
```
thibault-martinez: For now I'd revert

sdk/src/client/node_manager/node.rs
```diff
@@ -4,8 +4,7 @@
 use std::hash::Hash;
 
 use serde::{Deserialize, Serialize};
-
-use crate::Url;
+use url::Url;
```
thibault-martinez: same

sdk/src/wallet/migration/chrysalis.rs
```diff
@@ -310,6 +191,7 @@ fn decrypt_record(record_bytes: Vec<u8>, encryption_key: &[u8; 32]) -> crate::wa
     String::from_utf8(pt).map_err(|e| Error::Migration(format!("{:?}", e)))
 }
 
+#[allow(unused)]
```
thibault-martinez, line 194: Is it too convoluted to figure out under what combination(s) they are used ?

sdk/src/wallet/migration/migrate_4.rs
```diff
@@ -69,6 +69,7 @@ impl Migration<crate::client::stronghold::StrongholdAdapter> for Migrate {
     }
 }
 
+#[cfg(feature = "storage")]
 fn migrate_wallet(wallet: &mut serde_json::Value) -> Result<()> {
```
thibault-martinez, line 73: Could just move it where it's called since it's only used once but not a big deal

sdk/examples/wallet/participation.rs
DaughterOfMars: oh, thought I did

sdk/src/wallet/migration/chrysalis.rs
DaughterOfMars: No, the problem is the combinations are not useful. it would be rocksdb or (storage and stronghold). just makes more sense to do this

thibault-martinez: Fair enough

sdk/src/wallet/core/operations/stronghold_backup/stronghold_snapshot.rs
```diff
@@ -6,20 +6,19 @@ use std::{collections::HashMap, path::Path, sync::atomic::Ordering};
...
-            chrysalis::{migrate_from_chrysalis_data, to_chrysalis_key},
+            chrysalis::{migrate_from_chrysalis_data, to_chrysalis_key, CHRYSALIS_STORAGE_KEY},
             latest_backup_migration_version, migrate, MigrationData, MIGRATION_VERSION_KEY,
         },
-        storage::constants::{CHRYSALIS_STORAGE_KEY, WALLET_INDEXATION_KEY},
         ClientOptions, Error as WalletError, Wallet,
     },
 };
 
+pub(crate) const WALLET_INDEXATION_KEY: &str = "iota-wallet-account-manager";
```
thibault-martinez, line 22: That's a bit annoying, maybe we can just move the constants around idk

thibault-martinez (approved):

abdulmth (approved):

PR #1260, opened 2023-09-19
Python - Add `Irc27Metadata` and `Irc30Metadata`

### LINKED ISSUE #1170: Python - Add `Irc27Metadata` and `Irc30Metadata`
## Description

Add the equivalent of rust's `Irc27Metadata` and `Irc30Metadata` to python.

See https://github.com/iotaledger/tips/blob/main/tips/TIP-0027/tip-0027.md and https://github.com/iotaledger/tips/blob/main/tips/TIP-0030/tip-0030.md

Also see #1160

DaughterOfMars:
# Description of change

Adds the metadata helpers to python.

Closes #1170 

## Type of change


bindings/python/iota_sdk/types/irc_27.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2023 IOTA Stiftung
+# SPDX-License-Identifier: Apache-2.0
+
+from iota_sdk import utf8_to_hex
+from dataclasses import dataclass, field
+from typing import Optional, List
+
+@dataclass
```
Thoralf-M, line 25: I think here should also be the @json so it converts the fields to camelCase

bindings/python/iota_sdk/types/irc_30.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2023 IOTA Stiftung
+# SPDX-License-Identifier: Apache-2.0
+
+from iota_sdk.types.common import HexStr
+from iota_sdk import utf8_to_hex
+from dataclasses import dataclass, field
+from typing import Optional, List
+
+@dataclass
```
Thoralf-M, line 12: Same here also with @json above

Thoralf-M (changes requested):
Also mention the added classes in the changelog

Thoralf-M (dismissed):

Thoralf-M (dismissed):

Thoralf-M (dismissed):

Thoralf-M (approved):

Thoralf-M (changes requested):
ValueError: mutable default <class 'dict'> for field royalties is not allowed: use default_factory

Thoralf-M (dismissed):

bindings/python/examples/client/10_mint_nft.py
```diff
@@ -28,7 +35,7 @@
     nft_id='0x0000000000000000000000000000000000000000000000000000000000000000',
     amount=1000000,
     immutable_features=[
-        MetadataFeature(utf8_to_hex('Hello, World!'))
+        MetadataFeature(metadata.as_hex())
```
thibault-martinez: Would it be handy to also have an `as_feature()` method ?

DaughterOfMars: could be yeah

thibault-martinez: Also if ever we follow rust conventions in python, it would rather be `to` than `as` but whatever

bindings/python/iota_sdk/types/irc_27.py
thibault-martinez: @Thoralf-M I'd be interested to know why you resolved, I have the same question

DaughterOfMars: because the @json descriptor is defined at 2.0

thibault-martinez: Stop pinging that random person account :trollface:

thibault-martinez: But yeah right, forgot about that

bindings/python/examples/client/10_mint_nft.py
thibault-martinez: @DaughterOfMars Making sure you've seen the last comment, just resolve yourself if you want to ignore it

thibault-martinez (approved):

Thoralf-M (approved):

PR #1272, opened 2023-09-20
Don't return error with custom remainder address and ledger nano

### LINKED ISSUE #1099: Try to provide remainder Bip44 to ledger nano
## Description

https://github.com/iotaledger/iota-sdk/blob/4bf72ecb13c3540895d03cc62eb527241e41675a/sdk/src/client/api/block_builder/input_selection/core/remainder.rs#L26 currently just returns `None` for the `Bip44` value, but then we can't regenerate the address here https://github.com/iotaledger/iota-sdk/blob/4bf72ecb13c3540895d03cc62eb527241e41675a/sdk/src/client/secret/ledger_nano.rs#L297-L308

So we should search in the inputs (or maybe to the input selection provided addresses, if we can have the bip44 there?) for this address and return `Bip44` then if the address is known, instead of always `None`

## Motivation

So ledger nano users can verify the remainder address if no blind signing is needed

No.

Thoralf-M:
# Description of change

Don't return error with custom remainder address and ledger nano

Fixes #1099 

## Type of change

Running the simple transaction example modified with ledger nano
```
        const response = await account.send(amount, address, {
            remainderValueStrategy: {
                strategy: 'CustomAddress',
                value: {
                    address: 'rms1qqdnv60ryxynaeyu8paq3lp9rkll7d7d92vpumz88fdj4l0pn5mruskth6z', 
                    keyIndex: 0, 
                    internal: false, 
                    used: false
                }
            }
        });
```

### DIFF
--- a/bindings/nodejs/lib/types/wallet/transaction-options.ts
+++ b/bindings/nodejs/lib/types/wallet/transaction-options.ts
@@ -26,7 +26,7 @@ export interface TransactionOptions {
     /** Optional note, that is only stored locally. */
     note?: string;
     /** Whether to allow sending a micro amount. */
-    allowMicroAmount: boolean;
+    allowMicroAmount?: boolean;
 }
 
 /** The possible remainder value strategies. */
--- a/sdk/src/client/api/block_builder/input_selection/core/remainder.rs
+++ b/sdk/src/client/api/block_builder/input_selection/core/remainder.rs
@@ -24,7 +24,6 @@ impl InputSelection {
     fn get_remainder_address(&self) -> Option<(Address, Option<Bip44>)> {
         if let Some(remainder_address) = self.remainder_address {
             // Search in inputs for the Bip44 chain for the remainder address, so the ledger can regenerate it
-            #[cfg(feature = "ledger_nano")]
             for input in self.available_inputs.iter().chain(self.selected_inputs.iter()) {
                 let alias_transition = is_alias_transition(
                     &input.output,
--- a/sdk/src/client/api/block_builder/input_selection/core/remainder.rs
+++ b/sdk/src/client/api/block_builder/input_selection/core/remainder.rs
@@ -32,13 +32,12 @@ impl InputSelection {
                     self.outputs.as_slice(),
                     self.burn.as_ref(),
                 );
-                // PANIC: safe to unwrap as treasury outputs can't be used as input.
-                let required_address = input
-                    .output
-                    .required_and_unlocked_address(self.timestamp, input.output_id(), alias_transition)
-                    .unwrap()
-                    .0;
-                if remainder_address == required_address {
+                let required_address =
+                    input
+                        .output
+                        .required_and_unlocked_address(self.timestamp, input.output_id(), alias_transition);
+
+                if matches!(required_address, Ok(a) if a.0 == remainder_address) {
                     return Some((remainder_address, input.chain));
                 }
             }
--- a/sdk/src/client/api/block_builder/input_selection/core/remainder.rs
+++ b/sdk/src/client/api/block_builder/input_selection/core/remainder.rs
@@ -22,8 +22,27 @@ use crate::{
 impl InputSelection {
     // Gets the remainder address from configuration of finds one from the inputs.
     fn get_remainder_address(&self) -> Option<(Address, Option<Bip44>)> {
-        if self.remainder_address.is_some() {
-            return self.remainder_address.map(|address| (address, None));
+        if let Some(remainder_address) = self.remainder_address {
+            // Search in inputs for the Bip44 chain for the remainder address, so the ledger can regenerate it
+            #[cfg(feature = "ledger_nano")]
+            for input in self.available_inputs.iter().chain(self.selected_inputs.iter()) {
+                let alias_transition = is_alias_transition(
+                    &input.output,
+                    *input.output_id(),
+                    self.outputs.as_slice(),
+                    self.burn.as_ref(),
+                );
+                // PANIC: safe to unwrap as treasury outputs can't be used as input.
+                let required_address = input
+                    .output
+                    .required_and_unlocked_address(self.timestamp, input.output_id(), alias_transition)
+                    .unwrap()
+                    .0;
+                if remainder_address == required_address {
+                    return Some((remainder_address, input.chain));
+                }
+            }
+            return Some((remainder_address, None));
         }
 
         for input in &self.selected_inputs {
--- a/sdk/src/client/api/block_builder/input_selection/core/remainder.rs
+++ b/sdk/src/client/api/block_builder/input_selection/core/remainder.rs
@@ -21,7 +21,7 @@ use crate::{
 
 impl InputSelection {
     // Gets the remainder address from configuration of finds one from the inputs.
-    fn get_remainder_address(&self) -> Option<(Address, Option<Bip44>)> {
+    fn get_remainder_address(&self) -> Result<Option<(Address, Option<Bip44>)>, Error> {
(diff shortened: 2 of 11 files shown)


bindings/nodejs/lib/types/wallet/transaction-options.ts
```diff
@@ -26,7 +26,7 @@ export interface TransactionOptions {
     /** Optional note, that is only stored locally. */
     note?: string;
     /** Whether to allow sending a micro amount. */
-    allowMicroAmount: boolean;
+    allowMicroAmount?: boolean;
```
thibault-martinez, line 29: This is not breaking change, right ?

Thoralf-M: I think not, it would be only the other way around

sdk/src/client/api/block_builder/input_selection/core/remainder.rs
```diff
@@ -22,8 +22,27 @@ use crate::{
...
+            #[cfg(feature = "ledger_nano")]
+            for input in self.available_inputs.iter().chain(self.selected_inputs.iter()) {
+                let alias_transition = is_alias_transition(
+                    &input.output,
+                    *input.output_id(),
+                    self.outputs.as_slice(),
+                    self.burn.as_ref(),
+                );
+                // PANIC: safe to unwrap as treasury outputs can't be used as input.
+                let required_address = input
```
DaughterOfMars: Even if it's safe, why? Can't we just check
```rust
if matches!(remainder_address, Some(a) if a == required_address)) {
```

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -296,19 +296,23 @@ impl SecretManage for LedgerSecretManager {
...
             let (remainder_address, remainder_bip32): (Option<&Address>, LedgerBIP32Index) =
-                match &prepared_transaction.remainder {
-                    Some(a) => {
-                        let chain = a.chain.ok_or(Error::MissingBip32Chain)?;
-                        (
-                            Some(&a.address),
-                            LedgerBIP32Index {
-                                bip32_change: chain.change.harden().into(),
-                                bip32_index: chain.address_index.harden().into(),
+                prepared_transaction.remainder.as_ref().map_or_else(
```
DaughterOfMars: this is less easy to understand :c

Thoralf-M: Clippy complained otherwise, do you want me to put an `#[allow(`?

DaughterOfMars: I mean, I guess so. It's really horrible rn.

sdk/src/client/api/block_builder/input_selection/core/remainder.rs
Thoralf-M: Bit different but yes, changed it

sdk/src/client/secret/ledger_nano.rs
Thoralf-M: Done

DaughterOfMars (dismissed):

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -295,17 +295,21 @@ impl SecretManage for LedgerSecretManager {
...
+                        if let Some(chain) = a.chain {
+                            (
+                                Some(&a.address),
+                                LedgerBIP32Index {
+                                    bip32_change: chain.change.harden().into(),
+                                    bip32_index: chain.address_index.harden().into(),
+                                },
+                            )
+                        } else {
+                            (None, LedgerBIP32Index::default())
```
thibault-martinez, line 311: What does it mean exactly to have a remainder address but not the associated chain ?

Thoralf-M: Custom remainder address provided, but the address is not in the inputs (because from there we would get the chain)

thibault-martinez: So even if provided, we won't be able to use it ?

Thoralf-M: One can provide the address index and change in the wallet, but in the ISA we only accept an `Address`, so the information can't be directly provided without breaking change
But that's why I also changed the ledger nano part so it just doesn't count as remainder if the Bip44 chain is missing, will be just another output in the case we have no input with the corresponding address

sdk/src/client/api/block_builder/input_selection/core/remainder.rs
```diff
@@ -22,8 +22,26 @@ use crate::{
 impl InputSelection {
     // Gets the remainder address from configuration of finds one from the inputs.
     fn get_remainder_address(&self) -> Option<(Address, Option<Bip44>)> {
-        if self.remainder_address.is_some() {
-            return self.remainder_address.map(|address| (address, None));
+        if let Some(remainder_address) = self.remainder_address {
+            // Search in inputs for the Bip44 chain for the remainder address, so the ledger can regenerate it
+            #[cfg(feature = "ledger_nano")]
```
thibault-martinez: I wonder if we should just do it always

thibault-martinez: Could be useful for other contexts than ledger_nano idk

Thoralf-M:
```suggestion
```

Thoralf-M: True

sdk/src/client/api/block_builder/input_selection/core/remainder.rs
```diff
@@ -22,8 +22,25 @@ use crate::{
...
+                let alias_transition = is_alias_transition(
+                    &input.output,
+                    *input.output_id(),
+                    self.outputs.as_slice(),
+                    self.burn.as_ref(),
+                );
+                let required_address =
+                    input
+                        .output
+                        .required_and_unlocked_address(self.timestamp, input.output_id(), alias_transition);
```
thibault-martinez: I know the current error can only come from the Treasury thing but I'm just thinking that if ever we add some more faillible stuff to be, we'd be silently ignoring it, maybe we should just make get_remainder_address faillible and propagate. It's all private, it's not the end of the world I guess

Thoralf-M: Hm but even if there is another output in the future, we still wouldn't want to return an error, no?
I can also use `if let Ok(required_address)` if you prefer that

thibault-martinez: I was not necessarily speaking about other kinds of outputs, just if for some reason we had something else that make sit faillible then we'd wouldn't notice here. I don't really see a downside making the whole thing fallible, so I'd say let's be on the safe side

thibault-martinez: I'll approve anyway if you think it's not worth it

thibault-martinez (dismissed):

sdk/src/client/api/block_builder/input_selection/core/remainder.rs
Thoralf-M: Changed, but I think for other failure reasons we later might need extra handling then and ignore the error in this case

thibault-martinez (changes requested):
Conflict

thibault-martinez (dismissed):

thibault-martinez (approved):

kwek20 (approved):

PR #1298, opened 2023-09-22
Query parameter function

### LINKED ISSUE #1173: Make `get_request` query parameter easier to use
> Please also add the optional cursor as parameter https://github.com/iotaledger/tips-draft/pull/24/files#diff-04aff9a5004dd8023df5735c866f69254359635764a00e9d9215500a9676319bR2881-R2883 to `get_validators()`

I had a look and I think I'll do it in another PR. The `query` parameter of `get_request` is not very handy, it forces us to create the param string ourselves. This should take a vector and build the string inside.

_Originally posted by @thibault-martinez in https://github.com/iotaledger/iota-sdk/issues/1165#issuecomment-1710258522_

thibault-martinez:
Closes https://github.com/iotaledger/iota-sdk/issues/1173


sdk/src/client/node_api/mod.rs
```diff
@@ -13,3 +13,14 @@ pub mod mqtt;
 #[cfg_attr(docsrs, doc(cfg(feature = "participation")))]
 pub mod participation;
 pub mod plugin;
+
+pub(crate) fn query_tuples_to_query_string(
+    query: impl IntoIterator<Item = Option<(&'static str, String)>>,
```
kwek20: maybe make the String param a `ToString`? 

would reduce `page_size.into().map(|i| ("pageSize", i.to_string())),`
to `("pageSize", page_size.into())`

thibault-martinez: Why would that remove the first into? It's for the option

thibault-martinez: And `ToString` don't really allow you to call `into` so I can't do any of what you proposed

kwek20: sure then do `Into<String>`, its just a suggestion without checking out your code

thibault-martinez: Ok but just to be sure, it doesn't change into what you said. It changed `page_size.into().map(|i| ("pageSize", i.to_string()))` to `page_size.into().map(|i| ("pageSize", i.into()))` which doesn't seem like a big improvement.

kwek20: I see, tried it myself and the Into<Option<>> make it impossible to have clean

kwek20 (dismissed):

Thoralf-M (approved):

kwek20 (approved):

abdulmth (approved):

PR #1304, opened 2023-09-22
Node: Fix cyclic dep. in transaction payload

### LINKED ISSUE #1249: Node: Nested payload in TransactionDataPayload is not properly typed

abdulmth:
Fixes https://github.com/iotaledger/iota-sdk/issues/1249
replaces https://github.com/iotaledger/iota-sdk/pull/1253

The issue actually is not a cyclic dependency in class transformer, it's cyclic dependency in our code. Since `PayloadDiscriminator` depends on → `TransactionPayload` → `RegularTransactionEssence` → `PayloadDiscrminator` so the last would be replaced with 
`undefined`  by JavaScript.

### DIFF
--- a/bindings/nodejs/CHANGELOG.md
+++ b/bindings/nodejs/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
-## 1.0.12 - 2023-mm-dd
+## 1.0.12 - 2023-09-25
 
 ### Fixed
 
--- a/bindings/nodejs/CHANGELOG.md
+++ b/bindings/nodejs/CHANGELOG.md
@@ -19,6 +19,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 1.0.12 - 2023-mm-dd
+
+### Added
+
+- `RegularTransactionEssence.getParsedPayload()` method;
+
+### Changed
+
+- `RegularTransactionEssence.payload` type to `unknown`;
+
 ## 1.0.11 - 2023-09-14
 
 ### Fixed
--- a/bindings/nodejs/CHANGELOG.md
+++ b/bindings/nodejs/CHANGELOG.md
@@ -21,13 +21,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## 1.0.12 - 2023-mm-dd
 
-### Added
+### Fixed
 
-- `RegularTransactionEssence.getParsedPayload()` method;
-
-### Changed
-
-- `RegularTransactionEssence.payload` type to `unknown`;
+- Parsing of `RegularTransactionEssence.payload`;
 
 ## 1.0.11 - 2023-09-14
 
--- a/bindings/nodejs/lib/types/block/payload/transaction/essence.ts
+++ b/bindings/nodejs/lib/types/block/payload/transaction/essence.ts
@@ -76,7 +76,7 @@ class RegularTransactionEssence extends TransactionEssence {
         inputsCommitment: HexEncodedString,
         inputs: Input[],
         outputs: Output[],
-        payload: Payload | undefined,
+        payload: unknown | undefined,
     ) {
         super(TransactionEssenceType.Regular);
         this.networkId = networkId;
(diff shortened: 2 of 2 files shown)


bindings/nodejs/lib/types/block/payload/transaction/essence.ts
```diff
@@ -57,10 +60,7 @@ class RegularTransactionEssence extends TransactionEssence {
     })
     outputs: Output[];
 
-    @Type(() => Payload, {
-        discriminator: PayloadDiscriminator,
-    })
-    payload: Payload | undefined;
+    payload?: unknown;
```
thibault-martinez: I know nothing about TS but this looks sus :(

abdulmth: `unknown` is a special type in TS, from [w3schools](https://www.w3schools.com/typescript/typescript_special_types.php#:~:text=unknown%20is%20best%20used%20when,is%20of%20the%20casted%20type.) 

> unknown is best used when you don't know the type of data being typed. To add a type later, you'll need to cast it.
Casting is when we use the "as" keyword to say property or variable is of the casted type.

And this is what we want here because we don't know what type `payload` will be.  Also using `Payload` isn't really of any use because it's an abstract class that shouldn't be constructed but used internally to cast to the subclasses.

kwek20 (dismissed):

bindings/nodejs/lib/types/block/payload/transaction/essence.ts
```diff
@@ -39,6 +41,18 @@ abstract class TransactionEssence {
     }
 }
 
+const PayloadDiscriminator = {
```
Thoralf-M, line 45:
```suggestion
/**
 * PayloadDiscriminator for payloads inside of a TransactionEssence.
 */
const PayloadDiscriminator = {
```
so it doesn't get misused for payloads in a block

bindings/nodejs/lib/types/block/payload/transaction/essence.ts
```diff
@@ -39,6 +41,18 @@ abstract class TransactionEssence {
...
+    property: 'type',
+    subTypes: [
+        { value: MilestonePayload, name: PayloadType.Milestone as any },
+        { value: TaggedDataPayload, name: PayloadType.TaggedData as any },
+        {
+            value: TreasuryTransactionPayload,
+            name: PayloadType.TreasuryTransaction as any,
+        },
```
Thoralf-M:
```suggestion
        { value: TaggedDataPayload, name: PayloadType.TaggedData as any },
```

Thoralf-M (dismissed):

bindings/nodejs/CHANGELOG.md
```diff
@@ -19,6 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 1.0.12 - 2023-mm-dd
```
```suggestion
## 1.0.12 - 2023-09-25
```

thibault-martinez (approved):

kwek20 (approved):

Thoralf-M (approved):

PR #1305, opened 2023-09-23
Put IDs definition with outputs/NTs

thibault-martinez:
- Less cluttered outputs folder
- Delegation ID was already put together with the output and I think it makes sense


sdk/src/types/block/output/account.rs
```diff
@@ -33,6 +33,30 @@ use crate::types::{
...
+string_serde_impl!(AccountId);
+
+impl From<&OutputId> for AccountId {
+    fn from(output_id: &OutputId) -> Self {
+        Self::from(output_id.hash())
+    }
+}
+
+impl AccountId {
+    ///
```
Alex6323, line 48: docs?

thibault-martinez: Outside of the scope of this PR, will do later

sdk/src/types/block/output/foundry.rs
```diff
@@ -32,6 +33,54 @@ use crate::types::{
...
+        serial_number.pack(&mut packer).unwrap();
+        token_scheme_kind.pack(&mut packer).unwrap();
+
+        Self::new(bytes)
+    }
+
+    /// Returns the [`AccountAddress`] of the [`FoundryId`].
+    pub fn account_address(&self) -> AccountAddress {
+        // PANIC: the lengths are known.
+        AccountAddress::from(AccountId::new(self.0[1..AccountId::LENGTH + 1].try_into().unwrap()))
```
Alex6323, line 64: It's a bit unfortunate that `FoundryId` has to have exact knowledge about how `Address` packs its stuff (with 1 byte for the `AddressType` I assume). I know it's annoying but instead of the magic number `1` something like an `Address::OFFSET` constant would make sense.

thibault-martinez: Honestly I wouldn't add a constant just for that specific use case. I can throw in a size_of if you prefer but it doesn't change much 🤷🏻‍♂️

sdk/src/types/block/output/foundry.rs
```diff
@@ -32,6 +33,54 @@ use crate::types::{
...
+            self.0[AccountId::LENGTH + 1..AccountId::LENGTH + 1 + core::mem::size_of::<u32>()]
+                .try_into()
+                .unwrap(),
+        )
+    }
+
+    /// Returns the [`TokenScheme`](crate::types::block::output::TokenScheme) kind of the [`FoundryId`].
+    pub fn token_scheme_kind(&self) -> u8 {
+        // PANIC: the lengths are known.
+        *self.0.last().unwrap()
```
Alex6323, line 80: well, the panic comment doesn't really make sense here. Also ... using `last()` is elegant and probably fine for all eternity but not the strict definition of its position :smiley:

sdk/src/types/block/output/nft.rs
```diff
@@ -30,6 +30,30 @@ use crate::types::{
...
+string_serde_impl!(NftId);
+
+impl From<&OutputId> for NftId {
+    fn from(output_id: &OutputId) -> Self {
+        Self::from(output_id.hash())
+    }
+}
+
+impl NftId {
+    ///
```
Alex6323, line 45: doc or allow missing?

thibault-martinez: Later

sdk/src/types/block/output/foundry.rs
Alex6323: close if you think it's fine

Alex6323 (dismissed):

sdk/src/types/block/output/foundry.rs
thibault-martinez: I'll adapt the comment because it's indeed weird, but the last is fine until the protocol changes

abdulmth (approved):

Alex6323 (approved):

Thoralf-M (approved):

PR #1306, opened 2023-09-23
Finish Delegation TIP compliance

thibault-martinez:


sdk/src/types/block/output/delegation.rs
```diff
@@ -355,6 +357,53 @@ impl DelegationOutput {
             .locked_address(self.address(), context.essence.creation_slot())
             .unlock(unlock, inputs, context)
     }
+
+    // Transition, just without full ValidationContext.
+    pub(crate) fn transition_inner(current_state: &Self, next_state: &Self) -> Result<(), StateTransitionError> {
+        if !(current_state.delegation_id.is_null() && !next_state.delegation_id().is_null()) {
+            return Err(StateTransitionError::NonDelayedClaimingTransition);
+        }
```
Alex6323, line 364: can you confirm that the `!`s are correct? (one outside and one inside of the `(..)`) Not knowing the details but if they are correct maybe I would apply DeMorgan to make this term a bit easier.

thibault-martinez: It's correct and I would keep it like this, otherwise it would read different from the TIP and that would be annoying

sdk/src/types/block/error.rs
```diff
@@ -171,6 +171,7 @@ pub enum Error {
     UnsupportedOutputKind(u8),
     DuplicateOutputChain(ChainId),
     InvalidField(&'static str),
+    NullDelegationValidatorId,
```
Alex6323, line 174: Does `Null` mean `Empty` (as in Not-Provided)?

thibault-martinez: No, it means full of 0

sdk/src/types/block/output/delegation.rs
```diff
@@ -236,21 +239,22 @@ impl From<&DelegationOutput> for DelegationOutputBuilder {
...
 #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct DelegationOutput {
-    // Amount of IOTA tokens held by the output.
+    /// Amount of IOTA tokens to deposit with this output.
     amount: u64,
-    /// The amount of delegated coins.
+    /// Amount of delegated IOTA tokens.
     delegated_amount: u64,
-    /// Unique identifier of the Delegation Output, which is the BLAKE2b-256 hash of the Output ID that created it.
+    /// Unique identifier of the delegation output.
```
Alex6323, line 248: learning question: is the explanation not correct anymore?

thibault-martinez: Just unnecessary here, it's already in the type definition. No need to put it every time we have a delegation ID field.

Alex6323 (dismissed):

sdk/src/types/block/output/delegation.rs
Thoralf-M, line 364: This code should never be reached, since the next state is determined by the id, no?

thibault-martinez: It's more to prevent transitions from explicit ID to explicit ID as this doesn't seem to be allowed

sdk/src/types/block/output/delegation.rs
```diff
@@ -355,6 +357,53 @@ impl DelegationOutput {
...
+            return Err(StateTransitionError::NonDelayedClaimingTransition);
+        }
+
+        if current_state.delegated_amount != next_state.delegated_amount
+            || current_state.start_epoch != next_state.start_epoch
+            || current_state.validator_id != next_state.validator_id
+        {
+            return Err(StateTransitionError::MutatedImmutableField);
+        }
+
```
Thoralf-M: What about the end epoch rules?

Thoralf-M:
```suggestion
// TODO add end_epoch validation rules
```

thibault-martinez: https://github.com/iotaledger/iota-sdk/issues/1313

Thoralf-M (dismissed):

thibault-martinez:
Need to address https://github.com/iotaledger/tips-draft/pull/15/files/20a9a275206d60ca6358726a756fe16bbafd70dc..4c648a021e3a6dbee91c6f60344785ab3ada7e11

Thoralf-M (dismissed):

Thoralf-M (approved):

Alex6323 (approved):

abdulmth (approved):

PR #1314, opened 2023-09-25
Increase private tangle timeout

thibault-martinez:


Thoralf-M:
Workflow doesn't seem to use go 1.21.1 even though it should be available https://github.com/iotaledger/iota-sdk/actions/runs/6299326338/job/17099863481?pr=1314#step:7:711 ? With an older version, the snapshot creation will fail, it will copy nothing and therefore create a folder instead of a file and then the nodes crash

thibault-martinez:
Closing as timeout doesn't seem to be the issue.

PR #1337, opened 2023-09-27
SlotIndex and EpochIndex as u32

### LINKED ISSUE #1316: Make SlotIndex and EpochIndex u32
(will update when there is a TIP change)

thibault-martinez:
Closes https://github.com/iotaledger/iota-sdk/issues/1316


DaughterOfMars (commented):
good stuff

sdk/src/types/block/protocol.rs
```diff
@@ -163,8 +157,8 @@ impl ProtocolParameters {
     }
 
     /// Returns the slots per epoch of the [`ProtocolParameters`].
-    pub fn slots_per_epoch(&self) -> u64 {
-        2_u64.pow(self.slots_per_epoch_exponent() as u32)
+    pub fn slots_per_epoch(&self) -> u32 {
+        2_u32.pow(self.slots_per_epoch_exponent() as u32)
```
DaughterOfMars, line 161: Could do this
```suggestion
        1 << self.slots_per_epoch_exponent()
```

thibault-martinez: I'd rather keep the pow version tbh, intent is clearer

sdk/src/types/block/block_id.rs
```diff
@@ -34,7 +34,7 @@ pub struct BlockId {
 
 impl BlockId {
     /// The length of a [`BlockId`]
-    pub const LENGTH: usize = 40;
+    pub const LENGTH: usize = 36;
```
Thoralf-M, line 37: Since when is this the case?

thibault-martinez: Because slot index is part of the BlockId and it's now 4 bytes shorter

Thoralf-M: Ah sure makes sense, wen TIP updated

thibault-martinez: 🤷🏻

Thoralf-M (dismissed):

DaughterOfMars (dismissed):

sdk/src/types/block/protocol.rs
DaughterOfMars: This is how they do it in iota.go

Thoralf-M (approved):

DaughterOfMars (approved):

kwek20 (approved):

PR #1345, opened 2023-09-28
Separate block parent type defs

### LINKED ISSUE #1208: Per-block parents types with bounds
The following types are currently shared for all block variants (basic and validation) but these should have different bounds.
They need to be defined in each block module.

```rust
pub type StrongParents = Parents<1, 8>;
pub type WeakParents = Parents<0, 8>;
pub type ShallowLikeParents = Parents<0, 8>;
```

DaughterOfMars:
# Description of change

Now we get two christmases 🤷‍♀️ 

Closes #1208


sdk/src/types/api/core/response.rs
```diff
@@ -262,11 +262,11 @@ pub struct ValidatorResponse {
...
 pub struct IssuanceBlockHeaderResponse {
     /// Blocks that are strongly directly approved.
-    pub strong_parents: StrongParents,
+    pub strong_parents: basic::StrongParents,
     /// Blocks that are weakly directly approved.
-    pub weak_parents: WeakParents,
+    pub weak_parents: basic::WeakParents,
     /// Blocks that are directly referenced to adjust opinion.
-    pub shallow_like_parents: ShallowLikeParents,
+    pub shallow_like_parents: basic::ShallowLikeParents,
```
thibault-martinez: I don't know about these. That would mean we can't build a ValidationBlock from the API return. I don't think the number returned by the API should be validated at all. It's only when we try to build a Block with these that we should validated.

sdk/src/types/block/core/mod.rs
```diff
@@ -136,10 +136,10 @@ impl Packable for Block {
...
 }
 
-pub(crate) fn verify_parents(
-    strong_parents: &StrongParents,
-    weak_parents: &WeakParents,
-    shallow_like_parents: &ShallowLikeParents,
+pub(crate) fn verify_parents<const MAX: u8>(
+    strong_parents: &Parents<1, MAX>,
+    weak_parents: &Parents<0, MAX>,
+    shallow_like_parents: &Parents<0, MAX>,
```
thibault-martinez: This doesn't seem very flexible, same MAX for all, no variable MIN. But anyway, for this function, lengths don't matter, maybe we can make it operate on slices and not need generics at all.

sdk/src/types/block/core/mod.rs
thibault-martinez, line 1: Unrelated note but I kinda think we should move parents to core as it's where it should be defined and used by the block variants

sdk/src/types/block/core/mod.rs
```diff
@@ -21,8 +21,8 @@ pub use self::{
     validation::{ValidationBlock, ValidationBlockBuilder},
     wrapper::{BlockHeader, BlockWrapper},
 };
+use super::parent::Parents;
```
thibault-martinez: Not a big fan of mixing super and crate tbh, I don't see the logic here, why parents and not protocol? It's inconsistent with the rest of the codebase, I'd keep it in the crate import.

sdk/src/types/api/core/response.rs
DaughterOfMars: hmm, okay, I'll give it a try

sdk/src/types/block/core/mod.rs
DaughterOfMars: you right 👍

sdk/src/types/block/core/mod.rs
DaughterOfMars: done

sdk/src/types/block/core/mod.rs
DaughterOfMars: fixed

sdk/src/types/api/core/response.rs
```diff
@@ -262,17 +267,35 @@ pub struct ValidatorResponse {
...
-    pub shallow_like_parents: ShallowLikeParents,
+    pub shallow_like_parents: BTreeSet<BlockId>,
     /// The slot index of the latest finalized slot.
     pub latest_finalized_slot: SlotIndex,
     /// The most recent slot commitment.
     pub commitment: SlotCommitment,
 }
 
+impl IssuanceBlockHeaderResponse {
+    pub fn strong_parents<const MIN: u8, const MAX: u8>(
```
thibault-martinez, line 282: Very nice

sdk/src/types/block/core/mod.rs
```diff
@@ -137,15 +138,13 @@ impl Packable for Block {
 }
 
 pub(crate) fn verify_parents(
```
thibault-martinez: Move to the parents module ?

sdk/examples/client/get_block.rs
```diff
@@ -22,7 +22,7 @@ async fn main() -> Result<()> {
         .await?;
 
     // Fetch a block ID from the node.
-    let block_id = client.get_issuance().await?.strong_parents[0];
+    let block_id = client.get_issuance().await?.strong_parents.into_iter().next().unwrap();
```
thibault-martinez, line 25: Bit annoying but I guess we don't really have a cleaner way

DaughterOfMars: yeah not really

thibault-martinez (approved):

Thoralf-M (approved):

Alex6323 (approved):
:santa: :broken_heart:  :mrs_claus:

PR #1347, opened 2023-09-28
Add and use Epoch- and SlotIndex

### LINKED ISSUE #1191: Python: add SlotIndex and EpochIndex
And address https://github.com/iotaledger/iota-sdk/pull/1181/files#diff-f59e51f47c5db15e2a65128abf6ff590cf0c210bf5b97f412caa236c0469782bR98

And address https://github.com/iotaledger/iota-sdk/pull/1196/files#diff-f59e51f47c5db15e2a65128abf6ff590cf0c210bf5b97f412caa236c0469782bR123

Thoralf-M:
# Description of change

Add and use Epoch- and SlotIndex
Removed one `staking_unbonding_period` because it was duplicated, also put `subunit` at last position, because optional values needs to be last 

Closes #1191

### DIFF
--- a/bindings/python/iota_sdk/types/node_info.py
+++ b/bindings/python/iota_sdk/types/node_info.py
@@ -210,12 +210,12 @@ class ProtocolParameters:
     slot_duration_in_seconds: int
     slots_per_epoch_exponent: int
     mana_structure: ManaStructure
-    staking_unbonding_period: str
+    staking_unbonding_period: EpochIndex
     validation_blocks_per_slot: int
-    punishment_epochs: str
-    liveness_threshold: str
-    min_committable_age: str
-    max_committable_age: str
+    punishment_epochs: EpochIndex
+    liveness_threshold: SlotIndex
+    min_committable_age: SlotIndex
+    max_committable_age: SlotIndex
     epoch_nearing_threshold: SlotIndex
     congestion_control_parameters: CongestionControlParameters
     version_signaling: VersionSignaling
--- a/bindings/python/iota_sdk/types/node_info.py
+++ b/bindings/python/iota_sdk/types/node_info.py
@@ -212,7 +212,7 @@ class ProtocolParameters:
     mana_structure: ManaStructure
     staking_unbonding_period: EpochIndex
     validation_blocks_per_slot: int
-    punishment_epochs: EpochIndex
+    punishment_epochs: str
     liveness_threshold: SlotIndex
     min_committable_age: SlotIndex
     max_committable_age: SlotIndex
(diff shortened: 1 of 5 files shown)


bindings/python/iota_sdk/types/node_info.py
```diff
@@ -217,12 +213,10 @@ class ProtocolParameters:
     staking_unbonding_period: str
     validation_blocks_per_slot: int
     punishment_epochs: str
-    staking_unbonding_period: str
```
thibault-martinez, line 220: Wait that's allowed in Python? 2 fields with the same name?

thibault-martinez: Also, it should be an EpochIndex

bindings/python/iota_sdk/types/node_info.py
```diff
@@ -217,12 +213,10 @@ class ProtocolParameters:
     staking_unbonding_period: str
     validation_blocks_per_slot: int
     punishment_epochs: str
```
thibault-martinez, line 215: EpochIndex

bindings/python/iota_sdk/types/node_info.py
```diff
@@ -217,12 +213,10 @@ class ProtocolParameters:
...
-    staking_unbonding_period: str
     liveness_threshold: str
     min_committable_age: str
     max_committable_age: str
```
thibault-martinez, line 218: All 3 before are SlotIndex

DaughterOfMars (dismissed):

DaughterOfMars (dismissed):

DaughterOfMars (dismissed):

bindings/python/iota_sdk/types/node_info.py
thibault-martinez: Why not ?

DaughterOfMars: yeah I was confused about this too

Thoralf-M: It's just an amount of epochs, not an epoch index

Thoralf-M: In Rust we also don't use EpochIndex there

DaughterOfMars: fair enough

thibault-martinez: Kind of the same for committable age etc tbh. I actually think we should change it in Rust as well then? It's not an index sure, but it's a delta index, maybe it's interesting to make sure they are the same types because we end up doing operations on them. Idk 🤷🏻‍♂️

Thoralf-M: Would rather also remove the SlotIndex there, computation should be possible wit the inner value

thibault-martinez: As you prefer, as long as we're consistent

thibault-martinez (approved):

DaughterOfMars (approved):
not super happy about ditching the types but whatever

Alex6323 (approved):

PR #1369, opened 2023-10-02
feat/Node: id() on wrapper

kwek20:
# Description of change

the method `.id(protocolparams)` was missing on the blockwrapper type, which we have in rust

### DIFF
--- a/bindings/nodejs/lib/types/block/core/wrapper.ts
+++ b/bindings/nodejs/lib/types/block/core/wrapper.ts
@@ -1,7 +1,7 @@
 // Copyright 2023 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-import { IssuerId } from '../id';
+import { BlockId, IssuerId } from '../id';
 import { Signature, SignatureDiscriminator } from '../signature';
 import { SlotCommitmentId, SlotIndex } from '../slot';
 import { u64 } from '../../utils/type-aliases';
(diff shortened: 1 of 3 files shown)


bindings/nodejs/lib/types/block/core/wrapper.ts
```diff
@@ -73,6 +75,16 @@ class BlockWrapper {
         this.block = block;
     }
 
+    /**
+     * Compute the block ID (Blake2b256 hash of the block bytes).
+     *
+     * @param params The network protocol parameters.
+     * @returns The corresponding block ID.
+     */
+    id(params: ProtocolParameters): BlockId {
```
DaughterOfMars, line 84: Is this another scenario where we're going to be importing stuff we don't need if you just want the types?

bindings/nodejs/lib/types/utils/bridge/utils.ts
```diff
@@ -8,6 +8,8 @@ import {
     TokenSchemeType,
     Output,
     RentStructure,
```
Thoralf-M, line 10: Can't comment there, but 'Block' is defined but never used

bindings/nodejs/lib/utils/utils.ts
```diff
@@ -15,6 +15,8 @@ import {
     RentStructure,
```
Thoralf-M, line 15: Can't comment there, but 'Block' is defined but never used

bindings/nodejs/lib/types/block/core/wrapper.ts
kwek20: Seems to be so yes. :/

PR #1370, opened 2023-10-02
Fix set_client_options()

### LINKED ISSUE #1257: Wallet.set_client_options() doesn't update properly
## Description

When changing the client options to a node from another network, the bech32 HRPs of the addresses don't get updated
Might need 
```
        for account in self.accounts.write().await.iter_mut() {
            account.update_account_bech32_hrp().await?;
        }
```
here https://github.com/iotaledger/iota-sdk/blob/7aa1e7afbff4e468681ca0a805d72e80e9774d3f/sdk/src/wallet/core/operations/client.rs#L71, but also needs to have the bech32 hrp updated before, currently it seems to just use the one before or default (only node url provided when setting the client options)

So looks like the network info stays outdated, because if one sends a transaction, it also uses inputs from the other network which results in a conflicting transaction

Having a wallet connected to the testnet and then changing to a private tangle should update the addresses and not result in conflicting transactions.

No.

Thoralf-M:
# Description of change

Fix set_client_options() by getting the nodeinfo and updating the protocol params and then the accounts with a potential new bech32 hrp

Fixes #1257 

## Type of change

Running

```Rust
use iota_sdk::{
    client::{
        constants::SHIMMER_COIN_TYPE,
        secret::{stronghold::StrongholdSecretManager, SecretManager},
    },
    wallet::{ClientOptions, Result, Wallet},
};

#[tokio::main]
    // Setup Stronghold secret manager.
    // WARNING: Never hardcode passwords in production code.
    let secret_manager = StrongholdSecretManager::builder()
        .password("password".to_owned()) // A password to encrypt the stored data.
        .build("getting-started-db/vault.stronghold")?; // The path to store the account snapshot.

    let client_options = ClientOptions::new().with_node("https://api.shimmer.network")?;

    // Set up and store the wallet.
        .with_secret_manager(SecretManager::Stronghold(secret_manager))
        .with_client_options(client_options)
        .with_coin_type(SHIMMER_COIN_TYPE)
        .with_storage_path("getting-started-db")
        .finish()
        .await?;

    // Generate a mnemonic and store its seed in the Stronghold vault.
    // INFO: It is best practice to back up the mnemonic somewhere secure.
    let mnemonic = wallet.generate_mnemonic()?;
    println!("Mnemonic: {}", mnemonic.as_ref());
    wallet.store_mnemonic(mnemonic).await?;

    // Create an account.
    let account = wallet
        .create_account()
        .with_alias("Alice") // A name to associate with the created account.
        .finish()
        .await?;

    let first_address = &account.addresses().await?[0];
    println!("{}", first_address.address());

    let client_options = ClientOptions::new().with_node("https://api.testnet.shimmer.network")?;
    wallet.set_client_options(client_options).await?;

    let first_address = &account.addresses().await?[0];
    println!("{}", first_address.address());

    Ok(())
}
```
Got 
```
smr1qzkpx0wyhnepnaccagz5sdg8rxxktc6hy3ujqrarfu6shesxqp2fchhqcsh
rms1qzkpx0wyhnepnaccagz5sdg8rxxktc6hy3ujqrarfu6shesxqp2fcrstztw
````
where it without this change would print twice the same bech32 address

### DIFF
--- a/sdk/src/wallet/core/operations/client.rs
+++ b/sdk/src/wallet/core/operations/client.rs
@@ -37,7 +37,7 @@ where
             node_manager_builder,
             #[cfg(feature = "mqtt")]
             broker_options,
-            network_info,
+            mut network_info,
             api_timeout,
             remote_pow_timeout,
             #[cfg(not(target_family = "wasm"))]
@@ -48,7 +48,6 @@ where
         self.client
             .update_node_manager(node_manager_builder.build(HashMap::new()))
             .await?;
-        *self.client.network_info.write().await = network_info;
         *self.client.api_timeout.write().await = api_timeout;
         *self.client.remote_pow_timeout.write().await = remote_pow_timeout;
         #[cfg(not(target_family = "wasm"))]
@@ -61,6 +60,16 @@ where
         {
             *self.client.mqtt.broker_options.write().await = broker_options;
         }
+
+        // Update the protocol of the network_info to not have the default data, which can be wrong
+        let info = self.client.get_info().await?.node_info;
+        network_info.protocol_parameters = info.protocol.clone();
+        *self.client.network_info.write().await = network_info;
+
+        for account in self.accounts.write().await.iter_mut() {
+            account.update_account_bech32_hrp().await?;
+        }
+
         #[cfg(feature = "storage")]
         {
             WalletBuilder::from_wallet(self)
--- a/sdk/src/wallet/core/operations/client.rs
+++ b/sdk/src/wallet/core/operations/client.rs
@@ -62,8 +62,10 @@ where
         }
 
         // Update the protocol of the network_info to not have the default data, which can be wrong
-        let info = self.client.get_info().await?.node_info;
-        network_info.protocol_parameters = info.protocol.clone();
+        // Ignore errors, because there might be no node at all and then it should still not error
+        if let Ok(info) = self.client.get_info().await {
+            network_info.protocol_parameters = info.node_info.protocol;
+        }
         *self.client.network_info.write().await = network_info;
 
         for account in self.accounts.write().await.iter_mut() {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: sdk/src/wallet/core/operations/client.rs [hunk 1]
```diff
@@ -61,6 +60,16 @@ where
         {
             *self.client.mqtt.broker_options.write().await = broker_options;
         }
+
+        // Update the protocol of the network_info to not have the default data, which can be wrong
```
DaughterOfMars: I don't understand getting the info. Isn't this kind of up to to caller?

Thoralf-M: Usually people will just build the client options like this `ClientOptions::new().with_node("https://api.testnet.shimmer.network")?`, so basically only set the node field and everything else will be our default values. But the default values for protocol params are wrong if it's not the shimmer network, therefore the addresses will also have `smr` in the accounts, even though the network for which you provided the url, can have something else

DaughterOfMars: then we should change the protocol params field to be optional so we know whether it's been set manually by the user

Thoralf-M: But this would be breaking change

DaughterOfMars: Well, technically yes, but I wonder if it's necessary to fix the bug here. This way isn't correct, because it overwrites the user's manually defined options.

Thoralf-M: We overwrite this already anyways in the node syncing process. The background syncing just runs every x seconds and therefore highly unlikely at exactly this moment where we need to have it updated

DaughterOfMars (dismissed):

thibault-martinez (dismissed):

thibault-martinez (approved):

sdk/src/wallet/core/operations/stronghold_backup/mod.rs
```diff
@@ -161,6 +160,8 @@ impl Wallet {
             }
         }
 
+        let mut accounts = self.accounts.write().await;
```
Alex6323, line 163: :wen: no accounts? :trollface:

Alex6323 (approved):

PR #1376, opened 2023-10-02
Add mana calculations

### LINKED ISSUE #1366: Add mana calculations
https://github.com/iotaledger/tips-draft/blob/tip39/tips/TIP-0039/tip-0039.md

DaughterOfMars:
# Description of change

Adds mana calculations

Closes #1366


sdk/src/client/api/mana.rs
```diff
@@ -0,0 +1,81 @@
...
+
+use crate::{
+    client::Client,
+    types::block::{
+        protocol::{lower_bits, multiplication_and_shift, upper_bits},
+        slot::SlotIndex,
+    },
+};
+
+impl Client {
```
thibault-martinez: I wouldn't put this in the client. All mana calculations should be in the types module, if possible in the mana module.

DaughterOfMars: The problem is the necessity of the epochs iter. We need the client to do that calculation anyway

sdk/src/client/api/mana.rs
```diff
@@ -0,0 +1,81 @@
...
+        &self,
+        amount: u64,
+        slot_index_created: SlotIndex,
+        slot_index_target: SlotIndex,
+    ) -> crate::client::Result<u64> {
+        if slot_index_created >= slot_index_target {
+            return Ok(0);
+        }
+        let protocol_parameters_map = self.get_info().await?.node_info.protocol_parameters;
+        let params: &crate::types::block::protocol::ProtocolParameters = &protocol_parameters_map.latest().parameters;
```
Thoralf-M: We would need to get the correct one by `start_epoch` field and the current epoch and not just the latest one?

DaughterOfMars: Hmm, so then what if start and target epoch are using different protocol params? The whole calculation is potentially wrong in that case. 😱

Thoralf-M: Yeah I guess it needs to be adjusted for that

DaughterOfMars: But how can I handle that? The functions described in the TIP assume that the same mana structure is used for both

Thoralf-M: Maybe just have a function around this one, that loops over the protocol parameters with epoch index from when the output was created until this point and for each of them call this `potential_mana()`? Probably also worth to check what they did in go, for the node they need to calculate it also

DaughterOfMars: They simply don't handle the possibility of a protocol parameter change at all. This is their fn to convert from slot to epoch, which does not take into account a variable `slotsPerEpochExponent`.

```go
// EpochFromSlot calculates the EpochIndex from the given slot.
func (t *TimeProvider) EpochFromSlot(slot SlotIndex) EpochIndex {
	return EpochIndex(slot >> t.slotsPerEpochExponent)
}
```

So I'm really not sure what to do about this.

Thoralf-M: Ok apparently it's something we shouldn't care about now, because it's not planned to have new protocol parameters without bigger changes, which we would need to implement before anyways

sdk/src/client/api/mana.rs
DaughterOfMars: Fixed

sdk/src/types/block/mana/structure.rs
```diff
@@ -65,3 +108,103 @@ impl Default for ManaStructure {
         }
     }
 }
+
+impl ProtocolParameters {
+    /// Calculates the potential mana that is generated by holding `amount` tokens from `slot_index_created` to
+    /// `slot_index_target` and applies the decay to the result
+    pub fn potential_mana(&self, amount: u64, slot_index_created: SlotIndex, slot_index_target: SlotIndex) -> u64 {
```
kwek20: Into dyn index ?

DaughterOfMars: k

sdk/src/types/block/slot/index.rs
```diff
@@ -42,20 +41,16 @@ use crate::types::block::Error;
...
+pub struct SlotIndex(pub u64);
 
 impl SlotIndex {
-    /// Creates a new [`SlotIndex`].
-    pub fn new(index: u64) -> Self {
-        Self::from(index)
+    /// Gets the [`EpochIndex`] of this slot.
+    pub fn to_epoch_index(self, slots_per_epoch_exponent: u8) -> EpochIndex {
+        EpochIndex::from_slot_index(self, slots_per_epoch_exponent)
     }
```
Alex6323, line 50:
```suggestion
    /// Converts the slot index into the [`EpochIndex`] of this slot.
    pub fn to_epoch_index(self, slots_per_epoch_exponent: u8) -> EpochIndex {
        EpochIndex::from_slot_index(self, slots_per_epoch_exponent)
    }
```

sdk/src/types/block/slot/epoch.rs
```diff
@@ -32,52 +31,47 @@ use crate::types::block::Error;
...
-                    break;
-                }
-            }
-            res = *start_epoch + (slot_index >> exponent);
-            last = Some((*start_epoch, exponent));
-        }
-        Ok(Self(res))
+    pub fn from_slot_index(slot_index: SlotIndex, slots_per_epoch_exponent: u8) -> Self {
+        Self(*slot_index >> slots_per_epoch_exponent)
+    }
```
Alex6323, line 65: I know you didn't write the doc, but consider fixing it since it's not a getter.

sdk/src/types/block/mana/structure.rs
```diff
@@ -41,14 +45,53 @@ impl ManaStructure {
...
+            let epochs_to_decay = remaining_epochs.min(self.decay_factors().len() as u64);
+            remaining_epochs -= epochs_to_decay;
+
+            // Unwrap: Safe because the index is at most the length
+            let decay_factor = self.decay_factor_at(epochs_to_decay - 1).unwrap();
+
+            // apply the decay using fixed-point arithmetics.
+            (mana_hi, mana_lo) =
+                multiplication_and_shift(mana_hi, mana_lo, decay_factor, self.decay_factors_exponent());
+        }
```
Alex6323: why does this always terminate? Bc `self.decay_factors().len()` can never be `0` while `remaining_epochs > 0`?

sdk/src/types/block/slot/index.rs
DaughterOfMars: I don't think that's better

sdk/src/types/block/slot/epoch.rs
DaughterOfMars: I don't see an issue?

sdk/src/types/block/mana/structure.rs
DaughterOfMars: `epochs_to_decay` will either be the length of the decay factors lookup, or it will be `== remaining_epochs`, in which case subtracting it will make `remaining_epochs == 0`

sdk/src/types/block/mana/structure.rs
```diff
@@ -65,3 +108,109 @@ impl Default for ManaStructure {
...
+            let potential_mana_0 = c + mana_structure.generate_mana(amount, slots_since_epoch_start.0 as u32)
+                - (c >> mana_structure.generation_rate_exponent());
+            potential_mana_0 - potential_mana_n_1 + potential_mana_n
+        }
+    }
+}
+
+/// Returns the upper 32 bits of a u64 value.
+fn upper_bits(v: u64) -> u64 {
+    v >> 32
```
Alex6323, line 208: hmm ... isn't the correct computation: `(v >> 32) & u32::MAX as u64`?

DaughterOfMars: doesn't make any difference?

sdk/src/types/block/slot/epoch.rs
Alex6323: I think "Creates the epoch index given ..." would be SLIGHTLY better since it creates a `Self` and `Gets` usually indicates a getter, but I don't care that much. Just a small detail I couldn't unsee anymore :smile:

sdk/src/types/block/mana/structure.rs
Alex6323: sorry, you're right ... it doesn't make a difference for unsigned.

sdk/src/types/block/mana/structure.rs
```diff
@@ -65,3 +108,106 @@ impl Default for ManaStructure {
...
+}
+
+/// Returns the lower n bits of a u64 value.
+fn lower_n_bits(v: u64, n: u8) -> u64 {
+    debug_assert!(n <= 64);
```
Alex6323, line 213:
```suggestion
    debug_assert!(n > 0 && n <= 64);
```

sdk/src/types/block/slot/index.rs
Alex6323: not from me though :joy: ...it's from what's already there, but okay

kwek20 (dismissed):

sdk/src/types/block/mana/structure.rs
kwek20: Yea this is needed, ran into it on my local test with prot_params::default(), we gotta get better values for manastruct

DaughterOfMars: I don't think this is what we want. it's possible to get the lowest 0 bits which is always 0

Alex6323 (dismissed):

Thoralf-M (approved):

Alex6323 (approved):

PR #1382, opened 2023-10-03
Nodejs: Add restricted and implicit address types

### LINKED ISSUE #1221: Add new address types - Nodejs
https://github.com/iotaledger/tips-draft/pull/26

DaughterOfMars:
# Description of change

Adds the new address types from TIP 50 to nodejs. Does not add any functionality for the `Capabilities` flags.

Closes #1221


thibault-martinez (changes requested):
Failing CI

thibault-martinez (changes requested):
Still failing

bindings/nodejs/lib/types/block/address.ts
```diff
@@ -118,12 +133,81 @@ class NftAddress extends Address {
...
+        allowed_capabilities: Uint8Array,
+    ): RestrictedAddress {
+        this.allowed_capabilities = allowed_capabilities;
+        return this;
+    }
+
+    toString(): string {
+        let hex = this.address.toString();
+        if (this.allowed_capabilities.some((c) => c != 0)) {
+            let cap = Buffer.from(
```
Thoralf-M:
```suggestion
            const cap = Buffer.from(
```

bindings/nodejs/lib/types/block/address.ts
```diff
@@ -118,12 +133,81 @@ class NftAddress extends Address {
...
+ */
+class RestrictedAddress extends Address {
+    /**
+     * The inner address.
+     */
+    readonly address: Address;
+    /**
+     * The allowed capabilities bitflags.
+     */
+    allowed_capabilities: Uint8Array = new Uint8Array();
```
Thoralf-M: I guess this will be a prefix hex string, if not there needs to be at least some conversion if we deserialize the type?

thibault-martinez: According to Phillip, this will be an array of indexes when serialised

thibault-martinez: Sorry he just corrected himself. It is indeed hexstring.
Here is an example `{"type":40,"address":{"type":0,"pubKeyHash":"0x4f172a5c322267154e25256a0c02702376107c3d593e7b67485b45706e4b752e"},"allowedCapabilities":"0x0203"}`

thibault-martinez: Just ignore my deleted comment, I'm tired and misread

bindings/nodejs/lib/types/block/address.ts
```diff
@@ -118,12 +133,89 @@ class NftAddress extends Address {
...
+}
+
+/**
+ * An address with restricted capabilities.
+ */
+class RestrictedAddress extends Address {
+    /**
+     * The inner address.
+     */
+    readonly address: Address;
```
kwek20, line 177: this needs to be parsed again but will be a recursive issue. can look at the regulartransactionessence payload to see how we did it

DaughterOfMars: That was only a problem because the type was defined separately tho, you sure it's a problem here?

thibault-martinez: @kwek20

bindings/nodejs/lib/types/block/address.ts
```diff
@@ -118,12 +133,89 @@ class NftAddress extends Address {
...
+    /**
+     * The allowed capabilities bitflags.
+     */
+    private allowed_capabilities: HexEncodedString = '0x00';
+    /**
+     * @param address An address.
+     */
+    constructor(address: Address) {
+        super(AddressType.Nft);
```
Alex6323:
```suggestion
        super(AddressType.Restricted);
```

bindings/nodejs/lib/types/block/address.ts
```diff
@@ -118,12 +133,89 @@ class NftAddress extends Address {
...
+ */
+class RestrictedAddress extends Address {
+    /**
+     * The inner address.
+     */
+    readonly address: Address;
+    /**
+     * The allowed capabilities bitflags.
+     */
+    private allowed_capabilities: HexEncodedString = '0x00';
```
Alex6323: can we not use either `Uint8Array` or a bitmask (Uint16 or Uint32?)  internally and only convert to string when necessary?

bindings/nodejs/lib/types/block/address.ts
DaughterOfMars: How did I miss this 😆

bindings/nodejs/lib/types/block/address.ts
DaughterOfMars: The problem is that I think the serialization just takes the class as a dictionary and it doesn't have additional logic afaik

bindings/nodejs/lib/types/block/address.ts
kwek20: well it still needs the annotation, and not sure its a problem no

kwek20: im referring to
```
@Type(() => Address, {
        discriminator: AddressDiscriminator,
    })
'''

**Review by Thoralf-M — DISMISSED** [review]

File: bindings/nodejs/lib/types/block/address.ts (hunk 3 above)
Comment (DaughterOfMars) [reply bindings/nodejs/lib/types/block/address.ts:177]: Since the inner address cannot be every address type anyway, I just created a new discriminator for it

**Review by Thoralf-M — DISMISSED** [review]

File: bindings/nodejs/lib/types/block/address.ts [hunk 6]
```diff
@@ -118,12 +133,105 @@ class NftAddress extends Address {
...
+}
+
+const RestrictedAddressDiscriminator = {
+    property: 'type',
+    subTypes: [
+        { value: Ed25519Address, name: AddressType.Ed25519 as any },
+        { value: AccountAddress, name: AddressType.Account as any },
+        { value: NftAddress, name: AddressType.Nft as any },
+        {
+            value: ImplicitAccountCreationAddress,
```
thibault-martinez: Restricted can't be ImplicitAccountCreation

thibault-martinez (changes requested):

bindings/nodejs/lib/types/block/address.ts
DaughterOfMars: oh ok

bindings/nodejs/lib/types/block/address.ts
```diff
@@ -118,12 +133,101 @@ class NftAddress extends Address {
...
 }
 
+/**
+ * An implicit account creation address.
+ */
+class ImplicitAccountCreationAddress extends Address {
+    /**
+     * The Ed25519 address.
+     */
+    readonly address: Ed25519Address;
```
thibault-martinez: This is fine in Rust because we have a DTO but I'm not sure it's fine in Js? The serialisation of that is supposed to be `pubKeyHash`, wouldn't this nest it in an address field?

thibault-martinez: Oh, I just saw the toString sorry 🤦🏻

thibault-martinez: Wait, will this actually be used for JSON ?

bindings/nodejs/lib/types/block/address.ts
```diff
@@ -118,12 +133,101 @@ class NftAddress extends Address {
...
+     * The inner address.
+     */
+    @Type(() => Address, {
+        discriminator: RestrictedAddressDiscriminator,
+    })
+    readonly address: Address;
+    /**
+     * The allowed capabilities bitflags.
+     */
+    private allowed_capabilities: HexEncodedString = '0x00';
```
thibault-martinez: camelCase?

bindings/nodejs/lib/types/block/address.ts
DaughterOfMars: @kwek20 ?

bindings/nodejs/lib/types/block/address.ts
DaughterOfMars: oopsie

bindings/nodejs/lib/types/block/address.ts
kwek20: well theres no dicts in js, but it just calls .toJSON on each object. We can implement a custom one and itll be picked up.
There is nothing for parsing this though, making it difficult to actually use 2 way 😆

bindings/nodejs/lib/types/block/address.ts
kwek20: youre missing @Type(() => Ed25519Address)

bindings/nodejs/lib/types/block/address.ts
kwek20: (so it say not worth the effort)

bindings/nodejs/lib/types/block/address.ts
thibault-martinez: That would handle the JSOn serialisation as if it was a Ed25519Address and not contained one?

DaughterOfMars: fixed it. FYI this is because the `class-transformer` lib requires nested classes to specify their types for...some reason (it's because javascript is a terrible language)

thibault-martinez: I'd still like to have a confirmation from @kwek20 that this will skip the `address` key itself

DaughterOfMars: oh, no, that's not what it will do. I misunderstood the issue

thibault-martinez: Either we find a serialisation trick, or we don't use a `Ed25519Address` but directly the inner field

kwek20: I vote for private inner field and a getter to get the Ed25519Address

bindings/nodejs/lib/types/block/address.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright 2023 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-import { plainToInstance } from 'class-transformer';
+import { plainToInstance, Type, Transform, Expose } from 'class-transformer';
```
thibault-martinez: 'Expose' is defined but never used

bindings/nodejs/lib/types/block/address.ts
```diff
@@ -118,12 +133,107 @@ class NftAddress extends Address {
...
+/**
+ * An implicit account creation address.
+ */
+class ImplicitAccountCreationAddress extends Address {
+    /**
+     * The Ed25519 address.
+     */
+    @Transform(({ value }) => new Ed25519Address(value), { toClassOnly: true })
+    @Transform(({ value }) => value.toString(), { toPlainOnly: true })
+    private pubKeyHash: Ed25519Address;
```
thibault-martinez: Tbh at this point I think I agree with Brord, I'd have a HexEncodedString field and a Ed25519Address field.
This is just too convoluted for nothing.

bindings/nodejs/lib/types/block/address.ts
```diff
@@ -118,12 +133,105 @@ class NftAddress extends Address {
     }
 }
 
+/**
+ * An implicit account creation address.
+ */
+class ImplicitAccountCreationAddress extends Address {
+    /**
+     * The Ed25519 address.
```
thibault-martinez: Slightly wrong comment

bindings/nodejs/lib/types/block/address.ts
```diff
@@ -118,12 +133,105 @@ class NftAddress extends Address {
...
+
+    withAllowedCapabilities(
+        allowedCapabilities: Uint8Array,
+    ): RestrictedAddress {
+        this.setAllowedCapabilities(allowedCapabilities);
+        return this;
+    }
+
+    getAllowedCapabilities(): Uint8Array {
+        return Uint8Array.from(Buffer.from(this.allowedCapabilities, 'hex'));
```
thibault-martinez: Does this handle the leading 0x ?

bindings/nodejs/lib/types/block/address.ts
```diff
@@ -118,12 +133,105 @@ class NftAddress extends Address {
...
+    }
+
+    toString(): string {
+        return this.address.toString() + this.allowedCapabilities.substring(2);
```
thibault-martinez, line 219: Just making sure, this is never used by any JSON utility, right?

bindings/nodejs/lib/types/block/address.ts
DaughterOfMars: dang it

bindings/nodejs/lib/types/block/address.ts
DaughterOfMars: I don't think so, but not sure

thibault-martinez (approved):

Alex6323 (approved):

Issue #741, opened 2023-07-10
Add WriteStream, ReadStream, Converter and HexHelper


cpl121:
## Description

Currently `WriteStream`, `ReadStream`, `Converter` and `HexHelper` iota.js dependencies are being used in `Firefly` and they need to be added in the sdk

Also, the encodings have been modified to make transactions between layer 1 and layer 2, as can be seen in the [evm-toolkit](https://github.com/iotaledger/evm-toolkit/compare/0f1fa3d4aa973bde0212dd408ab1f75d9fac84d4...main).
To fit these changes, a new type `SpecialStream` has been created in [Firefly](https://github.com/iotaledger/firefly/blob/develop/packages/shared/lib/core/layer-2/classes/special-stream.class.ts#L5). It would be necessary to add this class together with `size64Encode`. 
On the other hand, although it is still in development, it would also be necessary to add the [SpecialStream](https://github.com/iotaledger/firefly/blob/fix/L2-transaction-metadata-decoding/packages/shared/lib/core/layer-2/classes/special-stream.class.ts#L17) to be able to `read` these new transactions, as well as their [decode](https://github.com/iotaledger/firefly/blob/fix/L2-transaction-metadata-decoding/packages/shared/lib/core/layer-2/classes/special-stream.class.ts#L130)

## Motivation

The iota.js dependencies used in Firefly need to be incorporated into the sdk, and the way to `encode/decode` transactions from layer 1 to layer 2 has been updated, thus the `WriteSpecialStream/ReadSpecialStream` classes needs to be added too

## Requirements

1. Add `WriteStream` and `WriteSpecialStream`
2. Add `ReadStream` and `ReadSpecialStream`
3. Add `Converter`
4. Add `HexHelper`

## Are you planning to do it yourself in a pull request?

No.

kwek20:
Hi @cpl121 is this still needed? If so, why? Adding a whole bunch of "custom" serialization really defeats the purpose of bindings and single source of truth. Is there a way we can change this so you dont need any of the streams?

As for Converter and HexHelper, we should have added all the missing utilities to our utils package. Let me know specifically what method you are missing and ill add it.

kwek20:
Meeting notes:
**HexHelper**: Not used
**Converter**: Methods have been moved to utils
[...]

PR #1390, opened 2023-10-04
feat/node: Add hex-byte util fns

kwek20:
# Description of change

Adds 2 ways to do hex-bytes and bytes-hex

Vote using emojis 
🚀 - Converter
🎉 - Plain JS

partially addresses #741


bindings/nodejs/lib/utils/converter.ts
```diff
@@ -0,0 +1,104 @@
...
+     * @param array The bytes to encode.
+     * @param includePrefix Include the 0x prefix on the returned hex.
+     * @param startIndex The index to start in the bytes.
+     * @param length The length of bytes to read.
+     * @param reverse Reverse the combine direction.
+     * @returns The array formated as hex.
+     */
+    public static bytesToHex(
+        array: ArrayLike<number>,
+        includePrefix: boolean = false,
```
thibault-martinez: sad CI noise

bindings/nodejs/lib/utils/index.ts
```diff
@@ -4,3 +4,34 @@
...
 export * from '../types/utils';
+
+/**
+ * Converts a byte array to a hexadecimal string.
+ *
+ * @param {Uint8Array} byteArray - The bytes to encode.
+ * @param {boolean} [prefix=false] - Whether to include the '0x' prefix in the resulting hexadecimal string.
+ * @returns {string} The hexadecimal representation of the input byte array.
+ */
+export const bytesToHex = (bytes: ArrayLike<number>, prefix = false) => {
```
Thoralf-M, line 15:
```suggestion
export const bytesToHex = (bytes: ArrayLike<number>, prefix = true) => {
```
Shouldn't the default be true? Since we always have it with prefix in the types?

kwek20: id like to, but firefly has it like this in the converter, so would be nicer for them to not have to change things right?

kwek20: arguably, hexToBytes removes the prefix, so the expected bytestohex input also has no prefix

bindings/nodejs/lib/utils/index.ts
```diff
@@ -4,3 +4,34 @@
...
+
+/**
+ * Converts a hexadecimal string to a Uint8Array byte array.
+ *
+ * @param {string} hexString - The hexadecimal string to be converted.
+ * @returns {Uint8Array} The Uint8Array byte array representation of the input hexadecimal string.
+ * @throws {Error} Will throw an error if the input string is not a valid hexadecimal string.
+ */
+export const hexToBytes = (hexString: string) => {
+    const hex = hexString.replace(/^0x/, '');
```
DaughterOfMars, line 31: probably better to use regex right?

DaughterOfMars: oh, this is regex lol nvm

DaughterOfMars (approved):
ooh this will clean up the capability code

bindings/nodejs/lib/utils/index.ts
kwek20: this is regex!

bindings/nodejs/lib/utils/utf8.ts
```diff
@@ -11,7 +11,7 @@ export const utf8ToBytes = (utf8: string) => {
 
 /** Convert hex encoded string to UTF8 string */
 export const hexToUtf8 = (hex: HexEncodedString) =>
-    decodeURIComponent(hex.replace('0x', '').replace(/[0-9a-f]{2}/g, '%$&'));
+    decodeURIComponent(hex.replace(/^0x/, '').replace(/[0-9a-f]{2}/g, '%$&'));
```
thibault-martinez, line 14: Just curious, does // mean regex in TS ? Without quote or anything?

thibault-martinez (approved):

bindings/nodejs/lib/utils/utf8.ts
kwek20: i cannot find out :D seems to not be, i think it needs some info in between, but "///" is also taken as a string

thibault-martinez: But this definitely works, right? No quote just seems super wild to me

kwek20: yep 100% wild stuff with nodejs guaranteed :D

kwek20: I doublechecked for ya, and this will turn into a RegExp type
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp

thibault-martinez: Thanks

Thoralf-M (approved):

PR #1394, opened 2023-10-04
Update python README.md

fijter:
Small fixes to the README for the Python bindings.

Do we still offer the `pip` package pre-built for this? If so we should probably make that part of the instructions to, saves a bunch of extra steps for most users that might just want to use it.


Alex6323 (approved):

Thoralf-M (approved):
Yes we upload wheels to pip, opened an issue to add it https://github.com/iotaledger/iota-sdk/issues/1396

PR #1395, opened 2023-10-04
Use `json` instead of `serde_json`

### LINKED ISSUE #1372: Consider switching from `serde_json` to `json`
## Description

Out current usage of `serde` is highly specific to JSON, and the lib generates a huge amount of code to support `serde`s flexibility. We could instead use [`json`](https://docs.rs/json/latest/json/) or a similar crate which is a lot more bare-bones. Keeping `serde` impls may still be a good idea for the community use, but would be much more generic, without renames and `with` impls.

DaughterOfMars:
# Description of change

This PR is a demonstration of a potential alternative serialization impl for json specifically.

If we proceed with the changes across all types, we would
1. remove specialized `serde` impls and simply derive for all types (perhaps with the exception of the IDs, which make sense as strings)
2. Remove `serde` feature from the core requirements
3. Remove all Dto types in favor of the `json!` macro and `ToJson`/`FromJson` impls

#1372

# Stats

The below are on `2.0`, NOT this branch

| profile | features | time | lib size |
| --- | --- | --- | --- |
| debug |  | 9.77s | 18.5 MB |
| debug | serde | 11.67s | 30.1 MB |
| release |  | 10.52s | 6 MB |
| release | serde | 13.15s | 13.2 MB |

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1610,6 +1610,7 @@ dependencies = [
  "iota-sdk",
  "iota_stronghold",
  "iterator-sorted",
+ "json",
  "lazy_static",
  "log",
  "num_cpus",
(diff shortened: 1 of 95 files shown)


DaughterOfMars:
Closing this as I think the effort is not worth the result.

thibault-martinez:
Not worth doing it this way or not worth moving out of serde at all? Depending on the answer, you may want to close the issue as well.

DaughterOfMars:
In order to make this viable, we would need to write our own procedural macro, as it's simply too much to create manual impls for every type that needs one. Unfortunately, the problems with using serde this way still remain an issue, but I don't see a simple solution for it right now.

PR #1398, opened 2023-10-05
refactor(types): vbyte -> storage score

### LINKED ISSUE #1352: Rename RentStructure to StorageScoreParameters
And update the fields https://github.com/iotaledger/tips/blob/973334ae4657974dff08385d2cb912e945736a37/tips/TIP-0048/openapi3-core.yaml#L2688

- [x] Rust https://github.com/iotaledger/iota-sdk/pull/1065
- [x] Python https://github.com/iotaledger/iota-sdk/pull/1673
- [x] NodeJs https://github.com/iotaledger/iota-sdk/pull/1673

Alex6323:
# Description of change

* Renames (and refactors) the `Rent` trait to `StorageScore` and provides 2 fns: `storage_score` and `rent_cost`.
* Renames `RentStructure` to `RentParameters` and changes its semantics to offsets.
* Introduces the new `RentStructure` type which wraps a `RentParameters` and something unmemorable but has to do with the implicit account

Closes #1352 

NOTE:
I turned this into a draft again, because the TIP isn't complete yet. So far this PR has implemented what was merged into `iota.go` here: https://github.com/iotaledger/iota.go/pull/537 (except for the closures that can be provided on some storage score implementors)


sdk/src/types/block/output/rent.rs
```diff
@@ -31,145 +36,153 @@ const DEFAULT_BYTE_COST_FACTOR_BLOCK_ISSUER_KEY: u8 = 1;
     serde(rename_all = "camelCase")
 )]
 pub struct RentStructure {
```
thibault-martinez: This has been renamed to RentParameters.
https://github.com/iotaledger/tips-draft/pull/25/files#diff-bf3729cb12f375685fc23f726a6be49461ce3df320037f6639b323dabfebf651R476
https://github.com/iotaledger/iota.go/blob/serix-iota-core/rent.go#L36

bindings/core/src/method_handler/utils.rs
```diff
@@ -8,8 +8,9 @@ use iota_sdk::{
         block::{
             address::{AccountAddress, Address, ToBech32Ext},
             input::UtxoInput,
-            output::{AccountId, FoundryId, InputsCommitment, NftId, Output, OutputId, Rent, TokenId},
+            output::{AccountId, FoundryId, InputsCommitment, NftId, Output, OutputId, TokenId},
             payload::{transaction::TransactionEssence, TransactionPayload},
+            rent::StorageScore,
```
DaughterOfMars: Maybe rename the module too?

cli/src/command/account.rs
```diff
@@ -721,12 +721,12 @@ pub async fn send_native_token_command(
     let address = address.convert()?;
     let transaction = if gift_storage_deposit.unwrap_or(false) {
         // Send native tokens together with the required storage deposit
-        let rent_structure = account.client().get_rent_structure().await?;
+        let rent_struct = account.client().get_rent_parameters().await?.into();
```
DaughterOfMars: rent_params?

Alex6323: no, it's converted `into` a `RentStructure`

DaughterOfMars: but why

Alex6323: downstream effect ... `RentStructure` (new) wraps the `RentParameters` (old RentStructure) and a new field regarding implicit account creation. Unfortunately it is not reflected in the TIP yet, but here you can see that's it's going to be:
https://github.com/iotaledger/iota.go/pull/537/files#diff-74ba3861618d2996e83bd87e90b7259060214740d314c9d10cc0ae99c47df9c5R51-R55

This new field must be computed at runtime to create the new `RentStructure` instance, and this is what we have to use in the trait to calculate the score and hence the rent cost correctly.

DaughterOfMars: Since it's calculated at runtime, why not just make it a method on `RentParameters`?

bindings/core/src/method_handler/utils.rs
Alex6323: Done

cli/src/command/account.rs
thibault-martinez: I think I would do that, not a big fan of the wrapper

thibault-martinez: (this will never make it into TIP btw, this is implementation detail)

Alex6323: Aight, but even if it won't become part of the TIP, from a technical PoV I think it's still a good idea to calculate it only once at runtime (when the rent parameters are fetched from the node) and keep it buffered somewhere. I think it's fine to use a wrapper struct in order to keep `RentParameters` identical with the TIP. This is the path forward they chose in `iota.go`. One can go a different route, but you have to/should buffer that computed score, and not re-compute it unnecessarily.

DaughterOfMars: I think there's probably a better way to cache the value, if we need to

Alex6323: Maybe, but since there's no immediately obvious way I would suggest to create an issue and someone from the team investigates a different approach in another/next sprint. I am strongly against caching the value globally or making everything overly complex just to prevent the wrapper. Wrappers are bread and butter in Rust so I don't really understand the discomfort with that :shrug:

DaughterOfMars: I think there's no immediately obvious reason to cache this tbh

Alex6323: hmm, I guess questioning that is fair! I'm changing it ...

thibault-martinez:
Closing in favour of https://github.com/iotaledger/iota-sdk/pull/1065

PR #1426, opened 2023-10-12
Print NFT, Alias, Foundry alias addresses

### LINKED ISSUE #1320: CLI: Add NftAddress and AliasAddress to `addresses` command
It currently only shows Ed25519 addresses but we could show the NFT/Alias ones and what they own.

Alex6323:
# Description of change

* Changes the `addresses` command to just print all account addresses as an indexed list (in a similar fashion as the outputs and transactions lists)
* Introduces a new `address <selector>` command which either takes an index of an address (as indexed by the `addresses` command) or accepts a `Bech32Address` and then displays its details (i.e. outputs, summed output amount, owned assets like NTs, NFTs, and Aliases)

Closes #1320

### DIFF
--- a/cli/src/command/account.rs
+++ b/cli/src/command/account.rs
@@ -958,7 +958,6 @@ pub async fn voting_output_command(account: &Account) -> Result<(), Error> {
 }
 
 async fn print_address(account: &Account, address: &Bech32Address) -> Result<(), Error> {
-    println_log_info!("Address");
     println_log_info!("{:<11}{}", "Bech32:", address);
     println_log_info!("{:<11}{}", "Hex:", address.inner());
     println_log_info!("{:<11}{}", "Type:", address.inner().kind_str());
@@ -1019,22 +1018,21 @@ async fn print_address(account: &Account, address: &Bech32Address) -> Result<(),
         }
     }
 
-    // base tokens table
-    println_log_info!("Base Tokens");
-    println_log_info!("{:<8}{}", "Amount:", amount);
+    // Coins table
+    println_log_info!("{:<11}{}", "Coins:", amount);
 
-    // outputs table
+    // Outputs table
     if !outputs.is_empty() {
-        println_log_info!("Outputs");
+        println_log_info!("Outputs:");
         for (id, kind) in outputs {
-            println_log_info!("  {kind}\t{id}");
+            println_log_info!("  {id}\t{kind}");
         }
     }
 
-    // native tokens table
+    // Native tokens table
     let native_tokens = native_tokens.finish_vec()?;
     if !native_tokens.is_empty() {
-        println_log_info!("Native Tokens");
+        println_log_info!("Native Tokens:");
         for (id, amount) in native_tokens
             .into_iter()
             .map(|nt| (*nt.token_id(), nt.amount().to_string()))
@@ -1046,7 +1044,7 @@ async fn print_address(account: &Account, address: &Bech32Address) -> Result<(),
 
     // NFT table
     if !nfts.is_empty() {
-        println_log_info!("NFTs");
+        println_log_info!("NFTs:");
         for id in nfts.into_iter() {
             println_log_info!("  {id}");
         }
@@ -1054,7 +1052,7 @@ async fn print_address(account: &Account, address: &Bech32Address) -> Result<(),
 
     // Aliases table
     if !aliases.is_empty() {
-        println_log_info!("Aliases");
+        println_log_info!("Aliases:");
         for id in aliases.into_iter() {
             println_log_info!("  {id}");
         }
--- a/cli/src/command/account.rs
+++ b/cli/src/command/account.rs
@@ -1116,7 +1116,7 @@ async fn get_addresses_sorted(account: &Account) -> Result<Vec<Bech32Address>, E
         .addresses()
         .await?
         .iter()
-        .map(|address| address.address().clone())
+        .map(|address| *address.address())
         .chain(
             account
                 .unspent_outputs(FilterOptions {
--- a/cli/src/command/account.rs
+++ b/cli/src/command/account.rs
@@ -957,20 +957,6 @@ pub async fn voting_output_command(account: &Account) -> Result<(), Error> {
     Ok(())
 }
 
-fn print_addresses(mut addresses: Vec<Bech32Address>) -> Result<(), Error> {
-    if addresses.is_empty() {
-        println_log_info!("No addresses found");
-    } else {
-        addresses.sort_unstable();
-
-        for (i, addr) in addresses.into_iter().enumerate() {
-            println_log_info!("{:<5}{}\t{}", i, addr, addr.kind_str());
-        }
-    }
-
-    Ok(())
-}
-
 async fn print_address(account: &Account, address: &Bech32Address) -> Result<(), Error> {
     let mut formatted_table = String::from("Address");
     formatted_table.push_str(&format!(
@@ -988,17 +974,15 @@ async fn print_address(account: &Account, address: &Bech32Address) -> Result<(),
         .unspent_outputs(None)
         .await?
         .into_iter()
-        .filter(|data| match &data.output {
-            Output::Basic(basic) => basic.address() == address.inner(),
-            Output::Nft(nft) => {
-                &Address::Nft(nft.nft_address(&data.output_id)) == address.inner() || &data.address == address.inner()
-            }
-            Output::Alias(alias) => {
-                &Address::Alias(alias.alias_address(&data.output_id)) == address.inner()
-                    || &data.address == address.inner()
-            }
-            Output::Foundry(foundry) => &Address::Alias(foundry.alias_address().clone()) == address.inner(),
-            _ => false,
+        .filter(|data| {
+            &data.address == address.inner()
+                || match &data.output {
+                    Output::Basic(basic) => basic.address() == address.inner(),
+                    Output::Nft(nft) => &Address::Nft(nft.nft_address(&data.output_id)) == address.inner(),
+                    Output::Alias(alias) => &Address::Alias(alias.alias_address(&data.output_id)) == address.inner(),
+                    Output::Foundry(foundry) => &Address::Alias(foundry.alias_address().clone()) == address.inner(),
+                    _ => false,
+                }
         })
         .collect::<Vec<_>>();
 
@@ -1083,6 +1067,20 @@ async fn print_address(account: &Account, address: &Bech32Address) -> Result<(),
     }
 
     println_log_info!("{formatted_table}");
+
+    Ok(())
+}
+
+fn print_addresses(mut addresses: Vec<Bech32Address>) -> Result<(), Error> {
+    if addresses.is_empty() {
+        println_log_info!("No addresses found");
+    } else {
+        addresses.sort_unstable();
+
+        for (i, addr) in addresses.into_iter().enumerate() {
+            println_log_info!("{:<5}{}\t{}", i, addr, addr.kind_str());
+        }
+    }
 
     Ok(())
 }
--- a/cli/src/command/account.rs
+++ b/cli/src/command/account.rs
@@ -957,9 +957,11 @@ pub async fn voting_output_command(account: &Account) -> Result<(), Error> {
 }
 
 async fn print_address(account: &Account, address: &Bech32Address) -> Result<(), Error> {
-    println_log_info!("{:<11}{}", "Bech32:", address);
-    println_log_info!("{:<11}{}", "Hex:", address.inner());
-    println_log_info!("{:<11}{}", "Type:", address.inner().kind_str());
+    let mut formatted_string = String::new();
+
+    formatted_string.push_str(&format!("{:<11}{}\n", "Bech32:", address));
+    formatted_string.push_str(&format!("{:<11}{}\n", "Hex:", address.inner()));
+    formatted_string.push_str(&format!("{:<11}{}\n", "Type:", address.inner().kind_str()));
 
     let unspent_outputs = account.unspent_outputs(None).await?;
     let current_time = iota_sdk::utils::unix_timestamp_now().as_secs() as u32;
@@ -1004,43 +1006,45 @@ async fn print_address(account: &Account, address: &Bech32Address) -> Result<(),
     }
 
     // Coins table
-    println_log_info!("{:<11}{}", "Coins:", amount);
+    formatted_string.push_str(&format!("{:<11}{}\n", "Coins:", amount));
 
     // Outputs table
     if !outputs.is_empty() {
-        println_log_info!("Outputs:");
+        formatted_string.push_str("Outputs:\n");
         for (id, kind) in outputs {
-            println_log_info!("  {id}\t{kind}");
+            formatted_string.push_str(&format!("  {id}\t{kind}\n"));
         }
     }
 
     // Native tokens table
     let native_tokens = native_tokens.finish_vec()?;
     if !native_tokens.is_empty() {
-        println_log_info!("Native Tokens:");
+        formatted_string.push_str("Native Tokens:\n");
         for (id, amount) in native_tokens
             .into_iter()
             .map(|nt| (*nt.token_id(), nt.amount().to_string()))
         {
-            println_log_info!("  {id}\t{amount}");
+            formatted_string.push_str(&format!("  {id}\t{amount}\n"));
         }
     }
 
     // NFT table
     if !nfts.is_empty() {
-        println_log_info!("NFTs:");
+        formatted_string.push_str("NFTs:\n");
         for id in nfts.into_iter() {
-            println_log_info!("  {id}");
+            formatted_string.push_str(&format!("  {id}\n"));
         }
     }
 
     // Aliases table
     if !aliases.is_empty() {
-        println_log_info!("Aliases:");
+        formatted_string.push_str("Aliases:\n");
         for id in aliases.into_iter() {
-            println_log_info!("  {id}");
+            formatted_string.push_str(&format!("  {id}\n"));
         }
     }
+
+    println_log_info!("{formatted_string}");
 
     Ok(())
 }
--- a/cli/src/command/account.rs
+++ b/cli/src/command/account.rs
@@ -971,8 +971,6 @@ async fn print_address(account: &Account, address: &Bech32Address) -> Result<(),
     let mut aliases = Vec::new();
 
     for (output_id, output_data) in unspent_outputs.into_iter().map(|data| (data.output_id, data)) {
-        outputs.push((output_id, output_data.output.kind_str().to_string()));
-
         // Panic: cannot fail for outputs belonging to an account.
         let (required_address, _) = output_data
             .output
@@ -980,14 +978,18 @@ async fn print_address(account: &Account, address: &Bech32Address) -> Result<(),
             .unwrap();
 
         if address.inner() == &required_address {
+            outputs.push((output_id, output_data.output.kind_str().to_string()));
+
             if let Some(nts) = output_data.output.native_tokens() {
                 native_tokens.add_native_tokens(nts.clone())?;
             }
+
             match &output_data.output {
                 Output::Alias(alias) => aliases.push(alias.alias_id_non_null(&output_id)),
                 Output::Nft(nft) => nfts.push(nft.nft_id_non_null(&output_id)),
                 Output::Basic(_) | Output::Foundry(_) | Output::Treasury(_) => {}
             }
+
             let sdr_amount = output_data
                 .output
                 .unlock_conditions()
--- a/cli/src/command/account.rs
+++ b/cli/src/command/account.rs
@@ -21,7 +21,7 @@ use iota_sdk::{
     wallet::{
         account::{
             types::{AccountIdentifier, OutputData, Transaction},
-            Account, ConsolidationParams, FilterOptions, OutputsToClaim, SyncOptions, TransactionOptions,
+            Account, ConsolidationParams, OutputsToClaim, SyncOptions, TransactionOptions,
         },
         CreateNativeTokenParams, MintNftParams, SendNativeTokensParams, SendNftParams, SendParams,
     },
@@ -963,11 +963,9 @@ fn print_addresses(mut addresses: Vec<Bech32Address>) -> Result<(), Error> {
     } else {
         addresses.sort_unstable();
 
-        let mut formatted_table = String::new();
         for (i, addr) in addresses.into_iter().enumerate() {
-            formatted_table.push_str(&format!("{:<5}{}\t{}\n", i, addr, addr.kind_str()));
+            println_log_info!("{:<5}{}\t{}", i, addr, addr.kind_str());
         }
-        println_log_info!("{formatted_table}");
     }
 
     Ok(())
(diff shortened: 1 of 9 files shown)


Thoralf-M (dismissed):

sdk/src/types/block/address/mod.rs
```diff
@@ -48,6 +48,16 @@ impl core::fmt::Debug for Address {
     }
 }
 
+impl core::fmt::Display for Address {
```
DaughterOfMars: derive_more?

DaughterOfMars:
What's the output look like?

thibault-martinez (changes requested):
Changelog

thibault-martinez:
Should we also list addresses if they don't own anything yet?

sdk/src/types/block/address/mod.rs
thibault-martinez: Done

DaughterOfMars (dismissed):

sdk/src/types/block/address/mod.rs
Alex6323: thanks

thibault-martinez:
Btw, there is no such thing as a Foundry address

Alex6323:
yeah, wasn't sure bc there's `alias_address` fn on it.

Alex6323:
@DaughterOfMars 
```sh
Account "alex": addresses
Address: 0
Bech32:  rms1qqtj7pvnl3lj9n9n6e9lc47mfutjfhjyprmprxtzz2g0uck8tr3gurtp7tq
Hex:     0x172f0593fc7f22ccb3d64bfc57db4f1724de4408f61199621290fe62c758e28e
Type:    Ed25519
Outputs: [
    OutputId(0xc2dd37d77efea83d65e2aad295751f27e20e3b3d17a5e99dffb24d27311afdf20000),
    OutputId(0xd349d87ed19d5a282fb2675ae6702c0c9f9eb83de166b16c23e4db6ab038bae70000),
    OutputId(0xd349d87ed19d5a282fb2675ae6702c0c9f9eb83de166b16c23e4db6ab038bae70100),
    OutputId(0xc2dd37d77efea83d65e2aad295751f27e20e3b3d17a5e99dffb24d27311afdf20100),
]
Base coin amount: 999947200
Native Tokens: [
    NativeToken {
        token_id: TokenId(0x085bb79eed2a046a8388d47b713bebc1ae02860d5356d366d315a93d9845b28b830100000000),
        amount: 100,
    },
]
NFTs: [
    (
        NftId(0x5c17e3651a787be20c4d09640c2a883f0842ea1afb28790cdea402e7370ce5b8),
        Bech32Address(rms1zpwp0cm9rfu8hcsvf5ykgrp23qlsssh2rtajs7gvm6jq9eehpnjmst5kf04),
    ),
]
Aliases: [
    (
        AliasId(0x5bb79eed2a046a8388d47b713bebc1ae02860d5356d366d315a93d9845b28b83),
        Bech32Address(rms1ppdm08hd9gzx4qug63ahzwltcxhq9psd2dtdxeknzk5nmxz9k29cxy8aseq),
    ),
]
Foundries: []
```
that would be the output atm.

thibault-martinez:
> @DaughterOfMars
> 
> ```shell
> Account "alex": addresses
> Address: 0
> Bech32:  rms1qqtj7pvnl3lj9n9n6e9lc47mfutjfhjyprmprxtzz2g0uck8tr3gurtp7tq
> Hex:     0x172f0593fc7f22ccb3d64bfc57db4f1724de4408f61199621290fe62c758e28e
> Type:    Ed25519
> Outputs: [
>     OutputId(0xc2dd37d77efea83d65e2aad295751f27e20e3b3d17a5e99dffb24d27311afdf20000),
>     OutputId(0xd349d87ed19d5a282fb2675ae6702c0c9f9eb83de166b16c23e4db6ab038bae70000),
>     OutputId(0xd349d87ed19d5a282fb2675ae6702c0c9f9eb83de166b16c23e4db6ab038bae70100),
>     OutputId(0xc2dd37d77efea83d65e2aad295751f27e20e3b3d17a5e99dffb24d27311afdf20100),
> ]
> Base coin amount: 999947200
> Native Tokens: [
>     NativeToken {
>         token_id: TokenId(0x085bb79eed2a046a8388d47b713bebc1ae02860d5356d366d315a93d9845b28b830100000000),
>         amount: 100,
>     },
> ]
> NFTs: [
>     (
>         NftId(0x5c17e3651a787be20c4d09640c2a883f0842ea1afb28790cdea402e7370ce5b8),
>         Bech32Address(rms1zpwp0cm9rfu8hcsvf5ykgrp23qlsssh2rtajs7gvm6jq9eehpnjmst5kf04),
>     ),
> ]
> Aliases: [
>     (
>         AliasId(0x5bb79eed2a046a8388d47b713bebc1ae02860d5356d366d315a93d9845b28b83),
>         Bech32Address(rms1ppdm08hd9gzx4qug63ahzwltcxhq9psd2dtdxeknzk5nmxz9k29cxy8aseq),
>     ),
> ]
> Foundries: []
> ```
> 
> that would be the output atm.

Tbh that's not really what I had in mind. I'd like to have the same output for these addresses than for Ed25519. This is why I've added the Type field.

Alex6323:
Changing it to look the same.

Alex6323:
current output:
```
Address:
  Key Index: 0
  Bech32:    rms1qqtj7pvnl3lj9n9n6e9lc47mfutjfhjyprmprxtzz2g0uck8tr3gurtp7tq
  Hex:       0x172f0593fc7f22ccb3d64bfc57db4f1724de4408f61199621290fe62c758e28e
  Type:      Ed25519

Outputs:
  Id:   0xc2dd37d77efea83d65e2aad295751f27e20e3b3d17a5e99dffb24d27311afdf20000
  Type: Alias
  Id:   0xd349d87ed19d5a282fb2675ae6702c0c9f9eb83de166b16c23e4db6ab038bae70000
  Type: Nft
  Id:   0xd349d87ed19d5a282fb2675ae6702c0c9f9eb83de166b16c23e4db6ab038bae70100
  Type: Basic
  Id:   0xc2dd37d77efea83d65e2aad295751f27e20e3b3d17a5e99dffb24d27311afdf20100
  Type: Foundry

Base Tokens:
  Amount: 999947200

Native Tokens:
  Id:     0x085bb79eed2a046a8388d47b713bebc1ae02860d5356d366d315a93d9845b28b830100000000
  Amount: 100

NFTs:
  Id:     0x5c17e3651a787be20c4d09640c2a883f0842ea1afb28790cdea402e7370ce5b8
  Bech32: rms1zpwp0cm9rfu8hcsvf5ykgrp23qlsssh2rtajs7gvm6jq9eehpnjmst5kf04

Aliases:
  Id:     0x5bb79eed2a046a8388d47b713bebc1ae02860d5356d366d315a93d9845b28b83
  Bech32: rms1ppdm08hd9gzx4qug63ahzwltcxhq9psd2dtdxeknzk5nmxz9k29cxy8aseq
```

thibault-martinez:
> current output:
> 
> ```
> Address:
>   Key Index: 0
>   Bech32:    rms1qqtj7pvnl3lj9n9n6e9lc47mfutjfhjyprmprxtzz2g0uck8tr3gurtp7tq
>   Hex:       0x172f0593fc7f22ccb3d64bfc57db4f1724de4408f61199621290fe62c758e28e
>   Type:      Ed25519
> 
> Outputs:
>   Id:   0xc2dd37d77efea83d65e2aad295751f27e20e3b3d17a5e99dffb24d27311afdf20000
>   Type: Alias
>   Id:   0xd349d87ed19d5a282fb2675ae6702c0c9f9eb83de166b16c23e4db6ab038bae70000
>   Type: Nft
>   Id:   0xd349d87ed19d5a282fb2675ae6702c0c9f9eb83de166b16c23e4db6ab038bae70100
>   Type: Basic
>   Id:   0xc2dd37d77efea83d65e2aad295751f27e20e3b3d17a5e99dffb24d27311afdf20100
>   Type: Foundry
> 
> Base Tokens:
>   Amount: 999947200
> 
> Native Tokens:
>   Id:     0x085bb79eed2a046a8388d47b713bebc1ae02860d5356d366d315a93d9845b28b830100000000
>   Amount: 100
> 
> NFTs:
>   Id:     0x5c17e3651a787be20c4d09640c2a883f0842ea1afb28790cdea402e7370ce5b8
>   Bech32: rms1zpwp0cm9rfu8hcsvf5ykgrp23qlsssh2rtajs7gvm6jq9eehpnjmst5kf04
> 
> Aliases:
>   Id:     0x5bb79eed2a046a8388d47b713bebc1ae02860d5356d366d315a93d9845b28b83
>   Bech32: rms1ppdm08hd9gzx4qug63ahzwltcxhq9psd2dtdxeknzk5nmxz9k29cxy8aseq
> ```

Why are we not treating Alias/NFT the same way as Ed25519? We need to see what they own.

cli/src/command/account.rs
```diff
@@ -938,25 +965,26 @@ pub async fn voting_output_command(account: &Account) -> Result<(), Error> {
     Ok(())
 }
 
-async fn print_address(account: &Account, address: &AccountAddress) -> Result<(), Error> {
+// TODO: display AccountAddress infos (key_index, change_address, etc) in the `addresses` command generated list?
```
Thoralf-M: Want to do in this PR?
If we want this information, I think it should just print the bip44 values as array to not bloat the output

Alex6323: I agree ... and yeah ... fine doing it in this PR :+1:

cli/src/command/account.rs
```diff
@@ -920,81 +965,108 @@ pub async fn voting_output_command(account: &Account) -> Result<(), Error> {
...
+        .into_iter()
+        .map(|nt| (*nt.token_id(), nt.amount().to_string()))
+    {
+        formatted_string.push_str(&format!("{0}{1:<8}{id}{0}{2:<8}{amount}", "\n  ", "Id:", "Amount:"));
+    }
+    formatted_string.push_str("\n");
+
+    formatted_string.push_str("\nNFTs:");
+    for (id, bech32) in nfts.into_iter() {
+        formatted_string.push_str(&format!("{0}{1:<8}{id}{0}{2:<8}{bech32}", "\n  ", "Id:", "Bech32:"));
```
thibault-martinez: I wouldn't print the Bech32 of NFT/Alias here tbh, not very relevant

cli/src/command/account.rs
```diff
@@ -920,81 +967,118 @@ pub async fn voting_output_command(account: &Account) -> Result<(), Error> {
...
-        nfts,
-        aliases,
-        foundries,
-    );
+    let mut row_count = 0;
+
+    // outputs table
+    formatted_table.push_str("\nOutputs:");
+    for (id, kind) in outputs {
+        formatted_table.push_str(&format!("{0}{1:<6}{id}{0}{2:<6}{kind}", "\n  ", "Id:", "Type:"));
```
thibault-martinez: I'd save some space and put the Type directly on the same line as the ID

cli/src/command/account.rs
```diff
@@ -920,81 +967,118 @@ pub async fn voting_output_command(account: &Account) -> Result<(), Error> {
...
+        row_count += 1;
+    }
+    formatted_table.push_str(if row_count == 0 { "\n  None\n" } else { "\n" });
+
+    // base token table
+    formatted_table.push_str("\nBase Tokens:");
+    formatted_table.push_str(&format!("{0}{1:<8}{2}\n", "\n  ", "Amount:", amount));
+
+    formatted_table.push_str("\nNative Tokens:");
```
thibault-martinez: Do you think we could not print the titles if empty ?

Alex6323: yes, if you prefer that ... we can not print them ... currently it would print `None`. I am not opinionated about it.

thibault-martinez: I tried it and it doesn't print `None`.  But I think it would be better without anyway.

cli/src/command/account.rs
```diff
@@ -306,15 +307,41 @@ impl FromStr for OutputSelector {
     }
 }
 
-/// `address` command
-pub async fn address_command(account: &Account, index: usize) -> Result<(), Error> {
-    let addresses = account.addresses().await?;
+/// Select by Bech32 address or list index
+#[derive(Debug, Copy, Clone)]
+pub enum AddressSelector {
+    // TODO: change to AccountAddress? (Note that AccountAddress isn't Copy)
```
DaughterOfMars: Neither is `Bech32Address` anymore :c

cli/src/command/account.rs
Alex6323: hmm ... I'ld rather have the table be of a consistent format rather save some space tbh? If I make them one-line it's also harder to extend this section in case we want to print something else for the outputs. The `address` section puts the address type on a separate line, so I would do it for the `outputs` as well. @Thoralf-M @DaughterOfMars @kwek20 opinions?

cli/src/command/account.rs
Alex6323: just to make sure ... this correct the bech32 address, tho, right? But you don't want to see it in the details list, but rather in the `addresses` list?

cli/src/command/account.rs
Alex6323: Done

cli/src/command/account.rs
Thoralf-M: I also prefer to save space and put it in the same line
Would also not have the empty new line before `Outputs` and so on

thibault-martinez: If you want consistency then I would definitely do one line as this is what we do for the outputs command

cli/src/command/account.rs
thibault-martinez: Yes I think this is redundant information that shouldn't be displayed here

sdk/src/wallet/account/operations/syncing/mod.rs
```diff
@@ -162,66 +162,36 @@ where
...
+        let mut new_outputs_data = outputs_data.clone();
 
+        loop {
             // Clear, so we only get new addresses
-            new_alias_and_nft_addresses = HashMap::new();
+            new_alias_and_nft_addresses.clear();
+
             // Add new alias and nft addresses
-            for output_data in new_outputs_data.iter() {
+            for output_data in new_outputs_data.drain(0..) {
```
Thoralf-M:
```suggestion
            for output_data in new_outputs_data.drain(..) {
```

sdk/src/wallet/account/operations/syncing/mod.rs
```diff
@@ -162,66 +162,36 @@ where
...
-                new_outputs_data
-            };
+        let bech32_hrp = self.client().get_bech32_hrp().await?;
+
+        let mut new_outputs_data = outputs_data.clone();
 
+        loop {
             // Clear, so we only get new addresses
-            new_alias_and_nft_addresses = HashMap::new();
+            new_alias_and_nft_addresses.clear();
```
Thoralf-M: We can just move the creation here `let mut new_alias_and_nft_addresses = HashMap::new();` and then don't need the clear anymore?

Alex6323: no, because then we would have to re-allocate in each iteration. This way ... we only allocate once. I guess `clear`ing is a lot cheaper.

Alex6323: oh, we can just drain this as well

sdk/src/wallet/account/operations/syncing/mod.rs
Alex6323: Changed

sdk/src/wallet/account/operations/syncing/mod.rs
Alex6323: Done

sdk/src/wallet/account/mod.rs
```diff
@@ -89,6 +90,8 @@ pub struct FilterOptions {
     pub foundry_ids: Option<HashSet<FoundryId>>,
     /// Return all nft outputs matching these IDs.
     pub nft_ids: Option<HashSet<NftId>>,
+    /// Return all outputs matching this address.
+    pub address: Option<Bech32Address>,
```
thibault-martinez: Isn't that breaking change ?

Alex6323: no, because it's additive

thibault-martinez: I just tried, it's breaking change. Because all the fields are public if someone is building the whole thing with no default for remaining fields, it will break.

DaughterOfMars: Although...do we really care about breaking changes in the CLI? It's a binary after all

thibault-martinez: This is not a CLI file

DaughterOfMars: oh whoops

Alex6323: ah, right. Good catch then!

cli/src/command/account.rs
Alex6323: I guess it doesn't make sense anymore with NFT and Alias addresses also listed there now, so let's resolve?

Thoralf-M: Well for Alias and NFT it would be just not existent, still interesting for the others
But I'm also fine not having it in this PR or at all, but then remove the TODO or open an issue :trollface:

cli/src/command/account.rs
Alex6323: This is done btw

cli/src/command/account.rs
Alex6323: forgot to say it ... done for some time now

cli/src/command/account.rs
Alex6323: chose to remove the TODO

cli/src/command/account.rs
```diff
@@ -920,81 +957,130 @@ pub async fn voting_output_command(account: &Account) -> Result<(), Error> {
...
-        "{log}\nOutputs: {:#?}\nBase coin amount: {}\nNative Tokens: {:#?}\nNFTs: {:#?}\nAliases: {:#?}\nFoundries: {:#?}\n",
-        output_ids,
-        amount,
-        native_tokens.finish_vec()?,
-        nfts,
-        aliases,
-        foundries,
-    );
+    // base tokens table
+    formatted_table.push_str("\nBase Tokens");
```
DaughterOfMars: could use `writeln!` instead of all the newline chars

thibault-martinez (changes requested):
Clippy

cli/src/command/account.rs
Alex6323: Hmm, not sure this is the right PR for this tbh bc we do it like this in various other places, and if changed, should probably be done consistently. Also ... I don't necessarily see the benefit of just getting rid of `\n` when we also use other special chars like `\t`. Another disadvantage would be the added compilation times due to expanding the macro itself in all those places.

thibault-martinez: I don't have an opinion on this but macro expansion is completely negligible, will not affect compilation time in any way

Alex6323: I agree ... that's why it's not the main point of the argument. I think if we change it to `writeln!` then this is not the right PR for this IMHO.

DaughterOfMars: I think it would just help clean this code a bit. format is already being used so using this macro instead is not going to change anything

Alex6323: well, format is not used everywhere but okay ... Thibault ... please make the call.

Thoralf-M (changes requested):
Clippy

Thoralf-M (dismissed):

DaughterOfMars (dismissed):

cli/src/command/account.rs
```diff
@@ -301,17 +307,48 @@ impl FromStr for OutputSelector {
     }
 }
 
-/// `addresses` command
-pub async fn addresses_command(account: &Account) -> Result<(), Error> {
-    let addresses = account.addresses().await?;
+/// Select by Bech32 address or list index
+#[derive(Debug, Copy, Clone)]
+pub enum AddressSelector {
+    // TODO: change to AccountAddress? (Note that AccountAddress isn't Copy)
```
thibault-martinez: Remove?

thibault-martinez: Or why do you want to use AccountAddress ?

cli/src/command/account.rs
```diff
@@ -920,81 +957,122 @@ pub async fn voting_output_command(account: &Account) -> Result<(), Error> {
...
-        output_ids,
-        amount,
-        native_tokens.finish_vec()?,
-        nfts,
-        aliases,
-        foundries,
-    );
+    // base tokens table
+    println_log_info!("Base Tokens");
+    println_log_info!("{:<8}{}", "Amount:", amount);
```
thibault-martinez: I'd put these two on the same line

cli/src/command/account.rs
```diff
@@ -1031,3 +1109,22 @@ fn outputs_ordering(a: &OutputData, b: &OutputData) -> std::cmp::Ordering {
...
+async fn get_addresses_sorted(account: &Account) -> Result<Vec<Bech32Address>, Error> {
+    let hrp = account.client().get_bech32_hrp().await?;
+    let mut addresses = account
+        .unspent_outputs(None)
+        .await?
+        .into_iter()
+        .filter_map(|data| match data.output {
+            Output::Alias(alias) => Some(Address::Alias(alias.alias_address(&data.output_id))),
+            Output::Nft(nft) => Some(Address::Nft(nft.nft_address(&data.output_id))),
+            Output::Basic(basic) => Some(*basic.address()),
```
thibault-martinez: Hmm that doesn't look right, surely this will create duplicates in the list?
You could collect into a HashSet to solve this but tbh I wouldn't use `unspent_outputs` for Ed25519, I would use the account method, otherwise we are not showing some addresses that have already been generated but not used. So account addresses method + unspent outputs for Alias/NFT.

Alex6323: yes, need to collect into a `HashSet` first, then convert to `Vec`, but yeah if you have a better way anyway then do it.

cli/src/command/account.rs
Alex6323: that's old and can be removed now. IIRC I wasn't sure how things should exactly be in the end.

cli/src/command/account.rs
thibault-martinez: I've done the other thing instead, collecting into a HashSet is not enough

cli/src/command/account.rs
Alex6323: yeah, fine with that now ... the table isn't format-consistent anymore anyway.

cli/src/command/account.rs
Alex6323: alright

Alex6323:
but now it will print all unspent outputs ... you also need to move `outputs.push` within the `required` check.

thibault-martinez:
> but now it will print all unspent outputs ... you also need to move `outputs.push` within the `required` check.

Right, didn't see it was done like that, thanks

Alex6323:
> > but now it will print all unspent outputs ... you also need to move `outputs.push` within the `required` check.
> 
> Right, didn't see it was done like that, thanks

It was actually done like this before this PR already. So now the table is different semantically, right, because we filter the outputs for owning/unlocking addresses, not just addresses associated with the output.

thibault-martinez:
> > > but now it will print all unspent outputs ... you also need to move `outputs.push` within the `required` check.
> > 
> > 
> > Right, didn't see it was done like that, thanks
> 
> It was actually done like this before this PR already. So now the table is different semantically, right, because we filter the outputs for owning/unlocking addresses, not just addresses associated with the output.

I think "associated" is a weird concept because some outputs definitely associated with the address weren't shown

thibault-martinez (dismissed):

Thoralf-M (dismissed):

cli/src/command/account.rs
```diff
@@ -920,81 +956,109 @@ pub async fn voting_output_command(account: &Account) -> Result<(), Error> {
...
+                Output::Alias(alias) => aliases.push(alias.alias_id_non_null(&output_data.output_id)),
+                Output::Nft(nft) => nfts.push(nft.nft_id_non_null(&output_data.output_id)),
+                Output::Basic(_) | Output::Foundry(_) | Output::Treasury(_) => {}
             }
+
+            let sdr_amount = output_data
+                .output
+                .unlock_conditions()
+                .unwrap()
+                .storage_deposit_return()
```
kwek20, line 1000: what if storage reposit return gets returned to one of your addresses? :) should we handle that, or is it taken into account elsewhere?

thibault-martinez: You can't really do anything about it, it's not yours to unlock. But I think we'll completely change this function when we have a proper balance per address feature and that would somewhat be included in it

Thoralf-M: In the account.balance() function we handle this, but here I'm not sure if we care that much, the address for which it is displayed can't do anything with the amount (unless the address in the SDRUC is the same as in the addressUC)

thibault-martinez (approved):

kwek20 (approved):

Thoralf-M (approved):

PR #1428, opened 2023-10-13
Split query parameters into multiple structs

### LINKED ISSUE #1248: Have specific structs for the QueryParameters
## Description

Have more specific enums for the QueryParameters, so for each route one can only provide query parameters that can actually be used there, like in the nodejs binding.

- [x] Rust
- [ ] Nodejs
- [ ] Python

## Motivation

https://github.com/iotaledger/iota-sdk/pull/1246#discussion_r1328690716

Yes.

Thoralf-M:
# Description of change

Split query parameters into multiple structs so users only can provide ones that are actually can be used for a certain endpoint and will not get an error during runtime if they provided a wrong one.
Also adjusted the query parameters for 2.0 since there only a single native token can be in one output the NT query parameters changed to this
```
    /// Filters outputs based on the presence of a native token.
    pub has_native_token: Option<bool>,
    /// Filters outputs based on the presence of a specific native token.
    pub native_token: Option<TokenId>,
```

#1248 

## Type of change

Just `cargo test`, can't test with a node yet


sdk/src/client/node_api/indexer/query_parameters.rs
```diff
@@ -8,69 +8,235 @@ use std::fmt;
...
-        query_parameters.dedup_by_key(|qp| qp.kind());
+impl_query_parameters_methods!(OutputsQueryParameters, OutputsQueryParameter);
 
-        Self(query_parameters)
-    }
+/// Query parameter for output requests.
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub enum OutputsQueryParameter {
```
DaughterOfMars: Should these just be structs?

Thoralf-M: Because each of the variants can be there at the same time and only once? Or why do you think would a struct be better?

DaughterOfMars: Yeah exactly. Feels like a struct to me

sdk/examples/client/02_address_balance.rs
```diff
@@ -43,12 +43,13 @@ async fn main() -> Result<()> {
 
     // Get output ids of outputs that can be controlled by this address without further unlock constraints
     let output_ids_response = client
-        .basic_output_ids([
-            BasicOutputsQueryParameters::Address(first_address.clone()),
-            BasicOutputsQueryParameters::HasExpiration(false),
-            BasicOutputsQueryParameters::HasTimelock(false),
-            BasicOutputsQueryParameters::HasStorageDepositReturn(false),
-        ])
+        .basic_output_ids(BasicOutputsQueryParameters {
```
DaughterOfMars: builder pattern?

sdk/src/client/node_api/indexer/query_parameters.rs
```diff
@@ -3,310 +3,240 @@
...
-    pub fn replace(&mut self, query_parameter: QueryParameter) {
-        match self
-            .0
-            .binary_search_by_key(&query_parameter.kind(), QueryParameter::kind)
-        {
-            Ok(pos) => self.0[pos] = query_parameter,
-            Err(pos) => self.0.insert(pos, query_parameter),
+macro_rules! impl_query_parameters_methods {
+    ($name:ty) => {
+        impl QueryParameter for $name {
```
DaughterOfMars, line 54: smart to do this using serde, but I'm pretty sure this doesn't need to be a macro, it can just be the default impl with `Self: Serialize` bounds

thibault-martinez:
Conflicts

sdk/src/client/node_api/indexer/query_parameters.rs
Thoralf-M: Changed

sdk/examples/client/02_address_balance.rs
Thoralf-M: Like this?
```Rust
BasicOutputsQueryParametersBuilder::default()
                .with_address(first_address.clone())
                .with_has_expiration(false)
                .with_has_timelock(false)
                .with_has_storage_deposit_return(false)
                .build()?
```
Can do it with derive_builder, but don't really see the benefit

sdk/src/client/node_api/indexer/query_parameters.rs
Thoralf-M: Changed to a struct :+1:

sdk/src/client/node_api/indexer/query_parameters.rs
```diff
@@ -9,41 +9,35 @@ use crate::types::block::{address::Bech32Address, output::TokenId, slot::SlotInd
 
 // https://github.com/iotaledger/inx-indexer/tree/develop/pkg/indexer
 
-pub trait QueryParameter {
+pub trait QueryParameter: Serialize + Send + std::marker::Sync {
```
DaughterOfMars: shouldn't need the Sync path?

sdk/examples/client/02_address_balance.rs
DaughterOfMars: Yeah that's what I was thinking. The benefit isn't huge, true. I don't mind really.

DaughterOfMars: Although if this is user-facing it may be worth it just for convenience.

sdk/src/client/node_api/indexer/query_parameters.rs
Thoralf-M: Yeah not sure, maybe a clippy bug, but without it I get this
```
error: future cannot be sent between threads safely
  --> sdk/src/client/node_api/indexer/mod.rs:24:10
   |
24 |     ) -> Result<OutputIdsResponse> {
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^ future returned by `get_output_ids` is not `Send`
   |
note: future is not `Send` as this value is used across an await
  --> sdk/src/client/node_api/indexer/mod.rs:45:18
   |
41 |                     query_parameters.to_query_string().as_deref(),
   |                     ---------------- has type `&impl QueryParameter` which is not `Send`
...
45 |                 .await?;
   |                  ^^^^^ - `query_parameters` is later dropped here
   |                  |
   |                  await occurs here, with `query_parameters` maybe used later
help: consider moving this into a `let` binding to create a shorter lived borrow
  --> sdk/src/client/node_api/indexer/mod.rs:41:21
   |
41 |                     query_parameters.to_query_string().as_deref(),
   |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: `impl QueryParameter` doesn't implement `std::marker::Sync`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#future_not_send
```

DaughterOfMars: No, I mean it should be just ```Send + Sync``` without the path

Thoralf-M: Ah, changed

sdk/src/client/node_api/indexer/query_parameters.rs
```diff
@@ -124,12 +118,12 @@ pub struct BasicOutputsQueryParameters {
...
 
-impl BasicOutputsQueryParametersBuilder {
+impl BasicOutputsQueryParameters {
     /// Sets `.address(address).has_expiration(false).has_expiration(false).has_storage_deposit_return(false)` to only
     /// get outputs that can be unlocked by the address without potential further restrictions.
-    pub fn only_address_unlock_condition(&mut self, address: impl Into<Bech32Address>) -> &mut Self {
+    pub fn only_address_unlock_condition(self, address: impl Into<Bech32Address>) -> Self {
         self.address(address.into())
             .has_expiration(false)
             .has_expiration(false)
```
DaughterOfMars, line 142: supposed to be `has_timelock`?

sdk/src/client/node_api/indexer/query_parameters.rs
```diff
@@ -211,12 +205,12 @@ pub struct NftOutputsQueryParameters {
...
 
-impl NftOutputsQueryParametersBuilder {
+impl NftOutputsQueryParameters {
     /// Sets `.address(address).has_expiration(false).has_expiration(false).has_storage_deposit_return(false)` to only
     /// get outputs that can be unlocked by the address without potential further restrictions.
-    pub fn only_address_unlock_condition(&mut self, address: impl Into<Bech32Address>) -> &mut Self {
+    pub fn only_address_unlock_condition(self, address: impl Into<Bech32Address>) -> Self {
         self.address(address.into())
             .has_expiration(false)
             .has_expiration(false)
```
DaughterOfMars, line 230: Same problem

sdk/src/client/node_api/indexer/query_parameters.rs
Thoralf-M: Yes thanks, not sure how I missed this :joy:

sdk/src/client/node_api/indexer/query_parameters.rs
Thoralf-M: Updated

sdk/src/client/node_api/indexer/query_parameters.rs
```diff
@@ -121,13 +121,13 @@ pub struct BasicOutputsQueryParameters {
 impl_query_parameters_methods!(BasicOutputsQueryParameters);
 
 impl BasicOutputsQueryParameters {
-    /// Sets `.address(address).has_expiration(false).has_expiration(false).has_storage_deposit_return(false)` to only
+    /// Sets `.address(address).has_expiration(false).has_storage_deposit_return(false).has_timelock(false)` to only
     /// get outputs that can be unlocked by the address without potential further restrictions.
     pub fn only_address_unlock_condition(self, address: impl Into<Bech32Address>) -> Self {
         self.address(address.into())
-            .has_expiration(false)
             .has_expiration(false)
```
DaughterOfMars, line 142: Do we really need these fns at all, since these are all the defaults? You can just call

```BasicOutputsQueryParameters::default().address(address)```

Comment (Thoralf-M) [reply sdk/src/client/node_api/indexer/query_parameters.rs:142]: `None` will be removed from the query_string, but we need to have it set to `false`

Comment (DaughterOfMars) [reply sdk/src/client/node_api/indexer/query_parameters.rs:142]: ahhh, I see

**Review by DaughterOfMars — DISMISSED** [review]

File: sdk/Cargo.toml [hunk 8]
```diff
@@ -22,6 +22,7 @@ bech32 = { version = "0.10.0-beta", default-features = false, features = [
     "alloc",
 ] }
 bitflags = { version = "2.4.0", default-features = false }
+derive_setters = { version = "0.1.6", default-features = false }
```
thibault-martinez: After `derive_more` 😬

DaughterOfMars (dismissed):

sdk/examples/client/quorum.rs
```diff
@@ -52,12 +52,13 @@ async fn main() -> Result<()> {
...
-            QueryParameter::HasTimelock(false),
-            QueryParameter::HasStorageDepositReturn(false),
-        ])
+        .basic_output_ids(
+            QueryParameter{address: Some(addresses[0]),
+            has_expiration: Some(false),
+            has_timelock: Some(false),
+            has_storage_deposit_return: Some(false),
+            ..Default::default()
+})
```
thibault-martinez: Update ?

sdk/src/wallet/account/operations/syncing/addresses/output_ids/basic.rs
```diff
@@ -37,11 +40,74 @@ where
...
+            output_ids.extend(
+                self.client()
+                    .basic_output_ids(BasicOutputQueryParameters::new().unlockable_by_address(bech32_address.clone()))
+                    .await?
+                    .items,
+            );
+            output_ids.extend(
+                self.client()
+                    .basic_output_ids(
+                        BasicOutputQueryParameters::new().storage_deposit_return_address(bech32_address.clone()),
```
thibault-martinez: Do we remove it then ?

sdk/examples/client/quorum.rs
Thoralf-M: It wasn't even tracked

DaughterOfMars (dismissed):

sdk/src/wallet/account/operations/syncing/addresses/output_ids/basic.rs
Thoralf-M: Yes

sdk/src/wallet/account/operations/syncing/addresses/output_ids/nft.rs
```diff
@@ -17,11 +25,70 @@ where
...
+            let mut output_ids = Vec::new();
+            output_ids.extend(
+                self.client()
+                    .nft_output_ids(NftOutputQueryParameters::new().unlockable_by_address(bech32_address.clone()))
+                    .await?
+                    .items,
+            );
+            output_ids.extend(
+                self.client()
+                    .nft_output_ids(NftOutputQueryParameters::new().storage_deposit_return_address(bech32_address))
```
thibault-martinez: :trollface:

sdk/src/client/node_api/indexer/query_parameters.rs
```diff
@@ -3,343 +3,295 @@
...
-}
-
-impl fmt::Display for QueryParameter {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let query_tuple = self.to_query_tuple();
-        write!(f, "{}={}", query_tuple.0, query_tuple.1)
+impl BasicOutputQueryParameters {
+    /// Sets `.address(address).has_expiration(false).has_storage_deposit_return(false).has_timelock(false)` to only
+    /// get outputs that can be unlocked by the address without potential further restrictions.
+    pub fn only_address_unlock_condition(self, address: impl Into<Bech32Address>) -> Self {
```
thibault-martinez: I feel like this should actually create a `BasicOutputQueryParameters`, it feels weird to call it on an existing one

sdk/src/client/node_api/indexer/query_parameters.rs
```diff
@@ -3,343 +3,295 @@
...
+    /// Returns outputs that are unlockable by the bech32 address.
+    pub unlockable_by_address: Option<Bech32Address>,
+    /// Filters outputs based on the presence of validated Sender (bech32 encoded).
+    pub sender: Option<Bech32Address>,
+    /// Filters outputs based on bech32-encoded issuer address.
+    pub issuer: Option<Bech32Address>,
+    /// Filters outputs based on bech32-encoded governor (governance controller) address.
+    pub governor: Option<Bech32Address>,
+    /// Filters outputs based on bech32-encoded state controller address.
+    pub state_controller: Option<Bech32Address>,
```
thibault-martinez: I would make all fields private. We only need to serialise the struct, no need to access the fields. Derive getters if needed. But as we've seen with Alex's PR, public fields like that are shooting ourselves in the foot as adding more fields becomes breaking change.

sdk/src/wallet/account/operations/syncing/addresses/output_ids/nft.rs
Thoralf-M: Removed

sdk/src/client/node_api/indexer/query_parameters.rs
Thoralf-M: Changed

sdk/src/client/node_api/indexer/query_parameters.rs
Thoralf-M: Changed

sdk/examples/how_tos/account_wallet/transaction.rs
```diff
@@ -54,7 +54,9 @@ async fn main() -> Result<()> {
     // Find first output unlockable by the account address
     let input = *account
         .client()
-        .basic_output_ids([QueryParameter::Address(account_address)])
```
thibault-martinez, line 57: Just making sure, does this example actually need only_address_unlock_condition ? Or should it be UnlockableByAddress?

sdk/src/wallet/account/operations/syncing/addresses/output_ids/nft.rs
```diff
@@ -17,10 +17,9 @@ where
         bech32_address: impl ConvertTo<Bech32Address>,
     ) -> crate::wallet::Result<Vec<OutputId>> {
         let bech32_address = bech32_address.convert()?;
-
         Ok(self
             .client()
-            .nft_output_ids([QueryParameter::UnlockableByAddress(bech32_address)])
+            .nft_output_ids(NftOutputQueryParameters::new().storage_deposit_return_address(bech32_address))
```
thibault-martinez: Wrong ?

sdk/src/wallet/account/operations/syncing/addresses/output_ids/nft.rs
```diff
@@ -17,10 +17,9 @@ where
         bech32_address: impl ConvertTo<Bech32Address>,
     ) -> crate::wallet::Result<Vec<OutputId>> {
         let bech32_address = bech32_address.convert()?;
```
thibault-martinez:
```suggestion
        let bech32_address = bech32_address.convert()?;

```

sdk/examples/how_tos/account_wallet/transaction.rs
Thoralf-M: only_address_unlock_condition() is the safer option, the other one could fail

sdk/src/wallet/account/operations/syncing/addresses/output_ids/nft.rs
Thoralf-M: Good catch

sdk/src/client/node_api/indexer/mod.rs
```diff
@@ -28,8 +28,11 @@ impl ClientInner {
             items: Vec::new(),
         };
 
-        // Return early with only a single page if a `QueryParameter::Cursor(_)` is provided.
-        let return_early = query_parameters.contains(QueryParameter::Cursor(String::new()).kind());
+        let query_string = query_parameters.to_query_string();
+        // Return early with only a single page if an empty string is provided as cursor.
+        let return_early = query_string
+            .map(|s| s.contains("cursor=&") || s.ends_with("cursor="))
```
thibault-martinez, line 34: Wouldn't `s.contains("cursor=")` be enough ?

thibault-martinez: Or maybe you just add `has_cursor` method idk

DaughterOfMars: regex?

DaughterOfMars (dismissed):

sdk/src/client/node_api/indexer/mod.rs
thibault-martinez: It's probably better that the check happens before the serialisation though

DaughterOfMars: True, could be part of the trait

Thoralf-M: If there is `cursor`, but with an actual specified value, then we don't want to return early, that's why just checking for `cursor=`  is not enough

sdk/Cargo.toml
```diff
@@ -130,6 +131,7 @@ pretty_assertions = { version = "1.4.0", default-features = false, features = [
 dotenvy = { version = "0.15.7", default-features = false }
 fern-logger = { version = "0.5.0", default-features = false }
 num_cpus = { version = "1.16.0", default-features = false }
+once_cell = { version = "1.18.0", default-features = false }
```
DaughterOfMars, line 134: ?

sdk/src/client/node_api/indexer/routes.rs
```diff
@@ -30,14 +26,9 @@ impl ClientInner {
     /// "createdBefore", "createdAfter", "cursor", "pageSize".
     /// Returns Err(Node(NotFound) if no results are found.
     /// api/indexer/v1/outputs
-    pub async fn output_ids(
-        &self,
-        query_parameters: impl Into<Vec<QueryParameter>> + Send,
-    ) -> Result<OutputIdsResponse> {
+    pub async fn output_ids(&self, query_parameters: OutputQueryParameters) -> Result<OutputIdsResponse> {
         let route = "api/indexer/v1/outputs";
```
thibault-martinez: bump the version?

sdk/Cargo.toml
Thoralf-M: Without it https://github.com/iotaledger/iota-sdk/blob/4a46228442a97c908abeff9ab0c1aa4b2b65b04d/sdk/src/types/block/mana/parameters.rs#L216 fails without the wallet feature

DaughterOfMars: good to know. but that's a test, so this should be in dev-dependencies if anything

Thoralf-M: It is a dev-dependency

thibault-martinez (approved):

sdk/src/client/node_api/indexer/query_parameters.rs
```diff
@@ -3,343 +3,306 @@
...
+
+        let mut basic_outputs_query_parameters = BasicOutputQueryParameters::new()
+            .address(
+                Bech32Address::try_from_str("atoi1qzt0nhsf38nh6rs4p6zs5knqp6psgha9wsv74uajqgjmwc75ugupx3y7x0r")
+                    .unwrap(),
+            )
+            .cursor("".into());
+        assert_eq!(
+            basic_outputs_query_parameters.to_query_string(),
+            Some("address=atoi1qzt0nhsf38nh6rs4p6zs5knqp6psgha9wsv74uajqgjmwc75ugupx3y7x0r&cursor=".into())
```
kwek20, line 299: is an empty cursor valid even?

sdk/src/client/node_api/indexer/mod.rs
```diff
@@ -28,8 +28,11 @@ impl ClientInner {
             items: Vec::new(),
         };
 
-        // Return early with only a single page if a `QueryParameter::Cursor(_)` is provided.
-        let return_early = query_parameters.contains(QueryParameter::Cursor(String::new()).kind());
+        let query_string = query_parameters.to_query_string();
+        // Return early with only a single page if an empty string is provided as cursor.
```
kwek20, line 32: oh i see :D

kwek20 (approved):
Too bad of the duplicate fields in the differnet query ones, but its probably better to keep em separate

PR #1429, opened 2023-10-13
Error fallback variant

### LINKED ISSUE #1374: Add a default boxed error as fallback variant

Alex6323:
# Description of change

Adds a `Fallback` variant to the block error enum.

Keeping it a `draft` for now bc I don't know whether that's the right enum, and whether other enums should have a fallback, too.

Closes #1374


DaughterOfMars:
I think we were planning to add it to the client error, but I'm wondering if we should have another discussion about non_exhaustive tbh

thibault-martinez:
Yeah adding one to all major error enums basically

thibault-martinez:
Will add non_exhaustive instead

PR #1448, opened 2023-10-16
Fix some mana calc bugs and simplify math

DaughterOfMars:
# Description of change

Some fixes that allow more mana calculation tests to pass as well as some simplification. There are still some bugs in the iota.go impl so we'll have to wait for those to be fixed before the last commented tests can pass.

### DIFF
--- a/sdk/src/types/block/mana/structure.rs
+++ b/sdk/src/types/block/mana/structure.rs
@@ -54,14 +54,10 @@ impl ManaStructure {
         (1 << self.bits_count) - 1
     }
 
-    fn decay(&self, mana: u64, epoch_delta: u32) -> u64 {
+    fn decay(&self, mut mana: u64, epoch_delta: u32) -> u64 {
         if mana == 0 || epoch_delta == 0 || self.decay_factors().is_empty() {
             return mana;
         }
-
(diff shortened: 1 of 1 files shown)


sdk/src/types/block/mana/structure.rs
```diff
@@ -73,13 +69,11 @@ impl ManaStructure {
             // Unwrap: Safe because the index is at most the length
             let decay_factor = self.decay_factor_at(epochs_to_decay - 1).unwrap();
 
-            // apply the decay using fixed-point arithmetics.
-            (mana_hi, mana_lo) =
-                multiplication_and_shift(mana_hi, mana_lo, decay_factor, self.decay_factors_exponent());
+            // apply the decay using fixed-point arithmetics
```
thibault-martinez, line 72: Is it still considered fixed point arithmetic?

thibault-martinez (approved):

sdk/src/types/block/mana/structure.rs
DaughterOfMars: I'm not entirely sure it was ever fixed point math 🤷‍♀️

Alex6323 (approved):

kwek20 (approved):

PR #1460, opened 2023-10-18
Use `pretty_assertions`

DaughterOfMars:
# Description of change

Uses the `pretty_assertions` as a dev dependency for better test assertions.

## Type of change

### DIFF
--- a/bindings/core/src/method/secret_manager.rs
+++ b/bindings/core/src/method/secret_manager.rs
@@ -73,6 +73,7 @@ pub enum SecretManagerMethod {
 
 #[cfg(test)]
 mod test {
+    use pretty_assertions::assert_eq;
     #[test]
     fn bip44_deserialization() {
         let signature_unlock_method: super::SecretManagerMethod = serde_json::from_str(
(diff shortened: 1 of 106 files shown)


bindings/core/src/method/secret_manager.rs
```diff
@@ -73,6 +73,7 @@ pub enum SecretManagerMethod {
 
 #[cfg(test)]
 mod test {
+    use pretty_assertions::assert_eq;
```
thibault-martinez, line 76:
```suggestion
    use pretty_assertions::assert_eq;

```

sdk/src/wallet/account/mod.rs
```diff
@@ -587,145 +587,150 @@ impl From<&AccountDetails> for AccountDetailsDto {
     }
 }
 
-#[test]
-fn serialize() {
```
thibault-martinez, line 591: lol

thibault-martinez (approved):

Alex6323 (approved):

PR #1463, opened 2023-10-18
CLI: enter doesn't show helper anymore

### LINKED ISSUE #1449: CLI: enter shouldn't print help
It should just do nothing

thibault-martinez:
Closes https://github.com/iotaledger/iota-sdk/issues/1449

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -34,7 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `transaction` and `transactions` indexed transactions in opposite order;
-- Enter doesn't show helper anymore;
+- Enter was showing the helper message;
 - Earlier trim of the input so that pre-commands (`help`, `clear`, `accounts`, ...) work even with leading/trailing spaces;
 
 ## 1.1.0 - 2023-09-29
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `transaction` and `transactions` indexed transactions in opposite order;
 - Enter doesn't show helper anymore;
+- Earlier trim of the input so that pre-commands (`help`, `clear`, `accounts`, ...) work even with leading/trailing spaces;
 
 ## 1.1.0 - 2023-09-29
 
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `transaction` and `transactions` indexed transactions in opposite order;
+- Enter doesn't show helper anymore;
 
 ## 1.1.0 - 2023-09-29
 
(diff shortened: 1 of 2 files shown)


DaughterOfMars (dismissed):

cli/src/account.rs
```diff
@@ -94,7 +96,7 @@ pub async fn account_prompt_internal(
                 }
                 _ => {
                     // Prepend `Account: ` so the parsing will be correct
-                    let command = format!("Account: {}", command.trim());
+                    let command = format!("Account: {}", command);
```
Thoralf-M:
```suggestion
                    let command = format!("Account: {command}");
```
Why doesn't clippy show this?

Thoralf-M (dismissed):

cli/CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `transaction` and `transactions` indexed transactions in opposite order;
+- Enter doesn't show helper anymore;
+- Earlier trim of the input so that pre-commands (`help`, `clear`, `accounts`, ...) work even with leading/trailing spaces;
```
Alex6323, line 38: I am not sure this (line 38) is a bug tbh and should be under the "Fixed" tab. To me it's more like a changed behavior? In Bash/Fish for example you can do `      clear` (so it trims), but you can't do `      cd..` (it does not trim the leading spaces) :shrug:

cli/CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `transaction` and `transactions` indexed transactions in opposite order;
+- Enter doesn't show helper anymore;
```
Alex6323: I thought we don't like using temporal terms like `now`, `anymore`, etc

Also ... I have a question ... Under the "Fixed" tab, should we describe what was wrong before like I did in line 36, or should we put a description of the new fixed state there like you did here, or should we not bother and do it both ways?

cli/CHANGELOG.md
Alex6323: I am not against trimming btw ... I am saying that the behavior will for most people diverge from what they are used to in their terminal...and those terminals probably don't trim leading (in all cases) and trailing whitespace for a (good?) reason.

thibault-martinez: You can definitely do `         cd ..` in fish, just tested it.

cli/src/account.rs
thibault-martinez: Not sure why it doesn't 😢

cli/CHANGELOG.md
thibault-martinez: Idk, this really seems like a bug to me, there is absolutely no reason not to trim?

cli/CHANGELOG.md
thibault-martinez: As long as it's understandable, I'd say it doesn't matter. I'll reword.

thibault-martinez: Done

cli/CHANGELOG.md
Alex6323: you're right ... when I tested I copied from here which didn't include the space in between :see_no_evil:

Alex6323 (approved):

DaughterOfMars (approved):

PR #1482, opened 2023-10-20
Refactor json decorator

### LINKED ISSUE #1473: Change _to_dict_custom() to class method?
## Description

Currently it's a static method which accepts a dict which gets created here https://github.com/iotaledger/iota-sdk/blob/346b6c897596607f9564483956ab6b7393fee04e/bindings/python/iota_sdk/types/common.py#L21-L29

Related comment https://github.com/iotaledger/iota-sdk/pull/1381/files/019aa6e73b16adc26cfafec138f8c6e6b5fb96fa..86d678be049d37c7e2e4b809a14a9641bb1266df#r1364408452

## Open questions (optional)

It should probably be just a class method which accepts `cls` as parameter, or is there any reason why it's not that?

No.

DaughterOfMars:
# Description of change

Refactors the `@json` decorator so that the methods can be overridden by defining a class method, and remove some unnecessary overrides.

Closes #1473


DaughterOfMars:
Waiting for #1481

thibault-martinez:
#1481 has been merged

thibault-martinez (changes requested):
CI failing

bindings/python/iota_sdk/types/token_scheme.py
```diff
@@ -22,20 +22,5 @@ class SimpleTokenScheme:
     maximum_supply: HexStr
     type: int = field(default_factory=lambda: 0, init=False)
 
-    @staticmethod
-    def _to_dict_custom(config):
```
Thoralf-M, line 26: Don't we want to keep this if we want to use `int` for the token amounts later?

Thoralf-M: I guess custom encoder/decoder would be better and we still have the git history for this, created a new issue https://github.com/iotaledger/iota-sdk/issues/1494

bindings/python/iota_sdk/types/common.py
```diff
@@ -12,21 +12,28 @@
...
         letter_case=LetterCase.CAMEL,
         undefined=Undefined.RAISE)(cls)
 
-    # Store original methods
-    original_to_dict = cls.to_dict
+    # Re-apply the original fn if it exists
+    if override_to_dict is not None:
+        setattr(cls, "_to_dict", override_to_dict)
+    else:
+        setattr(cls, "_to_dict", cls.to_dict)
```
Thoralf-M: Don't really get what's the point of this, if `to_dict` exists, you set it for `_to_dict`, just to set it as `to_dict` again inside of `custom_to_dict()`?

kwek20: yea i think we can remove line28-29, and remove _ from _to_dict() and were golden

DaughterOfMars: The reason is because there are two different definitions:
1. the fn that converts to a dict
2. the wrapper fn that calls the above and removes `None` values

Because of that, there must be two different fns in the class.

Thoralf-M (changes requested):
ImplicitAccountCreationAddress still has `_to_dict_custom`

bindings/python/iota_sdk/types/common.py
Thoralf-M: This here seems to just work fine
```Python
def json(cls):
    """Decorator to add to_dict and to_json methods to a dataclass."""

    # Get potential override method
    to_dict = getattr(cls, "to_dict", None)

    # Apply the dataclass_json decorator to get the default behavior
    cls = dataclass_json(
        letter_case=LetterCase.CAMEL,
        undefined=Undefined.RAISE)(cls)

    # If no custom one is defined, set the default from dataclass_json
    if to_dict is None:
        to_dict = cls.to_dict

    # Override to_dict to remove None values
    def custom_to_dict(self, *args, **kwargs):
        # pylint: disable=protected-access
        original_dict = to_dict(self, *args, **kwargs)

        result = {k: v for k, v in original_dict.items() if v is not None}
        return result

    def custom_to_json(self, *args, **kwargs):
        # Use the custom to_dict method for serialization
        return DataClassJsonMixin.to_json(self, *args, **kwargs)

    setattr(cls, "to_dict", custom_to_dict)
    setattr(cls, "to_json", custom_to_json)

    return cls
```

bindings/python/iota_sdk/types/common.py
```diff
@@ -12,21 +12,28 @@
...
 
-    # Override methods
+    # Override to_dict to remove None values
     def custom_to_dict(self, *args, **kwargs):
-        original_dict = original_to_dict(self, *args, **kwargs)
+        # pylint: disable=protected-access
+        original_dict = self._to_dict(self, *args, **kwargs)
```
Thoralf-M:
```Python
def test_to_dict():
    remainder_value_strategy = {
        "address": "iota...",
        "key_index": 0,
        "internal": False,
        "used": False
    }
    remainder_value_strategy_custom_address = from_dict(
        RemainderValueStrategyCustomAddress, remainder_value_strategy)
    print(f'{remainder_value_strategy_custom_address.to_dict()}')
    assert remainder_value_strategy_custom_address.to_dict() == {'strategy': 'CustomAddress', 'value': {
        'address': 'iota...', 'keyIndex': 0, 'internal': False, 'used': False}}


test_to_dict()
```
fails with 
```
Traceback (most recent call last):
  File "iota-sdk/bindings/python/tests/test_output.py", line 45, in <module>
    test_to_dict()
  File "iota-sdk/bindings/python/tests/test_output.py", line 40, in test_to_dict
    print(f'{remainder_value_strategy_custom_address.to_dict()}')
  File "/iota_sdk/types/common.py", line 34, in custom_to_dict
    original_dict = self._to_dict(self, *args, **kwargs)
TypeError: RemainderValueStrategyCustomAddress.to_dict() takes 1 positional argument but 2 were given
```

bindings/python/iota_sdk/types/common.py
DaughterOfMars: fair enough. python is weird

Thoralf-M (approved):

bindings/python/iota_sdk/types/send_params.py
```diff
@@ -71,18 +71,15 @@ class CreateNativeTokenParams():
         foundry_metadata: The foundry metadata of the native token.
         account_id: The ID of the corresponding account.
     """
-    circulating_supply: int
-    maximum_supply: int
+    circulating_supply: int = field(metadata=config(
+        encoder=str
+    ))
+    maximum_supply: int = field(metadata=config(
+        encoder=str
```
thibault-martinez, line 78: Are these going to be hex strings ?

Thoralf-M: No, I opened an issue for it https://github.com/iotaledger/iota-sdk/issues/1494

thibault-martinez: Hmm ok, I would have kept a custom dict because we are (temporarily) making this wrong. But if we won't forget to change it with the issue that's fine.

thibault-martinez (approved):

Alex6323 (approved):

PR #1496, opened 2023-10-24
Fix ledger nano remainder finder

thibault-martinez:


sdk/src/client/secret/ledger_nano.rs
```diff
@@ -296,26 +296,25 @@ impl SecretManage for LedgerSecretManager {
...
+                }
+                None => (None, LedgerBIP32Index::default()),
+            };
 
             let mut remainder_index = 0u16;
-            if let Some(remainder_address) = remainder_address {
+            if let Some(remainder_output) = remainder_output {
                 match &prepared_transaction.essence {
                     TransactionEssence::Regular(essence) => {
                         // find the index of the remainder in the essence
```
Thoralf-M: Please also update the comment next line, because `essences are sorted lexically` is wrong

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -326,21 +325,19 @@ impl SecretManage for LedgerSecretManager {
                         // The outputs in the essence already are sorted
                         // at this place, so we can rely on their order and don't have to sort it again.
                         'essence_outputs: for output in essence.outputs().iter() {
```
Thoralf-M: In the ISA we put the remainder at last position, so rev() should be faster
```suggestion
                        'essence_outputs: for output in essence.outputs().iter().rev() {
```

thibault-martinez: Ok but need to be careful with the index then

Thoralf-M: Ah yeah we can also just leave it as is, max is like ~20 rounds anyways

sdk/src/client/secret/ledger_nano.rs
thibault-martinez: Done

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -294,53 +294,46 @@ impl SecretManage for LedgerSecretManager {
...
                 match &prepared_transaction.essence {
                     TransactionEssence::Regular(essence) => {
-                        // find the index of the remainder in the essence
-                        // this has to be done because outputs in essences are sorted
-                        // lexically and therefore the remainder is not always the last output.
-                        // The index within the essence and the bip32 index will be validated
-                        // by the hardware wallet.
-                        // The outputs in the essence already are sorted
-                        // at this place, so we can rely on their order and don't have to sort it again.
-                        'essence_outputs: for output in essence.outputs().iter() {
```
thibault-martinez, line 328: The label didn't really seem needed?

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -294,53 +294,46 @@ impl SecretManage for LedgerSecretManager {
...
-                            if let Output::Basic(s) = output {
-                                if let Some(address) = s.unlock_conditions().address() {
-                                    if *remainder_address == *address.address() {
-                                        break 'essence_outputs;
-                                    }
-                                }
-                            } else {
+                        // Find the index of the remainder in the transaction this has to be done because the remainder
```
DaughterOfMars: punctuation. needed

thibault-martinez: Reworded a bit

Thoralf-M (approved):

DaughterOfMars (approved):

PR #1510, opened 2023-10-25
Improve syncing and fix deadlock

Thoralf-M:
# Description of change

Moved the claimable_outputs function call in syncing out of the loop, so it's now only called once.
And created `claimable_outputs` for the AccountDetails, so we can use it without needing to get the read lock (which caused the deadlock).

Deadlock happened basically like this:

Cargo.toml
```toml
[dependencies]
tokio = { version = "1.33.0", default-features = false, features = [
    "macros",
    "rt-multi-thread",
    "time",
    "sync",
] }
```
main.rs
```Rust
use tokio::{
    sync::RwLock,
    time::{sleep, Duration},
};

#[tokio::main]
async fn main() {
    let lock = RwLock::new(0);
    tokio::join!(read(&lock), write(&lock));
}

async fn read(lock: &RwLock<i32>) {
    println!("read");
    {
        sleep(Duration::from_millis(1)).await;
        let v = lock.read().await;
        println!("read {v}");

        sleep(Duration::from_millis(100)).await;

        // read again, while other branch is waiting to get write access -> deadlock
        let _v = lock.read().await;
    }
    println!("read end")
}

async fn write(lock: &RwLock<i32>) {
    println!("write");
    {
        sleep(Duration::from_millis(10)).await;
        let mut v = lock.write().await;
        println!("write +1");
        *v += 1;
    }
    println!("write end")
}
```

Be sure to reference any related issues by adding `fixes #issue_number`.

## Type of change

Created an account with 872 timelocked outputs and then modified the consolidation example to have background syncing and a few ms sleep before selecting inputs, so input selection and background syncing was running at the same moment. Without this change it resulted in a deadlock, after this change not anymore.

Also when the deadlock didn't occur the time to sync got reduced (~250 ms of this time was taken by the API request)
```
[SYNC] finished syncing in 489.46ms (before)
[SYNC] finished syncing in 311.41ms (after)
```


sdk/src/wallet/account/operations/balance.rs
```diff
@@ -63,13 +63,17 @@ where
...
+        let local_time = self.client().get_time_checked().await?;
         let mut balance = Balance::default();
         let mut total_rent_amount = 0;
         let mut total_native_tokens = NativeTokensBuilder::default();
 
         #[cfg(feature = "participation")]
         let voting_output = self.get_voting_output().await?;
 
+        let account_addresses = self.addresses().await?;
+        let claimable_outputs = account_details.claimable_outputs(OutputsToClaim::All, local_time)?;
```
kwek20, line 76: this can be `self.claimable_outputs(OutputsToClaim::All)`

thibault-martinez: Can it ? Seems like the whole point of the PR is to avoid having to call `self.claimable_outputs`

Thoralf-M: Have you read the description? :trollface: Because it was like that before, it has led  to a deadlock (because internally it does `self.read()`)

kwek20: But so does self.addresses(), which isnt a problem? I suspect the real gain is not calling the lock twice per iteration (time and claimable_outputs)

Thoralf-M: True, self.addresses() can actually have the same problem and self.get_voting_output() as well
For that we would need to update also some more things that are called internally

thibault-martinez: I think you're right on this, should call addresses on the account_details?

Thoralf-M: Updated

kwek20 (dismissed):

sdk/src/wallet/account/operations/participation/mod.rs
```diff
@@ -232,13 +232,7 @@ where
     /// If multiple outputs with this tag exist, the one with the largest amount will be returned.
     pub async fn get_voting_output(&self) -> Result<Option<OutputData>> {
         log::debug!("[get_voting_output]");
```
thibault-martinez: I would put the log in the other one, otherwise it's not always logged

sdk/src/wallet/account/operations/output_claiming.rs
```diff
@@ -123,6 +121,26 @@ where
...
+    crate::wallet::Error: From<S::Error>,
+{
+    /// Get basic and nft outputs that have
+    /// [`ExpirationUnlockCondition`](crate::types::block::output::unlock_condition::ExpirationUnlockCondition),
+    /// [`StorageDepositReturnUnlockCondition`] or
+    /// [`TimelockUnlockCondition`](crate::types::block::output::unlock_condition::TimelockUnlockCondition) and can be
+    /// unlocked now and also get basic outputs with only an [`AddressUnlockCondition`] unlock condition, for
+    /// additional inputs
+    pub async fn claimable_outputs(&self, outputs_to_claim: OutputsToClaim) -> crate::wallet::Result<Vec<OutputId>> {
+        log::debug!("[OUTPUT_CLAIMING] claimable_outputs");
```
thibault-martinez: Maybe unnecessary

sdk/src/wallet/account/operations/participation/mod.rs
Thoralf-M: But the other one has no lock, so its not that interesting I think

sdk/src/wallet/account/operations/output_claiming.rs
Thoralf-M: Maybe

sdk/src/wallet/account/operations/participation/mod.rs
thibault-martinez: Why not that interesting? claimable_outputs won't log it anymore then, we're losing information?

Thoralf-M: Not that interesting because there is no lock involved and the logic there is also pretty simple

sdk/src/wallet/account/operations/output_claiming.rs
thibault-martinez: Because already logged in account_details.claimable_outputs I meant

sdk/src/wallet/account/operations/participation/mod.rs
thibault-martinez: I wouldn't say the logs are here because of locks... It's just interesting to know what has been called, when, in what order etc

sdk/src/wallet/account/operations/output_claiming.rs
Thoralf-M: I know, but could be interesting to know what has been called in what order etc :trollface:
Whatever, removed because now with this PR this place is fine

thibault-martinez (approved):

Alex6323 (approved):
maybe the PR title should indicate that a bug was fixed as well :shrug:

kwek20 (approved):

PR #1516, opened 2023-10-26
Semantic validation cleanup

thibault-martinez:


sdk/src/types/block/semantic.rs
```diff
@@ -209,294 +210,279 @@ impl<'a> ValidationContext<'a> {
...
+                if let Some(storage_deposit_return) = unlock_conditions.storage_deposit_return() {
+                    let amount = self
+                        .storage_deposit_returns
+                        .entry(storage_deposit_return.return_address().clone())
+                        .or_default();
 
-                *amount = amount
-                    .checked_add(storage_deposit_return.amount())
-                    .ok_or(Error::StorageDepositReturnOverflow)?;
+                    *amount = amount
```
kwek20, line 271: now that i see this again, are we sure this updates the correct amount variable?

thibault-martinez: Looks good to me

kwek20 (dismissed):

sdk/src/types/block/semantic.rs
```diff
@@ -147,11 +147,10 @@ impl TryFrom<u8> for TransactionFailureReason {
 }
 
 ///
-pub struct ValidationContext<'a> {
+pub struct SemanticValidationContext<'a> {
```
Alex6323, line 150: maybe add a doc comment now?

thibault-martinez: I think it's way too soon to add documentation on things that are 100% going to change, I can add a super generic one if you want but I don't really see the point

sdk/src/types/block/semantic.rs
```diff
@@ -209,294 +210,279 @@ impl<'a> ValidationContext<'a> {
...
-                0,
-                None,
-                output.unlock_conditions(),
-            ),
-        };
 
-        if let Err(conflict) = conflict {
-            return Ok(Some(conflict));
-        }
+    ///
```
Alex6323, line 214: why is there no clippy that disallows such docs? :trollface:

sdk/src/types/block/semantic.rs
```diff
@@ -209,294 +210,279 @@ impl<'a> ValidationContext<'a> {
...
+                let addresses = unlock_conditions
+                    .iter()
+                    .filter_map(|uc| match uc {
+                        UnlockCondition::Address(uc) => Some(uc.address()),
+                        UnlockCondition::Expiration(uc) => Some(uc.return_address()),
+                        UnlockCondition::StateControllerAddress(uc) => Some(uc.address()),
+                        UnlockCondition::GovernorAddress(uc) => Some(uc.address()),
+                        _ => None,
+                    })
+                    .filter_map(Address::as_restricted_opt);
```
Alex6323, line 346: it should be possible to merge those two `filter_map`s?

sdk/src/types/block/semantic.rs
```diff
@@ -209,294 +210,279 @@ impl<'a> ValidationContext<'a> {
...
-                if unlock_conditions.timelock().is_some()
-                    && !address.has_capability(AddressCapabilityFlag::OutputsWithTimelock)
-                {
-                    // TODO: add a variant https://github.com/iotaledger/iota-sdk/issues/1430
-                    return Ok(Some(TransactionFailureReason::SemanticValidationFailed));
-                }
+            self.output_amount = self
+                .output_amount
+                .checked_add(amount)
+                .ok_or(Error::CreatedAmountOverflow)?;
```
Alex6323, line 402: would be kinda neat to rename this error to `CreatedCoinOverflow` which nicely differentiates it from `CreatedManaOverflow` IMO.

sdk/src/types/block/semantic.rs
thibault-martinez: That just moves the line up a bit but it's the same, do you think it's better?

sdk/src/types/block/semantic.rs
thibault-martinez: I would keep amount as it's the field name

sdk/src/types/block/semantic.rs
```diff
@@ -209,294 +210,279 @@ impl<'a> ValidationContext<'a> {
...
+            // TODO: add a variant https://github.com/iotaledger/iota-sdk/issues/1430
+            return Ok(Some(TransactionFailureReason::SemanticValidationFailed));
         }
-    }
 
-    // Validation of amounts.
-    if context.input_amount != context.output_amount {
-        return Ok(Some(TransactionFailureReason::SumInputsOutputsAmountMismatch));
-    }
+        let mut native_token_ids = HashSet::new();
```
Alex6323: since we know the required max capacity maybe we should make use of that?

sdk/src/types/block/semantic.rs
thibault-martinez: Actually I'm going to keep it like that. There is an incoming PR where address capabilities check will be made syntactic instead of semantic. I am planning on moving this code directly to the UCs and provide `addresses` and `restricted_addresses` methods so it needs to be split.

sdk/src/types/block/semantic.rs
```diff
@@ -209,294 +210,279 @@ impl<'a> ValidationContext<'a> {
...
+        // Validation of state transitions and destructions.
+        for (chain_id, current_state) in self.input_chains.iter() {
+            if Output::verify_state_transition(
+                Some(current_state),
+                self.output_chains.get(chain_id).map(core::ops::Deref::deref),
+                &self,
+            )
+            .is_err()
+            {
+                return Ok(Some(TransactionFailureReason::InvalidChainStateTransition));
```
Alex6323, line 469: I feel like this error should receive the `chain id`? issue maybe?

thibault-martinez: It can't, this is an integer enum

sdk/src/types/block/semantic.rs
Alex6323: no, leave it as it is then if you think that can't be properly described yet.

sdk/src/types/block/semantic.rs
thibault-martinez: Done

sdk/src/types/block/semantic.rs
Alex6323: okay

sdk/src/types/block/semantic.rs
Alex6323: okidoki

sdk/src/types/block/semantic.rs
Alex6323: lol yeah

Alex6323 (approved):

kwek20 (approved):

PR #1525, opened 2023-10-29
Add new TransactionFailureReason

### LINKED ISSUE #1430: Add a state transition error / transaction failure reason for missing capabilities
## Description

We need an error variant for the case when a state transition is invalid because of missing capabilities.

See https://github.com/iotaledger/iota-core/issues/428

thibault-martinez:
Fixes https://github.com/iotaledger/iota-sdk/issues/1430

### DIFF
--- a/bindings/nodejs/lib/types/models/transaction-failure-reason.ts
+++ b/bindings/nodejs/lib/types/models/transaction-failure-reason.ts
@@ -110,32 +110,32 @@ export enum TransactionFailureReason {
     /**
      * Burning of native tokens is not allowed in the transaction capabilities.
      */
-    TransactionCapabilityNativeTokenBurningNotAllowed = 21,
+    transactionCapabilityNativeTokenBurningNotAllowed = 21,
 
     /**
      * Burning of mana is not allowed in the transaction capabilities.
      */
-    TransactionCapabilityManaBurningNotAllowed = 22,
+    transactionCapabilityManaBurningNotAllowed = 22,
 
     /**
      * Destruction of accounts is not allowed in the transaction capabilities.
      */
-    TransactionCapabilityAccountDestructionNotAllowed = 23,
+    transactionCapabilityAccountDestructionNotAllowed = 23,
 
     /**
      * Destruction of anchors is not allowed in the transaction capabilities.
      */
-    TransactionCapabilityAnchorDestructionNotAllowed = 24,
+    transactionCapabilityAnchorDestructionNotAllowed = 24,
 
     /**
      * Destruction of foundries is not allowed in the transaction capabilities.
      */
-    TransactionCapabilityFoundryDestructionNotAllowed = 25,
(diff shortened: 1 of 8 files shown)


Thoralf-M (dismissed):

bindings/nodejs/lib/types/models/transaction-failure-reason.ts
```diff
@@ -107,6 +107,36 @@ export enum TransactionFailureReason {
...
+    /**
+     * Destruction of foundries is not allowed in the transaction capabilities.
+     */
+    TransactionCapabilityFoundryDestructionNotAllowed = 25,
+
+    /**
+     * Destruction of nfts is not allowed in the transaction capabilities.
+     */
+    TransactionCapabilityNftDestructionNotAllowed = 26,
+
```
Alex6323: first char should be lower case?

bindings/python/iota_sdk/types/block/metadata.py
```diff
@@ -138,6 +144,12 @@ class TransactionFailureReason(Enum):
     MissingStakingFeature = 18
     FailedToClaimStakingReward = 19
     FailedToClaimDelegationReward = 20
+    TransactionCapabilityNativeTokenBurningNotAllowed = 21
+    TransactionCapabilityManaBurningNotAllowed = 22
+    TransactionCapabilityAccountDestructionNotAllowed = 23
+    TransactionCapabilityAnchorDestructionNotAllowed = 24
+    TransactionCapabilityFoundryDestructionNotAllowed = 25
+    TransactionCapabilityNftDestructionNotAllowed = 26
```
Alex6323, line 152: can we remove the prefix "TransactionCapability"? I think it should be clear from the context and the indexation (21...26) that those are related to transaction capabilities?

bindings/nodejs/lib/types/models/transaction-failure-reason.ts
thibault-martinez: Fixed

bindings/python/iota_sdk/types/block/metadata.py
thibault-martinez: I would prefer to keep them as explicit as possible.
Also not sure why the indexes would give any indication that it's about tx caps?

Thoralf-M (approved):

bindings/python/iota_sdk/types/block/metadata.py
Alex6323: because it's ordered by their corresponding index, not by name. So they will always be grouped like this.

thibault-martinez: They won't be grouped when/if we add more later

Alex6323: mmh, okay then

Alex6323 (approved):

PR #1526, opened 2023-10-29
ISA: handle restricted addresses

thibault-martinez:


sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -234,7 +235,11 @@ impl InputSelection {
                 .unwrap()
                 .0;
 
-            self.addresses.contains(&required_address)
+            if let Address::Restricted(restricted_address) = required_address {
+                self.addresses.contains(restricted_address.address())
+            } else {
+                self.addresses.contains(&required_address)
+            }
```
Alex6323, line 242: that's weird tbh , but I guess there's no other way?

thibault-martinez: What is weird exactly?

Alex6323: I would like to prevent doing the pattern matching on the user side of that enum, but instead let  the `Address` type have a method that returns the address to compare. Is that possible?

Thoralf-M (approved):

sdk/src/client/api/block_builder/input_selection/mod.rs
thibault-martinez: I guess it's possible but tbh I would keep it like that, more explicit, better shows intent, easier to understand when reading

Alex6323 (approved):

PR #1535, opened 2023-10-30
Fix `StateMetadataOutput`'s constructor not setting the `stateMetadata` field

thibault-martinez:


Thoralf-M (approved):

kwek20 (approved):

schierlm:
In case you want to complete other fields "forgotten" by other constructors of types (indirectly) referenced by block Payload subclases (as of 1.1.1-WASM), you can check

https://github.com/iota-laboratory/iota-laboratory.github.io/blob/03fd537e0a5d3be7d40b5b6a9657356b87cf4652/block_payload_types.js#L3-L178

for `cf:` (*constructor fields*) values:

- **TreasuryTransactionPayload**: No exposed constructor at all, but the hidden one has all fields
- **AliasOutput**: Misses (after this commit) `nativeTokens`, `features`, `immutableFeatures`
- **BasicOutput**: Misses `nativeTokens`, `features`
- **FoundryOutput**: Misses `nativeTokens`, `features`, `immutableFeatures`
- **MigratedFunds**, **NativeToken**: No exposed constructor at all
- **MilestonePayload**: Exposed constructor has no fields at all
- **NftOutput**: Misses `nativeTokens`, `features`, `immutableFeatures`

There is no urge to do that from my side, though. My workaround to dynamically tweak creation of those objects by setting the properties manually works fine for now.

// cc: @Thoralf-M

PR #1551, opened 2023-11-01
Add some checks for capabilities

### LINKED ISSUE #1524: Make sure empty Address/Transaction capabilities (de)serialize correctly
Continuation of https://github.com/iotaledger/iota-sdk/pull/1509

- [x] Rust
- [x] Nodejs
- [x] Python

### LINKED ISSUE #1545: Add checks for trailing zero bytes in capabilities
https://github.com/iotaledger/iota.go/pull/593

DaughterOfMars:
# Description of change

- [x] Trims trailing zeroes in rust when building capabilities
- [x] Checks for trailing zeroes when deserializing/unpacking
- [x] Skips empty bytes in rust and bindings

Closes #1524
Closes #1545


bindings/nodejs/lib/types/block/address.ts
```diff
@@ -239,13 +239,20 @@ class RestrictedAddress extends Address {
     }
 
     getAllowedCapabilities(): Uint8Array {
-        return Uint8Array.from(
-            Buffer.from(this.allowedCapabilities.substring(2), 'hex'),
-        );
+        return this.allowedCapabilities !== undefined
+            ? Uint8Array.from(
+                  Buffer.from(this.allowedCapabilities.substring(2), 'hex'),
```
Thoralf-M, line 244: Any reason to still store `allowedCapabilities` with the hex prefix? Looks like we add it in `setAllowedCapabilities()`, just to always remove when we use it

DaughterOfMars: because of the serialization

Thoralf-M (dismissed):

sdk/src/types/block/capabilities.rs
```diff
@@ -60,30 +91,28 @@ impl<Flag: CapabilityFlag> Capabilities<Flag> {
...
     }
 
     /// Enables a given set of flags.
     pub fn add_capabilities(&mut self, flags: impl IntoIterator<Item = Flag>) -> &mut Self {
         for flag in flags {
-            self.add_capability(flag);
+            self.inner_add_capability(flag, false);
         }
-        self
+        self.trim()
```
thibault-martinez: Do we actually need to trim? Adding capabilities should never end up adding an empty byte? Or am I missing something?

DaughterOfMars: 🤔 I think you're right

Thoralf-M (dismissed):

Thoralf-M (dismissed):

sdk/src/types/block/capabilities.rs
```diff
@@ -173,25 +181,22 @@ impl<Flag: 'static> Packable for Capabilities<Flag> {
     type UnpackVisitor = ();
 
     fn pack<P: packable::packer::Packer>(&self, packer: &mut P) -> Result<(), P::Error> {
-        if !self.is_none() {
-            self.bytes.pack(packer)?;
-        } else {
-            0_u8.pack(packer)?;
-        }
+        self.bytes.pack(packer)?;
```
thibault-martinez, line 200: Is that equivalent? I remember asking to switch to that at some point and there was a reason not to

sdk/src/types/block/capabilities.rs
```diff
@@ -21,11 +23,17 @@ pub struct Capabilities<Flag> {
 }
 
 impl<Flag> Capabilities<Flag> {
-    pub(crate) fn from_bytes(bytes: BoxedSlicePrefix<u8, u8>) -> Self {
-        Self {
+    /// Try to create capabilities from serialized bytes. Bytes with trailing zeroes are invalid.
+    pub(crate) fn from_bytes(bytes: BoxedSlicePrefix<u8, u8>) -> Result<Self, Error> {
+        if let Some(idx) = bytes.iter().rposition(|c| 0.ne(c)) {
+            if idx + 1 < bytes.len() {
```
thibault-martinez: Not too sure about that, it could find a non-trailing 0 byte

sdk/src/types/block/capabilities.rs
DaughterOfMars: It is now, since there's no longer the possibility of having trailing zeroes in the bytes

sdk/src/types/block/capabilities.rs
DaughterOfMars: It's searching for `v != 0` from the right

thibault-martinez: Let's say we have `[5, 12, 30, 5, 0]`, length is 5, idx found is 3 and 3 + 1 < 5 ?
Or am I too tired for this

DaughterOfMars: yes so that would be an error

thibault-martinez: Yeah sorry I completely brainfarted.
Maybe rework into using `slice::last` would make it easier to understand.

thibault-martinez (dismissed):

sdk/src/types/block/capabilities.rs
DaughterOfMars: it has to be this way because there can be multiple trailing zeroes

thibault-martinez: Sure but multiple always include the last one, no need to check more than that

DaughterOfMars: Hm...that's a really good point 😆

sdk/src/types/block/capabilities.rs
```diff
@@ -207,3 +228,163 @@ pub trait CapabilityFlag {
...
+    impl TestFlag {
+        const VAL_1: u8 = 0b00000001;
+        const VAL_2: u8 = 0b00000010;
+        const VAL_3: u8 = 0b00000100;
+        const VAL_4: u8 = 0b00001000;
+        const VAL_5: u8 = 0b00010000;
+        const VAL_6: u8 = 0b00100000;
+        const VAL_7: u8 = 0b01000000;
+        const VAL_8: u8 = 0b10000000;
+        const VAL_9: u8 = 0b00000001;
```
Thoralf-M, line 289: Why is this the same as VAL_1?

DaughterOfMars: Because that's how the flags work, they have to be one of the 8 possible bits, and each is assigned to an index in the byte array

Thoralf-M: Ah in index() it's used different, thanks

Thoralf-M (dismissed):

thibault-martinez (dismissed):

sdk/src/types/block/error.rs
```diff
@@ -397,6 +405,7 @@ impl fmt::Display for Error {
             Self::InvalidEpochDelta { created, target } => {
                 write!(f, "invalid epoch delta: created {created}, target {target}")
             }
+            Self::TrailingCapabilityBytes => write!(f, "capabilities bytes has trailing zeroes"),
```
tadzik: Minor:

```suggestion
            Self::TrailingCapabilityBytes => write!(f, "capability bytes have trailing zeroes"),
```

tadzik (dismissed):

thibault-martinez (approved):

Thoralf-M (approved):

tadzik (approved):

PR #1552, opened 2023-11-01
Use ints for U256 values in Python

### LINKED ISSUE #1494: Python use int for u128 values
## Description

Python use int for u128 values, similar to https://github.com/iotaledger/iota-sdk/issues/1163, just that u128 is encoded as HexStr instead of a normal string

- [ ] https://github.com/iotaledger/iota-sdk/pull/1482/files#r1370711780

## Motivation

Easier to use

No.

DaughterOfMars:
# Description of change

Uses `int` for big numbers in python.

Closes #1494


bindings/python/iota_sdk/types/token_scheme.py
```diff
@@ -17,9 +18,15 @@ class SimpleTokenScheme:
         maximum_supply: The maximum supply of the token.
         type: The type code of the token scheme.
     """
-    minted_tokens: HexStr
-    melted_tokens: HexStr
-    maximum_supply: HexStr
+    minted_tokens: int = field(metadata=config(
+        encoder=str
```
thibault-martinez: Does that actually encode to prefixed hex ?

DaughterOfMars: oh, no, I didn't realize it was hex

Thoralf-M: decoding from hex works by default?

thibault-martinez (changes requested):
`CreateNativeTokenParams` also needs to be adapted

bindings/python/iota_sdk/types/token_scheme.py
DaughterOfMars: I don't think so? but tbh I have no idea

DaughterOfMars: it did not, but I fixed it

Thoralf-M (dismissed):

bindings/python/iota_sdk/types/balance.py
```diff
@@ -62,8 +62,14 @@ class NativeTokensBalance:
...
     token_id: HexStr
-    total: HexStr
-    available: HexStr
+    total: int = field(metadata=config(
+        encoder=hex,
+        decoder=lambda v: int(v, 0)
+    ))
+    available: int = field(metadata=config(
+        encoder=hex,
+        decoder=lambda v: int(v, 0)
```
thibault-martinez: There is no `hex` for decoder ?

DaughterOfMars: the decoder defaults to `int` because that's the type, but it cannot decode hex without the second argument (radix, which is 0 to indicate that python should figure that part out)

Thoralf-M (dismissed):

bindings/python/iota_sdk/types/send_params.py
```diff
@@ -72,10 +72,12 @@ class CreateNativeTokenParams():
...
     circulating_supply: int = field(metadata=config(
-        encoder=str
+        encoder=hex,
+        decoder=lambda v: int(v, 0)
     ))
     maximum_supply: int = field(metadata=config(
-        encoder=str
+        encoder=hex,
+        decoder=lambda v: int(v, 0)
     ))
```
thibault-martinez, line 81: I wonder if we could store the whole `field` in a helper function somewhere and just do `maximum_supply: int = hex_int` or something?

kwek20 (dismissed):

bindings/python/iota_sdk/types/send_params.py
DaughterOfMars: hmm, yeah, probably

kwek20 (approved):

thibault-martinez (approved):

Alex6323 (approved):

Thoralf-M (approved):

PR #1557, opened 2023-11-02
Refactor hex decoding

### LINKED ISSUE #1555: Avoid Python U256 decoder repetition?
As seen in https://github.com/iotaledger/iota-sdk/pull/1552, things like 
```python
    amount: int = field(metadata=config(
        encoder=hex,
        decoder=lambda v: int(v, 16)
    ))
```
should probably be factored.
Maybe something like 
```python
def HexField(**args):
    return field(metadata=config(
        encoder=hex,
        decoder=lambda v: int(v, 16),
        **args,
    ))
```

tadzik:
# Description of change

Factors out the hex string decoding. I'm following the pattern established by `opt_int_encoder` rather than a full-blown custom field like in the issue description, to keep up with existing conventions and not introduce too much magic.

Fixes #1555 

## Type of change

- Refactoring

I didn't get `tox` to work yet, so I'm hoping the CI will run the tests for me :)

## Change checklist

### DIFF
--- a/bindings/python/iota_sdk/types/common.py
+++ b/bindings/python/iota_sdk/types/common.py
@@ -113,7 +113,7 @@ def opt_int_encoder(value):
 
 
 def hex_str_decoder(value: str) -> int:
-    """ Parses a given string as a hexadecimal integer """
+    """Parses a given string as a hexadecimal integer."""
     return int(value, 16)
 
 
--- a/bindings/python/iota_sdk/types/common.py
+++ b/bindings/python/iota_sdk/types/common.py
@@ -112,6 +112,11 @@ def opt_int_encoder(value):
     return None
 
 
+def hex_str_decoder(value: str) -> int:
+    """ Parses a given string as a hexadecimal integer """
+    return int(value, 16)
+
+
 @json
 @dataclass
 class AddressAndAmount():
--- a/bindings/python/iota_sdk/types/balance.py
+++ b/bindings/python/iota_sdk/types/balance.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from typing import List, Optional
 from dataclasses import dataclass, field
 from dataclasses_json import config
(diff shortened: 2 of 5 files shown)


bindings/python/iota_sdk/types/common.py
```diff
@@ -112,6 +112,11 @@ def opt_int_encoder(value):
     return None
 
 
+def hex_str_decoder(value: str) -> int:
+    """ Parses a given string as a hexadecimal integer """
```
Thoralf-M: Any reason for the extra spaces?
```suggestion
    """Parses a given string as a hexadecimal integer."""
```

tadzik: Just my aesthetic choice. ~~`pep8`~~`pycodestyle` doesn't care either way it seems, but I'm happy to follow existing conventions.

tadzik:
Testing deferred to https://github.com/iotaledger/iota-sdk/issues/1558 while the bindings are borked.

We have existing unit tests that sort of cover this, but a) they don't run b) I can't check if they'd break if this did. Will come back to it later.

Thoralf-M (approved):

DaughterOfMars (approved):

thibault-martinez:
Just checking in, did you come to the conclusion it was not worth to do it for the whole `field`?

Alex6323 (approved):

tadzik:
> Just checking in, did you come to the conclusion it was not worth to do it for the whole `field`?

I didn't see any prior art (in our code or otherwise) in building custom `field`s like that, but we (and the `dataclass_json` docs) both use the thingy with custom encoders and decoders – so I figured I'd stick to that. It also avoids the problem of potentially having to extend the field initializer in the future were we to customize it somehow, which @DaughterOfMars brought up yesterday.

thibault-martinez (approved):

PR #1560, opened 2023-11-03
Bindings core: merge wallet methods into single enum

Thoralf-M:
# Description of change

Bindings core: merge wallet methods into single enum, because they are now all called on the wallet and not on a separated account anymore

## Type of change

`cargo test`


bindings/core/src/method/wallet.rs
```diff
@@ -25,13 +43,6 @@ use crate::OmittedDebug;
 #[serde(tag = "name", content = "data", rename_all = "camelCase")]
 #[non_exhaustive]
 pub enum WalletMethod {
```
thibault-martinez, line 45: We could take the opportunity to sort it but we can do that later

thibault-martinez (dismissed):

bindings/core/src/method/wallet.rs
Thoralf-M: Already thought you will comment on that :trollface: But yes I would wait a bit until we decided what we want to keep/remove

Alex6323 (dismissed):

kwek20 (changes requested):
Missing the Response comment changes, like here:
https://github.com/iotaledger/iota-sdk/blob/54228655694e397d5a8f0107cf56250876c45b77/bindings/core/src/response.rs#L240C22-L240C22

kwek20 (approved):

thibault-martinez (approved):

Alex6323 (approved):

PR #1563, opened 2023-11-03
Fail tests early when "storage" not enabled

tadzik:
# Description of change

The tests currently fail in confusing ways, due to "storage" not being enabled by default in bindings. Curiously, it does seem to be enabled in the wallet (running [this code](https://github.com/iotaledger/iota-sdk/blob/d6f6250ba3d14003dfb252fd504d0adcd842b436/sdk/src/wallet/core/builder.rs#L135-L156), and causing test failures when ran concurrently), breaking the functionality of each test using a custom, unique storage location.

This makes the test fail early (in compile time) if storage is not enabled, which makes sense given how heavily they rely on it. README now also documents how the tests should be run (we cannot force it, [apparently](https://github.com/rust-lang/cargo/issues/2911)).

## Type of change

- Bug fix – in case storage was disabled in bindings but used in the code
- Documentation Fix

`cargo test` now passes in rootdir when "storage" is enabled
 
## Change checklist


README.md
```diff
@@ -104,6 +104,14 @@ brew install cmake openssl@1.1
 You can install `cmake`, `clang`, and `openssl` with your distro's package manager or download them from their websites.
 On Debian and Ubuntu, you will also need the `build-essential` and `libudev-dev` packages.
 
+### Testing
+
+When working on the SDK itself, run its tests using `cargo` with `storage` feature enabled:
```
Thoralf-M: Not sure about this, I mean tests should also work fine without enabling this feature

tadzik: Yeah, I'm not thrilled about this either. Then again, my gut feeling is that our tests should exercise everything by default, and not leave us in the situation where we only test some of the stuff unless we remember to enable the entire featureset. For that reason, I'd rather have them (visibly) not work out of the box, rather than quietly skip stuff we want tested.

Thoralf-M: But then it should rather suggest testing with --all-features, no? Because we have more features not enabled by default than the storage (also rocksdb has an own feature, so this is also different to storage alone)

Alex6323: Why not just suggest running what the CI runs, so `cargo ci-test`?

thibault-martinez: I think I agree with Thoralf, we should suggest `-all-features`. And if there has been a gap in the feature set, it's a CI issue that we need to fix.

tadzik: I like @Alex6323's suggestion, any reason to not direct people to `ci-test` (after instructing them to install `nextest`)?

thibault-martinez: I think it's just confusing to locally run something that is called CI. We could rename it idk 🤷🏻‍♂️

thibault-martinez: But IMO I'd just give the whole command and be explicit

Thoralf-M (approved):

thibault-martinez (approved):

PR #1566, opened 2023-11-03
Add {BasicOutput, Output}::is_implicit_account

thibault-martinez:
https://github.com/iotaledger/tips/blob/tip42/tips/TIP-0042/tip-0042.md#creation-semantic-transaction-validation-rules

### DIFF
--- a/sdk/src/types/block/output/basic.rs
+++ b/sdk/src/types/block/output/basic.rs
@@ -312,9 +312,9 @@ impl BasicOutput {
     /// Simple deposit outputs are basic outputs with only an address unlock condition, no native tokens and no
     /// features. They are used to return storage deposits.
     pub fn simple_deposit_address(&self) -> Option<&Address> {
-        if let [UnlockCondition::Address(address)] = self.unlock_conditions().as_ref() {
+        if let [UnlockCondition::Address(uc)] = self.unlock_conditions().as_ref() {
             if self.mana == 0 && self.native_tokens.is_empty() && self.features.is_empty() {
-                return Some(address.address());
+                return Some(uc.address());
             }
         }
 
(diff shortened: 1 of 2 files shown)


Thoralf-M (dismissed):

kwek20 (dismissed):

sdk/src/types/block/output/basic.rs
```diff
@@ -320,6 +320,14 @@ impl BasicOutput {
 
         None
     }
+
+    /// Checks whether the basic output is an implicit account.
+    pub fn is_implicit_account(&self) -> bool {
+        self.unlock_conditions.len() == 1
```
Alex6323: any unlock condition does suffice?

thibault-martinez: You can't build a BasicOutput without an address so if there is only one UC, it has to be an address

sdk/src/types/block/output/mod.rs
```diff
@@ -252,6 +252,15 @@ impl Output {
         }
     }
 
+    /// Checks whether the output is an implicit account.
+    pub fn is_implicit_account(&self) -> bool {
+        if let Output::Basic(output) = self {
+            output.is_implicit_account()
+        } else {
+            false
+        }
```
Alex6323, line 261: could be made into a one-liner I think:
```rust
matches!(self, Output::Basic(output) if output.is_implicit_account())
```

thibault-martinez: Are you strongly opinionated about that? I think the if/let version is easier to humanly parse

sdk/src/types/block/output/basic.rs
Alex6323: sorry, wasn't clear enough, the question was whether we should be explicit here rather then just checking that there is a UC (which then has to be an address UC)

sdk/src/types/block/output/mod.rs
Alex6323: no, not at all, just wanted to mention it if you'ld prefer it a bit shorter. I try to remember your preference.

Alex6323 (dismissed):

kwek20 (approved):

Alex6323 (approved):

PR #1568, opened 2023-11-03
Allow filtering anchors/delegations and make filtering infallible

thibault-martinez:
- Make all filtering methods infallible
- Add filtering options for anchors and delegations

### DIFF
--- a/sdk/src/wallet/core/mod.rs
+++ b/sdk/src/wallet/core/mod.rs
@@ -373,7 +373,7 @@ where
                     }
                 }
 
-                // If ids are provided, only return them and no other outputs.
+                // Include the output if we're not filtering by IDs.
                 if filter.account_ids.is_none()
                     && filter.anchor_ids.is_none()
                     && filter.foundry_ids.is_none()
--- a/sdk/src/wallet/core/mod.rs
+++ b/sdk/src/wallet/core/mod.rs
@@ -296,7 +296,7 @@ where
         &self,
         outputs: impl Iterator<Item = &'a OutputData>,
         filter: impl Into<Option<FilterOptions>>,
-    ) -> Result<Vec<OutputData>> {
+    ) -> Vec<OutputData> {
         let filter = filter.into();
 
         if let Some(filter) = filter {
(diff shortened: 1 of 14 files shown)


tadzik (dismissed):

kwek20 (dismissed):

sdk/src/wallet/core/mod.rs
```diff
@@ -358,50 +379,75 @@ where
                 }
```
Thoralf-M, line 384: Can't comment there, L377 needs additional anchors, delegations .is_none()

thibault-martinez: Good catch, done

Thoralf-M (dismissed):

sdk/src/wallet/core/mod.rs
```diff
@@ -353,55 +374,85 @@ where
...
                 // If ids are provided, only return them and no other outputs.
-                if filter.account_ids.is_none() && filter.foundry_ids.is_none() && filter.nft_ids.is_none() {
+                if filter.account_ids.is_none()
+                    && filter.anchor_ids.is_none()
+                    && filter.foundry_ids.is_none()
+                    && filter.nft_ids.is_none()
+                    && filter.delegation_ids.is_none()
+                {
                     filtered_outputs.push(output.clone());
                 }
```
Alex6323, line 384: A bit weird this `if` tbh. Also ... seems like something that can easily introduce a bug. I wonder if there's a better way. Also ... is the comment correct?

thibault-martinez: Comment looks correct to me yeah

Alex6323 (dismissed):

sdk/src/wallet/core/mod.rs
tadzik: > seems like something that can easily introduce a bug

Yeah, I believe that's what https://github.com/iotaledger/iota-sdk/pull/1568/files#r1382091913 was.

I can't think of a way to make this compile-time-impossible to forget, but we can at least make sure that we have a unit test that takes this path to maximize our chances.

tadzik: The comment's correct, but it did confuse me too due to describing an `else` of that `if`, if that makes sense. The code itself reads more like something like

`// Always include the output if we're not filtering by IDs`

to me.

tadzik (dismissed):

sdk/src/wallet/core/mod.rs
thibault-martinez: I'll put your comment but without the always as it's a bit too strong, it still depends on the other filtering options

Alex6323 (approved):

tadzik (approved):

Thoralf-M (approved):

kwek20 (approved):

Issue #1565, opened 2023-11-03
Refactor semantic errors


thibault-martinez:
- [x] Some variants of `TransactionFailureReason` should be remapped to `StateTransitionError`
- [x] Do we need both ?
- [x] Do we need to respect the TIP for the client facing error ?
- [x] https://github.com/iotaledger/iota-sdk/pull/1572#pullrequestreview-1757551310

https://github.com/iotaledger/iota-core/blob/develop/pkg/retainer/retainer/errors.go

thibault-martinez:
Closed by https://github.com/iotaledger/iota-sdk/pull/2058

PR #1572, opened 2023-11-05
Semantic validation optional unlocks

### LINKED ISSUE #1570: Have semantic_validation() for tx essence/payload without unlocks
## Description

Have semantic_validation() for tx essence/payload with the unlocks optional. And make it available to the bindings.

## Motivation

Currently semantic_validation() accepts the signed transaction, but it would also make sense to have an option to validate the tx before signing, so one doesn't have to resign if something is wrong.

No.

thibault-martinez:
Closes https://github.com/iotaledger/iota-sdk/issues/1570


bindings/core/src/method_handler/utils.rs
```diff
@@ -111,18 +111,26 @@ pub(crate) fn call_utils_method_internal(method: UtilsMethod) -> Result<Response
...
+        } => {
+            let transaction = Transaction::try_from_dto(transaction)?;
+            let inputs = &inputs
+                .into_iter()
+                .map(InputSigningData::try_from)
+                .collect::<iota_sdk::client::Result<Vec<InputSigningData>>>()?;
+            let inputs = inputs
+                .iter()
+                .map(|input| (input.output_id(), &input.output))
+                .collect::<Vec<(&OutputId, &Output)>>();
```
Thoralf-M, line 121: Why iterate twice?

sdk/src/types/block/semantic.rs
```diff
@@ -342,6 +313,26 @@ impl<'a> SemanticValidationContext<'a> {
                     .checked_add(consumed_native_token.amount())
                     .ok_or(Error::ConsumedNativeTokensAmountOverflow)?;
             }
+
+            if let Some(unlocks) = self.unlocks {
+                if unlocks.len() != self.inputs.len() {
+                    return Ok(Some(TransactionFailureReason::InvalidInputUnlock));
```
Thoralf-M, line 319: This error is not really correct if one is just missing

thibault-martinez: There is no better error 🤷🏻‍♂️

Thoralf-M: I don't like this limited API error things

bindings/core/src/method_handler/utils.rs
thibault-martinez: The first one is to collect and return the conversion error, I didn't want to bother much because we're supposedly removing the DTO soon

Thoralf-M (dismissed):

Thoralf-M (approved):

kwek20 (commented):
cant we return Response::ok instead of TransactionFailureReason(None) ?

thibault-martinez:
> cant we return Response::ok instead of TransactionFailureReason(None) ?

I have added your comment in https://github.com/iotaledger/iota-sdk/issues/1565 because this is linked. Let's keep as is for now.

sdk/src/types/block/semantic.rs
```diff
@@ -342,6 +313,26 @@ impl<'a> SemanticValidationContext<'a> {
...
+                }
+
+                let unlock = &unlocks[index];
```
DaughterOfMars, line 322: Should we do a safe get here?

thibault-martinez: There is a check that unlocks and inputs have the same length right before. And the index comes from iterating the inputs.  Do you still prefer get + unwrap?

Alex6323 (approved):

kwek20 (approved):

sdk/src/types/block/semantic.rs
DaughterOfMars: Ah, that's fine then

DaughterOfMars (approved):

PR #1573, opened 2023-11-05
Add MultiAddress

thibault-martinez:
Implements https://github.com/iotaledger/tips/blob/tip52/tips/TIP-0052/tip-0052.md without unlocks.


sdk/src/types/block/address/bech32.rs
```diff
@@ -172,11 +176,15 @@ impl Bech32Address {
...
 impl core::fmt::Display for Bech32Address {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(
-            f,
-            "{}",
-            bech32::encode::<bech32::Bech32>(self.hrp.0, &self.inner.pack_to_vec(),).unwrap()
-        )
+        let bytes = if self.inner.is_multi() {
+            let mut bytes = vec![MultiAddress::KIND];
+            bytes.extend(Blake2b256::digest(self.inner.pack_to_vec()));
```
thibault-martinez: I don't really like that extend, any idea ?

sdk/src/types/block/address/multi.rs
```diff
@@ -0,0 +1,225 @@
...
+                threshold: *threshold,
+            });
+        }
+    }
+    Ok(())
+}
+
+impl fmt::Display for MultiAddress {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", prefix_hex::encode(self.pack_to_vec()))
```
thibault-martinez: We do that for RestrictedAddress already, idk if it's really ideal

sdk/src/types/block/address/bech32.rs
DaughterOfMars: you could have fun with rshift 🤷‍♀️

DaughterOfMars: although I think the best solution may be a chain and collect since the packed vec will need to re-allocate anyway most likely

thibault-martinez: Wouldn't that be the same amount of allocations?

sdk/src/types/block/address/multi.rs
```diff
@@ -0,0 +1,226 @@
...
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct MultiAddress {
+    /// The weighted unlocked addresses.
+    addresses: BoxedSlicePrefix<WeightedAddress, WeightedAddressCount>,
+    /// The threshold that needs to be reached by the unlocked addresses in order to unlock the multi address.
+    threshold: u16,
+}
+
+impl MultiAddress {
+    /// The [`Address`](crate::types::block::address::Address) kind of an [`MultiAddress`].
```
DaughterOfMars:
```suggestion
    /// The [`Address`](crate::types::block::address::Address) kind of a [`MultiAddress`].
```

sdk/src/types/block/address/multi.rs
```diff
@@ -0,0 +1,226 @@
...
+    pub const KIND: u8 = 40;
+    /// The allowed range of inner [`Address`]es.
+    pub const ADDRESSES_COUNT: RangeInclusive<u8> = 1..=10;
+
+    /// Creates a new [`MultiAddress`].
+    #[inline(always)]
+    pub fn new(addresses: Vec<WeightedAddress>, threshold: u16) -> Result<Self, Error> {
```
DaughterOfMars: impl Iter instead? then collect to boxed slice?

sdk/src/types/block/address/multi.rs
DaughterOfMars: what do you mean?

thibault-martinez: Idk, it's kind of a useless Display tbh, not a format anyone can do something with

sdk/src/types/block/address/bech32.rs
DaughterOfMars: maybe. not sure if the ```let mut bytes = vec![MultiAddress::KIND]``` will need to reallocate because you didn't specify the capacity.

thibault-martinez: Pretty sure extend will reallocate yes. You're saying that wouldn't be necessary with collect because of size hints ?

sdk/src/types/block/address/multi.rs
thibault-martinez: Iterator or IntoIterator?

sdk/src/types/block/address/bech32.rs
DaughterOfMars: I believe so yes. But that may actually be true with extend as well, the compiler is a mystery sometimes

sdk/src/types/block/address/multi.rs
DaughterOfMars: Oh, I see. Well then let's change it!

sdk/src/types/block/address/multi.rs
DaughterOfMars: probably IntoIterator

sdk/src/types/block/address/bech32.rs
thibault-martinez: I changed to chain, I prefer it anyway

sdk/src/types/block/address/multi.rs
thibault-martinez: To what ? I feel like Bech32 would be a bit inconsistent as for the other Addresses, Display prints the inner ID/PubKeyHash

sdk/src/types/block/address/multi.rs
thibault-martinez: Done

sdk/src/types/block/address/multi.rs
DaughterOfMars: Maybe just a list of the inner addresses?

Thoralf-M: Or how about displaying the dto?

thibault-martinez: Hmm idk, I think the full JSON of this is way too detailed for Display, think I'll go with array of inner addresses

thibault-martinez: Looks like that
```
[0x52fdfc072182654f163f5f0f9a621d729566c74d10037c4d7bbb0407d1e2c649, 0x53fdfc072182654f163f5f0f9a621d729566c74d10037c4d7bbb0407d1e2c649, 0x54fdfc072182654f163f5f0f9a621d729566c74d10037c4d7bbb0407d1e2c649, 0x55fdfc072182654f163f5f0f9a621d729566c74d10037c4d7bbb0407d1e2c649, 0x56fdfc072182654f163f5f0f9a621d729566c74d10037c4d7bbb0407d1e2c649]
```

sdk/tests/types/address/multi.rs
```diff
@@ -0,0 +1,59 @@
...
+              "nftId": "0x56fdfc072182654f163f5f0f9a621d729566c74d10037c4d7bbb0407d1e2c649"
+            },
+            "weight": 3
+          }
+        ],
+        "threshold": 2
+    });
+    let multi_address = serde_json::from_value::<Address>(multi_address_json).unwrap();
+
+    println!("{}", multi_address);
```
Thoralf-M: Want to keep this?
```suggestion
```

sdk/src/types/block/address/multi.rs
Thoralf-M: Not sure if I like it, one can't even see the address type

sdk/tests/types/address/multi.rs
thibault-martinez: Whoops, removed

sdk/src/types/block/address/multi.rs
thibault-martinez: It just forwards to the display impls of the rest of the addresses. We Display only the hex IDs for the addresses. We may want to reconsider what we want as Display but not in this PR I'd say.

thibault-martinez: https://github.com/iotaledger/iota-sdk/issues/1587

Thoralf-M (dismissed):

sdk/src/types/block/address/bech32.rs
```diff
@@ -172,11 +176,15 @@ impl Bech32Address {
...
-        )
+        let bytes = if self.inner.is_multi() {
+            core::iter::once(MultiAddress::KIND)
+                .chain(Blake2b256::digest(self.inner.pack_to_vec()))
+                .collect()
+        } else {
+            self.inner.pack_to_vec()
+        };
+
+        write!(f, "{}", bech32::encode::<bech32::Bech32>(self.hrp.0, &bytes,).unwrap())
```
Alex6323:
```suggestion
        write!(f, "{}", bech32::encode::<bech32::Bech32>(self.hrp.0, &bytes).unwrap())
```

thibault-martinez: Oh no, why is fmt not removing this 😭

sdk/src/types/block/address/bech32.rs
```diff
@@ -172,11 +176,15 @@ impl Bech32Address {
...
-        )
+        let bytes = if self.inner.is_multi() {
+            core::iter::once(MultiAddress::KIND)
+                .chain(Blake2b256::digest(self.inner.pack_to_vec()))
+                .collect()
+        } else {
+            self.inner.pack_to_vec()
+        };
+
+        write!(f, "{}", bech32::encode::<bech32::Bech32>(self.hrp.0, &bytes).unwrap())
```
Alex6323, line 188: We never documented this unwrap btw. we don't care here?

Alex6323 (dismissed):

sdk/src/types/block/address/bech32.rs
thibault-martinez: documented it

DaughterOfMars (approved):

sdk/src/types/block/error.rs
```diff
@@ -261,6 +270,21 @@ impl fmt::Display for Error {
...
+                cumulative_weight,
+                threshold,
+            } => {
+                write!(
+                    f,
+                    "invalid multi address cumulative weight {cumulative_weight} < threshold {threshold}"
+                )
+            }
+            Self::InvalidWeightedAddressCount(count) => write!(f, "invalid weighted address count: {count}"),
+            Self::WeightedAddressesNotUniqueSorted => {
```
kwek20, line 285: didnt we address sorting using a ordered list or something before?

thibault-martinez: No, there was an attempt PR but it's been closed

Alex6323 (approved):

Thoralf-M (approved):

sdk/src/types/block/error.rs
kwek20: i see, treeset isnt boxed either, too bad

kwek20 (approved):

PR #1582, opened 2023-11-06
Allow some wallet data to be accessible by reference

DaughterOfMars:
# Description of change

Allows wallet data outputs and transactions to be accessible without cloning.


Thoralf-M (dismissed):

cli/src/wallet_cli/mod.rs
```diff
@@ -308,7 +307,7 @@ impl FromStr for OutputSelector {
 
 // `accounts` command
 pub async fn accounts_command(wallet: &Wallet) -> Result<(), Error> {
-    print_outputs(wallet.accounts().await, "Accounts:").await
+    print_outputs(wallet.data().await.accounts().cloned().collect(), "Accounts:")
```
kwek20, line 315: we really need to clone here?

kwek20: oef .cloned() all over, not nice :/

thibault-martinez: The clones were implicitly done by the non-iterative methods so yeah we need to explicitly clone now. But it means that it's way more efficient in the majority of the cases.

kwek20 (approved):

kwek20 (dismissed):

bindings/core/src/method_handler/wallet.rs
```diff
@@ -399,13 +422,28 @@ pub(crate) async fn call_wallet_method_internal(wallet: &Wallet, method: WalletM
...
+                    .filtered_unspent_outputs(filter)
+                    .map(OutputDataDto::from)
+                    .collect()
+            } else {
+                wallet_data
+                    .unspent_outputs()
+                    .values()
+                    .map(OutputDataDto::from)
+                    .collect()
+            })
```
Thoralf-M, line 430:
```suggestion
    let outputs = if let Some(filter) = filter_options {
        wallet_data
            .filtered_unspent_outputs(filter)
    } else {
        wallet_data
            .unspent_outputs()
            .values()
    };
    Response::OutputsData(outputs.map(OutputDataDto::from).collect())
```

DaughterOfMars: Don't think this will work as-is, because the iterator types are different

DaughterOfMars: I double checked and, indeed, this does not work :/

Thoralf-M (dismissed):

Thoralf-M (dismissed):

Alex6323 (dismissed):

kwek20 (dismissed):

sdk/src/wallet/operations/transaction/high_level/send_nft.rs
```diff
@@ -95,15 +95,13 @@ where
...
-        let unspent_outputs = self.unspent_outputs(None).await;
-
         let mut outputs = Vec::new();
 
         for SendNftParams { address, nft_id } in params {
             self.client().bech32_hrp_matches(address.hrp()).await?;
 
             // Find nft output from the inputs
-            if let Some(nft_output_data) = unspent_outputs.iter().find(|o| {
+            if let Some(nft_output_data) = self.data().await.unspent_outputs().values().find(|o| {
```
thibault-martinez: Looks like this could use `unspent_nft_output` instead ?

sdk/src/wallet/operations/syncing/mod.rs
```diff
@@ -98,11 +98,12 @@ where
         let wallet_address_with_unspent_outputs = AddressWithUnspentOutputs {
             address: self.address().await,
             output_ids: self
-                .unspent_outputs(None)
+                .data()
                 .await
-                .into_iter()
+                .unspent_outputs()
+                .values()
```
thibault-martinez: Just do `keys()` instead of values + map to OutputId ?

cli/src/wallet_cli/mod.rs
```diff
@@ -818,11 +821,12 @@ pub async fn sync_command(wallet: &Wallet) -> Result<(), Error> {
 
 /// `transaction` command
 pub async fn transaction_command(wallet: &Wallet, selector: TransactionSelector) -> Result<(), Error> {
-    let mut transactions = wallet.transactions().await;
+    let wallet_data = wallet.data().await;
     let transaction = match selector {
-        TransactionSelector::Id(id) => transactions.into_iter().find(|tx| tx.transaction_id == id),
+        TransactionSelector::Id(id) => wallet_data.transactions().find(|tx| tx.transaction_id == id),
```
thibault-martinez: I think it's a bit weird that `WalletData::outputs` return the hashmap and `WelletData::transactions` returns an iterator

Thoralf-M (approved):

thibault-martinez (approved):

Issue #1424, opened 2023-10-12
Update NodeJs binding to support "single account wallet"


Alex6323:

thibault-martinez:
Closed by https://github.com/iotaledger/iota-sdk/pull/1580

PR #1585, opened 2023-11-07
Update WalletOptions + some Clippy

Thoralf-M:
# Description of change

Update WalletOptions for single account wallet + some Clippy 

Related to #1424 

## Type of change

### DIFF
--- a/bindings/core/src/lib.rs
+++ b/bindings/core/src/lib.rs
@@ -17,6 +17,7 @@ use fern_logger::{logger_init, LoggerConfig, LoggerOutputConfigBuilder};
 pub use iota_sdk;
 use iota_sdk::{
     client::secret::{SecretManager, SecretManagerDto},
+    types::block::address::Bech32Address,
     wallet::{ClientOptions, Wallet},
 };
 use serde::Deserialize;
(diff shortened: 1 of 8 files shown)


bindings/core/src/lib.rs
```diff
@@ -42,14 +43,21 @@ pub fn init_logger(config: String) -> std::result::Result<(), fern_logger::Error
...
     pub secret_manager: Option<SecretManagerDto>,
+    pub storage_path: Option<String>,
 }
 
 impl WalletOptions {
+    pub fn with_address(mut self, address: impl Into<Option<Bech32Address>>) -> Self {
+        self.address = address.into();
+        self
+    }
+
```
Alex6323: why no `with_alias`?

Alex6323: also ... a bit nit-pickish ... but can you align the method order with the field order?

bindings/core/src/lib.rs
```diff
@@ -76,6 +84,15 @@ impl WalletOptions {
...
 
+        if let Some(address) = self.address {
+            builder = builder.with_address(address);
+        }
+        if let Some(alias) = self.alias {
+            builder = builder.with_alias(alias);
+        }
+        if let Some(bip_path) = self.bip_path {
+            builder = builder.with_bip_path(bip_path);
+        }
```
Alex6323: isn't that already handled in L85?

kwek20 (dismissed):

kwek20 (commented):
alex was better, unapproved from me :)

bindings/core/src/lib.rs
Thoralf-M: Wondering if we should remove them all, they're not used apart from tests
But added and aligned

bindings/core/src/lib.rs
Thoralf-M: Good catch

kwek20 (approved):

DaughterOfMars (approved):

Alex6323 (approved):

thibault-martinez (approved):

PR #1586, opened 2023-11-07
Remove ed25519 address generation from wallet

### LINKED ISSUE #1536: Remove generate_ed25519_address() from the wallet?
## Description

Move generate_ed25519_address() out of the wallet? Users can just generate with the secret manager before and then provide it to the wallet

## Motivation

Simplification and clearer logic, no secret manager required for the wallet at this point

## Open questions (optional)

Are there good reasons to not remove it?

No.

Alex6323:
# Description of change

* Removes Ed25519 address generation through the wallet
* ~Moves the single address generation to the secret manager~ (separate PR)
* Adds tests for the new method in the bindings

Note, that with this PR we remove emitting the event `LedgerAddressGeneration`, because this was done in `generate_ed25519_address` fn in the removed file `sdk/src/wallet/core/operations/address_generation.rs`. That isn't a problem though because it wasn't used by any wallet.

#1536


bindings/nodejs/lib/types/wallet/wallet.ts
```diff
@@ -15,8 +15,6 @@ import { Bech32Address, OutputId } from '../block';
 
 /** Options for the Wallet builder. */
 export interface WalletOptions {
-    /** The wallet address. */
-    address?: Bech32Address;
```
DaughterOfMars: unused import

bindings/core/src/lib.rs
```diff
@@ -44,7 +43,6 @@ pub fn init_logger(config: String) -> std::result::Result<(), fern_logger::Error
 #[derivative(Debug)]
 #[serde(rename_all = "camelCase")]
 pub struct WalletOptions {
-    pub address: Option<Bech32Address>,
```
Thoralf-M: This should stay, users should be able to provide the address or the bip_path

thibault-martinez (changes requested):
Conflicts

bindings/nodejs/lib/types/wallet/wallet.ts
Alex6323: Removed

bindings/core/src/lib.rs
thibault-martinez: Are we bringing this back then?

Alex6323: not sure why I removed it in the first place :thinking:

Alex6323: reverted

sdk/examples/wallet/offline_signing/1_prepare_transaction.rs
Thoralf-M, line 1: Undo the changes?

sdk/examples/wallet/offline_signing/0_generate_address.rs
Thoralf-M, line 1: Undo the changes, the address is needed in the file so the next example can load them, because it can't be generated there, since no secret manager is available in the online part

sdk/examples/wallet/offline_signing/1_prepare_transaction.rs
Alex6323: yes, you're right

sdk/examples/wallet/offline_signing/0_generate_address.rs
Alex6323: yup, thought going through the file wouldn't be necessary.

Alex6323: just for reference ... one other option would have been to just restore the offline wallet to get the address, but I guess that wouldn't fit that well with the intent of those examples.

Thoralf-M: You mean "transfer" the whole db so it can be restored in the online environment?

Alex6323: no, I just meant to cheat in the example, because most likely those are run on the same machine, so both ... offline and online dbs are just next to eachother.

Thoralf-M: Ah yeah wouldn't do it to keep it realistic (this way we can also detect issues related to this, for example at some point it wasn't possible to restore the wallet from provided addresses)

cli/src/cli.rs
```diff
@@ -1,7 +1,8 @@
 // Copyright 2020-2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{path::Path, str::FromStr};
+use core::str::FromStr;
+use std::path::Path;
```
Thoralf-M: Why this change? :D

sdk/tests/client/secret_manager/address_generation.rs
```diff
@@ -0,0 +1,113 @@
+// Copyright 2023 IOTA Stiftung
```
Thoralf-M:
```suggestion
// Copyright 2024 IOTA Stiftung
```

bindings/nodejs/examples/how_tos/wallet/create-wallet.ts
```diff
@@ -44,18 +44,7 @@ async function run() {
         // The mnemonic can't be retrieved from the Stronghold file, so make a backup in a secure place!
         await secretManager.storeMnemonic(process.env.MNEMONIC as string);
 
-        const walletAddress = await secretManager.generateEd25519Addresses({
```
DaughterOfMars, line 47: wait, why remove?

sdk/examples/wallet/ledger_nano.rs
```diff
@@ -38,32 +36,28 @@ async fn main() -> Result<()> {
...
-        .with_secret_manager(SecretManager::LedgerNano(secret_manager))
+        .with_secret_manager(secret_manager)
         .with_storage_path(&std::env::var("WALLET_DB_PATH").unwrap())
         .with_client_options(client_options)
         .with_bip_path(Bip44::new(SHIMMER_COIN_TYPE))
         .finish()
         .await?;
 
+    let recv_address = wallet.address().await;
+    println!("receive address: {recv_address}");
```
DaughterOfMars:
```suggestion
    println!("recipient address: {recv_address}");
```

sdk/examples/wallet/offline_signing/0_generate_address.rs
```diff
@@ -50,17 +51,18 @@ async fn main() -> Result<()> {
...
     println!("Generated a new wallet");
 
-    write_wallet_address_to_file(&wallet).await
+    write_wallet_address_to_file(&wallet.address().await).await?;
+
+    Ok(())
 }
 
-async fn write_wallet_address_to_file(wallet: &Wallet) -> Result<()> {
+// Backups the Bech32 wallet address to JSON file.
```
DaughterOfMars: why the change?

sdk/src/wallet/core/builder.rs
```diff
@@ -166,13 +166,17 @@ where
...
-            let secret_manager = loaded_wallet_builder
+            self.secret_manager = loaded_wallet_builder
                 .as_ref()
                 .and_then(|builder| builder.secret_manager.clone());
-
-            self.secret_manager = secret_manager;
+            if self.secret_manager.is_none() {
+                return Err(crate::wallet::Error::MissingParameter("secret_manager"));
+            }
         }
```
DaughterOfMars:
```suggestion
        if self.secret_manager.is_none() {
            self.secret_manager.replace(
                loaded_wallet_builder
                    .as_ref()
                    .ok_or(crate::wallet::Error::MissingParameter("secret_manager"))?
                    .secret_manager
                    .clone()
            );
        }
```

sdk/tests/client/common/constants.rs
```diff
@@ -4,3 +4,5 @@
 pub static NODE_LOCAL: &str = "http://localhost:8050";
 
 pub static FAUCET_URL: &str = "http://localhost:8091/api/enqueue";
+
+pub static DEFAULT_MNEMONIC: &str = "inhale gorilla deny three celery song category owner lottery rent author wealth penalty crawl hobby obtain glad warm early rain clutch slab august bleak";
```
DaughterOfMars, line 8: 🦍💨😮

bindings/nodejs/examples/how_tos/wallet/create-wallet.ts
thibault-martinez: Because with the recent wallet builder change, the address will be generated from the given `bipPath`, no need to provide both

thibault-martinez: I am not sure the default `bipPath` fields are the same as the one that were used for the address though, idk if it changes anything for the example

DaughterOfMars: yeah that's what I mean, the values are different. but 🤷‍♀️  if that's not a problem then fine

thibault-martinez: @Alex6323 is it not a problem?

sdk/tests/client/common/mod.rs
```diff
@@ -8,9 +8,9 @@ mod constants;
 //     api::GetAddressesOptions, constants::SHIMMER_COIN_TYPE, node_api::indexer::query_parameters::QueryParameter,
 //     request_funds_from_faucet, secret::SecretManager, Client, Result,
 // };
-use iota_sdk::client::Client;
+use iota_sdk::client::{Client, Result};
 
-pub use self::constants::{FAUCET_URL, NODE_LOCAL};
+pub use self::constants::*;
```
thibault-martinez: We usually avoid wildcards, I know it's tests so feel free to just resolve but I would personally revert for consistency

cli/src/cli.rs
thibault-martinez: Also curious about it

bindings/nodejs/examples/how_tos/wallet/create-wallet.ts
Thoralf-M: There is no default, but it's provided in the wallet options below (another one would also not be a problem)
```
            bipPath: {
                coinType: CoinType.IOTA,
            },
```

thibault-martinez: What about accountIndex? bech32Hrp? Irrelevant for this test? And for all tests where it's removed?

Thoralf-M: Here it's an example, account index is by default 0 and bech32HRP is used from the node, so even better
For the tests it needs to be checked case by case if it was relevant or not

sdk/tests/client/common/mod.rs
Alex6323: okay

cli/src/cli.rs
Alex6323: very important trick everyone should know :trollface: .... that was certainly imported automatically by VSCode lol (If I would have to guess I think I temporarily removed it then I misclicked something when adding the imports again)

sdk/examples/wallet/offline_signing/0_generate_address.rs
Alex6323: personal preference tbh ... the method name promises it will only require the wallet's address, so we're giving it just that, and not a reference to the wallet itself.

sdk/src/wallet/core/builder.rs
Alex6323: Sorry :trollface: "Applying suggestions on deleted lines is currently not supported. Please  and try again."

I'll add the change locally.

sdk/tests/client/common/mod.rs
thibault-martinez: wen

cli/src/cli.rs
Thoralf-M: So undo?

Alex6323: is already locally

sdk/tests/client/common/mod.rs
Alex6323: on it

Alex6323: wildcards terminated

sdk/src/wallet/core/builder.rs
Alex6323: Done

sdk/examples/wallet/offline_signing/0_generate_address.rs
Alex6323: Want me to change it back? I really don't mind if you think giving it a `&Wallet` is better. @DaughterOfMars

cli/src/cli.rs
Alex6323: Changed.

sdk/examples/wallet/offline_signing/0_generate_address.rs
DaughterOfMars: I don't really care, just wanted to be sure this was intentional 🤷‍♀️

sdk/examples/wallet/offline_signing/1_prepare_transaction.rs
```diff
@@ -35,7 +38,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let params = [SendParams::new(SEND_AMOUNT, RECV_ADDRESS)?];
 
-    // Recovers addresses from example `0_address_generation`.
+    // Recovers the address from example `0_generate_address.rs`.
```
DaughterOfMars:
```suggestion
    // Recovers the address generated by the example `0_generate_address.rs`.
```

sdk/examples/wallet/logger.rs
```diff
@@ -38,18 +39,17 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Restore a wallet
     let client_options = ClientOptions::new().with_node(&std::env::var("NODE_URL").unwrap())?;
-    let secret_manager = MnemonicSecretManager::try_from_mnemonic(std::env::var("MNEMONIC").unwrap())?;
-    let wallet = Wallet::builder()
-        .with_secret_manager(SecretManager::Mnemonic(secret_manager))
```
Thoralf-M, line 43: This shouldn't be removed?

sdk/tests/client/common/mod.rs
DaughterOfMars: `*` 🔫

sdk/examples/wallet/logger.rs
Alex6323: I think I have some time to review that code myself in a bit because I just can't remember why I did some of those changes.

sdk/src/client/api/address.rs
```diff
@@ -117,7 +139,31 @@ impl SecretManager {
...
+                self,
+                coin_type,
+                account_index,
+                address_index..address_index + 1,
+                options,
+            )
+            .await?
+            .pop()
+            // Panic: if the secret manager hasn't failed then there must be an address.
+            .unwrap()
```
thibault-martinez: Just use `[0]` like you did in `generate_ed25519_address`?

sdk/examples/wallet/logger.rs
Thoralf-M: The secret manager is needed to generate the wallet address

Alex6323: fixed :see_no_evil:

sdk/src/client/api/address.rs
Alex6323: Done

sdk/tests/client/common/mod.rs
Alex6323: @thibault-martinez Done

sdk/src/client/api/address.rs
```diff
@@ -97,7 +97,29 @@ impl Default for GetAddressesOptions {
...
+        Ok(SecretManage::generate_ed25519_addresses(
+            self,
+            coin_type,
+            account_index,
+            address_index..address_index + 1,
+            options,
+        )
+        // Panic: if the secret manager hasn't failed then there must be an address.
+        .await?[0]
+            .to_bech32(hrp))
```
Alex6323, line 129: I don't know why rustfmt is doing this tbh :man_shrugging:

sdk/src/client/api/address.rs
```diff
@@ -117,7 +150,29 @@ impl SecretManager {
...
+            SecretManage::generate_evm_addresses(
+                self,
+                coin_type,
+                account_index,
+                address_index..address_index + 1,
+                options,
+            )
+            // Panic: if the secret manager hasn't failed then there must be an address.
+            .await?[0]
+                .as_ref(),
```
Alex6323, line 171: are you okay, rustfmt?

bindings/core/src/method/secret_manager.rs
```diff
@@ -19,7 +19,22 @@ use crate::OmittedDebug;
 #[serde(tag = "name", content = "data", rename_all = "camelCase")]
 #[non_exhaustive]
 pub enum SecretManagerMethod {
-    /// Generate Ed25519 addresses.
+    // TODO: Should we refactor `GetAddressesOptions` so we can use it for this method as well?
+    /// Generate a single Ed25519 address.
+    #[serde(rename_all = "camelCase")]
+    GenerateEd25519Address {
```
Thoralf-M: This is already doing https://github.com/iotaledger/iota-sdk/issues/2144 ?

bindings/core/src/method/secret_manager.rs
```diff
@@ -19,7 +19,22 @@ use crate::OmittedDebug;
 #[serde(tag = "name", content = "data", rename_all = "camelCase")]
 #[non_exhaustive]
 pub enum SecretManagerMethod {
-    /// Generate Ed25519 addresses.
+    // TODO: Should we refactor `GetAddressesOptions` so we can use it for this method as well?
```
Thoralf-M: Do we have an issue for this or in which PR is this supposed to resolved?

bindings/core/src/method/secret_manager.rs
Thoralf-M: Also wondering, what was the motivation to put this here instead of just having it on the JS and Python bindings side where it calls GenerateEd25519Addresses and just returns the first address from the returned Vec? Params you could also change for the functions on the bindings side

bindings/core/src/method/secret_manager.rs
Alex6323: wanted to discuss this first, no issue yet

bindings/core/src/method/secret_manager.rs
Alex6323: In parts yes, the issue also addresses whether we should have `Bech32` in the name. At first I thought this should be addressed in a different PR, but I feel like it's fine to do it in this PR already (except for the rename).

bindings/core/src/method/secret_manager.rs
Thoralf-M: Bit weird to add the TODOs then I'd say, I think it would be better to first discuss and then add the TODO or open an issue that just references it without any TODO

bindings/core/src/method/secret_manager.rs
Thoralf-M: What about my other question? (more changes than needed like this just often require more discussion/even more changes and delay the merging of PRs...)

Alex6323: > Also wondering, what was the motivation to put this here instead of just having it on the JS and Python bindings side where it calls GenerateEd25519Addresses and just returns the first address from the returned Vec? Params you could also change for the functions on the bindings side

Yeah, this would also have been possible. But we had this method already as a wallet method, so it basically just moved.

bindings/core/src/method/secret_manager.rs
Alex6323: don't see the problem ... we did this so many times now ... if in the discussion we think it's worth an issue, then one is created and the comment is replaced with the issue number, otherwise I'll just remove the TODO. I won't create an issue beforehand if I am not very sure yet we all see this as an issue.

bindings/core/src/method/secret_manager.rs
Thoralf-M: But don't you think this would be better to change it?

Alex6323: With single wallet addresses I am not so sure anymore, but you tell me :man_shrugging: It feels like the "single address generation" is what is going to be the main use-case, so making the `multiple address generation` the only binding a bit odd, but maybe that's just me.

Thoralf-M: Not sure if I follow, my question is if you think that `GenerateEd25519Address` in this file is better than just having `GenerateEd25519Addresses` and calling this in `generateEd25519Address()` on the JS side
So for users they still would have both functions as right now, no difference there, only difference would be that we have one less method in this file

bindings/core/src/method/secret_manager.rs
Thoralf-M: With discussion you mean gh discussion?

bindings/core/src/method/secret_manager.rs
Alex6323: > more changes than needed like this just often require more discussion/even more changes and delay the merging of PRs...

that's unfortunately true, but I prefer that over merging things before we're happy with it. (Quality > Quantity)

bindings/core/src/method/secret_manager.rs
Alex6323: yup ... this way it's documented

Thoralf-M: Ah okay, that wasn't clear to me :+1: 
If we want both methods, I think we should change the options so they can be used for both

bindings/core/src/method/secret_manager.rs
Thoralf-M: The bindings secret manager methods have not really anything to do with the wallet? To me this has nothing to do with quality

Alex6323: > Not sure if I follow, my question is if you think that `GenerateEd25519Address` in this file is better than just having `GenerateEd25519Addresses` and calling this in `generateEd25519Address()` on the JS side So for users they still would have both functions as right now, no difference there, only difference would be that we have one less method in this file

Yeah, I got what you were saying. We can do that if you think that's better.

For context: My reasoning was that we might be able to deprecate `generateEd25519Addresses` because the functionality of generating multiple at once for a single private key is not that relevant anymore? Maybe I am wrong but the "plural" version seems like a remnant from older IOTA times, but I am not 100% certain :shrug: You probably know better :smile:

Alex6323: tbh ... I am fine with removing the single generation version, but strictly speaking this isn't what the PR is about. Just removing that method alltogether is not okay if it existed before and could be moved to the secret manager methods.

Alex6323: > The bindings secret manager methods have not really anything to do with the wallet? To me this has nothing to do with quality

huh ... this connection is exactly what we're trying to remove with this PR? Before this PR you could generate addresses through the wallet, right? But we still want to be able to generate a single address as before, just with the secret manager directly.

sdk/src/client/api/address.rs
DaughterOfMars: sometimes it just stops working for a file if something is wrong :/

bindings/core/src/method/secret_manager.rs
Alex6323: k ... I'll change it and make it a separate PR but as a matter of fact that method already existed and wasn't required to be removed :man_shrugging:

Thoralf-M: `generateEd25519Addresses` can also be used to generate a single address :shrug: 

> functionality of generating multiple at once for a single private key is not that relevant anymore

For a single wallet definitely not, multiple addresses one could still generate with a loop in a less efficient way from bindings side, personally I would actually be fine with only having the single address version
But again, I don't think this should be done in this PR as it already has 40 changed files and the SDK works perfectly fine without changing this as well

(saw your latest comment only after I sent this one)

Alex6323: Changed

Thoralf-M (approved):

bindings/nodejs/tests/secret_manager/secret_manager.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+describe('SecretManager', () => {
+    it('generate IOTA Ed25519 address', async () => {
+        const mnemonicSecretManager = {
+            mnemonic: "acoustic trophy damage hint search taste love bicycle foster cradle brown govern endless depend situate athlete pudding blame question genius transfer van random vast"
+        };
+
+        let bech32_hrp = Utils.iotaMainnetProtocolParameters().bech32Hrp;
+
+        const secretManager = SecretManager.create(mnemonicSecretManager);
+        const addresses = await secretManager.generateEd25519Addresses({
```
DaughterOfMars, line 22: Use singular version?

Alex6323: there's no singular version anymore (only for Python I could keep it)

DaughterOfMars: uhhh why?

Alex6323: well, technically it's still possible, but it would be inconsistent with the plural version which takes an object. I would rather make that arg object more generic to be used in both cases.

Alex6323: ... but that should be the follow-up PR. Do you think I should bring the method back here?

DaughterOfMars: It's fine for a follow up

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1590, opened 2023-11-07
Add MultiUnlock

thibault-martinez:
Implements https://github.com/iotaledger/tips/blob/tip52/tips/TIP-0052/tip-0052.md#unlocks


sdk/src/types/block/unlock/empty.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::types::block::Error;
+
+/// Used to maintain correct index relationship between addresses and signatures when unlocking a Multi Address where
```
Thoralf-M:
```suggestion
/// Used to maintain correct index relationship between addresses and signatures when unlocking a [`MultiAddress`](crate::types::block::address::MultiAddress) where
```
not sure if fmt is happy with the suggestion

sdk/src/types/block/unlock/mod.rs
```diff
@@ -120,40 +133,60 @@ impl Unlocks {
...
+            if index == 0 || anchor.index() >= index {
+                return Err(Error::InvalidUnlockAnchor(index));
+            }
+        }
+        Unlock::Nft(nft) => {
+            if index == 0 || nft.index() >= index {
+                return Err(Error::InvalidUnlockNft(index));
+            }
+        }
+        Unlock::Multi(_) => return Err(Error::MultiUnlockRecursion),
```
Thoralf-M, line 173: This makes sense if it's supposed to be only called inside of `verify_unlocks()`, but just looking at the function itself it's not clear why this would return a recursion error, so maybe add some comment here or above the function

sdk/src/types/block/unlock/multi.rs
```diff
@@ -0,0 +1,79 @@
...
+use alloc::{boxed::Box, vec::Vec};
+
+use derive_more::Deref;
+use packable::{prefix::BoxedSlicePrefix, Packable};
+
+use crate::types::block::{address::WeightedAddressCount, unlock::Unlock, Error};
+
+pub(crate) type UnlocksCount = WeightedAddressCount;
+
+/// Unlocks a Multi Address with a list of other unlocks.
```
Thoralf-M:
```suggestion
/// Unlocks a [`MultiAddress`](crate::types::block::address::MultiAddress) with a list of other unlocks.
```

sdk/src/types/block/unlock/mod.rs
thibault-martinez: Added a comment and renamed the function

Thoralf-M (dismissed):

DaughterOfMars (dismissed):

sdk/src/types/block/unlock/empty.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::types::block::Error;
```
Alex6323: move import to the `dto` module?

thibault-martinez: Done

DaughterOfMars (dismissed):

sdk/src/types/block/address/mod.rs
```diff
@@ -171,6 +171,24 @@ impl Address {
             (Self::ImplicitAccountCreation(address), _) => {
                 return Self::from(*address.ed25519_address()).unlock(unlock, context);
             }
+            (Self::Multi(address), Unlock::Multi(unlock)) => {
```
Alex6323: Would you be fine with calling this local variable `multi_address`? Then it doesn't need to get shadowed by `address` within the for-loop. Also `multi_address.len()` reads better than `address.len()` imo. Your choice though as this is just a preference.

thibault-martinez: Done

Alex6323 (approved):

DaughterOfMars (approved):

Thoralf-M (approved):

PR #1597, opened 2023-11-08
Allow ImplicitAccountCreationAddress unlocking

thibault-martinez:


sdk/src/types/block/address/mod.rs
```diff
@@ -168,6 +168,9 @@ impl Address {
             }
             // TODO maybe shouldn't be a semantic error but this function currently returns a TransactionFailureReason.
             (Self::Anchor(_), _) => return Err(TransactionFailureReason::SemanticValidationFailed),
+            (Self::ImplicitAccountCreation(ImplicitAccountCreationAddress(ed25519_address)), _) => {
+                Address::from(ed25519_address.clone()).unlock(unlock, context)?;
```
Thoralf-M: Doesn't need to be inserted in `context.unlocked_addresses`?

thibault-martinez: It will, but in the `Self::Ed25519` match of the inner address. Otherwise we're adding the ImplicitAccountCreationAddress as unlocked, which could cause issues

Thoralf-M: `but in the Self::Ed25519 match of the inner address.`
Can you point me to the code? Not sure what you mean here

thibault-martinez: L121

Alex6323: maybe it's worth a comment?

Thoralf-M: Ah it's calling the same function, I thought the unlock was on another type, all good then

Thoralf-M (dismissed):

Alex6323 (dismissed):

sdk/src/types/block/address/mod.rs
Alex6323:
```suggestion
                return Self::from(ed25519_address.clone()).unlock(unlock, context);
```
:thinking:

Thoralf-M (dismissed):

sdk/src/types/block/address/implicit_account_creation.rs
```diff
@@ -11,7 +11,7 @@ use crate::types::block::address::Ed25519Address;
 /// [`AccountOutput`](crate::types::block::output::AccountOutput).
 #[derive(Copy, Clone, Display, Eq, PartialEq, Ord, PartialOrd, Hash, FromStr, AsRef, Deref, From, Packable)]
 #[as_ref(forward)]
-pub struct ImplicitAccountCreationAddress(Ed25519Address);
+pub struct ImplicitAccountCreationAddress(pub(crate) Ed25519Address);
```
DaughterOfMars: Maybe just make it pub? Or add a getter?

thibault-martinez: Can't be a getter for the match, and not sure we want people to be able to change it when constructed tbh :/

thibault-martinez: I mean, it doesn't necessarily have to be part of the match

DaughterOfMars: Yeah but keep in mind this is supposed to be a public type but it's kind of unusable if you can't access the inner data somehow.

thibault-martinez: It has Deref and AsRef but fair enough, will add a getter 👍🏻

Alex6323 (dismissed):

sdk/src/types/block/address/implicit_account_creation.rs
thibault-martinez: Added a getter

Thoralf-M (dismissed):

sdk/src/types/block/address/mod.rs
```diff
@@ -168,8 +168,8 @@ impl Address {
             }
             // TODO maybe shouldn't be a semantic error but this function currently returns a TransactionFailureReason.
             (Self::Anchor(_), _) => return Err(TransactionFailureReason::SemanticValidationFailed),
-            (Self::ImplicitAccountCreation(ImplicitAccountCreationAddress(ed25519_address)), _) => {
-                return Self::from(ed25519_address.clone()).unlock(unlock, context);
+            (Self::ImplicitAccountCreation(address), _) => {
+                return Self::from(address.ed25519_address().clone()).unlock(unlock, context);
```
DaughterOfMars: This is `Copy`, clippy will complain

thibault-martinez: Removed

Thoralf-M (approved):

Alex6323 (approved):

DaughterOfMars (approved):

PR #1598, opened 2023-11-08
Implicit account transition

### LINKED ISSUE #1350: CLI: add AccountOutput bootstrapping

### LINKED ISSUE #1533: Add transition from implicit account to account

thibault-martinez:
Fixes #1533 & #1350


Alex6323:
how can I unlook now?

bindings/nodejs/lib/types/wallet/bridge/account.ts
thibault-martinez, line 1: Is this file still needed ?

Thoralf-M: Should be removed and the methods should be all in wallet.ts

sdk/src/wallet/operations/transaction/account.rs
```diff
@@ -0,0 +1,81 @@
...
+                BlockIssuerKeys::from_vec(vec![BlockIssuerKey::from(Ed25519BlockIssuerKey::from(public_key))])?,
+            )?])
+            .finish_output()?;
+
+        let transaction_options = TransactionOptions {
+            custom_inputs: Some(vec![*output_id]),
+            ..Default::default()
+        };
+
+        let prepared_transaction = self
```
Thoralf-M: There should be a method that returns this, so signing can be done independent
The Rust bindings should also only expose the prepare method and all in one can be done on bindings side as we do for all the other transaction methods

thibault-martinez: Ah, right

thibault-martinez: Done

bindings/python/iota_sdk/wallet/account.py
```diff
@@ -278,6 +278,24 @@ def implicit_account_creation_address(self) -> str:
...
+    def prepare_implicit_account_transition(
+            self, output_id: OutputId) -> PreparedTransaction:
+        """Prepares to transition an implicit account to an account.
+        """
+        prepared = self._call_account_method(
+            'implicitAccountTransition', {
+                'outputId': output_id
+            }
+        )
+        return PreparedCreateTokenTransaction(
```
Thoralf-M:
```suggestion
        return PreparedTransaction(
```

sdk/src/client/secret/mod.rs
```diff
@@ -553,7 +553,7 @@ where
                 // We can only sign ed25519 addresses and block_indexes needs to contain the account or nft
                 // address already at this point, because the reference index needs to be lower
                 // than the current block index
-                if !input_address.is_ed25519() {
+                if !input_address.is_ed25519() && !input_address.is_implicit_account_creation() {
```
Thoralf-M: Also want to add restricted addresses?

thibault-martinez: Sure

sdk/src/wallet/operations/transaction/account.rs
```diff
@@ -0,0 +1,83 @@
...
+            if implicit_account_data.output.is_implicit_account() {
+                implicit_account_data.output.as_basic()
+            } else {
+                return Err(Error::ImplicitAccountNotFound);
+            }
+        } else {
+            return Err(Error::ImplicitAccountNotFound);
+        };
+
+        let public_key = if let Some(bip_path) = self.bip_path().await {
```
Thoralf-M, line 45: I think the user should be able to provide the public key to the function as parameter

Thoralf-M: Probably somewhat related to the TODO issue in the else case

thibault-martinez: I guess I agree but there are more combinations we can do, I have opened an issue for it

sdk/src/client/secret/mod.rs
thibault-martinez: Done

bindings/python/iota_sdk/wallet/account.py
thibault-martinez: Whoops

Thoralf-M (dismissed):

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -432,25 +441,27 @@ impl InputSelection {
...
+                                log::debug!("validate_transitions error {err:?}");
+                                return Err(Error::UnfulfillableRequirement(Requirement::Account(
+                                    *account_output.account_id(),
+                                )));
+                            }
+                        }
+                        Output::Basic(_) => {
+                            // TODO https://github.com/iotaledger/iota-sdk/issues/1664
+                        }
+                        _ => unreachable!(),
```
Alex6323: I think Chloé's preference here would be to use `panic` instead since this execution branch hasn't been ruled out prior. It may be unreachable protocol-wise, but not programatically/logically.

thibault-martinez: This is 100% programatically unreachable

thibault-martinez: Or maybe I don't understand what you mean by programatically

sdk/src/client/secret/mod.rs
```diff
@@ -553,8 +553,10 @@ where
                 // We can only sign ed25519 addresses and block_indexes needs to contain the account or nft
                 // address already at this point, because the reference index needs to be lower
                 // than the current block index
-                if !input_address.is_ed25519() {
-                    Err(InputSelectionError::MissingInputWithEd25519Address)?;
+                match &input_address {
+                    Address::Ed25519(_) | Address::ImplicitAccountCreation(_) => {}
+                    Address::Restricted(restricted) if restricted.address().is_ed25519() => {}
+                    _ => Err(InputSelectionError::MissingInputWithEd25519Address)?,
```
Alex6323, line 559: Not sure I like how it is currently. What about?
```rust
if { match &input_address {
    Address::Ed25519(_) | Address::ImplicitAccountCreation(_) => true,
    Address::Restricted(restricted) if restricted.address().is_ed25519() => true,
    _ => false
} } { ... } else { Err(InputSelectionError::MissingInputWithEd25519Address)? }
```
Maybe I would consider binding the match first to a `let is_valid_input_address`, but I'ld find that overall easier to read.

sdk/src/client/api/block_builder/input_selection/mod.rs
Alex6323: like ... could it become suddenly reachable by a protocol change?

thibault-martinez: I guess, a massive protocol change that would break the whole codebase. I don't think that's a valid argument anyway, any change to the match could make an unreachable reachable 🤷🏻

thibault-martinez: I'll change it anyway as I don't really care but this is definitely unreachable

thibault-martinez: Changed

sdk/src/client/secret/mod.rs
thibault-martinez: Honestly think my version is more readable, same thing, less verbose 🤷🏻‍♂️

Alex6323: okay you can keep it, but I think code should be as explicit as possible and this is essentially a somewhat hidden `if else`.

Alex6323 (approved):

sdk/src/client/secret/mod.rs
thibault-martinez: I mean, literally all match are hidden if/else, that's syntactic sugar to make things shorter/prettier

PR #1602, opened 2023-11-09
Add missing `BlockState` and `TransactionState` variants

DaughterOfMars:
# Description of change

Adds some missing variants.


sdk/src/wallet/operations/syncing/transactions.rs
```diff
@@ -104,7 +104,9 @@ where
                         if let Some(tx_state) = metadata.transaction_state {
                             match tx_state {
                                 // TODO: Separate TransactionState::Finalized?
-                                TransactionState::Finalized | TransactionState::Confirmed => {
+                                TransactionState::Accepted
+                                | TransactionState::Confirmed
+                                | TransactionState::Finalized => {
                                     log::debug!(
                                         "[SYNC] confirmed transaction {transaction_id} in block {}",
                                         metadata.block_id
```
Alex6323, line 112: Can you explain why this applies to "accepted" as well? Before writing this comment I tried find the answer myself in the go code, but I am still not entirely sure tbh.

DaughterOfMars: I don't actually know 🤷‍♀️ was hoping someone could tell me

Thoralf-M: I wouldn't add it in the match here, because accepted could still get not finalized at the end
But probably good idea to put a TODO, since it might be interesting to match and do something in the accepted case already
From the indexer we also get outputs already when a tx gets accepted, but the outputs could also never get created at the end, so we definitely need to change some logic in the wallet around this

sdk/src/types/api/core.rs
```diff
@@ -327,6 +327,8 @@ pub struct SubmitBlockResponse {
 pub enum BlockState {
     // Stored but not confirmed.
     Pending,
+    // TODO
+    Accepted,
     // Confirmed with the first level of knowledge.
```
Thoralf-M, line 332: Also don't like the confirmed comment, `level of knowledge` says nothing without knowing exactly where this name comes from
Maybe something like the following would be better
```
accepted = Referenced by 67% of online validator committee.
confirmed = Referenced by 67% of entire validator committee.
```

DaughterOfMars:
Closing because apparently this is not needed https://github.com/iotaledger/tips/pull/153#issuecomment-1805257330

PR #1604, opened 2023-11-10
Remove `address` field from `OutputData` struct

### LINKED ISSUE #1514: Remove address from OutputData struct
## Description

Remove address from OutputData struct, since it doesn't really make sense there if we only have a single address in the wallet. In stardust it was used so the address with the output got added to the addresses_with_unspent_outputs

No.

Alex6323:
# Description of change

Removes said field from the `OutputData` struct. 

Cleans up and refactors `request_outputs_recursively` a bit.

Closes #1514


Thoralf-M (dismissed):

sdk/src/wallet/operations/syncing/mod.rs
```diff
@@ -168,102 +165,98 @@ where
...
-                        let nft_address = NftAddress::from(nft_output.nft_id_non_null(&output_data.output_id));
+        // Get the corresponding unspent output data
+        let mut new_outputs_data = self
+            .get_outputs_from_address_output_ids(addresses_to_sync_with_unspent_output_ids)
+            .await?;
 
-                        new_account_and_nft_addresses.insert(Address::Nft(nft_address), output_data.address);
+        loop {
+            // Try to discover new addresses
+            // See https://github.com/rust-lang/rust-clippy/issues/8539 regarding this lint.
```
DaughterOfMars, line 191: I feel like there's a better way 🤔

Alex6323: probably ... any ideas? Otherwise I would open an issue so we can investigate later. If you like the previous solution better where we consume and create a new binding we can also do that, but that would unnecessarily require a new allocation in each iteration.

Thoralf-M (dismissed):

sdk/src/wallet/operations/syncing/mod.rs
DaughterOfMars: I'll fiddle with it locally and get back to you

DaughterOfMars (dismissed):

Thoralf-M (dismissed):

Thoralf-M (approved):

thibault-martinez (approved):

PR #1606, opened 2023-11-10
CLI wallet: Ledger Nano support

### LINKED ISSUE #144: Cli Wallet: Ledger Nano support
## Description

Add ledger nano support in the cli wallet. When creating the wallet it should be possible to use a ledger nano instead of stronghold as secret manager

## Motivation

Allow ledger nano users to also use the cli wallet, also can help for some testing

## Requirements

1. Make it possible to create a wallet with Ledger NanoSecretManager
2. Listen to wallet events and print the address when a LedgerAddressGeneration event is emitted or `PreparedTransaction || PreparedTransactionEssenceHash` during tx creation, so users can compare the data with what's shown on the ledger nano

No.

Alex6323:
# Description of change

Adds support for Ledger Nano to the CLI wallet and fixes some usability issues.

Closes #144 

## Type of change

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
-## 1.3.0 - 2024-xx-xx
+## 1.3.0 - 2024-01-23
 
 ### Added
 
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -34,7 +34,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `UTC` suffix to the formatted date of `transactions`;
 - `addresses` handles NFT and Alias addresses;
 - `AccountCommand::Address` command that accepts either a list index or a `Bech32Address`;
-- Ledger Nano support;
 
 ### Changed
 
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
-## 1.3.0 - 2023-xx-xx
+## 1.3.0 - 2024-xx-xx
 
 ### Added
 
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -19,6 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 1.3.0 - 2023-xx-xx
+
+### Added
+
+- Ledger Nano support;
+
 ## 1.2.0 - 2023-10-26
 
 ### Added
@@ -28,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `UTC` suffix to the formatted date of `transactions`;
 - `addresses` handles NFT and Alias addresses;
 - `AccountCommand::Address` command that accepts either a list index or a `Bech32Address`;
+- Ledger Nano support;
 
 ### Changed
 
--- a/cli/src/command/wallet.rs
+++ b/cli/src/command/wallet.rs
@@ -7,7 +7,7 @@ use clap::{builder::BoolishValueParser, Args, CommandFactory, Parser, Subcommand
 use iota_sdk::{
     client::{
         constants::SHIMMER_COIN_TYPE,
-        secret::{ledger_nano::LedgerSecretManager, stronghold::StrongholdSecretManager, SecretManager},
+        secret::{stronghold::StrongholdSecretManager, SecretManager},
         stronghold::StrongholdAdapter,
         utils::Password,
     },
@@ -18,7 +18,7 @@ use log::LevelFilter;
 use crate::{
     error::Error,
     helper::{check_file_exists, generate_mnemonic, get_password, SecretManagerChoice},
-    println_log_error, println_log_info,
+    println_log_info,
 };
 
 const DEFAULT_LOG_LEVEL: &str = "debug";
@@ -33,13 +33,6 @@ pub struct WalletCli {
     /// Set the path to the wallet database.
     #[arg(long, value_name = "PATH", env = "WALLET_DATABASE_PATH", default_value = DEFAULT_WALLET_DATABASE_PATH)]
     pub wallet_db_path: String,
-    /// Set the secret manager to use.
-    #[arg(short, long, value_name = "SECRET_MANAGER", default_value = DEFAULT_SECRET_MANAGER)]
-    pub secret_manager: SecretManagerChoice,
-    /// Set the path to the stronghold snapshot file. Ignored if the <SECRET_MANAGER> is not a Stronghold secret
-    /// manager.
-    #[arg(long, value_name = "PATH", env = "STRONGHOLD_SNAPSHOT_PATH", default_value = DEFAULT_STRONGHOLD_SNAPSHOT_PATH)]
-    pub stronghold_snapshot_path: String,
     /// Set the account to enter.
     pub account: Option<AccountIdentifier>,
     /// Set the log level.
@@ -113,27 +106,15 @@ pub enum WalletCommand {
     Sync,
 }
 
-impl WalletCommand {
-    pub fn is_unlock_command(&self) -> bool {
-        match self {
-            Self::Accounts
-            | Self::Backup { .. }
-            | Self::ChangePassword { .. }
-            | Self::NewAccount { .. }
-            | Self::NodeInfo
-            | Self::SetNodeUrl { .. }
-            | Self::SetPow { .. }
-            | Self::Sync => true,
-            Self::Init(_)
-            | Self::MigrateStrongholdSnapshotV2ToV3 { .. }
-            | Self::Mnemonic { .. }
-            | Self::Restore { .. } => false,
-        }
-    }
-}
-
 #[derive(Debug, Clone, Args)]
 pub struct InitParameters {
+    /// Set the secret manager to use.
+    #[arg(short, long, value_name = "SECRET_MANAGER", default_value = DEFAULT_SECRET_MANAGER)]
+    pub secret_manager: SecretManagerChoice,
+    /// Set the path to the stronghold snapshot file. Ignored if the <SECRET_MANAGER> is not a Stronghold secret
+    /// manager.
+    #[arg(short = 't', long, value_name = "PATH", env = "STRONGHOLD_SNAPSHOT_PATH", default_value = DEFAULT_STRONGHOLD_SNAPSHOT_PATH)]
+    pub stronghold_snapshot_path: String,
     // TODO: remove this field to make `InitParameters` independ from the secret manager being used?
     /// Set the path to a file containing mnemonics. If empty, a mnemonic has to be entered or will be randomly
     /// generated. Only used by some secret managers.
@@ -150,6 +131,8 @@ pub struct InitParameters {
 impl Default for InitParameters {
     fn default() -> Self {
         Self {
+            secret_manager: SecretManagerChoice::Stronghold,
+            stronghold_snapshot_path: DEFAULT_STRONGHOLD_SNAPSHOT_PATH.to_string(),
             mnemonic_file_path: None,
             node_url: DEFAULT_NODE_URL.to_string(),
             coin_type: SHIMMER_COIN_TYPE,
@@ -193,13 +176,13 @@ pub async fn change_password_command(wallet: &Wallet, current_password: Password
 pub async fn init_command(
     storage_path: &Path,
     secret_manager: SecretManager,
-    parameters: InitParameters,
+    init_params: InitParameters,
 ) -> Result<Wallet, Error> {
     Ok(Wallet::builder()
         .with_secret_manager(secret_manager)
-        .with_client_options(ClientOptions::new().with_node(parameters.node_url.as_str())?)
+        .with_client_options(ClientOptions::new().with_node(init_params.node_url.as_str())?)
         .with_storage_path(storage_path.to_str().expect("invalid wallet db path"))
-        .with_coin_type(parameters.coin_type)
+        .with_coin_type(init_params.coin_type)
         .finish()
         .await?)
 }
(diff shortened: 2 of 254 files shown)


cli/src/helper.rs
```diff
@@ -343,3 +343,29 @@ pub async fn check_file_exists(path: &Path) -> Result<(), Error> {
...
+pub enum SecretManagerChoice {
+    Stronghold,
+    LedgerNano,
+}
+
+impl From<usize> for SecretManagerChoice {
+    fn from(value: usize) -> Self {
+        match value {
+            0 => Self::Stronghold,
+            1 => Self::LedgerNano,
```
Thoralf-M, line 359: What do you think about adding LedgerNanoSimulator here? Or would another place be better to select this option?
Because I think that would be very good if we can actually use a simulator as well with it for testing (so need to create `LedgerSecretManager::new(true)`)

Thoralf-M (commented):
Can you remove `let wallet = unlock_wallet(storage_path, None, None).await?;` from node_info_command()?
We don't need this for a stronghold wallet and certainly not for a ledger nano one

Please also check if there are other places where one is requested to insert a pw even if not needed (not needed at all or just not needed with a ledger nano)

cli/src/helper.rs
Alex6323: I like the idea. Will add it.

Alex6323: Added it ... I think ... Is that what you wanted?

Thoralf-M: Yes that's good

cli/src/helper.rs
```diff
@@ -343,3 +343,31 @@ pub async fn check_file_exists(path: &Path) -> Result<(), Error> {
...
+    LedgerNanoSimulator,
+}
+
+impl From<usize> for SecretManagerChoice {
+    fn from(value: usize) -> Self {
+        match value {
+            0 => Self::Stronghold,
+            1 => Self::LedgerNano,
+            2 => Self::LedgerNanoSimulator,
+            _ => unreachable!(),
```
DaughterOfMars: probably should use a `panic!` instead with a message, even if this shouldn't happen

DaughterOfMars (dismissed):

cli/src/helper.rs
Alex6323: okay (it truely is unreachable, but I don't mind)

DaughterOfMars: IMO `unreachable` should pretty much only be used when a branch is invalidated because of prior code *in the same scope*. This is a case where, in the context of this impl, the branch is definitely not invalidated.

Alex6323:
> Can you remove `let wallet = unlock_wallet(storage_path, None, None).await?;` from node_info_command()? We don't need this for a stronghold wallet and certainly not for a ledger nano one
> 
> Please also check if there are other places where one is requested to insert a pw even if not needed (not needed at all or just not needed with a ledger nano)

Are sure you need to enter a password to run the `node info` command? I don't. As you can see from the line you posted, password is `None`.

cli/src/helper.rs
```diff
@@ -356,7 +356,7 @@ impl From<usize> for SecretManagerChoice {
             0 => Self::Stronghold,
             1 => Self::LedgerNano,
             2 => Self::LedgerNanoSimulator,
-            _ => unreachable!(),
+            _ => panic!("invalid choice index"),
```
DaughterOfMars:
```suggestion
            _ => panic!("invalid secret manager choice index"),
```

cli/src/helper.rs
Alex6323: changed

Alex6323:
> > Can you remove `let wallet = unlock_wallet(storage_path, None, None).await?;` from node_info_command()? We don't need this for a stronghold wallet and certainly not for a ledger nano one
> > Please also check if there are other places where one is requested to insert a pw even if not needed (not needed at all or just not needed with a ledger nano)
> 
> Are sure you need to enter a password to run the `node info` command? I don't. As you can see from the line you posted, password is `None`.

Actually during current implementation of handling several secret managers I came to realize that this should require a password atm, because this command uses the node as stored in the wallet, not just any node. Hence the node used should be considered private information. An alternative would be to require the URL, but I guess that's not what we want. We want the node info of the node that specifically is used to sync the wallet.

cli/src/command/wallet.rs
```diff
@@ -111,10 +113,30 @@ pub enum WalletCommand {
...
+            | Self::MigrateStrongholdSnapshotV2ToV3 { .. }
+            | Self::Mnemonic { .. }
+            | Self::Restore { .. } => false,
+        }
+    }
+}
+
 #[derive(Debug, Clone, Args)]
 pub struct InitParameters {
+    // TODO: remove this field to make `InitParameters` independ from the secret manager being used?
```
DaughterOfMars: TODO

Thoralf-M (commented):
Running just `cargo run --release` has a different behavior (asks directly for stronghold pw) than running with `init` (shows secret manager selection)
Should both be like the init command?

Thoralf-M (changes requested):
Running a wallet command after it got initialized with ledger nano returns `Stronghold snapshot not found at './stardust-cli-wallet.stronghold'.` (get-info)  or it asks for a stronghold password, just to then fail with  wallet error: `no mnemonic has been stored into the Stronghold vault` (new-account)

Alex6323:
> Running a wallet command after it got initialized with ledger nano returns `Stronghold snapshot not found at './stardust-cli-wallet.stronghold'.` (get-info) or it asks for a stronghold password, just to then fail with wallet error: `no mnemonic has been stored into the Stronghold vault` (new-account)

how did you run the wallet? `Stronghold` is still the default secret manager if you don't pass `-s <SECRET_MANAGER>` to the binary.

Thoralf-M:
> > Running a wallet command after it got initialized with ledger nano returns `Stronghold snapshot not found at './stardust-cli-wallet.stronghold'.` (get-info) or it asks for a stronghold password, just to then fail with wallet error: `no mnemonic has been stored into the Stronghold vault` (new-account)
> 
> how did you run the wallet? `Stronghold` is still the default secret manager if you don't pass `-s <SECRET_MANAGER>` to the binary.

Just `cargo run --release get-info` (before initialized with ledger nano)

Alex6323:
> > > Running a wallet command after it got initialized with ledger nano returns `Stronghold snapshot not found at './stardust-cli-wallet.stronghold'.` (get-info) or it asks for a stronghold password, just to then fail with wallet error: `no mnemonic has been stored into the Stronghold vault` (new-account)
> > 
> > 
> > how did you run the wallet? `Stronghold` is still the default secret manager if you don't pass `-s <SECRET_MANAGER>` to the binary.
> 
> Just `cargo run --release get-info` (before initialized with ledger nano)

gonna fix the quirks ...

cli/src/command/wallet.rs
```diff
@@ -146,61 +152,37 @@ pub async fn accounts_command(storage_path: &Path, snapshot_path: &Path) -> Resu
     Ok(())
 }
 
-pub async fn backup_command(storage_path: &Path, snapshot_path: &Path, backup_path: &Path) -> Result<(), Error> {
-    let password = get_password("Stronghold password", !snapshot_path.exists())?;
-    let wallet = unlock_wallet(storage_path, snapshot_path, password.clone()).await?;
-    wallet.backup(backup_path.into(), password).await?;
+// TODO: should we allow the backup to have a different password?
```
Thoralf-M: Let's only add it if someone requests it

sdk/src/client/secret/mod.rs
```diff
@@ -195,6 +195,21 @@ impl Debug for SecretManager {
     }
 }
 
+impl fmt::Display for SecretManager {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            #[cfg(feature = "stronghold")]
+            Self::Stronghold(_) => write!(f, "Stronghold"),
+            #[cfg(feature = "ledger_nano")]
+            Self::LedgerNano(_) => write!(f, "LedgerNano"),
```
Thoralf-M: Would this be interesting?
```suggestion
            Self::LedgerNano(l) => {
                if l.is_simulator {
                    write!(f, "LedgerNanoSimulator")
                } else {
                    write!(f, "LedgerNano")
                }
            }
```

Alex6323: Probably! Done

cli/src/command/wallet.rs
Alex6323: Removed the TODO because achieving this (if desirable) requires more than just removing this field but a whole refactor of this struct.

cli/src/command/wallet.rs
Alex6323: Removed

sdk/src/wallet/core/operations/stronghold_backup/mod.rs
```diff
@@ -69,7 +69,7 @@ impl Wallet {
     /// coin type doesn't match
     /// if ignore_if_bech32_hrp_mismatch == Some("rms"), but addresses have something different like "smr", no accounts
     /// will be restored.
-    pub async fn restore_backup(
+    pub async fn restore_from_backup(
```
Thoralf-M: Can't do this breaking change in develop

cli/CHANGELOG.md
```diff
@@ -19,6 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 1.3.0 - 2023-xx-xx
```
Thoralf-M:
```suggestion
## 1.3.0 - 2024-xx-xx
```

cli/src/wallet.rs
```diff
@@ -73,43 +260,58 @@ pub async fn new_wallet(cli: WalletCli) -> Result<(Option<Wallet>, Option<Accoun
...
+                } => {
+                    println_log_error!(
+                        "Snapshot file for Stronghold secret manager linked with the wallet not found at '{}'",
+                        snapshot_path.display()
+                    );
+                    return Ok((None, None));
+                }
+                LinkedSecretManager::LedgerNano { is_connected: false } => {
+                    println_log_error!("Ledger Nano linked with the wallet not connected");
+                    return Ok((None, None));
```
Thoralf-M: That's annoying, why not still allow to open the wallet? To check balance or see the txs it's not necessary to have a ledger nano connected

Alex6323: yeah, it's annoying, but if we do it lazily (only check that it's connected when necessary), then we should do that as well for Stronghold? Only ask for the password for operations that require it?

Alex6323: should we make it a separate issue (and PR) that deals with removing that annoyance (because this PR is just concerned with Ledger Nano support)?

sdk/src/wallet/core/operations/stronghold_backup/mod.rs
Alex6323: ah, shoot, it's develop, right

cli/src/wallet.rs
Thoralf-M: I think it's a bit different here because you don't have to enter the password at a later point (if you just connect a ledger before doing a tx or creating a new address it will not fail, for stronghold it would without password), so you just have to not add this extra match?
If it's needed for whatever reason I don't see, feel free to close and for stronghold you can open an issue if you want, definitely not needed in this PR

cli/src/command/wallet.rs
DaughterOfMars: If you think it's worth doing, can we create an issue?

cli/src/command/wallet.rs
```diff
@@ -272,32 +250,28 @@ pub async fn restore_command(storage_path: &Path, snapshot_path: &Path, backup_p
...
-        "Wallet has been restored from the backup file \"{}\".",
-        backup_path.display()
-    );
-
-    Ok(wallet)
+    if let Err(e) = wallet.restore_backup(backup_path.into(), password, None, None).await {
+        // Clean up a failed restore (typically produces a wallet without a secret manager)
+        // TODO: a better way would be to not create any files/dirs in the first place when it's not clear yet whether
+        // the restore will be successful
+        std::fs::remove_dir_all(storage_path)?;
```
Thoralf-M: If there is still a db with accounts this will delete it, because it will error with `wallet error: backup failed can't restore backup when there are already accounts`
Result is nothing is restored but the db is gone

To reproduce:
1. create an account
2. create a backup
3. try to restore backup -> will fail and delete db

Alex6323: well, after that the restore will work :trollface:

sdk/src/client/stronghold/mod.rs
```diff
@@ -100,7 +100,7 @@ pub struct StrongholdAdapter {
     timeout_task: Arc<Mutex<Option<TaskHandle>>>,
 
     /// The path to a Stronghold snapshot file.
-    pub(crate) snapshot_path: PathBuf,
+    pub snapshot_path: PathBuf,
```
Thoralf-M: Why make this public?

Alex6323: I need it in the CLI. I could add a getter if you prefer.

Thoralf-M: I think a getter would be better, because otherwise one can change the path and then end up with multiple stronghold files that only have a part of the data?

Alex6323: Agree. Done.

Thoralf-M (dismissed):

Thoralf-M (dismissed):

DaughterOfMars (dismissed):

DaughterOfMars (dismissed):

Thoralf-M (dismissed):

cli/src/command/wallet.rs
Alex6323: not worth doing ... closing this if you don't mind ... if you do feel free to open again.

sdk/src/client/stronghold/mod.rs
```diff
@@ -230,6 +230,11 @@ impl StrongholdAdapterBuilder {
 }
 
 impl StrongholdAdapter {
+    /// Get the path to the corresponding Stronghold snapshot file.
+    pub fn snapshot_path(&self) -> &Path {
```
thibault-martinez, line 234: Changelog entry

sdk/src/client/secret/mod.rs
```diff
@@ -195,6 +195,27 @@ impl Debug for SecretManager {
     }
 }
 
+impl fmt::Display for SecretManager {
```
thibault-martinez, line 198: Changelog entry

Thoralf-M (dismissed):

DaughterOfMars (dismissed):

thibault-martinez:
(don't merge yet please, still having a look)

thibault-martinez (changes requested):
Conflict

DaughterOfMars (dismissed):

Thoralf-M (dismissed):

thibault-martinez (approved):

Thoralf-M (approved):

PR #1607, opened 2023-11-10
TIP BlockId tests

### LINKED ISSUE #1387: Python: fix some tests (and remove corresponding `pytest` ignore)
- [x] `test_address_generation_iota`
- [x] `test_address_generation_shimmer`
- [x] `test_basic_block_with_tx_payload`
- [x] `test_basic_block_with_tx_payload_all_output_types`
- [x] `test_basic_block_with_tx_payload_with_tagged_data_payload`
- [x] `test_sign_verify_ed25519`
- [x] `test_wallet_destroy`

### LINKED ISSUE #1391: Add TIP Block ID tests to bindings
Like in Rust https://github.com/iotaledger/iota-sdk/blob/2.0/sdk/tests/types/block_id.rs#L75

Alex6323:
# Description of change

* Uncomments the Block ID tests in Rust (2 of the 3 tests are ignored though until the TIP is updated with valid public keys)
* Adds the same tests to the bindings

Closes #1391 
Partially addresses #1387


bindings/nodejs/tests/fixtures/basic_block_tagged_data_payload.json
thibault-martinez, line 1: Just reuse the rust fixtures, not very handy to have to update 3 files

Alex6323: yeah, thought the same

bindings/nodejs/tests/types/block.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+        const block = basic_block_tagged_data_payload_json.block as unknown as Block;
+        expect(block.isBasic).toBe(true);
+        expect(block.body.isBasic).toBe(true);
+        expect(block.asBasic).toBeInstanceOf(BasicBlockBody);
+
+        const basic_block = block.body.asBasic();
+        expect(basic_block.payload).toBeInstanceOf(TaggedDataPayload);
+        const tagged_data_payload = basic_block.payload as unknown as TaggedDataPayload;
+        expect(tagged_data_payload.type).toBe(PayloadType.TaggedData);
+        expect(basic_block.maxBurnedMana).toEqual(864);
```
thibault-martinez: Why do all these expects? Hash would fail if there was an issue

Alex6323: yes, but it's less indicative. If we'ld do a change in the type, then the test would just fail with a different hash than expected, but left us no other clue. With the expects it's more likely to get a hint how to fix the issue.

thibault-martinez: But it seems so arbitrary, why `maxBurnedMana` only and not all fields ?

Alex6323: good point ... I took it from the previous tests (as they had been approved before)

thibault-martinez: Tbh I wouldn't bother too much, yes if hash fails we'd have to locate the issue but it's not that bad.

Alex6323: okay ... fine with me!

sdk/tests/types/fixtures/basic_block_tagged_data_payload.json
```diff
@@ -0,0 +1,42 @@
...
+             "0x525e5a4d011d6359253e3e6e32717a2f0b354478154e1f4426345608234e044e5b62187c",
+             "0x6d1859664875241d0c6a4c55146f1d491c0a1f5a090e785c4a7b691977730e75622a5d41",
+             "0x742c2e2d080a42165c52732b4e48245e795e61240b7832625d1c0f037c5937191f510f07"
+          ],
+          "weakParents":[
+             
+          ],
+          "shallowLikeParents":[
+             
+          ],
```
Thoralf-M: These entries shouldn't exist if empty?

thibault-martinez: I copy paste from the TIP, I'll comment there

thibault-martinez: But we should still be able to parse if present anyway, not a big deal

bindings/nodejs/tests/types/block.spec.ts
Alex6323: Done

bindings/nodejs/tests/types/block.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+        const expected_id = basic_block_transaction_payload_json.id as unknown as HexEncodedString;
+        // TODO: should we add an id() method on Block like we have in Python?
+        expect(Utils.blockId(block, params)).toEqual(expected_id);
+    });
+
+    it('compares validation block from a fixture', async () => {
+        const block = validation_block_json.block as unknown as Block;
+        const params = protocol_parameters.params as unknown as ProtocolParameters;
+        const expected_id = validation_block_json.id as unknown as HexEncodedString;
+        // TODO: should we add an id() method on Block like we have in Python?
```
thibault-martinez: Yeah sure, why not

bindings/nodejs/lib/types/block/core/block.ts
```diff
@@ -118,6 +119,14 @@ class Block {
     asValidation(): ValidationBlockBody {
         return this.body.asValidation();
     }
+
+    /**
+     * Gets the block id.
+     */
+    // id(params: ProtocolParameters): BlockId {
+    id(params: any): BlockId {
```
Thoralf-M: Why not this?
```suggestion
    id(params: ProtocolParameters): BlockId {
```

bindings/nodejs/lib/types/models/info/node-info-protocol.ts
```diff
@@ -54,15 +53,15 @@ export interface ProtocolParameters {
     /**
      * Current supply of base token.
      */
-    tokenSupply: u64;
+    tokenSupply: string;
```
Thoralf-M: Sure about this?
```Ts
(BigInt.prototype as any).toJSON = function () {
    return bigIntToHex(this);
};
```
should do the conversion in one way and for the nodeinfo it should be converted in getInfo()?

bindings/nodejs/lib/types/block/core/block.ts
Alex6323: forgot to clean up

bindings/nodejs/lib/types/models/info/node-info-protocol.ts
Alex6323: I am not sure at all :trollface: I am just glad atm that the tests pass.

bindings/nodejs/tests/types/block.spec.ts
Alex6323: Done

bindings/nodejs/lib/types/models/info/node-info-protocol.ts
Alex6323: seems like it's not trivial to achieve that. One direction BigInt->String is straightforward, but for the other direction there doesn't seem to be an easy way. One would have to convert the fields to bigint manually, right? Also ... this is coming from a `.json` file not `get-info` which means the solution needs to be generic over different sources?

Thoralf-M: > direction there doesn't seem to be an easy way. One would have to convert the fields to bigint manually, right?

Yes, that's what we did in the other functions that require this conversion
The .json file should have the same format as how get-info will be converted to if there is a bigint, so it will just not convert anything and be correct if you use it as parameter somewhere

Alex6323: Fixed by @kwek20 :pray:

Thoralf-M (changes requested):
Conflicting

Thoralf-M (changes requested):
Test still fails with 
```
called `Result::unwrap()` on an `Err` value: Error("cannot deserialize basic block body: data did not match any variant of untagged enum PayloadDto", line: 0, column: 0)
```

Alex6323:
~@Thoralf-M I know ... it's because I think the unlocks aren't serialized correctly anymore probably since the DTO removal. workign on a fix...~ it wasn't that.

bindings/nodejs/tests/types/block.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    // // FIXME:
+    // // "error": "cannot deserialize basic block body: data did not match any variant of untagged enum PayloadDto at line 1 column 2942",
+    // it('compares basic block transaction payload from a fixture', async () => {
+    //     const block = parseBlock(basic_block_transaction_payload_json.block);
+    //     expect(block).toBeInstanceOf(Block);
+    //     const params = protocol_parameters_json.params;
+    //     const expected_id = basic_block_transaction_payload_json.id as BlockId;
+    //     expect(block.id(params)).toEqual(expected_id);
+    // });
```
Thoralf-M: Can be uncommented now?

bindings/nodejs/lib/client/client.ts
```diff
@@ -425,8 +425,8 @@ export class Client {
     /**
      * Get the token supply.
      */
-    async getTokenSupply(): Promise<u64> {
-        return (await this.getProtocolParameters()).tokenSupply;
+    async getTokenSupply(): Promise<bigint> {
```
thibault-martinez: Why? Isn't u64 a bigint already?

sdk/tests/types/protocol_parameters.rs
```diff
@@ -8,7 +8,7 @@ use pretty_assertions::assert_eq;
 // Test from https://github.com/iotaledger/tips/blob/tip49/tips/TIP-0049/tip-0049.md#protocol-parameter-hash
 #[test]
 fn serde_packable_hash() {
-    let file = std::fs::read_to_string("./tests/types/fixtures/protocol_parameters.json").unwrap();
+    let file = std::fs::read_to_string("./tests/types/fixtures/protocol_parameters3.json").unwrap();
```
thibault-martinez: revert?

sdk/tests/types/block_id.rs
```diff
@@ -60,275 +66,60 @@ fn memory_layout() {
...
+    assert_eq!(prefix_hex::encode(&block_bytes), json["bytes"]);
+    assert_eq!(block, Block::unpack_unverified(block_bytes).unwrap());
+    assert_eq!(
+        serde_json::Value::String(block.id(&protocol_parameters()).to_string()),
+        json["id"]
+    );
+}
 
-//     let block_id = block.id(&protocol_parameters()).to_string();
+// TODO waiting for the ContextInput discussion
```
thibault-martinez: remove

bindings/nodejs/tests/types/block.spec.ts
Alex6323: closing it

bindings/nodejs/lib/client/client.ts
Alex6323: yes it is ... IIRC this (bigint) is just an alias for u64 ... I probably did it for consistency, but need to check again.

bindings/nodejs/tests/types/block.spec.ts
```diff
@@ -0,0 +1,38 @@
+// Copyright 2023 IOTA Stiftung
```
Thoralf-M:
```suggestion
// Copyright 2024 IOTA Stiftung
```

sdk/tests/types/block_id.rs
Alex6323: Done

sdk/tests/types/protocol_parameters.rs
Alex6323: Done

bindings/python/tox.ini
```diff
@@ -15,7 +15,7 @@ passenv = *
 deps = -r requirements-dev.txt
 commands =
     pip install .
-    pytest
+    pytest {posargs}
```
Thoralf-M: What's this for?

bindings/nodejs/lib/client/client.ts
Thoralf-M: u64 is an alias for BigInt

bindings/python/tox.ini
DaughterOfMars: chatGPT artifact?

thibault-martinez (changes requested):
Conflicts

bindings/python/tox.ini
Alex6323: @Thoralf-M I wanted to have more control over which tests run, either I wanted only a singe test to run or the ones in a module, in this case I often times ran `tox -e tests -- "tests/test_block.py"` which kept me from forgetting to update the venv. Ended up just using `pytest` directly, but found it useful in the beginning.

Alex6323: removed it again

bindings/nodejs/lib/types/block/core/block.ts
```diff
@@ -118,6 +119,13 @@ class Block {
     asValidation(): ValidationBlockBody {
         return this.body.asValidation();
     }
+
+    /**
+     * Gets the block id.
+     */
+    id(params: ProtocolParameters): BlockId {
+        return Utils.blockId(this, params);
```
Thoralf-M: This should be added like here in lib.rs https://github.com/iotaledger/iota-sdk/blob/1539710218d4d8d71c8400da0d6257ef2074d027/bindings/nodejs/lib/index.ts#L19-L34
otherwise webpack will fail since it depends on the binary

bindings/python/iota_sdk/types/block/block.py
```diff
@@ -64,13 +80,25 @@ class Block:
...
 
+    def to_dict(self):
+        """Custom dict conversion.
+        """
+
+        return {
+            "header": self.header.to_dict(),
+            "body": self.body.to_dict(),
+            "signature": self.signature.to_dict(),
+        }
```
Thoralf-M: I think this shouldn't be needed, `to_dict()` gets called automatically if possible, no?

bindings/nodejs/lib/types/block/core/block.ts
DaughterOfMars: I thought we only did that because it was a breaking change to move it to a different module?

bindings/python/iota_sdk/types/block/block.py
Alex6323: it is needed

bindings/nodejs/lib/client/client.ts
Alex6323: should we use the `u64` alias consistently? there are a few place where we don't currently, `wallet.ts` for example.

bindings/nodejs/lib/types/block/core/block.ts
Thoralf-M: To which module would you move it?

bindings/python/iota_sdk/types/block/block.py
Thoralf-M: For what? I removed it and it's still working fine

bindings/nodejs/lib/types/block/core/block.ts
Thoralf-M: We moved it to this place so it's higher up in the dependency structure, when importing just the type, this doesn't get loaded

thibault-martinez (changes requested):
Conflicts

bindings/python/iota_sdk/types/unlock.py
```diff
@@ -37,59 +37,66 @@ class UnlockType(IntEnum):
...
     """An unlock which must reference a previous unlock which unlocks the NFT that the input is locked to.
     """
-    reference: int
     type: int = field(default_factory=lambda: int(UnlockType.Nft), init=False)
+    reference: int
 
 
 # pylint: disable=missing-function-docstring,unused-argument
+
+def serialize_unlocks(unlock: List[Unlock]) -> List[Dict[str, Any]]:
```
DaughterOfMars: huh?

Alex6323: remnant ... gone

bindings/python/iota_sdk/types/payload.py
```diff
@@ -33,12 +38,94 @@ class TaggedDataPayload:
...
+    """
+    payload_type = d['type']
+    if payload_type == PayloadType.TaggedData:
+        return TaggedDataPayload.from_dict(d)
+    raise Exception(f'invalid payload type: {payload_type}')
+
+
+@json
+@dataclass
+class Transaction:
```
Thoralf-M, line 89: Why didn't you leave this in transaction.py?

bindings/python/iota_sdk/utils.py
```diff
@@ -170,8 +169,8 @@ def block_id(block: Block, params: ProtocolParameters) -> HexStr:
         """ Return a block ID (Blake2b256 hash of block bytes) from a block.
         """
         return _call_method('blockId', {
-            'block': block,
-            'protocol_parameters': params,
+            'block': block.to_dict(),
```
Thoralf-M:
```suggestion
            'block': block,
```
This shouldn't be necessary

bindings/python/iota_sdk/types/payload.py
Alex6323: Because the cyclic import fix didn't work anymore with the changes. It might have changed now with the latest one though. Want me to try to bring it back?

bindings/python/iota_sdk/utils.py
Alex6323: that's certainly true now

bindings/python/iota_sdk/types/payload.py
Thoralf-M: Yes please

Thoralf-M: Had a quick look myself and it's not just about the typings which could be solved with `TYPE_CHECKING`, maybe there is no proper way then, will approve as it is

Thoralf-M (dismissed):

thibault-martinez (approved):

Thoralf-M (approved):

PR #1609, opened 2023-11-12
Use packable struct-level verify-with

thibault-martinez:


kwek20 (dismissed):
So basically the plan is every struct verify_with should be called in the end of the struct::new fn?

thibault-martinez:
> So basically the plan is every struct verify_with should be called in the end of the struct::new fn?

It'll be called after the struct is fully unpacked. Unpack just construct the struct, it doesn't call new because there is not always a new function.

DaughterOfMars (dismissed):
hey, nice

DaughterOfMars (approved):

Thoralf-M (approved):

kwek20 (approved):

PR #1235, opened 2023-09-15
Add work score trait and impls

Alex6323:
# Description of change

This PR introduces the `WorkScore` trait and impls it for all required types (`BlockWrapper` et al).

## Links to any relevant issues

Closes #1018

## Type of change

Choose a type of change, and delete any options that are not relevant.

- Enhancement (a non-breaking change which adds functionality)

## How the change has been tested

Not (yet)

PR #1610, opened 2023-11-13
Calculate and set `max_burned_mana` field

### LINKED ISSUE #1438: Add max_burned_mana utilities to BasicBlockBuilder
Set values / compute value

Alex6323:
# Description of change

Calculates and sets the `max_burned_mana` value for a `BasicBlockBody`. Currently blocked by #1235.

Closes #1438


sdk/src/client/api/block_builder/mod.rs
```diff
@@ -34,6 +35,17 @@ impl ClientInner {
 
         let protocol_params = self.get_protocol_parameters().await?;
 
+        let mut basic_block_body = BlockBody::build_basic(issuance.strong_parents()?)
+            .with_weak_parents(issuance.weak_parents()?)
+            .with_shallow_like_parents(issuance.shallow_like_parents()?)
+            .with_payload(payload)
+            .finish()?;
+
+        basic_block_body.set_max_burned_mana_to_minimum(
```
thibault-martinez: Can we not make it part of the builder? This is very inconsistent with the codebase, we don't have a single mutable method on the built types.

sdk/src/types/block/core/basic.rs
```diff
@@ -152,6 +152,23 @@ impl BasicBlockBody {
...
     }
+
+    /// Sets the `max_burned_mana` field to the minimum possible mana amount for the block to get accepted by the
+    /// network.
+    pub fn set_max_burned_mana_to_minimum(&mut self, params: WorkScoreParameters, reference_mana_cost: u64) {
+        self.max_burned_mana = self.work_score(params) as u64 * reference_mana_cost;
+    }
+}
+
+impl WorkScore for BasicBlockBody {
```
thibault-martinez: Remove

sdk/src/client/api/block_builder/mod.rs
Alex6323: we could pass the params and the rmc to the `finish` method. I also tried that but found it very ugly :shrug:

thibault-martinez: I don't think it's ugly, it's less code and doesn't break our non-mutability promise. It's very similar to what we did with StorageScore I guess.

Thoralf-M: Having it like for the outputs with the minimum amount would probably be good
So two `new` methods, one where the user provides the amount and one where the user provides the WorkScoreParameters with the reference_mana_cost which then calculates the minimum?

thibault-martinez: I agree, it should be the exact same thing

Alex6323: aight

sdk/src/types/block/core/basic.rs
Alex6323: you mean the method, not the trait impl I guess

thibault-martinez: I mean the trait impl. Why are you adding it again?

sdk/src/client/api/block_builder/mod.rs
```diff
@@ -43,12 +44,14 @@ impl ClientInner {
...
                 issuer_id,
             ),
-            // TODO: burned mana calculation
-            BlockBody::build_basic(issuance.strong_parents()?, 0)
+            BlockBody::build_basic(issuance.strong_parents()?)
                 .with_weak_parents(issuance.weak_parents()?)
                 .with_shallow_like_parents(issuance.shallow_like_parents()?)
                 .with_payload(payload)
-                .finish_block_body()?,
+                .finish_block_body_with_minimum_mana_amount(
```
thibault-martinez: I would still do as Thoralf suggested as this doesn't allow passing a specific value

sdk/src/types/block/core/basic.rs
Alex6323: lol, that wasn't me (git merge :angry: )

sdk/src/types/block/core/basic.rs
```diff
@@ -84,10 +84,42 @@ impl BasicBlockBodyBuilder {
         })
     }
 
+    /// Finishes the builder into a [`BasicBlockBody`] with the minimum amount of mana required for the block to get
+    /// accepted by the network.
+    pub fn finish_with_minimum_mana_amount(
```
DaughterOfMars: Should this be part of the block builder?

sdk/src/client/api/block_builder/mod.rs
Alex6323: lol, I only see the previous 2 comments of you guys just now after refreshing the page ... god I hate this :joy:

Alex6323: I had it with another `new` function, but then we need optional fields in the builder just for them. It makes much more sense to pass them in the finish method to prevent this.

sdk/src/client/api/block_builder/mod.rs
Alex6323: there is `with_max_burned_mana`

Thoralf-M: Specific value could be provided with `with_max_burned_mana(mana)` and then `.finish_block_body()`
Not sure what I like more

Thoralf-M: Probably easier to forget setting this value to anything with the functions like in this PR now

thibault-martinez: Ah yeah I forgot `with_max_burned_mana`, but I agree, super easy to not set it and then it defaults to 0 which is bad. I would actually remove `with_max_burned_mana`.

sdk/src/types/block/core/basic.rs
thibault-martinez: What do you mean exactly?

sdk/src/client/api/block_builder/mod.rs
thibault-martinez: I mean, maybe not remove it as we still have `with_amount` for outputs but you get me

sdk/src/types/block/core/basic.rs
DaughterOfMars: Well, whenever you're building a block, you'll want this value to be set automatically when you `finish`, right?

thibault-martinez: Yes

DaughterOfMars: So then this should probably be part of that process rather than a fn in the client

thibault-martinez: This is not in the client :trollface:

sdk/src/client/api/block_builder/mod.rs
Alex6323: yeah, I wouldn't remove it

sdk/src/types/block/core/basic.rs
DaughterOfMars: oh man I was clearly confused

DaughterOfMars: shouldn't this all just be `finish` though? I mean it's invalid to not set this right

thibault-martinez: I would very much prefer `finish` yeah. It can't not be set yes, but it can be set to a specific value or to minimum, like the storage deposit.

thibault-martinez: Maybe a bit more annoying than storage deposit as it needs 2 params though

Alex6323: okay ... changing it

sdk/src/types/block/rand/block.rs
```diff
@@ -45,7 +45,7 @@ pub fn rand_basic_block_body_with_strong_parents(strong_parents: basic::StrongPa
 pub fn rand_basic_block_body_builder_with_strong_parents(
     strong_parents: basic::StrongParents,
 ) -> BasicBlockBodyBuilder {
-    BlockBody::build_basic(strong_parents, rand_number())
+    BlockBody::build_basic(strong_parents, basic::MaxBurnedManaAmount::Amount(0))
```
Thoralf-M: Does this work?
```suggestion
    BlockBody::build_basic(strong_parents, basic::MaxBurnedManaAmount::Amount(rand_number()))
```

sdk/src/types/block/core/basic.rs
```diff
@@ -64,24 +99,35 @@ impl BasicBlockBodyBuilder {
         self
     }
 
-    /// Adds max burned mana to a [`BasicBlockBodyBuilder`].
-    #[inline(always)]
-    pub fn with_max_burned_mana(mut self, max_burned_mana: u64) -> Self {
-        self.max_burned_mana = max_burned_mana;
-        self
-    }
+    // TODO: keep or remove?
```
Thoralf-M: Currently I don't see a use case for this, so I guess it's fine to remove

Thoralf-M (dismissed):

sdk/src/client/api/block_builder/mod.rs
Alex6323: dealt with using the `enum` approach

sdk/src/client/api/block_builder/mod.rs
Alex6323: decided with @Thoralf-M to remove it

sdk/src/types/block/core/basic.rs
Alex6323: Done

sdk/src/client/api/block_builder/mod.rs
thibault-martinez, line 1: Remove the unused imports of this file

sdk/src/client/api/block_builder/mod.rs
```diff
@@ -43,12 +44,18 @@ impl ClientInner {
...
+                issuance.strong_parents()?,
+                MaxBurnedManaAmount::MinimumAmount {
+                    params: protocol_params.work_score_parameters,
+                    reference_mana_cost: issuance.commitment.reference_mana_cost(),
+                },
+            )
+            .with_weak_parents(issuance.weak_parents()?)
+            .with_shallow_like_parents(issuance.shallow_like_parents()?)
+            .with_payload(payload)
+            .finish()?
```
thibault-martinez: Can you use `finish_block_body` instead of finish + into pls

Alex6323: Done

sdk/src/types/block/core/mod.rs
```diff
@@ -81,7 +82,10 @@ impl BlockBody {
 
     /// Creates a new [`BasicBlockBodyBuilder`].
     #[inline(always)]
-    pub fn build_basic(strong_parents: self::basic::StrongParents, max_burned_mana: u64) -> BasicBlockBodyBuilder {
+    pub fn build_basic(
```
thibault-martinez, line 85: Maybe I'm annoying now but I would do the same as the output and provide two builders. Having a `MaxBurnedManaAmount` as param is not a very great UX IMO.

Alex6323: You're right. Changed to something more convenient and less verbose.

thibault-martinez: That's better than what I had in mind, maybe we could do the same for outputs

thibault-martinez: (another PR if ever)

thibault-martinez: Even though it's still nice to be explicit on the name

Alex6323: yeah ... only problem with that is you'ld have to adapt (possibly add) builders fns when the enum changes.

Thoralf-M (approved):

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1617, opened 2023-11-14
Fix private tangle action

thibault-martinez:


.github/actions/private-tangle/setup/action.yml
```diff
@@ -18,13 +18,13 @@ runs:
 
         # remove the last 5 lines and add docker compose with `-d` to run it in the background
         sed -i -n -e :a -e '1,5!{P;N;D;};N;ba' run.sh
-        echo "docker compose -f \$DOCKER_COMPOSE_FILE up -d" >> run.sh
+        echo "docker compose -f docker-compose.yml up -d" >> run.sh
```
thibault-martinez, line 21: Looks like this env var is never defined, not sure where this is supposed to happen

Thoralf-M: In the run.sh file before https://github.com/iotaledger/iota-core/blob/5851ebd9b1fe272be6a0f86e92f0841f457bbfd5/tools/docker-network/run.sh#L17

Thoralf-M: Not sure why it doesn't like it anymore, but just directly using `docker-compose.yml` should be fine

Thoralf-M (approved):

.github/actions/private-tangle/setup/action.yml
thibault-martinez: But the env var is evaluated in the shell before being written in run.sh, not sure how that ever worked...

Thoralf-M: With the `\` it shouldn't get evaluated

thibault-martinez: Should we just remove the file name? They don't have it in the original docker compose up

thibault-martinez: Ahhhh but they have removed `DOCKER_COMPOSE_FILE` from run.sh I see

DaughterOfMars (approved):

tadzik (approved):

Alex6323 (approved):

PR #1619, opened 2023-11-15
Remove SDRUC supply check

thibault-martinez:
Turns out this check never existed in https://github.com/iotaledger/tips/blob/tip38/tips/TIP-0038/tip-0038.md#storage-deposit-return-unlock-condition ?

### DIFF
--- a/bindings/core/src/method_handler/wallet.rs
+++ b/bindings/core/src/method_handler/wallet.rs
@@ -297,12 +297,11 @@ pub(crate) async fn call_wallet_method_internal(wallet: &Wallet, method: WalletM
             Response::PreparedTransaction(PreparedTransactionDataDto::from(&data))
         }
         WalletMethod::PrepareTransaction { outputs, options } => {
-            let token_supply = wallet.client().get_token_supply().await?;
             let data = wallet
                 .prepare_transaction(
                     outputs
                         .into_iter()
-                        .map(|o| Ok(Output::try_from_dto_with_params(o, token_supply)?))
+                        .map(|o| Ok(Output::try_from(o)?))
                         .collect::<Result<Vec<Output>>>()?,
                     options,
                 )
@@ -342,12 +341,11 @@ pub(crate) async fn call_wallet_method_internal(wallet: &Wallet, method: WalletM
             Response::SentTransaction(TransactionWithMetadataDto::from(&transaction))
         }
         WalletMethod::SendOutputs { outputs, options } => {
-            let token_supply = wallet.client().get_token_supply().await?;
             let transaction = wallet
                 .send_outputs(
                     outputs
                         .into_iter()
-                        .map(|o| Ok(Output::try_from_dto_with_params(o, token_supply)?))
+                        .map(|o| Ok(Output::try_from(o)?))
                         .collect::<iota_sdk::wallet::Result<Vec<Output>>>()?,
                     options,
                 )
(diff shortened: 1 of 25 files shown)


Thoralf-M (dismissed):
Love it

Alex6323 (dismissed):

sdk/src/types/block/output/account.rs
```diff
@@ -676,7 +676,7 @@ pub(crate) mod dto {
         type Dto = AccountOutputDto;
         type Error = Error;
 
-        fn try_from_dto_with_params_inner(dto: Self::Dto, params: ValidationParams<'_>) -> Result<Self, Self::Error> {
+        fn try_from_dto_with_params_inner(dto: Self::Dto, _params: ValidationParams<'_>) -> Result<Self, Self::Error> {
```
DaughterOfMars: now this should just be a `TryFrom`

thibault-martinez: Right

thibault-martinez: Done

Thoralf-M (dismissed):

bindings/core/src/method_handler/wallet.rs
```diff
@@ -297,12 +297,11 @@ pub(crate) async fn call_wallet_method_internal(wallet: &Wallet, method: WalletM
             Response::PreparedTransaction(PreparedTransactionDataDto::from(&data))
         }
         WalletMethod::PrepareTransaction { outputs, options } => {
-            let token_supply = wallet.client().get_token_supply().await?;
             let data = wallet
                 .prepare_transaction(
                     outputs
                         .into_iter()
-                        .map(|o| Ok(Output::try_from_dto_with_params(o, token_supply)?))
+                        .map(|o| Ok(Output::try_from(o)?))
```
DaughterOfMars: clippy gonna complain

thibault-martinez: Why? What can I do because of the Ok ?

DaughterOfMars: pretty sure this can be reduced to ```.map(Output::try_from)```

thibault-martinez: No, we need to `?` to convert the error

thibault-martinez: At least for this one, I'll try the others

DaughterOfMars: That can happen in the collect instead since they are all the same error

thibault-martinez: There was one

Thoralf-M (dismissed):

bindings/core/src/method_handler/wallet.rs
thibault-martinez: No they are not the same error

DaughterOfMars: No, I mean all the collected results have the same error

thibault-martinez: Ok so maybe I'm not understanding what you mean. Also in a hurry at the airport :trollface: 
Can you show me because simply `.map(Output::try_from)` doesn't work.

DaughterOfMars:
```
.map(Output::try_from)
      .collect::<Result<Vec<Output>, _>>()?,
```

I removed the `Result` import and manually used `crate::Result` in the return type, but that's all

thibault-martinez: I see, thanks, done

thibault-martinez: If it was only me I would remove these Result typedefs, I hate them

DaughterOfMars (approved):

Thoralf-M (approved):

Alex6323 (approved):

PR #1630, opened 2023-11-16
Remove token supply from validation params

DaughterOfMars:
# Description of change

Simply not needed anymore.


sdk/src/types/mod.rs
```diff
@@ -74,36 +74,17 @@ impl<'a, T> From<&'a Boo<'a, T>> for Boo<'a, T> {
 #[derive(Clone, Default, Debug)]
 pub struct ValidationParams<'a> {
```
thibault-martinez: Do we even still need this at all?

DaughterOfMars: probably not, we could just take a protocol parameters directly

thibault-martinez (approved):

Thoralf-M (approved):

Alex6323 (approved):

PR #1642, opened 2023-11-19
Change *_ignore_errors to *_ignore_not_found

thibault-martinez:


sdk/src/client/node_api/core/mod.rs
```diff
@@ -33,14 +33,12 @@ impl Client {
 
     /// Requests outputs by their output ID in parallel, ignoring failed requests.
     /// Useful to get data about spent outputs, that might not be pruned yet.
-    pub async fn get_outputs_ignore_errors(&self, output_ids: &[OutputId]) -> Result<Vec<Output>> {
-        Ok(
-            futures::future::join_all(output_ids.iter().map(|id| self.get_output(id)))
-                .await
-                .into_iter()
-                .filter_map(Result::ok)
```
Thoralf-M, line 41: Shouldn't we instead filter only `crate::client::node_api::error::Error::NotFound(_)` and return other errors?

thibault-martinez: Maybe, but they should be called _ignore_not_found then

Thoralf-M: Yeah fine to have them renamed, but the idea was to just not fail if some outputs were already pruned. If it fails for another reason, we should still return an error I think

thibault-martinez: Ok I see, that's not what they were currently doing, they were never failing atm.

Thoralf-M: Probably also fine like this in 99,9% of the cases, but there is a small change that it could be a problem so we should change it

thibault-martinez: Changed

Thoralf-M (dismissed):

Thoralf-M (approved):

sdk/src/client/node_api/core/mod.rs
```diff
@@ -31,51 +31,45 @@ impl Client {
...
-        Ok(
-            futures::future::join_all(output_ids.iter().map(|id| self.get_output_metadata(id)))
-                .await
-                .into_iter()
-                .filter_map(Result::ok)
-                .collect(),
-        )
+    /// Requests metadata for outputs by their output ID in parallel, ignoring outputs not found.
+    pub async fn get_outputs_metadata_ignore_not_found(&self, output_ids: &[OutputId]) -> Result<Vec<OutputMetadata>> {
+        futures::future::join_all(output_ids.iter().map(|id| self.get_output_metadata(id)))
```
Alex6323, line 51: hmmm ... may as well want to use `try_join_all` here? Otherwise the behavior is a bit different and not just ignoring outputs not found.

thibault-martinez: No, `try_join_all` will stop at the first error

sdk/src/client/node_api/core/mod.rs
```diff
@@ -31,51 +31,45 @@ impl Client {
...
         output_ids: &[OutputId],
     ) -> Result<Vec<OutputWithMetadata>> {
-        Ok(
-            futures::future::join_all(output_ids.iter().map(|id| self.get_output_with_metadata(id)))
-                .await
-                .into_iter()
-                .filter_map(Result::ok)
-                .collect(),
-        )
+        futures::future::join_all(output_ids.iter().map(|id| self.get_output_with_metadata(id)))
```
Alex6323, line 69: `try_join_all` so we notice failed requests?

thibault-martinez: Same answer

sdk/src/client/node_api/core/mod.rs
Alex6323: yeah, that's fine but probably should described in the doc that it also ignores failed requests?

thibault-martinez: Idk I think it's stated already `ignoring outputs not found`. What do you want to add ?

Alex6323 (approved):

PR #1655, opened 2023-11-20
Add SecretManage::generate_ed25519_public_keys

thibault-martinez:


Thoralf-M (dismissed):

sdk/src/client/secret/mod.rs
```diff
@@ -66,6 +70,15 @@ use crate::{
 pub trait SecretManage: Send + Sync {
     type Error: std::error::Error + Send + Sync;
 
+    /// Generates public keys.
```
Alex6323, line 73: Maybe add the same link for details on the coin type since it also applies for this method?

thibault-martinez: I've done it quick and dirty to unblock myself because this is all going to change anyway. But I've added it.

Alex6323 (approved):

Thoralf-M (approved):

PR #1662, opened 2023-11-21
Fix SlotCommitment JSON

thibault-martinez:


Thoralf-M (dismissed):

bindings/nodejs/lib/types/utils/bridge/utils.ts
```diff
@@ -197,8 +197,8 @@ export interface __OutputHexBytes__ {
 // Modified `SlotCommitment` with bigint types converted to strings.
 type SlotCommitmentConverted = Omit<
     SlotCommitment,
-    'index' | 'cumulativeWeight'
-> & { index: string; cumulativeWeight: string };
+    'cumulativeWeight' | 'referenceManaCost'
+> & { cumulativeWeight: string, referenceManaCost: string };
```
thibault-martinez: Is this the right way to do? It's the first time I see this

Thoralf-M: It's the only type where we do this, seems like it shouldn't be needed because we do the conversion for bigint in https://github.com/iotaledger/iota-sdk/blob/b23d0e23e65315332498694bec8ff31d8a300160/bindings/nodejs/lib/index.ts#L15 and utils/utils.ts `computeSlotCommitmentId()`, but not sure either

Thoralf-M (approved):

Alex6323 (approved):

PR #1680, opened 2023-11-24
Add support for MultiAddress in ISA (selected inputs only)

thibault-martinez:


sdk/src/client/api/block_builder/input_selection/requirement/sender.rs
```diff
@@ -42,6 +42,30 @@ impl InputSelection {
...
+                            .output
+                            .required_and_unlocked_address(self.slot_index, input.output_id())?
+                            .0;
+
+                        if &required_address == weight_address.address() {
+                            cumulative_weight += weight_address.weight() as u16;
+                            break;
+                        }
+                    }
+                }
```
Alex6323: could be done more efficiently I think (but depends on how costly the fn call and how likely the worst case scenario is)

thibault-martinez: How? Without a substantial change to ISA I'm not sure. Also this is a temporary feature as suggested by the TODO.

Alex6323: I would probably also check the threshold in the loop so we can break if the threshold is surpassed already, unless we really need the exact amount of cumulative weight.

thibault-martinez: Will do

thibault-martinez: Done

Alex6323 (approved):

PR #1682, opened 2023-11-24
Python: small fixes

kwek20:
# Description of change

Ran node info, works now.

Also did an fmt on rust, how did that pass :D


bindings/python/iota_sdk/client/client.py
```diff
@@ -321,6 +322,13 @@ def get_node(self) -> Dict[str, Any]:
         """
         return self._call_method('getNode')
 
+    def get_info(self) -> NodeInfoWrapper:
```
thibault-martinez: We already have this method in `bindings/python/iota_sdk/client/_node_core_api.py` ?

Thoralf-M: Yes and even without `print(f'{result}')` :trollface:

kwek20: i swear, i looked everywhere and used the search :D it didnt show!!

bindings/python/iota_sdk/client/client.py
```diff
@@ -14,6 +14,7 @@
 from iota_sdk.types.block.signed_block import UnsignedBlock
 from iota_sdk.types.common import HexStr, Node
 from iota_sdk.types.feature import Feature
+from iota_sdk.types.node_info import NodeInfoWrapper
```
Thoralf-M: Remove as well?

Thoralf-M (approved):

thibault-martinez (approved):

PR #1702, opened 2023-11-27
Add block issuer key source for implicit account transition

### LINKED ISSUE #1666: Add more PublicKey generation fallbacks
- [x] https://github.com/iotaledger/iota-sdk/pull/1598/files#diff-bb97bc42ed78814b7c87abe4a31292fcfa12055d3bf69e2dd7ed05666178fbb0R51

DaughterOfMars:
# Description of change

Adds block issuer key source override to implicit account transition which can be a key or a bip path.

Closes #1666


Thoralf-M (dismissed):

bindings/core/src/method/wallet.rs
```diff
@@ -194,7 +195,11 @@ pub enum WalletMethod {
     /// Prepares to transition an implicit account to an account.
     /// Expected response: [`PreparedTransaction`](crate::Response::PreparedTransaction)
     #[serde(rename_all = "camelCase")]
-    PrepareImplicitAccountTransition { output_id: OutputId },
+    PrepareImplicitAccountTransition {
+        output_id: OutputId,
+        public_key: Option<PublicKey>,
+        bip_path: Option<Bip44>,
```
Alex6323: So if `public_key` is set the value in `bip_path` is ignored no matter if it's Some or None?

Could this not already be using the enum:
```rust
key_source: Option<BlockIssuerKeySource>
```

DaughterOfMars: Yeah, that is correct. I didn't want to make that utility enum public really, and this seemed easier for users. I forgot to put default tags on them though.

Alex6323: okay :+1:

Alex6323 (dismissed):

Thoralf-M (dismissed):

bindings/core/src/method_handler/wallet.rs
```diff
@@ -210,8 +210,18 @@ pub(crate) async fn call_wallet_method_internal(wallet: &Wallet, method: WalletM
...
+            output_id,
+            public_key,
+            bip_path,
+        } => {
+            let data = if public_key.is_some() {
+                wallet
+                    .prepare_implicit_account_transition(&output_id, public_key)
+                    .await?
+            } else {
+                wallet.prepare_implicit_account_transition(&output_id, bip_path).await?
```
thibault-martinez, line 220: Is there nothing we can do with an Option method to do one or the other without the if ?

bindings/core/src/method/wallet.rs
```diff
@@ -194,7 +195,13 @@ pub enum WalletMethod {
     /// Prepares to transition an implicit account to an account.
     /// Expected response: [`PreparedTransaction`](crate::Response::PreparedTransaction)
     #[serde(rename_all = "camelCase")]
-    PrepareImplicitAccountTransition { output_id: OutputId },
+    PrepareImplicitAccountTransition {
+        output_id: OutputId,
+        #[serde(default)]
+        public_key: Option<PublicKey>,
```
thibault-martinez: Since this is not our type, how does this serialise ? prefixed hex ?

cli/src/wallet_cli/mod.rs
```diff
@@ -584,8 +586,18 @@ pub async fn implicit_account_creation_address_command(wallet: &Wallet) -> Resul
...
+    wallet: &Wallet,
+    output_id: OutputId,
+    public_key: Option<String>,
+) -> Result<(), Error> {
+    let public_key = public_key
+        .map(|s| {
+            PublicKey::try_from_bytes(prefix_hex::decode(s).map_err(|e| Error::Miscellaneous(e.to_string()))?)
+                .map_err(|e| Error::Miscellaneous(e.to_string()))
+        })
+        .transpose()?;
```
thibault-martinez, line 598: No way to make this prettier 😭 ?

sdk/src/wallet/operations/transaction/account.rs
```diff
@@ -81,3 +103,9 @@ where
             .await
     }
 }
+
+#[derive(From)]
+pub enum BlockIssuerKeySource {
+    Key(PublicKey),
+    Path(Bip44),
```
thibault-martinez: I'd maybe call the variant BipPath, Path is a bit too generic IMO

bindings/core/src/method_handler/wallet.rs
DaughterOfMars: no, because the static typing is different

bindings/core/src/method/wallet.rs
DaughterOfMars: HMMMM I doubt it. probably as bytes. What do we want?

cli/src/wallet_cli/mod.rs
DaughterOfMars: I can try I guess, but I don't think so

sdk/src/wallet/operations/transaction/account.rs
DaughterOfMars: alrighty

bindings/core/src/method/wallet.rs
thibault-martinez: prefixed_hex I'd say, but maybe this is already the case, have to check

bindings/core/src/method_handler/wallet.rs
thibault-martinez: sad reacts only

bindings/core/src/method_handler/wallet.rs
```diff
@@ -204,8 +205,20 @@ pub(crate) async fn call_wallet_method_internal(wallet: &Wallet, method: WalletM
...
-        WalletMethod::PrepareImplicitAccountTransition { output_id } => {
-            let data = wallet.prepare_implicit_account_transition(&output_id).await?;
+        WalletMethod::PrepareImplicitAccountTransition {
+            output_id,
+            public_key,
+            bip_path,
+        } => {
+            let data = if let Some(public_key_str) = public_key {
+                let public_key = PublicKey::try_from_bytes(prefix_hex::decode(public_key_str)?)
+                    .map_err(iota_sdk::wallet::Error::other)?;
```
thibault-martinez: Why is this needed? Wallet has a From impl for crypto errors.

thibault-martinez: Ah wait PublicKey is our type

thibault-martinez: Hmm I think I would add a Types variant in Wallet + a from tbh, seems like something useful

thibault-martinez: Ahhh there is `Block` but no from for it, maybe just add it?

bindings/core/src/method/wallet.rs
DaughterOfMars: done

cli/src/wallet_cli/mod.rs
DaughterOfMars: it cannot

thibault-martinez: Alright, thanks

bindings/core/src/method_handler/wallet.rs
DaughterOfMars: I'm confused, you mean add an error variant?

thibault-martinez: I'm confused as well, why can't we just `?` since the wallet error has a from impl for crypto error ?

DaughterOfMars: it didn't help much but I fixed it

thibault-martinez: But for my understanding, why can't we use ? directly ?

DaughterOfMars: Because the error type returned here is local to the bindings, not the wallet error

thibault-martinez (commented):
Conflicts

Thoralf-M (approved):

sdk/src/wallet/error.rs
```diff
@@ -129,6 +129,12 @@ pub enum Error {
     ImplicitAccountNotFound,
 }
 
+impl Error {
+    pub fn other<E: 'static + std::error::Error + Send + Sync>(err: E) -> Self {
```
thibault-martinez, line 133: Do you want to keep this even if not used atm?

DaughterOfMars: Hmm. I think so. We have the variant even though we don't use it, but if we do I'm going to want this method to exist. But technically we could add this method at any time without a breaking change so 🤷‍♀️

thibault-martinez (approved):

PR #1712, opened 2023-11-29
Make output dtos private

Thoralf-M:
# Description of change

Make output dtos private. Since the protocol parameters aren't needed to deserialize them, they don't need to be public anymore

## Type of change

cargo test


sdk/src/client/api/types.rs
```diff
@@ -160,7 +160,7 @@ impl TryFrom<RemainderDataDto> for RemainderData {
 
     fn try_from(dto: RemainderDataDto) -> Result<Self, Self::Error> {
         Ok(Self {
-            output: Output::try_from(dto.output)?,
```
thibault-martinez, line 163: Can be a from now ?

Thoralf-M: Yes, but I would do other types in an own PR so this here doesn't get bigger. Already opened an issue https://github.com/iotaledger/iota-sdk/issues/1713

Thoralf-M: I mean we can remove the whole dto here

thibault-martinez: Ok sir

sdk/src/client/api/types.rs
```diff
@@ -169,7 +169,7 @@ impl TryFrom<RemainderDataDto> for RemainderData {
 impl From<&RemainderData> for RemainderDataDto {
```
DaughterOfMars, line 169: Maybe just make this ```impl From<RemainderData> for RemainderDataDto``` now, and remove the clones?

sdk/src/client/secret/types.rs
```diff
@@ -179,7 +179,7 @@ impl TryFrom<InputSigningDataDto> for InputSigningData {
 impl From<&InputSigningData> for InputSigningDataDto {
```
DaughterOfMars, line 179: Same

DaughterOfMars: And others

sdk/src/types/block/output/mod.rs
```diff
@@ -439,55 +440,12 @@ pub trait MinimumOutputAmount: StorageScore {
...
-                OutputDto::Anchor(o) => Self::Anchor(AnchorOutput::try_from(o)?),
-                OutputDto::Foundry(o) => Self::Foundry(FoundryOutput::try_from(o)?),
-                OutputDto::Nft(o) => Self::Nft(NftOutput::try_from(o)?),
-                OutputDto::Delegation(o) => Self::Delegation(DelegationOutput::try_from(o)?),
-            })
-        }
-    }
-
-    impl<'de> Deserialize<'de> for OutputDto {
+    impl<'de> Deserialize<'de> for Output {
```
DaughterOfMars: We *could* just make this an untagged enum derive now

DaughterOfMars: or, rather, just derive both

sdk/src/types/block/output/account.rs
```diff
@@ -617,7 +617,7 @@ pub(crate) mod dto {
     /// Describes an account in the ledger that can be controlled by the state and governance controllers.
     #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
     #[serde(rename_all = "camelCase")]
-    pub struct AccountOutputDto {
+    pub(crate) struct AccountOutputDto {
```
thibault-martinez, line 620: This could be private. I know it's for the test but maybe it could just do a JSON back and forth and it would achieve the same. Maybe later.

sdk/src/types/block/output/mod.rs
thibault-martinez: For reference, `Feature` does that and it works fine

sdk/src/types/block/output/delegation.rs
```diff
@@ -449,7 +449,7 @@ pub(crate) mod dto {
 
     #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
     #[serde(rename_all = "camelCase")]
-    pub struct DelegationOutputDto {
+    struct DelegationOutputDto {
```
thibault-martinez, line 452: whole mod can be private ?

sdk/src/client/api/types.rs
Thoralf-M: Well but we can just remove the dto completely, no? https://github.com/iotaledger/iota-sdk/issues/1713

sdk/src/types/block/output/mod.rs
Thoralf-M: Ah it seems serde is checking the next variant if one fails, I thought there could be issues if they have the same fields
Removed

thibault-martinez (dismissed):

sdk/src/types/block/output/mod.rs
thibault-martinez: It does check the next variant but the variant are serialised with their type so even if they have the same fields the type would different

thibault-martinez: At least I think

sdk/src/types/block/output/account.rs
Thoralf-M: Currently a bit annoying to change with the `try_from_dtos()` part below, but we can probably remove all that 2 PRs down the line

sdk/src/client/api/types.rs
DaughterOfMars: Indeed!

sdk/src/types/block/output/mod.rs
DaughterOfMars: Yes, that is due to the checking in the individual Dto impls created by the macro

sdk/src/types/block/output/delegation.rs
```diff
@@ -433,7 +433,7 @@ fn verify_unlock_conditions_packable<const VERIFY: bool>(
 }
 
 #[cfg(feature = "serde")]
-pub(crate) mod dto {
+pub mod dto {
```
DaughterOfMars: pub still?

thibault-martinez: Oops

Thoralf-M: Fixed

sdk/src/client/secret/types.rs
Thoralf-M: Same as above, would rather do in another PR

thibault-martinez (approved):

DaughterOfMars (approved):

PR #1714, opened 2023-11-29
Some API changes

thibault-martinez:
https://github.com/iotaledger/iota.go/pull/625


sdk/src/wallet/operations/syncing/outputs.rs
```diff
@@ -60,7 +60,7 @@ where
                     output_id: output_with_meta.metadata().output_id().to_owned(),
                     metadata: *output_with_meta.metadata(),
                     output: output_with_meta.output().clone(),
-                    is_spent: output_with_meta.metadata().is_spent(),
+                    is_spent: output_with_meta.metadata().spent().is_some(),
```
Thoralf-M: How about adding `.is_spent()` which calls `.spent().is_some()` inside?

sdk/src/wallet/update.rs
```diff
@@ -69,7 +69,8 @@ where
                     wallet_data.unspent_outputs.remove(&output_id);
                     // Update spent data fields
                     if let Some(output_data) = wallet_data.outputs.get_mut(&output_id) {
-                        output_data.metadata.set_spent(true);
+                        // TODO wat do?
```
Thoralf-M: Guess we just don't update it here

sdk/src/wallet/operations/syncing/outputs.rs
thibault-martinez: Sure

sdk/src/types/api/core.rs
```diff
@@ -173,6 +174,31 @@ pub struct BaseTokenResponse {
...
+    /// The total stake of the pool, including delegators.
+    #[serde(with = "string")]
+    pool_stake: u64,
+    /// The stake of a validator.
+    #[serde(with = "string")]
+    validator_stake: u64,
+    /// The fixed cost of the validator, which it receives as part of its Mana rewards.
+    #[serde(with = "string")]
+    fixed_cost: u64,
+    /// Shows whether validator was active recently.
```
Alex6323:
```suggestion
    /// Shows whether the validator was active recently.
```

sdk/src/wallet/operations/syncing/outputs.rs
thibault-martinez: Done

sdk/src/wallet/update.rs
thibault-martinez: Is that not going to cause issues somewhere else? Do we rely on this or it's purely informative ?

Thoralf-M: I think we rely on the is_spent value below, but yeah should be double checked to be sure

Thoralf-M: Looks like we only use it here https://github.com/iotaledger/iota-sdk/blob/9387a3535dc74978fa1e170f8ac5e9152033a01e/sdk/src/wallet/operations/syncing/transactions.rs#L248 and there I'm not sure if it should be changed or not
Or we have to set some 0x0000 values for the case where the output is spent and pruned so we don't get this information anymore

maybe leave the TODO for now

Thoralf-M (dismissed):

sdk/src/wallet/update.rs
thibault-martinez: Let me create an issue and update the TODO then

sdk/src/types/api/core.rs
```diff
@@ -173,6 +174,31 @@ pub struct BaseTokenResponse {
...
+    #[serde(with = "string")]
+    validator_stake: u64,
+    /// The fixed cost of the validator, which it receives as part of its Mana rewards.
+    #[serde(with = "string")]
+    fixed_cost: u64,
+    /// Shows whether the validator was active recently.
+    active: bool,
+    /// The latest protocol version the validator supported.
+    latest_supported_protocol_version: u8,
+    // The protocol hash of the latest supported protocol of the validator.
```
Alex6323:
```suggestion
    /// The protocol hash of the latest supported protocol of the validator.
```

Thoralf-M (approved):

sdk/src/types/api/core.rs
```diff
@@ -173,6 +174,31 @@ pub struct BaseTokenResponse {
     pub decimals: u32,
 }
 
+/// Information of a validator.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ValidatorResponse {
```
Alex6323, line 180: `ValidatorInfoResponse` would be a kinda better name, similar to `NodeInfo` but :shrug:

Alex6323: not important

Alex6323 (approved):

PR #1725, opened 2023-12-01
Fix CI doc warnings

thibault-martinez:


bindings/core/src/method/wallet.rs
```diff
@@ -133,7 +133,7 @@ pub enum WalletMethod {
     /// Get wallet balance information.
     /// Expected response: [`Balance`](crate::Response::Balance)
     GetBalance,
-    /// Get the [`Output`](iota_sdk::types::block::output::Output) that minted a native token by its TokenId
+    /// Get the [`Output`] that minted a native token by its TokenId
```
Alex6323:
```suggestion
    /// Get the [`Output`] that minted a native token by its TokenId.
```

Alex6323: maybe wait before committing to batch them

Alex6323 (dismissed):

Alex6323 (approved):

PR #1730, opened 2023-12-03
Two more Packable derive

thibault-martinez:


Thoralf-M (approved):

sdk/src/types/block/core/block.rs
```diff
@@ -111,60 +115,29 @@ impl BlockHeader {
...
-        Ok(Self {
-            protocol_version,
-            network_id,
-            issuing_time,
-            slot_commitment_id,
-            latest_finalized_slot,
-            issuer_id,
-        })
+fn verify_network_id<const VERIFY: bool>(network_id: &u64, params: &ProtocolParameters) -> Result<(), Error> {
+    if VERIFY && *network_id != params.network_id() {
```
Alex6323, line 133:
```suggestion
    if VERIFY && network_id != &params.network_id() {
```
not that it mattered, but I would prefer this.

sdk/src/types/block/core/block.rs
```diff
@@ -111,60 +115,29 @@ impl BlockHeader {
...
-        self.slot_commitment_id.pack(packer)?;
-        self.latest_finalized_slot.pack(packer)?;
-        self.issuer_id.pack(packer)?;
-
-        Ok(())
+fn verify_protocol_version<const VERIFY: bool>(
+    protocol_version: &u8,
+    params: &ProtocolParameters,
+) -> Result<(), Error> {
+    if VERIFY && *protocol_version != params.version() {
```
Alex6323, line 122:
```suggestion
    if VERIFY && protocol_version != &params.version() {
```
my preference

sdk/src/types/block/core/block.rs
thibault-martinez: We mostly dereference in verify functions so that would be inconsistent

Alex6323: hmm, okay

Alex6323 (approved):

PR #1733, opened 2023-12-04
Add /api/core/v3/outputs/{outputId}/full route

### LINKED ISSUE #1688: Add `/api/core/v3/outputs/{outputId}/full` route

Thoralf-M:
# Description of change

Add /api/core/v3/outputs/{outputId}/full route

Closes https://github.com/iotaledger/iota-sdk/issues/1688

## Type of change

Running the added example


sdk/src/types/api/core.rs
```diff
@@ -525,3 +525,58 @@ pub struct UtxoChangesResponse {
...
+pub struct OutputResponse {
+    pub(crate) output: Output,
+    pub(crate) output_id_proof: OutputIdProof,
+}
+
+/// The proof of the output identifier.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OutputIdProof {
+    pub(crate) output: Output,
```
thibault-martinez: The output is again in the proof ?

Thoralf-M: No, don't review drafts :trollface:

sdk/src/types/api/core.rs
```diff
@@ -525,3 +526,84 @@ pub struct UtxoChangesResponse {
...
+        )
+    }
+}
+
+/// The proof of the output commitment.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct HashableNode {
+    #[serde(rename = "type")]
+    pub kind: u8,
+    pub l: Box<OutputCommitmentProof>,
```
DaughterOfMars: r should be optional, right?

Thoralf-M: Why optional? Would be just a LeafHash then, no?

DaughterOfMars: well one of these can be empty, iirc

Thoralf-M: Don't see in which case that would be and also don't see anything optional here
https://github.com/iotaledger/iota.go/blob/2934050abf37c9ece6a35a9c5c96f87c648164b9/merklehasher/merkle_proof.go#L36

Alex6323: I think Chloé is referring to the Rust POI impl in Chronicle, but that is a bit different. I think the reference implementation changed, since the test vector hash in the TIP is now different than it was back then despite being the same block ids.

Alex6323 (dismissed):

sdk/src/types/api/core.rs
DaughterOfMars: still, is it not possible to have only one child?

Alex6323: you can build a merkle tree over a single element...it used to be prefixed with `l` (left) though, now it's `0`:
https://github.com/iotaledger/iota.go/blob/2934050abf37c9ece6a35a9c5c96f87c648164b9/merklehasher/merkle_hasher.go

which explains why the test vector hash changed.

DaughterOfMars: I see!

DaughterOfMars (dismissed):

sdk/src/types/api/core.rs
```diff
@@ -528,3 +529,84 @@ pub struct UtxoChangesResponse {
     pub created_outputs: Vec<OutputId>,
     pub consumed_outputs: Vec<OutputId>,
 }
+
+/// Contains the generic [`Output`] with associated [`OutputIdProof`].
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OutputResponse {
```
thibault-martinez: What's that for ?

sdk/src/wallet/types/mod.rs
```diff
@@ -43,6 +43,8 @@ pub struct OutputData {
     pub metadata: OutputMetadata,
     /// The actual Output
     pub output: Output,
+    /// The OutputIdProof
```
thibault-martinez:
```suggestion
    /// The output ID proof
```

sdk/src/types/api/core.rs
Thoralf-M: That's for the normal outputs route, but not used yet

DaughterOfMars (dismissed):

sdk/src/types/api/core.rs
```diff
@@ -528,3 +529,85 @@ pub struct UtxoChangesResponse {
...
+                    serde_json::from_value::<ValueHash>(value)
+                        .map_err(|e| serde::de::Error::custom(format!("cannot deserialize value hash: {e}")))?,
+                ),
+                _ => return Err(serde::de::Error::custom("invalid output commitment proof")),
+            },
+        )
+    }
+}
+
+/// The proof of the output commitment.
```
thibault-martinez: :trollface:

sdk/src/wallet/types/mod.rs
```diff
@@ -43,6 +43,8 @@ pub struct OutputData {
     pub metadata: OutputMetadata,
     /// The actual Output
     pub output: Output,
+    /// The output ID proof
+    pub output_id_proof: OutputIdProof,
```
thibault-martinez, line 46: Do we actually need it here?

thibault-martinez (changes requested):
Common Feature Sets failing

sdk/src/wallet/types/mod.rs
Thoralf-M: Not sure, might be useful before signing so one can verify again that the inputs are actually what one thinks

Thoralf-M: Also thought that we could just have `OutputWithMetadata` here instead of the fields separated

Thoralf-M:
> Common Feature Sets failing

There is a small problem that asks for changes in 55 files :trollface: 
https://github.com/Thoralf-M/iota-sdk/compare/outputs-full-route...Thoralf-M:iota-sdk:small-change
In this PR here I only modified OutputMetadata, but we also had a OutputMetadataResponse which is now redundant
And then is the question if we want to store the OutputIdProof everywhere or if we say we don't need it

sdk/src/types/block/output/mod.rs
```diff
@@ -75,17 +78,27 @@ pub enum OutputBuilderAmount {
...
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(rename_all = "camelCase")
+)]
 pub struct OutputWithMetadata {
-    pub(crate) output: Output,
-    pub(crate) metadata: OutputMetadata,
+    pub output: Output,
+    pub output_id_proof: OutputIdProof,
+    pub metadata: OutputMetadata,
```
Alex6323, line 90: `OutputWithProofAndMetadata`? :trollface:

Alex6323: out of interest: why couldn't we put the proof inside of the metadata from a TIP perspective, or ... if we wanted to keep it separate have a distinct API call to query the proof?

sdk/src/wallet/types/mod.rs
Alex6323: can you remined me? Why do we store the `output_id` in the `OutputData` if it's already in the also contained metadata?

sdk/src/types/block/output/mod.rs
Thoralf-M: One could have also structured it in a different way, but it made more sense this way because the output and the proof is static data and the metadata is dynamic, also the proof can only be used together with the output, having it in the metadata without the output itself doesn't help (the proof is also returned in the route with the output alone)

Alex6323 (dismissed):

sdk/src/types/block/output/mod.rs
Alex6323: I see.

sdk/src/wallet/types/mod.rs
Thoralf-M: It was added so it's available without potential error from calculating it (because in the old metadata it needed to be created from tx id and output index which could fail)
Now if we trust the metadata we could remove it I guess, but if we compute it from the output with proof, then this could still fail 
We could maybe have a builder for the output data which compares the output id from the proof with the one from the metadata :thinking:

Thoralf-M: Opened an issue https://github.com/iotaledger/iota-sdk/issues/1759

thibault-martinez:
> > Common Feature Sets failing
> 
> There is a small problem that asks for changes in 55 files :trollface: [Thoralf-M/iota-sdk@outputs-full-route...Thoralf-M:iota-sdk:small-change](https://github.com/Thoralf-M/iota-sdk/compare/outputs-full-route...Thoralf-M:iota-sdk:small-change) In this PR here I only modified OutputMetadata, but we also had a OutputMetadataResponse which is now redundant And then is the question if we want to store the OutputIdProof everywhere or if we say we don't need it

I don't understand your answer. This is unrelated to the CI failing?

Thoralf-M:
@thibault-martinez CI is failing because I imported an API type to types module which is not no_std compatible, so I moved the OutputWithMetadata out of the types and replaced the OutputWithMetadataResponse
And because OutputWithMetadata is used all over the place, it all needed to be updated

thibault-martinez (dismissed):

Alex6323 (approved):

thibault-martinez (approved):

PR #1736, opened 2023-12-04
Update mana tests with fixed iota.go data

DaughterOfMars:
# Description of change

iota.go was fixed so these tests work now

### DIFF
--- a/sdk/src/types/block/mana/parameters.rs
+++ b/sdk/src/types/block/mana/parameters.rs
@@ -351,30 +351,29 @@ mod test {
         );
     }
 
-    // TODO: Uncomment when iota.go fixes their calculation + test data
-    // #[test]
-    // fn test_potential_mana_lookup_len_delta() {
-    //     assert_eq!(
-    //         params().potential_mana(
-    //             i64::MAX as u64,
-    //             params().first_slot_of(1),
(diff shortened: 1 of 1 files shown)


sdk/src/types/block/mana/parameters.rs
```diff
@@ -351,30 +351,29 @@ mod test {
...
-    //         params().potential_mana(
-    //             i64::MAX as u64,
-    //             params().first_slot_of(1),
-    //             params().first_slot_of(3 * params().mana_parameters().decay_factors().len() as u32 + 1)
-    //         ),
-    //         Ok(410192222442040018)
-    //     );
-    // }
+    #[test]
+    fn test_potential_mana_lookup_len_delta() {
```
thibault-martinez: Can we remove the test_ prefix ? We usually don't do that, I personally find it a bit redundant

Thoralf-M (dismissed):

sdk/src/types/block/mana/parameters.rs
DaughterOfMars: sure

thibault-martinez (approved):

Thoralf-M (approved):

PR #1742, opened 2023-12-05
Add capabilities to TransactionOptions

thibault-martinez:


Thoralf-M (dismissed):

sdk/src/wallet/operations/transaction/account.rs
```diff
@@ -102,12 +103,17 @@ where
         // TODO https://github.com/iotaledger/iota-sdk/issues/1740
         let issuance = self.client().get_issuance().await?;
 
+        // TODO this needs to be removed when Mana is properly handled
```
Alex6323: issue so we don't forget?

DaughterOfMars (commented):
are the capabilities actually used anywhere?

thibault-martinez:
> are the capabilities actually used anywhere?

Apart from the temporary use, not right now no. But we still need support for it.

sdk/src/wallet/operations/transaction/account.rs
thibault-martinez: Done

DaughterOfMars (approved):

Thoralf-M (approved):

PR #1756, opened 2023-12-06
Split Secret Manager traits

### LINKED ISSUE #1490: Reconsider what should be part of SecretManager or not
Should this be part of the Trait? Seems kind of weird. Maybe we should just remove both.

_Originally posted by @DaughterOfMars in https://github.com/iotaledger/iota-sdk/pull/1483#discussion_r1368646716_

DaughterOfMars:
# Description of change

This PR splits the `SecretManage` trait and refactors the trait to unify the essential functions for a wallet.

Closes #1490


bindings/core/src/method/secret_manager.rs
```diff
@@ -21,11 +19,13 @@ use crate::OmittedDebug;
 pub enum SecretManagerMethod {
     /// Generate Ed25519 addresses.
     GenerateEd25519Addresses {
```
Thoralf-M, line 22:
```suggestion
    #[serde(rename_all = "camelCase")]
    GenerateEd25519Addresses {
```

bindings/core/src/method_handler/secret_manager.rs
```diff
@@ -1,63 +1,45 @@
...
     },
 };
 
 use crate::{method::SecretManagerMethod, response::Response, Result};
 
 /// Call a secret manager method.
-pub(crate) async fn call_secret_manager_method_internal<S: SecretManage + DowncastSecretManager>(
-    secret_manager: &S,
+pub(crate) async fn call_secret_manager_method_internal(
+    secret_manager: &SecretManager,
```
Thoralf-M, line 26: Why is this not the trait anymore?

thibault-martinez (changes requested):
Please fix conflicts

bindings/core/src/method_handler/secret_manager.rs
DaughterOfMars: Since we have to store it, it has to be the enum anyway so I figured we should just use it here. Still, we could change it if it is necessary to support custom SMs

Thoralf-M: We don't need to store it? But yes we need deserialize unless we change something to support custom SMs :+1:

sdk/src/wallet/operations/syncing/transactions.rs
```diff
@@ -202,14 +198,15 @@ where
...
 
-        for mut transaction in transactions_to_reissue {
-            log::debug!("[SYNC] reissue transaction");
-            let reissued_block = self
-                .submit_signed_transaction(transaction.payload.clone(), None)
-                .await?;
-            transaction.block_id.replace(reissued_block);
-            updated_transactions.push(transaction);
-        }
+        // TODO: this is a problem
```
Thoralf-M: I guess we shouldn't do this automatically for now (or never), since it also requires more mana which doesn't get allotted if we just submit the same payload

bindings/core/src/method_handler/secret_manager.rs
DaughterOfMars: Sorry, I meant we need to define the possible variants in the bindings.

sdk/src/wallet/operations/syncing/transactions.rs
DaughterOfMars: Should I just remove all the related code then?

Thoralf-M: I think we should do it in an own PR, because it affects more than what you commented

bindings/core/src/method_handler/secret_manager.rs
Thoralf-M: Worth to keep the trait then for this potential future? If not feel free to keep as it is

bindings/core/src/method/secret_manager.rs
```diff
@@ -36,40 +37,40 @@ pub enum SecretManagerMethod {
     SignatureUnlock {
         /// Transaction signing hash
         transaction_signing_hash: String,
-        /// Chain used to sign the hash
-        #[serde(with = "Bip44Def")]
-        chain: Bip44,
+        /// Options used to sign the hash
+        signing_options: serde_json::Value,
```
Thoralf-M, line 42: Should all the signing options be optional? For something like a PrivateKeySecretManager that can't have multiple keys one otherwise needs to provide something empty just so this field is not missing?

DaughterOfMars: I'm going to make them default to `Value::Null`, which will deserialize to `()` correctly (I double checked).

sdk/examples/wallet/offline_signing/1_prepare_transaction.rs
```diff
@@ -41,11 +42,11 @@ async fn main() -> Result<()> {
     let client_options = ClientOptions::new().with_node(&std::env::var("NODE_URL").unwrap())?;
 
     // Create the wallet with the secret_manager and client options
-    let wallet = Wallet::builder()
+    let wallet = WalletBuilder::new()
         .with_secret_manager(SecretManager::Placeholder)
         .with_storage_path(ONLINE_WALLET_DB_PATH)
         .with_client_options(client_options.clone())
-        .with_bip_path(Bip44::new(SHIMMER_COIN_TYPE))
+        // .with_public_key_options(())
```
Thoralf-M: Can be removed?

sdk/examples/wallet/offline_signing/1_prepare_transaction.rs
```diff
@@ -41,11 +42,11 @@ async fn main() -> Result<()> {
     let client_options = ClientOptions::new().with_node(&std::env::var("NODE_URL").unwrap())?;
 
     // Create the wallet with the secret_manager and client options
-    let wallet = Wallet::builder()
+    let wallet = WalletBuilder::new()
         .with_secret_manager(SecretManager::Placeholder)
```
Thoralf-M: Is this still needed if we don't need to sign?

DaughterOfMars: don't think so

sdk/examples/wallet/offline_signing/1_prepare_transaction.rs
DaughterOfMars: removed

sdk/src/client/secret/mnemonic.rs
```diff
@@ -147,6 +206,26 @@ impl MnemonicSecretManager {
         let seed = Seed::from_bytes(bytes.as_ref());
         Ok(Self(seed))
     }
+
+    /// Generate a random mnemonic to use for the secret manager.
+    pub fn generate_random() -> Result<Self, Error> {
```
Thoralf-M: This might be nice for testing, but looks dangerous to have public, because one has no access to the mnemonic/seed, so when it's dropped all control over potential generated addresses is gone

sdk/src/wallet/core/operations/stronghold_backup/mod.rs
```diff
@@ -98,212 +96,65 @@ impl Wallet {
...
             } else {
                 false
             }
         });
 
         if !ignore_backup_values {
-            self.data_mut().await.bip_path = read_bip_path;
+            if let Some(opts) = loaded_pub_key_opts {
+                // TODO
+                // self.secret_data.public_key_options = opts.clone();
```
Thoralf-M: Want to handle in this PR or open an issue?

sdk/src/client/secret/mnemonic.rs
DaughterOfMars: Isn't it already publicly available?

sdk/src/wallet/core/operations/stronghold_backup/mod.rs
DaughterOfMars: well the problem is that this field is not mutable, and I'm not sure whether it should be

sdk/src/client/secret/mnemonic.rs
Thoralf-M: I only see it added in this PR

DaughterOfMars: I moved it to the tests common module

sdk/src/client/secret/manager.rs
```diff
@@ -0,0 +1,469 @@
...
+                    <LedgerSecretManager as Generate<Vec<ed25519::PublicKey>>>::Options::deserialize(options)?;
+                Ok(l.generate(&options).await?)
+            }
+            SecretManager::Mnemonic(m) => {
+                let options =
+                    <MnemonicSecretManager as Generate<Vec<ed25519::PublicKey>>>::Options::deserialize(options)?;
+                Ok(m.generate(&options).await?)
+            }
+            #[cfg(feature = "private_key_secret_manager")]
+            SecretManager::PrivateKey(_) => todo!(),
```
Alex6323, line 121: TODO

sdk/src/client/secret/manager.rs
```diff
@@ -0,0 +1,469 @@
...
+            }
+            #[cfg(feature = "ledger_nano")]
+            SecretManager::LedgerNano(_) => Err(ledger_nano::Error::UnsupportedOperation.into()),
+            SecretManager::Mnemonic(m) => {
+                let options =
+                    <MnemonicSecretManager as Generate<secp256k1_ecdsa::PublicKey>>::Options::deserialize(options)?;
+                Ok(m.generate(&options).await?)
+            }
+            #[cfg(feature = "private_key_secret_manager")]
+            SecretManager::PrivateKey(_) => todo!(),
```
Alex6323, line 160: TODO

sdk/src/client/secret/manager.rs
```diff
@@ -0,0 +1,469 @@
...
+            SecretManager::LedgerNano(_) => Err(ledger_nano::Error::UnsupportedOperation.into()),
+            SecretManager::Mnemonic(m) => {
+                let options =
+                    <MnemonicSecretManager as Generate<Vec<secp256k1_ecdsa::PublicKey>>>::Options::deserialize(
+                        options,
+                    )?;
+                Ok(m.generate(&options).await?)
+            }
+            #[cfg(feature = "private_key_secret_manager")]
+            SecretManager::PrivateKey(_) => todo!(),
```
Alex6323, line 200: TODO

sdk/src/wallet/core/operations/stronghold_backup/mod.rs
Thoralf-M: It needs to be, unless we change the restore backup in such a way that it creates a new wallet instead of updating the existing one (not sure if that makes sense, but definitely no in this PR)

sdk/src/client/secret/mod.rs
```diff
@@ -66,613 +57,504 @@ use crate::{
     },
 };
 
-/// The secret manager interface.
-#[async_trait]
-pub trait SecretManage: Send + Sync {
-    type Error: std::error::Error + Send + Sync;
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct PublicKeyOptions {
```
Thoralf-M, line 62: What's the reason again to have this instead of Bip44 directly?

sdk/src/client/secret/mod.rs
```diff
@@ -66,613 +57,504 @@ use crate::{
...
+    pub fn with_address_index(mut self, address_index: u32) -> Self {
+        self.address_index = address_index;
+        self
     }
 }
 
-impl From<MnemonicSecretManager> for SecretManager {
-    fn from(secret_manager: MnemonicSecretManager) -> Self {
-        Self::Mnemonic(secret_manager)
+impl From<Bip44> for PublicKeyOptions {
```
Thoralf-M, line 99: Maybe also add a `.to_bip44()` method?

sdk/src/client/secret/mod.rs
Alex6323: `Bip44` is not our type, but we could also make it a `PublicKeyOptions(Bip44)`?

sdk/src/wallet/core/operations/stronghold_backup/mod.rs
DaughterOfMars: tbh that may be better than dealing with even more lock contention

sdk/src/client/secret/mod.rs
DaughterOfMars: I've thought about changing it, but of course then there's the mismatch between generating a single key and multiple

sdk/src/client/secret/mod.rs
DaughterOfMars: sure

Thoralf-M: Or impl Into, not sure what's better

sdk/src/client/secret/manager.rs
DaughterOfMars: So all of these are here because I'm not sure what we want to do if you try to generate multiple keys using a private key secret manager (which can only ever give you one key). Should we give that key, even if it doesn't match the provided config? Or return an error?

sdk/src/client/secret/mod.rs
Thoralf-M: Okay, but should we maybe name it `Bip44PublicKeyOptions` to make its purpose clearer?

sdk/src/wallet/core/operations/stronghold_backup/mod.rs
Thoralf-M: I did it initially as part of the wallet builder and then changed in this commit https://github.com/iotaledger/wallet.rs/pull/1019/commits/037e5c3241a7e267c1e84d317a6681eef439a908, so the restore method can be called on an existing wallet instead of having a different builder (if I remember correctly, Firefly team wanted this change, because it felt a bit weird to them to have the method not available on the wallet, but I think it's not that bad)

sdk/src/client/secret/mod.rs
```diff
@@ -66,613 +57,504 @@ use crate::{
...
+                Address::ImplicitAccountCreation(implicit) => Address::Ed25519(*implicit.ed25519_address()),
+                Address::Restricted(restricted) => restricted.address().clone(),
+                _ => required_address,
+            };
+
+            // Check if we already added an [Unlock] for this address
+            match block_indexes.get(&required_address) {
+                // If we already have an [Unlock] for this address, add a [Unlock] based on the address type
+                Some(block_index) => match required_address {
+                    Address::Ed25519(_) | Address::ImplicitAccountCreation(_) => {
```
Alex6323, line 245: Can this actually still be `Address::ImplicitAccountCreation`? It was converted in L227, right?

sdk/src/client/secret/mod.rs
```diff
@@ -66,613 +57,504 @@ use crate::{
...
+    }
+
+    async fn sign<S>(&self, msg: &[u8], options: &Self::Options) -> crate::client::Result<S>
+    where
+        Self: Sign<S>,
+    {
+        Sign::<S>::sign(self, msg, options).await
     }
 }
+impl<T> SecretManageExt for T {}
```
Alex6323, line 357: `T: SecretManage` bound?

sdk/src/client/secret/mod.rs
DaughterOfMars: Hmm..I'll think about it.

sdk/src/wallet/operations/transaction/sign_transaction.rs
```diff
@@ -40,46 +27,48 @@ where
...
+        //                 self.emit(WalletEvent::TransactionProgress(
+        //                     TransactionProgressEvent::PreparedTransaction(Box::new(PreparedTransactionDataDto::from(
+        //                         prepared_transaction_data,
+        //                     ))),
+        //                 ))
+        //                 .await;
+        //             }
+        //         }
+        //     }
+        // }
```
Thoralf-M: What about this?

sdk/src/client/secret/ledger_nano.rs
```diff
@@ -309,19 +375,13 @@ impl SecretManage for LedgerSecretManager {
...
-                        )
-                    } else {
-                        (None, LedgerBIP32Index::default())
-                    }
-                }
+                [remainder] => (
+                    Some(&remainder.output),
+                    LedgerBIP32Index {
+                        bip32_change: chain.change.harden().into(),
+                        bip32_index: chain.address_index.harden().into(),
```
Thoralf-M, line 382: Before this was part of the remainder, but now it's just provided to this function which I think can be wrong if the remainder address isn't an Ed25519 address
So if remainder address != Ed25519, return None for the remainder_bip32 and below only go in this part `if let Some(remainder_output) = remainder_output {` if it's Some

sdk/src/client/secret/types.rs
```diff
@@ -141,9 +120,6 @@ pub struct InputSigningData {
     pub output: Output,
     /// The output metadata
     pub output_metadata: OutputMetadata,
-    /// The chain derived from seed, only for ed25519 addresses
-    #[serde(with = "option_bip44", default)]
-    pub chain: Option<Bip44>,
```
Thoralf-M, line 146: So do we bring this back for signing with multiple different addresses in inputs?

sdk/src/client/secret/ledger_nano.rs
DaughterOfMars: Since you understand this better, would you push the change?

sdk/src/client/secret/types.rs
DaughterOfMars: might have to :c

sdk/src/wallet/operations/transaction/sign_transaction.rs
DaughterOfMars: I put it back

sdk/src/client/secret/mod.rs
Thoralf-M: It's not very important, so you can resolve whether you do or not

cli/src/cli.rs
```diff
@@ -416,11 +427,22 @@ pub async fn init_command(
         .map(|addr| Bech32Address::from_str(&addr))
         .transpose()?;
 
-    Ok(Wallet::builder()
+    let public_key_options = init_params.bip_path.map(|bip| {
+        LedgerOptions::new(
+            PublicKeyOptions::new(bip.coin_type)
+                .with_account_index(bip.account)
+                .with_internal(bip.change != 0)
+                .with_address_index(bip.address_index),
```
Thoralf-M, line 435: Can now just be `PublicKeyOptions::from(bip)`, also in more places

sdk/src/client/secret/ledger_nano.rs
Thoralf-M: I think it depends on https://github.com/iotaledger/iota-sdk/pull/1756#discussion_r1484276282, if we bring this back, then we can also do this part different

sdk/src/client/secret/mod.rs
DaughterOfMars: why?

sdk/src/client/secret/ledger_nano.rs
DaughterOfMars: that's fair

cli/src/cli.rs
DaughterOfMars: ah, good point, thanks

thibault-martinez:
Working on a better solution.

PR #1764, opened 2023-12-11
Update MetadataFeature

### LINKED ISSUE #1693: Refactor MetadataFeature
https://github.com/iotaledger/tips/blob/tip38/tips/TIP-0038/tip-0038.md#metadata-feature

Thoralf-M:
# Description of change

Refactor MetadataFeature to be compatible with https://github.com/iotaledger/tips/blob/tip38/tips/TIP-0038/tip-0038.md#metadata-feature

Closes #1693 

## Type of change

Running tests


Thoralf-M:
Please review only sdk/src/types/block/output/feature/metadata.rs for now

Alex6323:
> Please review only sdk/src/types/block/output/feature/metadata.rs for now

Why not keep it a draft then? :thinking:  (You know we can also review that file if it's a draft :trollface: )

thibault-martinez:
> Please review only sdk/src/types/block/output/feature/metadata.rs for now

Looks overall fine to me

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -1,24 +1,56 @@
...
+        BoxedSlicePrefix<u8, MetadataFeatureValueLength>,
+        MetadataFeatureLength,
+    >,
+) -> Result<(), Error> {
+    if VERIFY {
+        for key in map.keys() {
+            if !key.is_ascii() {
+                return Err(Error::NonAsciiMetadataKey(key.to_vec()));
+            }
+        }
```
Alex6323:
```rust
if map.keys().any(|key| !key.is_ascii()) {
    return Err(...)
}
```

Alex6323: a bit more readable, rusty and less nested IMO

Alex6323: Is that really all that needs to be verified? What about the uniqueness of the keys and the minimum number of entries count?

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +59,78 @@ impl MetadataFeature {
...
-    pub fn new(data: impl Into<Vec<u8>>) -> Result<Self, Error> {
-        Self::try_from(data.into())
+    pub fn new(data: impl Into<BTreeMap<Vec<u8>, Vec<u8>>>) -> Result<Self, Error> {
+        let data = data.into();
+
+        for key in data.keys() {
+            if !key.is_ascii() {
+                return Err(Error::NonAsciiMetadataKey(key.to_vec()));
+            }
+        }
```
Alex6323: same as above

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +59,78 @@ impl MetadataFeature {
...
 impl core::fmt::Display for MetadataFeature {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(f, "{}", prefix_hex::encode(self.data()))
+        write!(
+            f,
+            "{:?}",
+            self.0
+                .keys()
+                // Safe to unwrap, keys must be ascii
+                .map(|k| alloc::str::from_utf8(k).unwrap())
```
Alex6323, line 229: use `from_utf8_unchecked`?

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +59,78 @@ impl MetadataFeature {
...
     }
 }
 
 impl core::fmt::Debug for MetadataFeature {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(f, "MetadataFeature({self})")
+        let dbg_map: BTreeMap<&str, String> = self
+            .0
+            .iter()
+            .map(|(k, v)| (alloc::str::from_utf8(k).unwrap(), prefix_hex::encode(v.as_ref())))
```
Alex6323: `from_utf8_unchecked`?

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +59,78 @@ impl MetadataFeature {
...
+                .map(|k| alloc::str::from_utf8(k).unwrap())
+                .collect::<Vec<&str>>()
+        )
     }
 }
 
 impl core::fmt::Debug for MetadataFeature {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(f, "MetadataFeature({self})")
+        let dbg_map: BTreeMap<&str, String> = self
```
Alex6323: bit inconsistent ... here you create a binding while for `Display` you don't.

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -434,35 +469,74 @@ pub(crate) mod irc_30 {
...
+    impl From<&MetadataFeature> for MetadataFeatureDto {
+        fn from(value: &MetadataFeature) -> Self {
+            let mut entries = BTreeMap::new();
+            for (k, v) in value.0.iter() {
+                entries.insert(
+                    // Safe to unwrap, keys must be ascii
+                    alloc::str::from_utf8(k.as_ref()).expect("invalid ascii").to_string(),
+                    prefix_hex::encode(v.as_ref()),
+                );
+            }
```
Alex6323: map collect instead of mut/for/insert?

sdk/src/types/block/output/feature/metadata.rs
thibault-martinez: It was probably done with the `for` to have access to the key

sdk/src/types/block/output/feature/metadata.rs
thibault-martinez: I think I would personally keep the failing one tbh. Doesn't really hurt and at least gives us more guarantee than adding an unsafe block and never knowing about the issue.

sdk/src/types/block/output/feature/metadata.rs
thibault-martinez: Done

sdk/src/types/block/output/feature/metadata.rs
thibault-martinez: Done

sdk/src/types/block/output/feature/metadata.rs
Alex6323: yes, somehow overlooked that :+1:

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -43,7 +43,7 @@ fn verify_keys_packable<const VERIFY: bool>(
 ) -> Result<(), Error> {
     if VERIFY {
         for key in map.keys() {
-            if !key.is_ascii() {
```
Alex6323: :+1: :+1:

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -43,7 +43,7 @@ fn verify_keys_packable<const VERIFY: bool>(
...
+            if !key.iter().all(|b| b.is_ascii_graphic()) {
```
Alex6323: :+1: :+1:

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -1,24 +1,56 @@
...
 
+fn verify_keys_packable<const VERIFY: bool>(
+    map: &BTreeMapPrefix<
+        BoxedSlicePrefix<u8, MetadataFeatureKeyLength>,
+        BoxedSlicePrefix<u8, MetadataFeatureValueLength>,
+        MetadataFeatureLength,
+    >,
+) -> Result<(), Error> {
+    if VERIFY {
+        for key in map.keys() {
```
Alex6323: check (according to TIP) for at least one entry in the map not necessary?

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -434,35 +471,78 @@ pub(crate) mod irc_30 {
...
+    }
+
+    impl<'de> Deserialize<'de> for MetadataFeature {
+        fn deserialize<D: Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+            let value = Value::deserialize(d)?;
+            Ok(
+                match value
+                    .get("type")
+                    .and_then(Value::as_u64)
+                    .ok_or_else(|| de::Error::custom("invalid metadata type"))? as u8
```
Alex6323: if the type is out of u8 bounds for some (malicious or accidental) reason then this method could still panic instead of returning an error and allowing to continue on, so I guess we need to do a bounds check before performing `as u8`  because there's no `Value::as_u8` (in serde_json) that would do it for us?

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Doesn't `LENGTH_RANGE` already ensure that during unpacking? Or do you see a way how it can pass that without a key?

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: `Value::as_u8` doesn't exist, but I used `u8::try_from` now :+1:

sdk/src/types/block/output/feature/metadata.rs
Alex6323: It could pass it with just the key length prefix (but not the actual key)?

Thoralf-M: Don't think so, since the key also is also bounded `MetadataFeatureKeyLength = BoundedU8<1, { u8::MAX }>` and then during unpacking it would fail since it's missing >=1 byte

Alex6323: hmm, I guess in combination of both you're right and this is enough to ensure there's at least 1 entry. Still I wonder if we should make the validation checks a bit more explicit, or at least mention in some comment that all validation rules are met (considering we usually try to be as 1:1 with the TIPs) :shrug:

Thoralf-M: Added

Alex6323 (dismissed):

sdk/examples/how_tos/outputs/features.rs
```diff
@@ -56,12 +56,18 @@ async fn main() -> Result<()> {
         // with metadata feature block
         nft_output_builder
             .clone()
-            .add_feature(MetadataFeature::new("Hello, World!")?)
+            .add_feature(MetadataFeature::new(std::collections::BTreeMap::from_iter(vec![(
```
thibault-martinez: That's annoying, I would change the `new` function to take a `impl IntoIterator<Item = (Vec<u8>, Vec<u8>)>` and collect to a BTreeMap inside

sdk/tests/client/input_selection/nft_outputs.rs
```diff
@@ -1210,67 +1210,68 @@ fn transitioned_zero_nft_id_no_longer_is_zero() {
     });
 }
 
-#[test]
-fn changed_immutable_metadata() {
-    let protocol_parameters = protocol_parameters();
-    let nft_id_1 = NftId::from_str(NFT_ID_1).unwrap();
+// TODO: enable again when MetadataFeature is cleared up
```
thibault-martinez: What needs to be cleared up exactly?

Thoralf-M: MetadataFeature now not anymore, but Irc27Metadata can't be converted the same way as before, so this would need updates now

Alex6323 (dismissed):

sdk/examples/client/output/build_account_output.rs
```diff
@@ -49,9 +49,12 @@ async fn main() -> Result<()> {
     // Account id needs to be null the first time
     let account_output = AccountOutputBuilder::new_with_minimum_amount(storage_score_params, AccountId::null())
         .add_feature(SenderFeature::new(address.clone()))
-        .add_feature(MetadataFeature::new(metadata)?)
+        .add_feature(MetadataFeature::new([(metadata.to_vec(), metadata.to_vec())])?)
         .add_immutable_feature(IssuerFeature::new(address.clone()))
-        .add_immutable_feature(MetadataFeature::new(metadata)?)
+        .add_immutable_feature(MetadataFeature::new(std::collections::BTreeMap::from_iter(vec![(
```
thibault-martinez: BTreeMap can be removed?

cli/src/wallet_cli/mod.rs
```diff
@@ -775,13 +772,23 @@ pub async fn mint_nft_command(
...
-    let nft_options = MintNftParams::new()
+    let mut nft_options = MintNftParams::new()
         .with_address(address)
-        .with_immutable_metadata(immutable_metadata)
-        .with_metadata(metadata)
         .with_tag(tag)
         .with_sender(sender)
         .with_issuer(issuer);
+    if let Some(metadata) = metadata {
+        nft_options = nft_options.with_metadata(MetadataFeature::new(std::collections::BTreeMap::from_iter(vec![(
```
thibault-martinez: remove?

sdk/examples/client/output/build_account_output.rs
Thoralf-M: Yeah missed that, done

cli/src/wallet_cli/mod.rs
```diff
@@ -775,13 +772,20 @@ pub async fn mint_nft_command(
...
-        .with_metadata(metadata)
         .with_tag(tag)
         .with_sender(sender)
         .with_issuer(issuer);
+    if let Some(metadata) = metadata {
+        nft_options = nft_options.with_metadata(MetadataFeature::new([(metadata.clone(), metadata)])?);
+    }
+    if let Some(immutable_metadata) = immutable_metadata {
+        nft_options = nft_options.with_immutable_metadata(MetadataFeature::new(
+            std::collections::BTreeMap::from_iter(vec![(immutable_metadata.clone(), immutable_metadata)]),
```
thibault-martinez: :trollface:

sdk/src/wallet/operations/transaction/high_level/minting/create_native_token.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use alloc::collections::BTreeMap;
```
thibault-martinez: remove?

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +64,105 @@ impl MetadataFeature {
...
+    fn try_from(data: Vec<(Vec<u8>, Vec<u8>)>) -> Result<Self, Error> {
+        let mut res = BTreeMap::<
+            BoxedSlicePrefix<u8, MetadataFeatureKeyLength>,
+            BoxedSlicePrefix<u8, MetadataFeatureValueLength>,
+        >::new();
+        for (k, v) in data {
+            if !k.iter().all(|b| b.is_ascii_graphic()) {
+                return Err(Error::NonAsciiMetadataKey(k.to_vec()));
             }
+            res.insert(
```
Thoralf-M: Should this return an error if there was already an entry?

sdk/tests/client/input_selection/nft_outputs.rs
Thoralf-M: Didn't saw you updated Irc27Metadata already, so enabled it again now

sdk/src/wallet/operations/transaction/high_level/minting/create_native_token.rs
```diff
@@ -170,7 +170,10 @@ where
                     )));
 
                     if let Some(foundry_metadata) = params.foundry_metadata {
-                        foundry_builder = foundry_builder.add_immutable_feature(MetadataFeature::new(foundry_metadata)?)
+                        foundry_builder = foundry_builder.add_immutable_feature(MetadataFeature::new(
+                            // TODO: what hardcoded key or let user provide the full metadata?
+                            [(b"foundry".to_vec(), foundry_metadata)],
```
thibault-martinez: When I asked Max before the holidays he said `data`, but I didn't ask specifically for Foundry metadata, maybe ask him

sdk/src/types/block/rand/output/feature.rs
```diff
@@ -29,8 +35,38 @@ pub fn rand_issuer_feature() -> IssuerFeature {
 
 /// Generates a random [`MetadataFeature`].
 pub fn rand_metadata_feature() -> MetadataFeature {
-    let bytes = rand_bytes(rand_number_range(MetadataFeature::LENGTH_RANGE) as usize);
-    MetadataFeature::new(bytes).unwrap()
+    let mut map = BTreeMap::new();
+    let mut total_size = 0;
+    for _ in 0..10 {
```
thibault-martinez, line 45:
```suggestion

    for _ in 0..10 {
```

sdk/src/types/block/error.rs
```diff
@@ -56,6 +60,7 @@ pub enum Error {
     InvalidAccountIndex(<UnlockIndex as TryFrom<u16>>::Error),
     InvalidAnchorIndex(<UnlockIndex as TryFrom<u16>>::Error),
     InvalidBlockBodyKind(u8),
+    NonAsciiMetadataKey(Vec<u8>),
```
thibault-martinez: Maybe change to `NonGraphicAsciiMetadataKey` since it's been changed

cli/src/wallet_cli/mod.rs
```diff
@@ -775,13 +772,21 @@ pub async fn mint_nft_command(
...
 
-    let nft_options = MintNftParams::new()
+    let mut nft_options = MintNftParams::new()
         .with_address(address)
-        .with_immutable_metadata(immutable_metadata)
-        .with_metadata(metadata)
         .with_tag(tag)
         .with_sender(sender)
         .with_issuer(issuer);
+    if let Some(metadata) = metadata {
```
thibault-martinez, line 798:
```suggestion

    if let Some(metadata) = metadata {
```

cli/src/wallet_cli/mod.rs
```diff
@@ -775,13 +772,22 @@ pub async fn mint_nft_command(
...
+    let mut nft_options = MintNftParams::new()
         .with_address(address)
-        .with_immutable_metadata(immutable_metadata)
-        .with_metadata(metadata)
         .with_tag(tag)
         .with_sender(sender)
         .with_issuer(issuer);
+
+    if let Some(metadata) = metadata {
+        nft_options = nft_options.with_metadata(MetadataFeature::new([(metadata.clone(), metadata)])?);
```
thibault-martinez: Same key and value?

cli/src/wallet_cli/mod.rs
```diff
@@ -775,13 +772,22 @@ pub async fn mint_nft_command(
...
+    }
+    if let Some(immutable_metadata) = immutable_metadata {
+        nft_options = nft_options.with_immutable_metadata(MetadataFeature::new([(
+            immutable_metadata.clone(),
```
Alex6323: if it needs to be changed (to some hardcoded key?) then probably here as well

cli/src/wallet_cli/mod.rs
Thoralf-M: Well no, how should a user provide it? Do we want to support entries or only a single one?

sdk/src/wallet/operations/transaction/high_level/minting/create_native_token.rs
Thoralf-M: Nothing defined yet, I just changed it so the user can define it here

cli/src/wallet_cli/mod.rs
Thoralf-M: We could do this for a single entry, but value would then only be used if key is provided
```Rust
    immutable_metadata_key: Option<String>,
    immutable_metadata_value: Option<Vec<u8>>,
    metadata_key: Option<String>,
    metadata_value: Option<Vec<u8>>,
```

sdk/examples/client/output/build_account_output.rs
```diff
@@ -49,9 +49,9 @@ async fn main() -> Result<()> {
     // Account id needs to be null the first time
     let account_output = AccountOutputBuilder::new_with_minimum_amount(storage_score_params, AccountId::null())
         .add_feature(SenderFeature::new(address.clone()))
-        .add_feature(MetadataFeature::new(metadata)?)
+        .add_feature(MetadataFeature::new([(metadata.to_vec(), metadata.to_vec())])?)
```
DaughterOfMars: this is kinda wierd especially in an example

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +64,105 @@ impl MetadataFeature {
 
     /// Creates a new [`MetadataFeature`].
     #[inline(always)]
-    pub fn new(data: impl Into<Vec<u8>>) -> Result<Self, Error> {
-        Self::try_from(data.into())
+    pub fn new(data: impl IntoIterator<Item = (Vec<u8>, Vec<u8>)>) -> Result<Self, Error> {
```
DaughterOfMars: Maybe we should just wrap the inner map methods instead and have `new` create an empty map

sdk/src/types/block/output/feature/metadata.rs
DaughterOfMars: I would say so

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: > Maybe we should just wrap the inner map methods instead

What do you mean by that?

cli/src/wallet_cli/mod.rs
thibault-martinez: In the CLI I would support a single entry (at least for now), and I would even use a default key

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -1,24 +1,61 @@
...
 
+fn verify_keys_packable<const VERIFY: bool>(
+    map: &BTreeMapPrefix<
+        BoxedSlicePrefix<u8, MetadataFeatureKeyLength>,
+        BoxedSlicePrefix<u8, MetadataFeatureValueLength>,
+        MetadataFeatureLength,
+    >,
+) -> Result<(), Error> {
+    if VERIFY {
+        if map.is_empty() {
```
thibault-martinez: This is already covered by `MetadataFeatureLength`?

cli/src/wallet_cli/mod.rs
Thoralf-M: Default means still have the option for the user to set it? And default should be `data`?

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Yes https://github.com/iotaledger/iota-sdk/pull/1764#discussion_r1444338286 do you want it removed? @Alex6323 :trollface:

thibault-martinez: Seems useless to me, if this fails then we have an issue in BTreeMapPrefix

Alex6323: well, didn't I agree eventually with you that it's not needed to make a code change? You can revert :+1:  @Thoralf-M

Alex6323 (dismissed):

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Done

cli/src/wallet_cli/mod.rs
Thoralf-M: Added something

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -1,24 +1,56 @@
...
+    prefix::{BTreeMapPrefix, BoxedSlicePrefix},
+};
 
 use crate::types::block::{output::StorageScore, protocol::WorkScore, Error};
 
 pub(crate) type MetadataFeatureLength =
     BoundedU16<{ *MetadataFeature::LENGTH_RANGE.start() }, { *MetadataFeature::LENGTH_RANGE.end() }>;
 
+pub(crate) type MetadataFeatureKeyLength = BoundedU8<1, { u8::MAX }>;
+pub(crate) type MetadataFeatureValueLength = BoundedU16<0, { u16::MAX }>;
```
thibault-martinez, line 26: Shouldn't this use LENGTH_RANGE ?

thibault-martinez: Or is LENGTH_RANGE for the complete serialised thing? Not specific limits for key/value in the TIP?

cli/src/wallet_cli/mod.rs
```diff
@@ -145,12 +142,18 @@ pub enum WalletCommand {
     MintNft {
         /// Address to send the NFT to, e.g. rms1qztwng6cty8cfm42nzvq099ev7udhrnk0rw8jt8vttf9kpqnxhpsx869vr3.
         address: Option<Bech32Address>,
+        /// Immutable metadata key, e.g. --immutable-metadata-key data.
+        #[arg(long, group = "immutable_metadata", default_value = "data")]
```
thibault-martinez: I don't think it should be part of the group? Group means only one of them I think

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +59,110 @@ impl MetadataFeature {
 
     /// Creates a new [`MetadataFeature`].
     #[inline(always)]
-    pub fn new(data: impl Into<Vec<u8>>) -> Result<Self, Error> {
-        Self::try_from(data.into())
+    pub fn new(data: impl IntoIterator<Item = (Vec<u8>, Vec<u8>)>) -> Result<Self, Error> {
+        let data: BTreeMap<Vec<u8>, Vec<u8>> = data.into_iter().collect();
+
+        for key in data.keys() {
```
thibault-martinez: I'd construct it first and then call `verify_keys_packable` to avoid the code duplication

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -205,15 +270,17 @@ pub(crate) mod irc_27 {
...
         }
     }
 
     impl TryFrom<Irc27Metadata> for MetadataFeature {
         type Error = Error;
+
         fn try_from(value: Irc27Metadata) -> Result<Self, Error> {
-            Self::new(value.to_bytes())
+            // TODO: is this hardcoded key correct or should users provide it?
+            Self::new([("irc-27".as_bytes().to_vec(), value.to_bytes())])
```
thibault-martinez: At least for the IRC ones, this is what Max told me to do before the holidays

Thoralf-M: Want to have the TODO removed? Otherwise I would keep it until we have something official to reference

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: > Not specific limits for key/value in the TIP?

No, just the overall size is limited. We can also limit values by that (or even less because of the required  type, entries count and key bytes), but doesn't change much?

cli/src/wallet_cli/mod.rs
Thoralf-M: A group can also be configured, but yeah easier to just not have it in there, removed it

sdk/src/types/block/output/feature/metadata.rs
thibault-martinez: Nah it's fine

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +59,107 @@ impl MetadataFeature {
 
     /// Creates a new [`MetadataFeature`].
     #[inline(always)]
-    pub fn new(data: impl Into<Vec<u8>>) -> Result<Self, Error> {
-        Self::try_from(data.into())
+    pub fn new(data: impl IntoIterator<Item = (Vec<u8>, Vec<u8>)>) -> Result<Self, Error> {
+        let data: BTreeMap<Vec<u8>, Vec<u8>> = data.into_iter().collect();
+
```
thibault-martinez, line 83:
```suggestion
```

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +59,107 @@ impl MetadataFeature {
...
+        let metadata = Self::try_from(data)?;
+        verify_keys_packable::<true>(&metadata.0)?;
+        Ok(metadata)
```
thibault-martinez:
```suggestion
    Ok(metadata)
```

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +59,108 @@ impl MetadataFeature {
...
-macro_rules! impl_from_vec {
-    ($type:ty) => {
-        impl TryFrom<$type> for MetadataFeature {
-            type Error = Error;
+impl TryFrom<Vec<(Vec<u8>, Vec<u8>)>> for MetadataFeature {
+    type Error = Error;
 
-            fn try_from(value: $type) -> Result<Self, Self::Error> {
-                Vec::<u8>::from(value).try_into()
+    fn try_from(data: Vec<(Vec<u8>, Vec<u8>)>) -> Result<Self, Error> {
```
thibault-martinez: Maybe there is a way to use the other try_from inside to avoid code duplication

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +59,108 @@ impl MetadataFeature {
...
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(f, "{}", prefix_hex::encode(self.data()))
+        write!(
+            f,
+            "{:?}",
+            self.0
+                .keys()
+                // Safe to unwrap, keys must be ascii
+                .map(|k| alloc::str::from_utf8(k).unwrap())
+                .collect::<Vec<&str>>()
```
thibault-martinez, line 230: Only showing keys ?

thibault-martinez (dismissed):

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: They're at least human readable, values would be hex encoded, not sure if we want them or not

thibault-martinez: Hmm, I don't have an opinion, happy to keep it like that

Alex6323 (dismissed):

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Just moved it to a new function that accepts IntoIterator

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +59,97 @@ impl MetadataFeature {
...
-    type Err = Error;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        Self::new(prefix_hex::decode::<Vec<u8>>(s).map_err(Error::Hex)?)
+fn metadata_feature_from_iter(data: impl IntoIterator<Item = (Vec<u8>, Vec<u8>)>) -> Result<MetadataFeature, Error> {
+    let mut res = BTreeMap::<
+        BoxedSlicePrefix<u8, MetadataFeatureKeyLength>,
+        BoxedSlicePrefix<u8, MetadataFeatureValueLength>,
+    >::new();
+    for (k, v) in data {
```
thibault-martinez:
```suggestion
    
    for (k, v) in data {
```

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -27,77 +59,98 @@ impl MetadataFeature {
...
+                v.into_boxed_slice()
+                    .try_into()
+                    .map_err(Error::InvalidMetadataFeatureValueLength)?,
+            )
+            .is_some()
+        {
+            return Err(Error::InvalidMetadataFeature("Duplicated metadata key".to_string()));
+        };
     }
+    Ok(MetadataFeature(
```
thibault-martinez:
```suggestion
    
    Ok(MetadataFeature(
```

thibault-martinez (dismissed):

cli/src/wallet_cli/mod.rs
```diff
@@ -145,12 +142,18 @@ pub enum WalletCommand {
     MintNft {
         /// Address to send the NFT to, e.g. rms1qztwng6cty8cfm42nzvq099ev7udhrnk0rw8jt8vttf9kpqnxhpsx869vr3.
         address: Option<Bech32Address>,
+        /// Immutable metadata key, e.g. --immutable-metadata-key data.
+        #[arg(long, default_value = "data")]
```
DaughterOfMars, line 149: do we need group tags?

thibault-martinez: No, the data group is because it can come from the CLI or a file, the key only comes from the CLI

sdk/src/types/block/output/feature/metadata.rs
DaughterOfMars: I meant provide an API like this:

```rust
impl MetadataFeature {
  pub fn new() -> Self {
    Self(Default::default())
  }

  pub fn with_value(mut self, key: &str, value: impl Into<Vec<u8>>) -> Result<Self, Error> {
    if self.0
        .deref_mut()
        .insert(
            key.as_bytes().to_vec().into_boxed_slice()
                .try_into()
                .map_err(Error::InvalidMetadataFeatureKeyLength)?,
            value.into().into_boxed_slice()
                .try_into()
                .map_err(Error::InvalidMetadataFeatureValueLength)?,
        )
        .is_some()
    {
        return Err(Error::InvalidMetadataFeature("Duplicated metadata key".to_string()));
    }
    Ok(self)
  }
}

...

MetadataFeature::new().with_value("data1", metadata1)?.with_value("data2", metadata2)?
```

cli/src/wallet_cli/mod.rs
```diff
@@ -1243,7 +1257,10 @@ pub async fn prompt_internal(
                                 wallet,
                                 circulating_supply,
                                 maximum_supply,
-                                bytes_from_hex_or_file(foundry_metadata_hex, foundry_metadata_file).await?,
+                                bytes_from_hex_or_file(foundry_metadata_hex, foundry_metadata_file)
+                                    .await?
+                                    // TODO: Let user specify key or the full metadata
```
kwek20: Todo, issue?

Thoralf-M: Allowed to set it instead

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: To me this looks more complicated to use, what are your opinions @Alex6323 @kwek20 @thibault-martinez ?

thibault-martinez: I don't really mind, I guess we'll get more experience with it as we use it and we'll see what makes sense in terms of API. I'm not sure why we'd need to construct an empty map if we already have the values, but a function to add values (if we don't have one) is a good idea.

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -82,6 +103,76 @@ impl StorageScore for MetadataFeature {}
...
+            BoxedSlicePrefix<u8, MetadataFeatureKeyLength>,
+            BoxedSlicePrefix<u8, MetadataFeatureValueLength>,
+            MetadataFeatureLength,
+        > = map
+            .try_into()
+            .map_err(|_| Error::InvalidMetadataFeature("invalid metadata feature".to_string()))
+            .map_err(UnpackError::Packable)?;
+
+        verify_keys_packable::<true>(&r).map_err(UnpackError::Packable)?;
+        verify_length_packable::<true>(&r).map_err(UnpackError::Packable)?;
```
DaughterOfMars: Can't this be done before creating the structure, using the bytes?

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -82,6 +103,76 @@ impl StorageScore for MetadataFeature {}
...
+        Ok(())
+    }
+
+    fn unpack<U: Unpacker, const VERIFY: bool>(
+        unpacker: &mut U,
+        visitor: &Self::UnpackVisitor,
+    ) -> Result<Self, UnpackError<Self::UnpackError, U::Error>> {
+        let len = u16::unpack::<_, VERIFY>(unpacker, &()).coerce()?;
+
+        let mut map = BTreeMap::<
```
thibault-martinez: Why do we need all that stuff ? We should be able to still use packable on the inner BTreeMapPrefix

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Just tried and deref_mut() is not implemented for the type, so we don't have a way to do this?

DaughterOfMars: that seems like a big problem for the usability of the type, probably need to update packable

thibault-martinez: We can't have DerefMut there, it would breach the length restrictions

DaughterOfMars: alternative: use `BTreeSet<String, Vec<u8>>` in the feature and only use the packable types when packing/unapcking. We have to do it manually anyway so it's not such a big deal imo

Thoralf-M: I think it we removed it since the bounds aren't checked if you just insert some value
For a safe way we would probably need custom wrapper methods that do the validation?

thibault-martinez: Tbh I think we should just merge as is and revisit later 🤷🏻‍♂️

DaughterOfMars: alright 🤷‍♀️

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: BTreeMapPrefix has no public insert method, inner type can't be mutated

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Not sure, on the Unpacker trait I don't see a method to get the length of the bytes, do you know a way?

sdk/src/types/block/output/feature/metadata.rs
thibault-martinez: Why do you need to insert ?

sdk/src/types/block/output/feature/metadata.rs
DaughterOfMars: Looks like the only way currently is how it's done for the block:

```rust
let start_opt = unpacker.read_bytes();

...

if VERIFY {
    let block_len = if let (Some(start), Some(end)) = (start_opt, unpacker.read_bytes()) {
        end - start
    } else {
        block.packed_len()
    };
    if block_len > Self::LENGTH_MAX {
        return Err(UnpackError::Packable(Error::InvalidBlockLength(block_len)));
    }
}
```

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Well not necessarily insert, but there is also no TryFrom impl we can use apart from BTreeMap, so how else can we build it?

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: As far as I see this depends on a CounterUnpacker, because otherwise `read_bytes()` just always returns None
Do you think it's really worth it to use the CounterUnpacker just for this?

thibault-martinez (dismissed):

sdk/src/types/block/output/feature/metadata.rs
DaughterOfMars: yeah tbh

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Just derived packable now as you suggested :smile:

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Just derived packable now again and moved it to the verify method where we can't do this as far as I understand

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -51,6 +53,24 @@ fn verify_keys_packable<const VERIFY: bool>(
...
+        BoxedSlicePrefix<u8, MetadataFeatureValueLength>,
+        MetadataFeatureLength,
+    >,
+) -> Result<(), Error> {
+    if VERIFY {
+        let len = map.packed_len();
+        if !MetadataFeature::LENGTH_RANGE
+            .contains(&u16::try_from(len).map_err(|e| Error::InvalidMetadataFeature(e.to_string()))?)
+        {
+            return Err(Error::InvalidMetadataFeature(len.to_string()));
```
Alex6323: Maybe add a bit more context to the string?, like `format!("Out of bounds byte Length: {}")`. Just close if you disagree or find it unnecessary.

Alex6323 (dismissed):

thibault-martinez (changes requested):
CI failing

thibault-martinez (dismissed):

sdk/src/types/block/output/feature/metadata.rs
DaughterOfMars: yeah I really don't like that solution, since it means we call the pack impl on every unpack for no real reason

Thoralf-M: So you prefer manual impl with CounterUnpacker?

DaughterOfMars: yes

DaughterOfMars: the manual impl can be extremely simple, as you can call unpack on the `BTreeSetPrefix` type instead of doing all the work manually.

Thoralf-M: Hm I fear this is actually not possible without a change in packable
https://github.com/iotaledger/iota-sdk/blob/2b44b886538b3c5dbbaf55a8ace38e8f44273bfe/sdk/src/types/block/core.rs#L169 creates the CounterUnpacker in another method which accepts the bytes and only then calls the unpack() method.
Inside of the `unpack()` method there seems to be no way to create it (we need `slice: &'a [u8]` for the SliceUnpacker and if we would have that, we could also just use its length already

DaughterOfMars: We should be able to do this:

```rust
fn unpack<U: Unpacker, const VERIFY: bool>(
        unpacker: &mut U,
        visitor: &Self::UnpackVisitor,
    ) -> Result<Self, UnpackError<Self::UnpackError, U::Error>> {
  let mut unpacker = CounterUnpacker::new(unpacker);
  ...
}
```

since the `CounterUnpacker` just takes any `U: Unpacker` and `Unpacker` is implemented for `&mut U`

sdk/examples/how_tos/nft_collection/01_mint_collection_nft.rs
```diff
@@ -63,7 +66,7 @@ async fn main() -> Result<()> {
     let nft_mint_params = (0..NFT_COLLECTION_SIZE)
         .map(|index| {
             MintNftParams::new()
-                .with_immutable_metadata(get_immutable_metadata(index).to_bytes())
+                .with_immutable_metadata(MetadataFeature::try_from(get_immutable_metadata(index)).unwrap())
```
kwek20, line 69: in the build_nft_output we set this using ::new("IRC27".as_bytes().to_vec(), tip_27_immutable_metadata.

Any reason you chose not to here and below? (mint_nft.rs)

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Right thanks this works

sdk/examples/how_tos/nft_collection/01_mint_collection_nft.rs
Thoralf-M: tip_27_immutable_metadata is a string, here it's the Irc27Metadata struct

kwek20: yea but we should still be consistent with the key naming i guess? It doesnt actually technically matter ofc.
Unless there will be a lookup table for the metadata keys? 

I guess resolve if you decide the key is fine as is

Thoralf-M: Ah yeah it's nowhere official defined yet, so maybe it will change, but I aligned it :+1:

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -1,103 +1,250 @@
...
+            builder.insert(k, v);
+        }
+        builder.finish()
+    }
+
+    pub fn build() -> MetadataFeatureMap {
+        Default::default()
+    }
+
+    pub fn to_map(&self) -> MetadataFeatureMap {
```
Alex6323, line 90: should `build()` and `to_map` be documented?

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -1,103 +1,250 @@
...
+        // Unsafe: acceptable because of type level checking, i.e. keys must be ASCII
+        unsafe {
+            MetadataFeatureMap(
+                self.0
+                    .iter()
+                    .map(|(k, v)| {
+                        (
+                            String::from_utf8_unchecked(k.as_ref().to_owned()),
```
Alex6323: consider this answer to what I also suggested doing:
https://github.com/iotaledger/iota-sdk/pull/1764#discussion_r1453281606
in order to decide to keep the unsafe block. We should be doing it consistently I guess.

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -1,103 +1,250 @@
...
+    #[inline(always)]
+    pub fn new() -> Self {
+        Self(Default::default())
+    }
 
-macro_rules! impl_from_vec {
-    ($type:ty) => {
-        impl TryFrom<$type> for MetadataFeature {
-            type Error = Error;
+    pub fn with_key_value(mut self, key: &str, value: impl Into<Vec<u8>>) -> Self {
```
Alex6323, line 133: no warnings for missing docs here (so it's not part of the sdk's API)?

Thoralf-M: We just have #[allow(missing_docs)]

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: The link doesn't work for me, just shows `There are no files selected for viewing`

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -1,103 +1,250 @@
...
 
     /// Creates a new [`MetadataFeature`].
     #[inline(always)]
-    pub fn new(data: impl Into<Vec<u8>>) -> Result<Self, Error> {
-        Self::try_from(data.into())
+    pub fn new(data: impl IntoIterator<Item = (String, Vec<u8>)>) -> Result<Self, Error> {
+        let mut builder = Self::build();
+        for (k, v) in data {
+            builder.insert(k, v);
+        }
```
Alex6323:
```suggestion
        builder.extend(data.into_iter());
```

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Ok was also not exposed

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: Done

sdk/src/types/block/output/feature/metadata.rs
Alex6323: updated the link

Thoralf-M: Not sure how that's related, but I removed the unsafe block

Alex6323: sorry, misclicked
https://github.com/iotaledger/iota-sdk/pull/1764#discussion_r1431495860

sdk/src/types/block/output/feature/metadata.rs
```diff
@@ -1,103 +1,247 @@
...
-pub struct MetadataFeature(
-    // Binary data.
-    pub(crate) BoxedSlicePrefix<u8, MetadataFeatureLength>,
-);
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct MetadataFeature(MetadataBTreeMapPrefix);
+
+fn verify_packable<const VERIFY: bool>(feature: &MetadataFeature) -> Result<(), Error> {
+    verify_keys_packable::<VERIFY>(feature)?;
+    verify_byte_length_packable::<VERIFY>(feature.packed_len())?;
```
DaughterOfMars, line 43:
```suggestion
    // Add feature type byte
    verify_byte_length_packable::<VERIFY>(feature.packed_len() + 1)?;
```

DaughterOfMars: or was this fixed some other way?

sdk/examples/client/output/build_account_output.rs
```diff
@@ -49,9 +49,9 @@ async fn main() -> Result<()> {
     // Account id needs to be null the first time
     let account_output = AccountOutputBuilder::new_with_minimum_amount(storage_score_params, AccountId::null())
         .add_feature(SenderFeature::new(address.clone()))
-        .add_feature(MetadataFeature::new(metadata)?)
+        .add_feature(MetadataFeature::build().with_key_value("data", metadata).finish()?)
         .add_immutable_feature(IssuerFeature::new(address.clone()))
-        .add_immutable_feature(MetadataFeature::new(metadata)?)
+        .add_immutable_feature(MetadataFeature::build().with_key_value("data", metadata).finish()?)
```
thibault-martinez, line 54: What is different in this example that needs to use `build().with_key_value()` and not just `new`?

DaughterOfMars: I just sprinkled the alternate usage in the examples. I didn't feel like updating them all 🤷‍♀️

thibault-martinez: Ok but as a general question, why is it better?

DaughterOfMars: It's just a better syntax imo, users don't have to deal with three levels of parentheses due to the array of tuples situation. looks cleaner and easier to understand what's happening

cli/src/wallet_cli/mod.rs
```diff
@@ -1240,14 +1255,17 @@ pub async fn prompt_internal(
...
                             foundry_metadata_file,
                         } => {
                             create_native_token_command(
                                 wallet,
                                 circulating_supply,
                                 maximum_supply,
-                                bytes_from_hex_or_file(foundry_metadata_hex, foundry_metadata_file).await?,
+                                bytes_from_hex_or_file(foundry_metadata_hex, foundry_metadata_file)
+                                    .await?
+                                    .map(|d| MetadataFeature::new([(foundry_metadata_key, d)]).unwrap()),
```
thibault-martinez: Just `?` instead of unwrap?

sdk/examples/client/output/build_account_output.rs
thibault-martinez: Alright

sdk/src/types/block/output/feature/metadata.rs
Thoralf-M: I read the TIP again and the range is only for the map inside without the type, so it's not needed (also asked Philipp and he confirmed this)

DaughterOfMars: ah, perfect!

DaughterOfMars (approved):

thibault-martinez (approved):

Alex6323 (approved):

PR #1773, opened 2023-12-13
Address::{is_ed25519_backed, backing_ed25519}

thibault-martinez:


sdk/src/types/block/address/mod.rs
```diff
@@ -113,6 +113,25 @@ impl Address {
...
+
+    /// Returns the backing [`Ed25519Address`], if any.
+    pub fn backing_ed25519(&self) -> Option<&Ed25519Address> {
+        match self {
+            Self::Ed25519(ed25519) => Some(ed25519),
+            Self::ImplicitAccountCreation(implicit) => Some(implicit.ed25519_address()),
+            Self::Restricted(restricted) => restricted.address().as_ed25519_opt(),
+            _ => None,
+        }
+    }
```
Alex6323, line 133: what about multi addresses? :thinking: (especially if it only contains a single weighted ed25519 address)

thibault-martinez: I actually think MultiAddress shouldn't be allowed to be a single address. I spoke with Phillip and he'll talk to the team to disallow or not. Anyway, I wouldn't really consider multi addresses here as there is no clear mapping to a single address being able to sign.

Alex6323 (approved):

kwek20 (commented):
👍 was checking if we could use this somewhere already but doesnt seem so easily

thibault-martinez:
> 👍 was checking if we could use this somewhere already but doesnt seem so easily

I actually think we can but I'd do that when refactoring some parts.

PR #1775, opened 2023-12-13
Fix OutputData::input_signing_data

thibault-martinez:


Alex6323 (dismissed):

sdk/src/wallet/types/mod.rs
```diff
@@ -66,20 +66,18 @@ impl OutputData {
...
-        } else if required_address.is_ed25519() {
-            if wallet_data.address.inner() == &required_address {
-                // TODO #1279: do we need a check to make sure that `wallet_data.address` and `wallet_data.bip_path` are
-                // never conflicting?
+        // TODO #1279: do we need a check to make sure that `wallet_data.address` and `wallet_data.bip_path` are never
+        // conflicting?
+
+        let chain = self.chain.or_else(|| {
+            if &required_address == wallet_data.address.inner()
+                || required_address.backing_ed25519() == wallet_data.address.inner().as_ed25519_opt()
```
Alex6323: edge case: what if both or `None`? :thinking:

kwek20 (commented):
we used to return Ok(None);, now no more. do we really need to send Result<Option<>> still?

sdk/src/wallet/types/mod.rs
Thoralf-M: `wallet_data.address.inner().as_ed25519_opt()` can not be None at this point?

Thoralf-M (dismissed):

sdk/src/wallet/types/mod.rs
Alex6323: chain implies ed25519, right? So can we remove the first part of the if?

Also to make extra sure we match `Some`s we could then just do:
`if matches!((required_address.backing_ed25519(), wallet_data.address.inner().as_ed25519_opt()), (Some(a), Some(b)) if a == b) { ... }`

thibault-martinez: Honestly I don't like these `matches!` with an included `if`, I find them extremely hard to read

Alex6323: you can't do that unfortunately :shrug: :

```rust
if let (Some(a), Some(b)) = (required_address.backing_ed25519(), wallet_data.address.inner().as_ed25519_opt()) && a == b { ... }
```

still ... the first part of the `if` isn't necessary, right?

thibault-martinez: What first part? `required_address == wallet_data.address.inner()`? It's so we also check address that are not directly Ed25519

Thoralf-M (approved):

Alex6323 (approved):

PR #1776, opened 2023-12-13
Add UTXO changes full API

thibault-martinez:
https://github.com/iotaledger/iota.go/pull/638

### DIFF
--- a/bindings/core/src/method/client.rs
+++ b/bindings/core/src/method/client.rs
@@ -273,6 +273,12 @@ pub enum ClientMethod {
         /// Commitment ID of the commitment to look up.
         commitment_id: SlotCommitmentId,
     },
+    /// Get all full UTXO changes of a given slot by Commitment ID.
+    #[serde(rename_all = "camelCase")]
+    GetUtxoChangesFull {
+        /// Commitment ID of the commitment to look up.
+        commitment_id: SlotCommitmentId,
+    },
     /// Look up a commitment by a given commitment index.
     GetCommitmentByIndex {
         /// Index of the commitment to look up.
@@ -280,6 +286,11 @@ pub enum ClientMethod {
     },
     /// Get all UTXO changes of a given slot by commitment index.
     GetUtxoChangesByIndex {
+        /// Index of the commitment to look up.
+        index: SlotIndex,
+    },
+    /// Get all full UTXO changes of a given slot by commitment index.
+    GetUtxoChangesFullByIndex {
         /// Index of the commitment to look up.
         index: SlotIndex,
     },
--- a/bindings/core/src/response.rs
+++ b/bindings/core/src/response.rs
@@ -18,7 +18,8 @@ use iota_sdk::{
             core::{
                 BlockMetadataResponse, BlockWithMetadataResponse, CommitteeResponse, CongestionResponse,
                 InfoResponse as NodeInfo, IssuanceBlockHeaderResponse, ManaRewardsResponse, OutputWithMetadataResponse,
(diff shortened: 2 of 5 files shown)


bindings/core/src/method/client.rs
```diff
@@ -273,6 +273,12 @@ pub enum ClientMethod {
...
         commitment_id: SlotCommitmentId,
     },
+    /// Get all full UTXO changes of a given slot by Commitment ID.
+    #[serde(rename_all = "camelCase")]
+    GetUtxoChangesFull {
+        /// Commitment ID of the commitment to look up.
+        commitment_id: SlotCommitmentId,
+    },
     /// Look up a commitment by a given commitment index.
     GetCommitmentByIndex {
```
kwek20, line 283: order! move this one up to behind here please:
```
GetCommitment {
        /// Commitment ID of the commitment to look up.
        commitment_id: SlotCommitmentId,
    },
```

thibault-martinez: I'd rather not, this is the order of the TIP

kwek20: why wouldnt we have it alphabetical? what order does the tip even follow?

thibault-martinez: They group by parameter, first SlotCommitmentId then SlotIndex

Thoralf-M (approved):

kwek20 (approved):

sdk/src/client/node_api/core/routes.rs
```diff
@@ -333,6 +334,17 @@ impl ClientInner {
         self.get_request(path, None, false, false).await
     }
 
+    /// Get all full UTXO changes of a given slot by slot commitment ID.
+    /// GET /api/core/v3/commitments/{commitmentId}/utxo-changes/full
+    pub async fn get_utxo_changes_full_by_slot_commitment_id(
```
Alex6323, line 339: I would suggest the name `get_full_utxo_changes_by_slot_commitment_id` tbh, but ignore if you disagree.
Should you agree, then the response type should probably be renamed too.

bindings/core/src/response.rs
```diff
@@ -142,6 +143,10 @@ pub enum Response {
     /// - [`GetUtxoChangesByIndex`](crate::method::ClientMethod::GetUtxoChangesByIndex)
     UtxoChanges(UtxoChangesResponse),
     /// Response for:
+    /// - [`GetUtxoChangesFull`](crate::method::ClientMethod::GetUtxoChangesFull)
+    /// - [`GetUtxoChangesFullByIndex`](crate::method::ClientMethod::GetUtxoChangesFullByIndex)
+    UtxoChangesFull(UtxoChangesFullResponse),
```
Alex6323, line 148: btw ... I think for the variant it's reasonable to have `full` suffixed as it is here so both variants can be next to each other, although maybe we should just no alphabetially sort enum variants, but group them logically :shrug:

sdk/src/client/node_api/core/routes.rs
thibault-martinez: I would keep this way so it's more consistent with the URL

bindings/core/src/response.rs
thibault-martinez: They are grouped logically, by parameter. See Brord's comment.

thibault-martinez: Oh that's the response file actually, that's a bit different. Don't think it matters that much...

Alex6323 (approved):

PR #1782, opened 2023-12-14
Fix ProtocolParameters::slot_index

### LINKED ISSUE #1781: Fix block id computation
Currently for some reason node returns "0x00c62140075ce5ecb1aa85d3a24f9807a41ccd4599b70c81f3905c9a067f2410575a0000", but we return
"0x00c62140075ce5ecb1aa85d3a24f9807a41ccd4599b70c81f3905c9a067f2410525a0000", so `7` is replaced by `2` for some reason

```TS
        let info = await client.getInfo();
        const block = await client.getBlock("0x00c62140075ce5ecb1aa85d3a24f9807a41ccd4599b70c81f3905c9a067f2410575a0000");
        console.log(Utils.blockId(block, info.nodeInfo.protocolParameters[0].parameters))
```

thibault-martinez:


Thoralf-M (approved):

kwek20 (approved):
So genesis slot can be more than 0 right?

Thoralf-M:
> So genesis slot can be more than 0 right?

Correct

PR #1787, opened 2023-12-18
Update protocol parameters test

thibault-martinez:


Alex6323:
Do we really need this PR? This is also done in #1607.

thibault-martinez:
> Do we really need this PR? This is also done in #1607.

Ah I didn't know. I don't like the formatting in yours though, but we can change it there.

PR #1790, opened 2023-12-18
MultiAddress Display update

### LINKED ISSUE #1587: Reconsider what Address Display should be
It just forwards to the display impls of the rest of the addresses. We Display only the hex IDs for the addresses. We may want to reconsider what we want as Display but not in this PR I'd say.

_Originally posted by @thibault-martinez in https://github.com/iotaledger/iota-sdk/pull/1573#discussion_r1384874421_

kwek20:
# Description of change

Changes MultiAddress Display to show the hashed total of its addresses instead of the addresses themselves

Fixes #1587


thibault-martinez (dismissed):

sdk/tests/types/address/multi.rs
```diff
@@ -55,6 +55,10 @@ fn json_packable_bech32() {
     let multi_address_unpacked = Address::unpack_verified(multi_address_bytes, &()).unwrap();
 
     assert_eq!(multi_address, multi_address_unpacked);
+    assert_eq!(
+        format!("{}", multi_address.as_multi()),
```
Alex6323: `.to_string()`?

sdk/src/types/block/address/multi.rs
```diff
@@ -121,6 +122,17 @@ impl MultiAddress {
     pub fn threshold(&self) -> u16 {
         self.threshold
     }
+
+    /// Hash the [`MultiAddress`] with BLAKE2b-256.
+    #[inline(always)]
+    pub fn hash(&self) -> [u8; 32] {
```
Alex6323, line 128: why `pub`? users gonna need to work with the hash of it?

sdk/tests/types/address/multi.rs
thibault-martinez: Done

sdk/src/types/block/address/multi.rs
thibault-martinez: I asked myself the same but since it's public anyway through the display I think it's fair to have the method pub

kwek20: yea, figured it wouldnt hurt to have available

thibault-martinez (approved):

PR #1791, opened 2023-12-18
Some TIP cleanup

thibault-martinez:


sdk/src/types/block/payload/tagged_data.rs
```diff
@@ -1,7 +1,8 @@
 // Copyright 2020-2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-//! Module describing the tagged data payload.
+//! A basic payload type that allows the addition of arbitrary data.
+//! [TIP-53: Tagged Data](https://github.com/iotaledger/tips/blob/tip53/tips/TIP-0053/tip-0053.md).
```
Alex6323, line 5: I feel like this doc should be on the type itself, and I would keep the module doc as it was.

sdk/src/types/block/address/multi.rs
```diff
@@ -97,8 +98,9 @@ impl MultiAddress {
     pub fn new(addresses: impl IntoIterator<Item = WeightedAddress>, threshold: u16) -> Result<Self, Error> {
         let mut addresses = addresses.into_iter().collect::<Box<[_]>>();
 
-        addresses.sort_by(|a, b| a.address().cmp(b.address()));
+        addresses.sort_by(|a, b| a.address().pack_to_vec().cmp(&b.address().pack_to_vec()));
```
DaughterOfMars: I recognize that this is not likely to present a significant performance concern, but we have really no idea how many times this closure will be called. It would be better to generate a map (BTreeMap, probably) with the keys being the packed representation, then iterate through the map in order to create the sorted list.

sdk/src/types/block/payload/tagged_data.rs
thibault-martinez: I'd rather keep it like that tbh, because I'll do the same for them all so it's consistent. The content of a TIP may affect more than just the type so I think module wise make sense.

sdk/src/types/block/output/storage_score.rs
```diff
@@ -1,6 +1,10 @@
 // Copyright 2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+//! Storage deposit is a concept that creates a monetary incentive to keep the ledger state small. This is achieved by
+//! enforcing a minimum IOTA coin deposit in every output based on the actually used disk space of the output itself.
```
DaughterOfMars:
```suggestion
//! enforcing a minimum IOTA coin deposit in every output based on the disk space that will actually be used to store it.
```

sdk/src/types/block/address/multi.rs
thibault-martinez: Fair enough

sdk/src/types/block/payload/tagged_data.rs
DaughterOfMars: I'm fine with it as-is

Alex6323: okay

sdk/src/types/block/address/multi.rs
Alex6323: agree ... and out of interest: can you maybe explain to me what this change corresponds to in the TIP?

thibault-martinez: https://github.com/iotaledger/tips/blob/tip52/tips/TIP-0052/tip-0052.md#additional-syntactic-transaction-validation-rules
> The addresses must be lexically ordered and unique based on the contained serialized address.

thibault-martinez: Done

DaughterOfMars (approved):

Thoralf-M (approved):

PR #1793, opened 2023-12-18
feat(2.0, nodejs): Silently request to stop background syncing when `WalletMethodHandler` is dropped

marc2332:
# Description of change

This will make sure that the background syncing is stopped when the `WalletMethodHandler` is dropped. You might ask why this is useful as there is already `destroyWallet` method. Well, this is more like a "fallback" plan for when that method is not called yet the wallet is dropped. For example, I encountered that everytime I edited some frontend code in Firefly, that triggered webpack to reload the frontend, and thus removing from memory all the Wallets instances, without their `stopBackgroundSync` method called. This caused that if I wanted to log into a profile whose `stopBackgroundSync` method was not called before, now gave me a rocksdb error, because, the old wallet instance was still living in the background syncing thread, and thus the DB still locked.

## Type of change

- Enhancement

In firefly, but you can test it by starting the background syncing and never stopping, but destroying the wallet.

## Change checklist


marc2332:
A potential case where a new Wallet instance is created quickly enough to throw a `no available locks` error, just after a previous wallet instance has been dropped without manually calling stopBackgroundSync and thus calling it asynchronously when it drops will be improved by https://github.com/iotaledger/iota-sdk/pull/1797

bindings/nodejs/src/wallet.rs
```diff
@@ -16,26 +16,52 @@ use crate::{
...
 
-pub type WalletMethodHandler = Arc<RwLock<Option<Wallet>>>;
+pub struct WalletMethodHandlerInner(Option<Wallet>);
+
+impl Drop for WalletMethodHandlerInner {
+    fn drop(&mut self) {
+        log::debug!("drop WalletMethodHandlerInner");
+        // Request to stop the background syncing silently if this wallet hasn't been destroyed yet
+        if let Some(wallet) = self.0.take() {
+            wallet.request_stop_background_syncing();
```
Thoralf-M, line 29: Don't you want to call this in `destroy_wallet()`?

marc2332: > Don't you want to call this in `destroy_wallet()`?

The point here is that you might not have control over how the wallet is destroyed, in our case this happens when reloading (ctrl-r) Firefly in development mode. The wallets instances get destroyed, but their `stopBackgroundSync` methods never get called. This is kind of like a fallback so you don't have "zombie" threads locking the db.

marc2332: > Don't you want to call this in `destroy_wallet()`?

That would be great too though :) But it doesn't fix our issue

Thoralf-M (dismissed):

thibault-martinez (changes requested):
Please rebase to remove the python errors and apply format 🙏🏻

sdk/src/wallet/core/operations/background_syncing.rs
```diff
@@ -80,7 +86,7 @@ where
             return Ok(());
         }
         // send stop request
-        self.background_syncing_status.store(2, Ordering::Relaxed);
+        self.request_stop_background_syncing();
         // wait until it stopped
         while self.background_syncing_status.load(Ordering::Relaxed) != 0 {
```
DaughterOfMars, line 91: Upon seeing this code again, it strikes me that this is a really error-prone way to do this, and that we should be using a different sync type here that we can `await` instead of just sleeping.

marc2332: I got you https://github.com/iotaledger/iota-sdk/pull/1797  😉

Thoralf-M (dismissed):

sdk/src/wallet/core/operations/background_syncing.rs
DaughterOfMars: Ah, good!

bindings/nodejs/src/wallet.rs
```diff
@@ -16,26 +19,54 @@ use crate::{
...
 };
 
-pub type WalletMethodHandler = Arc<RwLock<Option<Wallet>>>;
+pub struct WalletMethodHandlerInner(Option<Wallet>);
+
+impl Drop for WalletMethodHandlerInner {
+    fn drop(&mut self) {
+        log::debug!("drop WalletMethodHandlerInner");
+        // Request to stop the background syncing silently if this wallet hasn't been destroyed yet
+        if let Some(wallet) = self.0.take() {
```
DaughterOfMars, line 28: So, if `destroy_wallet` is called before dropping, this will not stop the syncing. Perhaps we should repeat this code in that fn?

thibault-martinez: @marc2332 Just making sure you've seen this question :)

marc2332: It was also suggested in https://github.com/iotaledger/iota-sdk/pull/1793#discussion_r1433757234, I can do it if you want :)

marc2332: > @marc2332 Just making sure you've seen this question :)

Nice timing :joy:

thibault-martinez: Does that answer your question @DaughterOfMars ?

DaughterOfMars: That seems to be a different question. I'm saying that this check will do nothing in the Drop impl if destroy_wallet was called first. So if you do this order of operations:
1. Start background syncing
2. Call `destroyWallet`
3. Drop the memory

the background syncing will still continue.

thibault-martinez (dismissed):

bindings/nodejs/src/wallet.rs
marc2332: > That seems to be a different question. I'm saying that this check will do nothing in the Drop impl if destroy_wallet was called first. So if you do this order of operations:
> 
>     1. Start background syncing
> 
>     2. Call `destroyWallet`
> 
>     3. Drop the memory
> 
> 
> the background syncing will still continue.

That's exactly what I am talking about, I could also make it call `request_stop_background_syncing` in `destroyWallet` so it stops the background syncing

DaughterOfMars: Oh, I thought that other comment was about moving the call from the drop impl. But yes, I think it needs to be called in both places

Thoralf-M (approved):

thibault-martinez (approved):
Thank you 🙏🏻

PR #1798, opened 2023-12-19
Align ContextInput type bytes

thibault-martinez:
https://github.com/iotaledger/iota.go/pull/644

### DIFF
--- a/bindings/core/src/method/utils.rs
+++ b/bindings/core/src/method/utils.rs
@@ -76,7 +76,7 @@ pub enum UtilsMethod {
     ComputeNftId { output_id: OutputId },
     /// Computes the output ID from transaction id and output index
     ComputeOutputId { id: TransactionId, index: u16 },
-    /// Computes a tokenId from the aliasId, serial number and token scheme type.
+    /// Computes a tokenId from the accountId, serial number and token scheme type.
     #[serde(rename_all = "camelCase")]
     ComputeTokenId {
         account_id: AccountId,
--- a/sdk/src/types/block/context_input/mod.rs
+++ b/sdk/src/types/block/context_input/mod.rs
@@ -144,14 +144,12 @@ mod tests {
         assert_eq!(reward.as_reward().index(), 10);
 
         // Test wrong type returns error.
-        let reward_serialization_result: Result<ContextInput, _> = serde_json::from_str(
-            r#"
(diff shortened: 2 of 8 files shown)


bindings/core/src/method/utils.rs
```diff
@@ -76,7 +76,7 @@ pub enum UtilsMethod {
     ComputeNftId { output_id: OutputId },
     /// Computes the output ID from transaction id and output index
     ComputeOutputId { id: TransactionId, index: u16 },
-    /// Computes a tokenId from the aliasId, serial number and token scheme type.
+    /// Computes a tokenId from the accountId, serial number and token scheme type.
```
Alex6323, line 79: `output ID`, `transaction id`, and `accountId` :see_no_evil:

thibault-martinez: Feel free to open a PR to harmonise the whole codebase but I won't do it here :carlos:

sdk/src/types/block/context_input/mod.rs
```diff
@@ -135,7 +135,7 @@ mod tests {
     fn reward() {
         let reward: ContextInput = serde_json::from_value(serde_json::json!(
             {
-                "type": 3,
+                "type": 2,
```
Alex6323, line 138: I noticed that the supposed-to-be-failing test down below uses `serde_json::from_str(<raw_string>)` instead of how we do it above? This could be made consistent,right?

sdk/tests/types/address/mod.rs
```diff
@@ -17,7 +17,7 @@ use iota_sdk::types::block::{
 use pretty_assertions::assert_eq;
 
 const ED25519_ADDRESS: &str = "0xebe40a263480190dcd7939447ee01aefa73d6f3cc33c90ef7bf905abf8728655";
-const ALIAS_ID: &str = "0xe9ba80ad1561e437b663a1f1efbfabd544b0d7da7bb33e0a62e99b20ee450bee";
+const ACCOUNT_ID: &str = "0xe9ba80ad1561e437b663a1f1efbfabd544b0d7da7bb33e0a62e99b20ee450bee";
```
Alex6323, line 20: `...bee` :trollface:

sdk/src/types/block/context_input/mod.rs
thibault-martinez: Done

Alex6323 (approved):

PR #1812, opened 2024-01-08
Evaluate lock contention areas

### LINKED ISSUE #1543: Re-evaluate places where we are taking locks across await points
## Description

We need to evaluate the places where we are calling Wallet::data and data_mut and holding the lock across await points

## Motivation

Possible risk of dead locks.

DaughterOfMars:
# Description of change

There are some places where we need to be very cautious about holding locks. I have done some small refactoring to try to ensure that locks do not occur, and evaluated the places where locks seem likely.

Closes #1543


sdk/src/wallet/core/mod.rs
```diff
@@ -94,7 +94,7 @@ pub struct WalletInner<S: SecretManage = SecretManager> {
     #[cfg(feature = "storage")]
     pub(crate) storage_options: StorageOptions,
     #[cfg(feature = "storage")]
-    pub(crate) storage_manager: tokio::sync::RwLock<StorageManager>,
+    pub(crate) storage_manager: StorageManager,
```
thibault-martinez, line 100: Why is it not needed anymore?

sdk/src/wallet/core/operations/stronghold_backup/mod.rs
```diff
@@ -79,15 +79,20 @@ impl Wallet {
...
 
-        let mut secret_manager = self.secret_manager.as_ref().write().await;
-        // Get the current snapshot path if set
-        let new_snapshot_path = if let SecretManager::Stronghold(stronghold) = &mut *secret_manager {
-            stronghold.snapshot_path.clone()
-        } else {
-            PathBuf::from("wallet.stronghold")
-        };
+        // We don't want to overwrite a possible existing wallet
+        // TODO not too sure about this, it used to check the presence of accounts, this is not 100% equivalent
```
Thoralf-M: I think this is the best way now

sdk/src/wallet/core/mod.rs
DaughterOfMars: All access is immutable. Figured it's better not to have the lock at all in that case. However, if you think there's a valid reason to leave it, let me know.

Alex6323 (dismissed):

sdk/src/wallet/core/operations/stronghold_backup/mod.rs
```diff
@@ -103,16 +107,24 @@ impl Wallet {
...
             if ignore {
                 // TODO: #1279 okay that if both are none we always load the backup values?
-                wallet_data.bip_path != read_bip_path
+                curr_bip_path != read_bip_path
             } else {
                 false
             }
         });
 
+        // TODO: Do we even need to do this, since it should get set when we restore below?
```
Thoralf-M: I think it's a bit different, it only gets set below `if restore_wallet {` which depends on the bech32 hrp

Thoralf-M (dismissed):

Alex6323 (dismissed):

thibault-martinez (changes requested):
Format

thibault-martinez (approved):

Alex6323 (approved):

PR #1815, opened 2024-01-09
CLI: better UX for accounts and implicit_accounts commands

thibault-martinez:


Thoralf-M (approved):

cli/src/wallet_cli/mod.rs
```diff
@@ -314,7 +314,25 @@ impl FromStr for OutputSelector {
...
+        let account_id = account.output.as_account().account_id_non_null(&output_id);
+        let account_address = account_id.to_bech32(wallet.client().get_bech32_hrp().await?);
+
+        println_log_info!(
+            "{:<16} {output_id}\n{:<16} {account_id}\n{:<16} {account_address}\n",
+            "Output ID:",
+            "Account ID:",
+            "Account Address:"
+        );
+    }
```
Alex6323, line 333: can this be factored out into a function bc you do the same for the implicit accounts, right?

thibault-martinez: It's not the same

kwek20 (approved):

PR #1818, opened 2024-01-09
Use bytes for Ed25519Signature public key

DaughterOfMars:
# Description of change

Looks like using the `PublicKey` type here was a mistake, as the validation is not intended to be at the syntactic level. Unfortunately this is a breaking change, and there's no way to avoid that.

https://github.com/iotaledger/inx-chronicle/issues/1308

## Type of change

### DIFF
--- a/bindings/core/src/method_handler/utils.rs
+++ b/bindings/core/src/method_handler/utils.rs
@@ -1,4 +1,4 @@
-// Copyright 2023 IOTA Stiftung
+// Copyright 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
 use crypto::keys::bip39::Mnemonic;
@@ -97,7 +97,7 @@ pub(crate) fn call_utils_method_internal(method: UtilsMethod) -> Result<Response
             use iota_sdk::types::block::Error;
             let signature = Ed25519Signature::try_from(signature)?;
             let message: Vec<u8> = prefix_hex::decode(message)?;
-            Response::Bool(signature.verify(&message).map_err(Error::from)?)
+            Response::Bool(signature.try_verify(&message).map_err(Error::from)?)
         }
         UtilsMethod::VerifySecp256k1EcdsaSignature {
             public_key,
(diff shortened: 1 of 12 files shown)


Thoralf-M (changes requested):
Missing changelog entry and version update or don't you need a new release now?

DaughterOfMars:
Yeah, I figured I'd wait until we decide whether this is even happening 😆

sdk/CHANGELOG.md
```diff
@@ -19,6 +19,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 1.1.4 - 2024-MM-DD
+
+### Added
+
+- `Ed25519Signature` methods `new_from_bytes`, `public_key_bytes`, `from_bytes`, `try_verify`;
+
+### Changed
```
thibault-martinez: There is a Deprecated section in the template so I'd use it

bindings/core/src/method_handler/utils.rs
```diff
@@ -94,9 +94,10 @@ pub(crate) fn call_utils_method_internal(method: UtilsMethod) -> Result<Response
             Response::Ok
         }
         UtilsMethod::VerifyEd25519Signature { signature, message } => {
+            use iota_sdk::types::block::Error;
```
thibault-martinez: Move to the other imports pls

sdk/src/types/block/signature/ed25519.rs
```diff
@@ -31,24 +31,44 @@ impl Ed25519Signature {
...
+        }
     }
 
     /// Returns the public key of an [`Ed25519Signature`].
+    #[deprecated(since = "1.1.4", note = "use Ed25519Signature::public_key_bytes instead")]
     pub fn public_key(&self) -> &PublicKey {
+        panic!("deprecated method: use Ed25519Signature::public_key_bytes instead")
+    }
+
+    /// Returns the public key of an [`Ed25519Signature`].
```
Alex6323:
```suggestion
    /// Returns the unvalidated public key bytes of an [`Ed25519Signature`].
```

Thoralf-M (approved):

Alex6323 (approved):

PR #1819, opened 2024-01-09
CLI: add congestion command

thibault-martinez:


sdk/src/types/api/core.rs
```diff
@@ -325,7 +325,7 @@ pub struct CongestionResponse {
     pub reference_mana_cost: u64,
     /// The Block Issuance Credits of the requested account.
     #[serde(with = "string")]
-    pub block_issuance_credits: u64,
+    pub block_issuance_credits: i128,
```
thibault-martinez, line 328: Not too sure about this but it can actually be negative...

Thoralf-M (dismissed):

cli/src/wallet_cli/mod.rs
```diff
@@ -478,6 +480,28 @@ pub async fn claimable_outputs_command(wallet: &Wallet) -> Result<(), Error> {
...
+        .or(wallet_data
+            .accounts()
+            .next()
+            .map(|o| o.output.as_account().account_id_non_null(&o.output_id)))
+        .or(wallet_data
+            .implicit_accounts()
+            .next()
+            .map(|o| AccountId::from(&o.output_id)))
+        .ok_or(WalletError::NoAccountToIssueBlock)?;
+    drop(wallet_data);
```
DaughterOfMars: Just a tiny bit better I think
```suggestion
    let account_id = {
      let wallet_data = wallet.data().await;
      account_id
          .or_else(|| wallet_data
              .accounts()
              .next()
              .map(|o| o.output.as_account().account_id_non_null(&o.output_id)))
          .or_else(|| wallet_data
              .implicit_accounts()
              .next()
              .map(|o| AccountId::from(&o.output_id)))
          .ok_or(WalletError::NoAccountToIssueBlock)?
    };
```

sdk/src/types/api/core.rs
DaughterOfMars: 👀

Thoralf-M (approved):

sdk/src/types/api/core.rs
thibault-martinez: We still need the full amplitude of an u64 but it can also go negative so I guess i128 will do

DaughterOfMars (approved):

PR #1824, opened 2024-01-10
Align method serialization in Python

### LINKED ISSUE #650: Align Python serialization
## Description

Currently in the wallet we have MyEncoder https://github.com/iotaledger/iota-sdk/blob/4c25849ff8cc472d561f4494bc4d2542f0c4dbfc/bindings/python/iota_sdk/wallet/common.py#L15 which converts the types to a dict, but in the client not https://github.com/iotaledger/iota-sdk/blob/4c25849ff8cc472d561f4494bc4d2542f0c4dbfc/bindings/python/iota_sdk/client/client.py#L142-L153, so in the client methods we make sure that the types are converted to dicts if required already before like here https://github.com/iotaledger/iota-sdk/blob/4c25849ff8cc472d561f4494bc4d2542f0c4dbfc/bindings/python/iota_sdk/client/client.py#L246-L257
For some types like the amount where we change the type from int to str it might be required, but to convert classes to dicts it shouldn't be necessary to do that manually.

## Motivation

Aligned/cleaner code

## Open questions (optional)

What's the best way to do it?

No.

Thoralf-M:
# Description of change

Align method serialization in Python
Fix TimelockUnlockConditionDto

Closes #650 

## Type of change

Running examples


bindings/python/iota_sdk/client/_node_core_api.py
```diff
@@ -179,8 +179,6 @@ def call_plugin_route(self, base_plugin_path: str, method: str,
             query_params: The parameters of the query.
             request: The request object sent to the endpoint of the plugin.
         """
-        if query_params is None:
```
DaughterOfMars: Wouldn't this still be needed?

Thoralf-M: No, because I changed it in the parameter `= []` assigns the default value if nothing was provided

DaughterOfMars: But what if you pass in `None`? Will it break?

DaughterOfMars: Maybe I'm overthinking this actually

DaughterOfMars (dismissed):
I won't pretend to understand all of the python nonsense, but looks ok to me

thibault-martinez (changes requested):
Tests failing

bindings/python/iota_sdk/client/_node_core_api.py
Thoralf-M: I think you're right that it would break if None is provided and lint complained so it's reverted now

bindings/python/iota_sdk/common.py
```diff
@@ -0,0 +1,51 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import json
+from json import dumps, JSONEncoder
+from enum import Enum
+import humps
+
+
+def custom_encoder(func, *args, **kwargs):
+    """The routine of dump json string.
```
Alex6323: I don't understand this sentence :smile:

bindings/python/iota_sdk/common.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    class MyEncoder(JSONEncoder):
+        """Custom encoder
+        """
+        # pylint: disable=too-many-return-statements
```
Alex6323, line 16: such a weird lint: What are you supposed to do to fix it if you really wanted to (asking for anybody, not just friends)?

bindings/python/iota_sdk/common.py
Thoralf-M: You already put the dot behind it in the past :trollface:
Is `Converts the parameters to a JSON string.` better?

bindings/python/iota_sdk/common.py
Thoralf-M: I guess the lint checker is just not really smart, for the example here they have a suggestion
https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/too-many-return-statements.html

bindings/python/iota_sdk/common.py
```diff
@@ -0,0 +1,51 @@
...
+                return o.__dict__
+            if isinstance(o, dict):
+                return o
+            if hasattr(o, "__dict__"):
+                obj_dict = o.__dict__
+                items_method = getattr(self, "items", None)
+                if callable(items_method):
+                    for k, v in obj_dict.items():
+                        obj_dict[k] = dumps(v, cls=MyEncoder)
+                        return obj_dict
```
Alex6323: is `return_obj_dict` correctly indented (~should it even be there?~)? feels like the loop should run through and not just return in the first iteration :thinking:

bindings/python/iota_sdk/common.py
```diff
@@ -0,0 +1,51 @@
...
+    for k, v in message.items():
+        if not isinstance(v, str):
+            message[k] = json.loads(dumps(v, cls=MyEncoder))
+
+    def remove_none(obj):
+        if isinstance(obj, (list, tuple, set)):
+            return type(obj)(remove_none(x) for x in obj if x is not None)
+        if isinstance(obj, dict):
+            return type(obj)((remove_none(k), remove_none(v))
+                             for k, v in obj.items() if k is not None and v is not None)
```
Alex6323, line 46: does it have to be indented that much?

bindings/python/iota_sdk/common.py
```diff
@@ -0,0 +1,51 @@
...
+        return obj
+    message_null_filtered = remove_none(message)
```
Alex6323, line 49:
```suggestion

    message_null_filtered = remove_none(message)
```

bindings/python/iota_sdk/common.py
Alex6323: yeah, I guess that would have helped understanding the method better :joy: It also recursively removes `None`s, right? I don't know if that should be mentioned somehow?

bindings/python/iota_sdk/common.py
Thoralf-M: Happens automatically for me, don't know why it's so much :joy:

bindings/python/iota_sdk/common.py
Thoralf-M: Hm yeah probably not needed at all because it's returned in the line below afterwards then, removed it

bindings/python/iota_sdk/common.py
Thoralf-M: Added it to the comment

bindings/python/iota_sdk/common.py
DaughterOfMars: Feels like a really weird thing to be strongly opinionated about

bindings/python/iota_sdk/common.py
DaughterOfMars: I think there's an extra parentheses at the end

Thoralf-M: It's the closing one for `return type(obj)(`, but probably that's why it's intended that much

DaughterOfMars: Ah, yes, that seems likely

Alex6323 (approved):

thibault-martinez (approved):

PR #1837, opened 2024-01-15
CLI: option-wrap congestion BIC

thibault-martinez:


Thoralf-M (dismissed):

cli/src/wallet_cli/mod.rs
```diff
@@ -330,14 +330,10 @@ pub async fn accounts_command(wallet: &Wallet) -> Result<(), Error> {
         let output_id = account.output_id;
         let account_id = account.output.as_account().account_id_non_null(&output_id);
         let account_address = account_id.to_bech32(wallet.client().get_bech32_hrp().await?);
-        let bic = wallet
-            .client()
-            .get_account_congestion(&account_id)
-            .await?
-            .block_issuance_credits;
+        let bic = wallet.client().get_account_congestion(&account_id).await.ok();
```
Alex6323: just noticed something very small and it's related to line 332 above: Can we make the client request for HRP just once before the loop? If you don't want to do this in this PR I can open one.

thibault-martinez: Doesn't change anything since it's cached but I've changed it anyway

Alex6323 (approved):

Thoralf-M (approved):

PR #1842, opened 2024-01-16
Add more tests from TIP

DaughterOfMars:
# Description of change

Adds tests from https://github.com/iotaledger/tips/commit/237e175b344d6dbf8d560f22679717d8a9bfb747


kwek20 (approved):
I know its from tip, but do we want some ManaGenerationTests with error?

DaughterOfMars:
> I know its from tip, but do we want some ManaGenerationTests with error?

Well, there's actually only one way it can fail, and we covered that.

thibault-martinez (approved):

Thoralf-M (approved):

PR #1852, opened 2024-01-18
Clippy allow single_element_loop, expect_fun_call in examples, bump h2

### LINKED ISSUE #1694: Clippy: specific allows for examples
Currently the examples have a lot of manual `#[allow(clippy::single_element_loop)]`
We could update when rust 1.75 is stable to use a `[lint]` section in `Cargo.toml`.
As this will cover the entire workspace member, we have to disallow the allowed lints in the sdk/src/lib.rs again.
(or wait until this works per folder)

An optional second task is update 
`std::env::var(var).unwrap_or_else(|_| panic!(".env variable '{var}' is undefined, see .env.example"));`
to
`std::env::var(var).expect(&format!(".env variable '{var}' is undefined, see .env.example"));`

which is currently not allowed due to `unwrap` 

issue stemming form #1676

Thoralf-M:
# Description of change

Clippy allow single_element_loop, expect_fun_call in examples, bump h2.

Closes https://github.com/iotaledger/iota-sdk/issues/1694.

## Type of change

Running `cargo clippy --examples`

### DIFF
--- a/sdk/examples/client/01_generate_addresses.rs
+++ b/sdk/examples/client/01_generate_addresses.rs
@@ -20,7 +20,7 @@ async fn main() -> Result<()> {
     dotenvy::dotenv().ok();
 
     for var in ["NODE_URL", "MNEMONIC"] {
-        std::env::var(var).unwrap_or_else(|_| panic!(".env variable '{var}' is undefined, see .env.example"));
+        std::env::var(var).expect(&format!(".env variable '{var}' is undefined, see .env.example"));
     }
 
     // Create a node client.
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -11,6 +11,8 @@
     clippy::significant_drop_in_scrutinee,
     clippy::significant_drop_tightening
 )]
+// Allowed in Cargo.toml for examples
+#![deny(clippy::single_element_loop, clippy::single_element_loop)]
 
 extern crate alloc;
 
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -12,7 +12,7 @@
     clippy::significant_drop_tightening
 )]
 // Allowed in Cargo.toml for examples
-#![deny(clippy::single_element_loop, clippy::single_element_loop)]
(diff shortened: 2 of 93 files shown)


sdk/src/lib.rs
```diff
@@ -11,6 +11,8 @@
     clippy::significant_drop_in_scrutinee,
     clippy::significant_drop_tightening
 )]
+// Allowed in Cargo.toml for examples
+#![deny(clippy::single_element_loop, clippy::single_element_loop)]
```
thibault-martinez: :trollface:

Thoralf-M: Thanks

thibault-martinez (approved):

sdk/examples/client/01_generate_addresses.rs
```diff
@@ -20,7 +20,7 @@ async fn main() -> Result<()> {
     dotenvy::dotenv().ok();
 
     for var in ["NODE_URL", "MNEMONIC"] {
-        std::env::var(var).unwrap_or_else(|_| panic!(".env variable '{var}' is undefined, see .env.example"));
+        std::env::var(var).expect(&format!(".env variable '{var}' is undefined, see .env.example"));
```
Alex6323, line 23: not really sure I like it ... why not just?
```
`.unwrap_or_else(|| panic!(".env variable '{var}' is undefined, see .env.example"))
```
it's a bit longer and not exactly the same as expect, yes but wouldn't we wouldn't have to do the weird enable/disable lint :shrug: 

there really should be an `expect_with` or smth in std that takes a closure.

Thoralf-M: > why not just? .... it's a bit longer

This was the only reason to have less distracting code in the examples, don't think the lints are that weird

But you can fight with wild @kwek20 about what's better :trollface:

thibault-martinez: I don't really have a strong opinion but I'm willing to do slightly unconventional things if it makes the examples a bit more straight to the point.

DaughterOfMars (approved):

PR #1855, opened 2024-01-18
Split Python CI

### LINKED ISSUE #1779: Binginds/python: Investigate GH action lint
## Description
The lint "Python bindings checks / Linter & Tests" takes 15 minutes to complete

## Motivation

 this should be way faster

## Requirements

1. Split up tests and lints like how autopep8 is split
2. Have lint actions finish in under 2 minutes of the action starting

## Open questions (optional)

run sequence or parallel?

Thoralf-M:
# Description of change

Split Python CI so it's faster and cleaner

Unfortunately, the linting can only be faster for the main package (finishes in 25 seconds) and not for the examples and tests, as these require the installation of the package first. (I haven't found a clean way to install it without compiling)

Closes #1779 

## Type of change

CI

### DIFF
--- a/.github/workflows/bindings-python.yml
+++ b/.github/workflows/bindings-python.yml
@@ -39,13 +39,41 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  lint:
+  format:
     name: Python PEP8 format
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9"]
+        python-version: ["3.10"]
+
+    steps:
+      - name: Checkout the Source Code
+        uses: actions/checkout@v3
+
+      - name: Set up Python ${{ matrix.python-version }} and Pip Cache
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: pip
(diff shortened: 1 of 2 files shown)


bindings/python/tox.ini
```diff
@@ -29,7 +29,6 @@ deps =
     -r requirements-dev.txt
     pylint
 commands =
-    pip install .
```
Alex6323, line 32: Why can this be removed? If I run `tox -e lint-sdk` locally this wouldn't use the latest changes?

Thoralf-M: Because it checks the source folder which doesn't get modified by installation, for tests and examples we need to build it, because there the sdk is imported

.github/workflows/bindings-python.yml
```diff
@@ -67,18 +67,57 @@ jobs:
         working-directory: bindings/python
         run: tox -e format
 
+  lint:
```
thibault-martinez, line 70: Should this also `needs: format` ?

Thoralf-M: Would delay it a few seconds, but I don't really care if we make it dependent or not

thibault-martinez (approved):

Alex6323 (approved):

PR #1866, opened 2024-01-22
Add block signature to block body max burned mana

### LINKED ISSUE #1861: BurnedMana based on Block workscore
And not only BlockBody workscore.
Waiting for a TIP update.
https://github.com/iotaledger/iota.go/pull/665

DaughterOfMars:
# Description of change

Unfortunately there's no way for us to use the `Block` `WorkScore` implementation, so this is a bit of a hack 🤷‍♀️ 

Closes #1861


Thoralf-M (dismissed):

sdk/src/types/block/core/basic.rs
```diff
@@ -106,7 +106,7 @@ impl BasicBlockBodyBuilder {
             MaxBurnedManaAmount::MinimumAmount {
                 params,
                 reference_mana_cost,
-            } => body.work_score(params) as u64 * reference_mana_cost,
+            } => (body.work_score(params) as u64 + params.signature_ed25519() as u64) * reference_mana_cost,
```
Alex6323, line 109: since it's a bit out of the ordinary ... maybe a short comment is warranted as a reminder (or maybe even something that should be somehow addressed in the future if possible). Might also just close the comment if you think it isn't worth it.

Alex6323 (dismissed):

sdk/src/types/block/core/basic.rs
thibault-martinez, line 109: I don't like it 😭

thibault-martinez: Let's try to find an alternative before we actually commit to this

DaughterOfMars: I have an idea

Thoralf-M (approved):

thibault-martinez (approved):

PR #1421, opened 2023-10-12
Update delegation creation and transition rules

Thoralf-M:
# Description of change

Add end_epoch delegation transition rules

## Links to any relevant issues

Closes #1313 

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How the change has been tested

Not tested, don't know how to test without having many other things done before

PR #1869, opened 2024-01-22
Add create, delay, and destroy delegation operations

### LINKED ISSUE #1307: Add delegation related methods
https://github.com/iotaledger/tips-draft/blob/tip40/tips/TIP-0040/tip-0040.md#consumed-outputs

- create
- delayed claiming
- destroy

DaughterOfMars:
# Description of change

Adds delegation wallet operations.

Closes #1307


sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
```diff
@@ -0,0 +1,194 @@
+// Copyright 2022 IOTA Stiftung
```
Thoralf-M:
```suggestion
// Copyright 2024 IOTA Stiftung
```

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
```diff
@@ -0,0 +1,194 @@
...
+        Wallet,
+    },
+};
+
+/// Params for `create_delegation_output()`
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CreateDelegationParams {
+    /// Bech32 encoded address which will control the delegation. Default will use the
+    /// ed25519 wallet address
```
Thoralf-M: Maybe put a TODO if the default shouldn't be the account address (would depend on setting one account as preferred/default account)

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
```diff
@@ -0,0 +1,194 @@
...
+        let past_bounded_epoch_index =
+            past_bounded_slot_index.to_epoch_index(protocol_parameters.slots_per_epoch_exponent);
+
+        let registration_slot = (past_bounded_epoch_index + 1).registration_slot(
+            protocol_parameters.slots_per_epoch_exponent,
+            protocol_parameters.epoch_nearing_threshold,
+        );
+
+        let delegation_output_builder = DelegationOutputBuilder::new_with_minimum_amount(
+            protocol_parameters.storage_score_parameters(),
```
Thoralf-M: Use the delegated_amount for the output amount, they must be equal at creation, probably also answers the TODO at L26

thibault-martinez (changes requested):
Please add to bindings core and bindings

bindings/nodejs/lib/types/wallet/create-delegation-transaction.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright 2021-2023 IOTA Stiftung
```
Thoralf-M:
```suggestion
// Copyright 2024 IOTA Stiftung
```

bindings/nodejs/lib/types/wallet/create-delegation-transaction.ts
```diff
@@ -0,0 +1,39 @@
...
+        super(preparedData.transaction, wallet);
+        this._delegationId = preparedData.delegationId;
+    }
+
+    /**
+     * The function returns the delegationId.
+     *
+     * Returns:
+     *
+     * The delegation id of the prepared transaction.
```
Thoralf-M:
```suggestion
     * @returns: The delegation id of the prepared transaction.
```

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
DaughterOfMars: I made this happen in the `DelegationOutputBuilder::finish` method. I don't know if it's correct to do that, but I can't see a scenario where these values are not the same 🤷‍♀️

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
DaughterOfMars: not sure what you mean

sdk/src/wallet/operations/transaction/high_level/delegation/mod.rs
```diff
@@ -0,0 +1,6 @@
+// Copyright 2024 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+pub(crate) mod create;
+mod delay;
```
thibault-martinez, line 5: Why the difference in privacy?

sdk/src/types/block/slot/index.rs
```diff
@@ -77,6 +77,20 @@ impl SlotIndex {
...
+    pub fn past_bounded_slot(self, max_commitable_age: u32) -> SlotIndex {
+        Self(self.0 + max_commitable_age)
+    }
+
+    /// Calculates the future bounded slot for the given slot.
+    /// Given any slot index of a commitment input, the result of this function is a slot index
+    /// that is at most equal to the slot of the block in which it was issued, or lower.
+    /// That means no commitment input can be chosen such that the index lies ahead of the slot index of the block,
+    /// hence the future is bounded.
+    pub fn future_bounded_slot(self, min_commitable_age: u32) -> SlotIndex {
```
thibault-martinez: These have been moved to SlotCommitmentId, remove them

thibault-martinez: Oh it's because we haven't merged the Delegation checks PR

sdk/src/wallet/operations/transaction/high_level/delegation/mod.rs
DaughterOfMars: One has a struct that is exported above.

sdk/src/types/block/slot/index.rs
DaughterOfMars: yep

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
Thoralf-M: I think we might want to have the address of an account output as remainder instead of the ed25519 address that controls the account

sdk/src/types/block/output/delegation.rs
```diff
@@ -187,9 +187,17 @@ impl DelegationOutputBuilder {
             unlock_conditions,
         };
 
-        output.amount = match self.amount {
-            OutputBuilderAmount::Amount(amount) => amount,
-            OutputBuilderAmount::MinimumAmount(params) => output.minimum_amount(params),
+        match self.amount {
+            OutputBuilderAmount::Amount(amount) => output.amount = amount,
```
Thoralf-M: Should `delegated_amount` be removed from the new method and be set here as well? Or should it accept `OutputBuilderAmount` instead of `u64` and the `with_amount()/with_minimum_amount()`  methods be removed?
Because `delegated_amount` gets set in the `new()` method of the builder and amount can be set to a different value in `with_amount()` so they can end up with different values

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
Thoralf-M: Ah missed that

sdk/src/wallet/operations/transaction/high_level/delegation/mod.rs
thibault-martinez: Ah I see, thanks

sdk/src/client/api/types.rs
```diff
@@ -74,6 +74,15 @@ impl TryFromDto<PreparedTransactionDataDto> for PreparedTransactionData {
     }
 }
 
+impl Serialize for PreparedTransactionData {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        PreparedTransactionDataDto::from(self).serialize(serializer)
```
thibault-martinez, line 83: We still need the DTO because we can't deserialize ?

bindings/nodejs/lib/types/wallet/index.ts
```diff
@@ -14,4 +14,6 @@ export * from './signed-transaction-data';
 export * from './transaction';
 export * from './transaction-options';
 export * from './prepared-create-token-transaction-data';
+export * from './create-delegation-transaction';
+export * from './create-delegation-transaction-data';
```
thibault-martinez: Does this really need to be two files ?

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
DaughterOfMars: do you think it should go like this then?
1. provided address
2. first wallet account address
3. wallet ed25519

bindings/nodejs/lib/types/wallet/index.ts
DaughterOfMars: No. I was kind of just mimicking the native token files. Shall I change both?

sdk/src/types/block/output/delegation.rs
DaughterOfMars: tbh I don't know. I think it depends a lot on whether the amounts should always match

sdk/src/client/api/types.rs
DaughterOfMars: Yeah, unfortunately

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    #[test]
+    fn create_delegation_params_serde() {
+        let params_none_1 = CreateDelegationParams {
+            address: None,
+            delegated_amount: 100,
+            validator_address: rand_account_address(),
+        };
+        let json_none = serde_json::to_string(&params_none_1).unwrap();
+        println!("{json_none}");
```
Thoralf-M:
```suggestion
```

sdk/src/types/block/output/delegation.rs
Thoralf-M: Ah right, during transition it can be different, but for creation they must match, otherwise it's invalid

Thoralf-M: So I think we shouldn't allow to set the delegated amount, first time it should be set to the amount and delayed claiming it must stay the same as for creation

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
Thoralf-M: Yes, I just opened an issue for this, so we can resolve it here https://github.com/iotaledger/iota-sdk/issues/1888

sdk/src/types/block/output/delegation.rs
DaughterOfMars: doesn't that mean it's always the same as the amount?

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
DaughterOfMars: lol whoops

sdk/src/types/block/output/delegation.rs
Thoralf-M: For delayed claiming, the amount can be reduced, but the delegated amount field must still stay as before

bindings/nodejs/lib/types/wallet/index.ts
DaughterOfMars: I changed them

sdk/src/types/block/output/delegation.rs
DaughterOfMars: then we must be able to construct the output with a lower amount :/

Thoralf-M: True

Thoralf-M: Sorry, will you revert or just change it a bit different?

sdk/src/wallet/operations/transaction/high_level/delegation/delay.rs
```diff
@@ -0,0 +1,134 @@
...
+        reclaim_excess: bool,
+    ) -> crate::wallet::Result<TransactionWithMetadata> {
+        let prepared_transaction = self
+            .prepare_delay_delegation_claiming(delegation_id, reclaim_excess)
+            .await?;
+
+        self.sign_and_submit_transaction(prepared_transaction, None, None).await
+    }
+
+    pub async fn prepare_delay_delegation_claiming(
```
Thoralf-M, line 36: Please add some doc comments that explain what this is doing with the reclaim_excess bool

sdk/src/types/block/slot/epoch.rs
```diff
@@ -74,6 +74,12 @@ impl EpochIndex {
     pub fn last_slot_index(self, slots_per_epoch_exponent: u8) -> SlotIndex {
         SlotIndex::from_epoch_index(self + 1, slots_per_epoch_exponent) - 1
     }
+
+    /// Returns the slot at the end of which the validator and delegator registration ends and the voting power
+    /// for the epoch with index epoch + 1 is calculated.
+    pub fn registration_slot(&self, slots_per_epoch_exponent: u8, epoch_nearing_threshold: u32) -> SlotIndex {
```
thibault-martinez: Reminder to remove when #1421 is merged

sdk/src/wallet/operations/transaction/high_level/delegation/delay.rs
DaughterOfMars: done

sdk/src/types/block/output/delegation.rs
```diff
@@ -179,18 +171,40 @@ impl DelegationOutputBuilder {
...
+                    amount
+                };
+            }
+            OutputBuilderAmount::MinimumAmount(params) => {
+                let min = output.minimum_amount(params);
+                output.delegated_amount = min;
+                output.amount = if let Some(builder_amount) = self.amount {
+                    if let OutputBuilderAmount::Amount(amount) = builder_amount {
+                        amount
+                    } else {
```
Alex6323, line 201: we allow the user to set it to something smaller than `min`?

DaughterOfMars: yeah, technically you can do this with all the outputs, but of course it will not be accepted

kwek20 (commented):
I think we decided on not updating the copyright headers a few months ago btw

bindings/python/iota_sdk/types/send_params.py
```diff
@@ -83,6 +84,23 @@ class CreateNativeTokenParams():
     account_id: Optional[str] = None
 
 
+@json
+@dataclass
+class CreateDelegationParams():
+    """Parameters for creating native tokens.
+
+    Attributes:
+        address: Bech32 encoded address which will control the delegation. By default, the ed25519 wallet address will be used.
```
kwek20: Order :(

sdk/src/wallet/operations/transaction/high_level/delegation/delay.rs
```diff
@@ -0,0 +1,143 @@
...
+                let min_basic_amount = BasicOutput::minimum_amount(
+                    delegation_output.output.as_delegation().address(),
+                    protocol_parameters.storage_score_parameters(),
+                );
+                delegation_output.output.amount() >= min_delegation_amount + min_basic_amount
+            } else {
+                false
+            };
+
+            // TODO: Replace with methods from https://github.com/iotaledger/iota-sdk/pull/1421
```
Thoralf-M: That's merged now

sdk/src/wallet/operations/transaction/high_level/delegation/delay.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// over the minimum storage requirements will be moved to a basic output that is unlockable by the same address
+    /// which controls the delegation.
+    /// Otherwise it will be added to a new delegation. In this manner, one can delegate for one epoch at a time and
+    /// never lose out on any rewards.
+    pub async fn prepare_delay_delegation_claiming(
+        &self,
+        delegation_id: DelegationId,
+        reclaim_excess: bool,
+    ) -> crate::wallet::Result<PreparedTransactionData> {
+        if let Some(delegation_output) = self.get_delegation_output(delegation_id).await {
```
Thoralf-M: I would prefer to have this and have the else block at the end (L137) removed
```suggestion
        let delegation_output =
            self.get_delegation_output(delegation_id)
                .await
                .ok_or(crate::wallet::Error::DelegationTransitionFailed(format!(
                    "no delegation output found with id {delegation_id}"
                )))?;
```

bindings/python/iota_sdk/types/send_params.py
DaughterOfMars: fixed

sdk/src/wallet/operations/transaction/high_level/delegation/delay.rs
DaughterOfMars: fixed

sdk/src/wallet/operations/transaction/high_level/delegation/delay.rs
DaughterOfMars: thanks, done

bindings/core/src/method/wallet.rs
```diff
@@ -318,6 +318,17 @@ pub enum WalletMethod {
...
+        params: CreateDelegationParams,
+        options: Option<TransactionOptions>,
+    },
+    #[serde(rename_all = "camelCase")]
+    PrepareDelayDelegationClaiming {
+        delegation_id: DelegationId,
+        reclaim_excess: bool,
+    },
+    #[serde(rename_all = "camelCase")]
+    PrepareDestroyDelegation { delegation_id: DelegationId },
```
Thoralf-M: Please also add comments here like we have for all the others

bindings/nodejs/lib/types/wallet/create-delegation-transaction.ts
```diff
@@ -0,0 +1,54 @@
...
+/** The result of preparing an operation to create a native token. */
+export class PreparedCreateDelegationTransactionData {
+    /** The id of the delegation */
+    delegationId!: DelegationId;
+    /** The prepared transaction which will create the delegation */
+    @Type(() => PreparedTransactionData)
+    transaction!: PreparedTransactionData;
+}
+
+/** The result of an operation to create a native token. */
```
Thoralf-M:
```suggestion
/** The result of preparing an operation to create a delegation. */
export class PreparedCreateDelegationTransactionData {
    /** The id of the delegation */
    delegationId!: DelegationId;
    /** The prepared transaction which will create the delegation */
    @Type(() => PreparedTransactionData)
    transaction!: PreparedTransactionData;
}

/** The result of an operation to create a delegation. */
```

bindings/core/src/method/wallet.rs
DaughterOfMars: done

sdk/src/types/block/slot/epoch.rs
DaughterOfMars: done

bindings/python/iota_sdk/types/send_params.py
```diff
@@ -83,6 +84,23 @@ class CreateNativeTokenParams():
     account_id: Optional[str] = None
 
 
+@json
+@dataclass
+class CreateDelegationParams():
+    """Parameters for creating native tokens.
```
Thoralf-M:
```suggestion
    """Parameters for creating a delegation output.
```

bindings/python/iota_sdk/wallet/prepared_transaction.py
```diff
@@ -55,32 +75,79 @@ def sign(self):
...
+    """
+    token_id: HexStr
+    transaction: PreparedTransactionData
 
-    """A prepared transaction for creating a native token.
 
-    Returns: The token id of the PreparedCreateTokenTransaction.
+@dataclass
+class PreparedCreateDelegationTransaction:
+    """A helper class for offline signing a create native token transaction.
```
Thoralf-M:
```suggestion
    """A helper class for offline signing to create a delegation transaction.
```

bindings/python/iota_sdk/wallet/prepared_transaction.py
```diff
@@ -55,32 +75,79 @@ def sign(self):
...
+            self.prepared_transaction_data.transaction)
+        CreateNativeTokenTransaction(
+            self.prepared_transaction_data.token_id, tx)
 
 
-class PreparedCreateTokenTransaction(PreparedTransaction):
+@json
+@dataclass
+class PreparedCreateTokenTransactionData:
+    """Prepared transaction data for creating a native token.
```
Thoralf-M, line 95:
```suggestion
    """Prepared transaction data for creating a delegation.
```

sdk/src/types/block/protocol/mod.rs
```diff
@@ -263,6 +263,62 @@ impl ProtocolParameters {
...
+    /// for the epoch is calculated.
+    pub fn registration_slot(&self, epoch_index: EpochIndex) -> SlotIndex {
+        epoch_index.registration_slot(
+            self.genesis_slot(),
+            self.slots_per_epoch_exponent(),
+            self.epoch_nearing_threshold(),
+        )
+    }
+
+    /// TODO what is a good name + description for these methods?
```
Thoralf-M: Maybe something like `delegation_start/end_epoch` or is this also interesting for something else?

sdk/src/wallet/operations/transaction/high_level/delegation/destroy.rs
```diff
@@ -0,0 +1,42 @@
...
+                "no delegation output found with id {delegation_id}"
+            ))
+        })?;
+        self.prepare_transaction(
+            [],
+            TransactionOptions {
+                custom_inputs: Some(vec![delegation_output.output_id]),
+                ..Default::default()
+            },
+        )
```
Thoralf-M: What did we say about the mana rewards, where will they be calculated?

sdk/src/wallet/operations/transaction/high_level/delegation/delay.rs
```diff
@@ -0,0 +1,110 @@
...
+        // Add a commitment context input with the latest slot commitment
+        let slot_commitment_id = self.client().get_info().await?.node_info.status.latest_commitment_id;
+        let context_input = ContextInput::from(CommitmentContextInput::new(slot_commitment_id));
+
+        let min_delegation_amount = delegation_output
+            .output
+            .minimum_amount(protocol_parameters.storage_score_parameters());
+
+        // In order to split the output, the amount must be at least twice the minimum for a delegation output
+        let can_split = delegation_output.output.amount() >= 2 * min_delegation_amount;
```
Thoralf-M: Should be `min_delegation_amount + min_basic_amount` instead?
So maybe just put the `min_basic_amount` definition from L63 above and use this value?

sdk/src/types/block/protocol/mod.rs
DaughterOfMars: Not that I'm aware of

sdk/src/wallet/operations/transaction/high_level/delegation/destroy.rs
DaughterOfMars: I think I can add it quickly actually

sdk/src/wallet/operations/transaction/high_level/delegation/delay.rs
DaughterOfMars: No because this will create two delegation outputs

Thoralf-M: Ah true, was confused by the `BasicOutput::minimum_amount()`, how about moving this line to L75?

DaughterOfMars: moved

Thoralf-M: Thanks

sdk/src/types/block/protocol/mod.rs
```diff
@@ -263,6 +263,62 @@ impl ProtocolParameters {
             max: self.max_committable_age(),
         }
     }
+
+    /// Calculates the past bounded slot for the given slot of the SlotCommitment.
+    /// Given any slot index of a commitment input, the result of this function is a slot index
+    /// that is at least equal to the slot of the block in which it was issued, or higher.
+    /// That means no commitment input can be chosen such that the index lies behind the slot index of the block,
+    /// hence the past is bounded.
+    pub fn past_bounded_slot(&self, slot_commitment_id: SlotCommitmentId) -> SlotIndex {
```
thibault-martinez, line 272: I would remove this and `future_bounded_slot`, they were move from here to `SlotCommitmentId` in the other PR, there is not much point having these

DaughterOfMars: I just prefer using this syntax
```rust
protocol_parameters.past_bounded_slot(slot_commitment_id)
```
rather than this
```rust
slot_commitment_id.past_bounded_slot(protocol_parameters.max_commitable_age())
```
and it matches the other usages as well

thibault-martinez: Yeah but it's a property of the CommitmentId, doesn't really exist outside of it. Whatever you end up choosing, I would remove one impl, doesn't make sense to have both.

DaughterOfMars: tbh I do think it makes sense to have both. the one on the type outlines the functionality, while the one on the params is a convenience function that ensures that you don't misuse the values (like passing in min_committable_age when the fn wants max)

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
```diff
@@ -0,0 +1,201 @@
...
+        let transaction = self.prepare_transaction([output], options).await?;
+
+        Ok(PreparedCreateDelegationTransaction {
+            delegation_id: DelegationId::from(&transaction.transaction.id().into_output_id(0)),
+            transaction,
+        })
+    }
+
+    /// Gets an existing delegation output.
+    pub(crate) async fn get_delegation_output(&self, delegation_id: DelegationId) -> Option<OutputData> {
```
thibault-martinez: We already have this, it's called `unspent_delegation_output`

sdk/src/types/block/protocol/mod.rs
thibault-martinez: Well if you insist. I just don't think we need convenience on such specific and low level computations at all. Just creates maintenance burden to duplicate.

bindings/nodejs/lib/types/wallet/create-delegation-transaction.ts
```diff
@@ -0,0 +1,54 @@
...
+import { PreparedTransaction } from './prepared-transaction';
+import { Type } from 'class-transformer';
+import { TransactionWithMetadata } from './transaction';
+import { PreparedTransactionData } from '../client/prepared-transaction-data';
+
+/*
+ * Represents prepared data for issuing a transaction to create a delegation.
+ */
+export class PreparedCreateDelegationTransaction extends PreparedTransaction {
+    readonly _delegationId: DelegationId;
```
thibault-martinez, line 14: does that work with serialisation ?

DaughterOfMars: Well this type doesn't get serialized

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
DaughterOfMars: changed

bindings/nodejs/lib/types/wallet/create-delegation-transaction.ts
thibault-martinez: Hmm ok, it's a bit confusing to me why we have both PreparedCreateDelegationTransaction and PreparedCreateDelegationTransactionData tbh

DaughterOfMars: one is a helper class which contains the wallet reference, so you can just call `send()` on it, and thus it cannot be deserialized

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
```diff
@@ -0,0 +1,182 @@
...
+/// Params for `create_delegation_output()`
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CreateDelegationParams {
+    /// Bech32 encoded address which will control the delegation.
+    /// By default, the ed25519 wallet address will be used.
+    // TODO: https://github.com/iotaledger/iota-sdk/issues/1888
+    pub address: Option<Bech32Address>,
+    /// The amount to delegate.
+    pub delegated_amount: u64,
```
thibault-martinez, line 24: IMO it's a tiny bit redundant to have `delegated_` here as what we primarily set is the amount and then delegated_amount is set to the same, and it's obviously delegated since it's in this struct

DaughterOfMars: I felt that it was clearer to indicate it this way, since I think of it more like we set `amount` to equal `delegated_amount` rather than the other way around

thibault-martinez: The opposite is closer to the TIP

DaughterOfMars: But the amount field can change whereas the delegation amount cannot. It just makes more sense to say that you're setting the delegation amount which implies the amount and later you might reduce the amount

bindings/python/iota_sdk/types/transaction_with_metadata.py
```diff
@@ -50,3 +50,29 @@ class TransactionWithMetadata:
...
+        token_id: The id of the created native token.
+        transaction: The transaction that created the native token.
+    """
+    token_id: HexStr
+    transaction: TransactionWithMetadata
+
+
+@json
+@dataclass
+class CreateDelegationTransaction:
```
thibault-martinez, line 70: Not really sure it should be in this file...

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
```diff
@@ -15,7 +15,6 @@ impl InputSelection {
                 input.output_id().transaction_id().slot_index(),
                 self.slot_index,
             )?;
-            // TODO rewards https://github.com/iotaledger/iota-sdk/issues/1310
```
thibault-martinez, line 18: Why are we not handling rewards here ?

sdk/src/client/api/block_builder/input_selection/remainder.rs
```diff
@@ -110,15 +110,14 @@ impl InputSelection {
 
         let native_tokens_diff = get_native_tokens_diff(&input_native_tokens, &output_native_tokens)?;
 
-        let mut input_mana = 0;
+        let mut input_mana = self.mana_rewards;
```
thibault-martinez: Isn't that dependent on the associated inputs to actually be burnt? Why do we add it always?

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
DaughterOfMars: there's just nothing needed here, or am I missing something?

sdk/src/client/api/block_builder/input_selection/remainder.rs
DaughterOfMars: No, the mana_rewards value is the value of rewards you are trying to claim

bindings/nodejs/lib/types/wallet/transaction-options.ts
```diff
@@ -90,3 +90,13 @@ export interface AccountOutputParams {
...
 }
+
+/** Options for delegation output creation */
+export interface CreateDelegationParams {
+    /** Bech32 encoded address which will control the delegation. By default, the ed25519 wallet address will be used. */
+    address?: Bech32Address;
+    /** The amount to delegate. */
+    delegatedAmount: NumericString;
+    /** The Account Address of the validator to which this output will delegate. */
+    validatorAddress: HexEncodedString;
```
thibault-martinez: Can/should it not be `AccountAddress` like in Rust and Python?

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
thibault-martinez: Shouldn't rewards contribute to meet the required output mana ?

sdk/src/types/block/protocol/mod.rs
thibault-martinez: I'll resolve but again, I think this is a bad idea

sdk/src/wallet/error.rs
```diff
@@ -37,6 +37,8 @@ pub enum Error {
     /// Custom input error
     #[error("custom input error {0}")]
     CustomInput(String),
+    #[error("delay delegation claiming failed {0}")]
```
thibault-martinez: Not only used in delayed claiming

sdk/src/wallet/error.rs
```diff
@@ -37,6 +37,8 @@ pub enum Error {
...
+    DelegationTransitionFailed(String),
```
thibault-martinez: Also, you only use it for delegation output not found, maybe make it a bit more specific to it

bindings/nodejs/lib/wallet/wallet.ts
```diff
@@ -1223,6 +1228,145 @@ export class Wallet {
         );
     }
 
+    /**
+     * Create a delegation.
+     *
+     * @param params The options for creating a delegation.
+     * @param transactionOptions Additional transaction options
+     * or custom inputs.
```
thibault-martinez: Does this really not fit on a line?

bindings/nodejs/lib/wallet/wallet.ts
```diff
@@ -1223,6 +1228,145 @@ export class Wallet {
...
+
+    /**
+     * Delay a delegation's claiming.
+     *
+     * @param delegationId The id of the delegation to delay.
+     * @returns The created transaction.
+     */
+    async delayDelegationClaiming(
+        delegationId: DelegationId,
+        reclaimExcess: boolean,
```
thibault-martinez, line 1280: missing doc

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
DaughterOfMars: They are added to the input total, and that causes them to be put in a remainder

bindings/python/iota_sdk/wallet/wallet.py
```diff
@@ -595,12 +594,61 @@ def prepare_send(self, params: List[SendParams],
...
+        ))
+        return PreparedTransaction(self, prepared)
+
+    def destroy_delegation(self, delegation_id: HexStr) -> TransactionWithMetadata:
+        """Destroy a delegation.
+        """
+        return self.prepare_destroy_delegation(delegation_id).send()
+
+    def prepare_destroy_delegation(self, delegation_id: HexStr) -> PreparedTransaction:
+        """Prepare to delay a delegation's claiming.
```
thibault-martinez: oopsie

sdk/src/wallet/error.rs
DaughterOfMars: true

bindings/python/iota_sdk/wallet/wallet.py
DaughterOfMars: oopsie

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
thibault-martinez: But that's a completely different process? Here we are trying to find inputs to match the requirement, if we don't consider rewards we may have to select more inputs than needed?

bindings/nodejs/lib/types/wallet/transaction-options.ts
DaughterOfMars: turns out it can :)

bindings/python/iota_sdk/types/transaction_with_metadata.py
DaughterOfMars: why not?

sdk/src/client/api/block_builder/input_selection/remainder.rs
Thoralf-M: But ISA doesn't know which output it has to select for this to work, so it only works if you have luck (because it selected the input that makes the rewards available) or if you provide these inputs as required inputs?

bindings/python/iota_sdk/types/transaction_with_metadata.py
thibault-martinez: Name suggest it's juste for the TransactionWithMetadata type, could make it a bit more generic if you want them all together. But I'd keep TransactionWithMetadata separate tbh

sdk/src/client/api/block_builder/input_selection/remainder.rs
thibault-martinez: Yeah but actually I think it's fine because she only adds rewards if the delegations are in custom/mandatory inputs, right Chloé?

DaughterOfMars: indeed

bindings/nodejs/lib/wallet/wallet.ts
thibault-martinez: There is more of it if ever you want to change them all...

sdk/src/client/api/block_builder/input_selection/remainder.rs
DaughterOfMars: actually, no, it just gets added always 😆

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -228,6 +230,12 @@ impl InputSelection {
         self
     }
 
+    /// Sets the total mana rewards.
```
Thoralf-M: Does this make sense then?
```suggestion
    /// Sets the total mana rewards for required inputs.
```

sdk/src/wallet/operations/transaction/high_level/delegation/destroy.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub async fn prepare_destroy_delegation(
+        &self,
+        delegation_id: DelegationId,
+    ) -> crate::wallet::Result<PreparedTransactionData> {
+        let delegation_output = self
+            .data()
+            .await
+            .unspent_delegation_output(&delegation_id)
+            .ok_or(crate::wallet::Error::MissingDelegation(delegation_id))?
+            .clone();
```
thibault-martinez: avoid the clone by just taking the ID?

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
```diff
@@ -0,0 +1,182 @@
...
+
+        let delegation_output_builder = DelegationOutputBuilder::new_with_amount(
+            params.delegated_amount,
+            DelegationId::null(),
+            params.validator_address,
+        )
+        .with_start_epoch(protocol_parameters.delegation_start_epoch(slot_commitment_id))
+        .add_unlock_condition(AddressUnlockCondition::new(address.clone()));
+
+        let output = delegation_output_builder.finish_output()?;
```
Thoralf-M: Any reason to do not build it directly?
```suggestion
        let output = DelegationOutputBuilder::new_with_amount(
            params.delegated_amount,
            DelegationId::null(),
            params.validator_address,
        )
        .with_start_epoch(protocol_parameters.delegation_start_epoch(slot_commitment_id))
        .add_unlock_condition(AddressUnlockCondition::new(address.clone()))
        .finish_output()?;
```

DaughterOfMars: done

sdk/src/wallet/operations/transaction/high_level/delegation/destroy.rs
DaughterOfMars: done

bindings/nodejs/lib/wallet/wallet.ts
DaughterOfMars: \>:(

Thoralf-M (dismissed):

bindings/python/iota_sdk/types/transaction_with_metadata.py
DaughterOfMars: the CreateNativeTokenTransaction is there too

DaughterOfMars: and these are just wrappers

thibault-martinez: Well it's here because you've added it, but it's fine, I'll resolve

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
DaughterOfMars: But we don't need to select inputs to get the reward mana, it is provided in a totally different way

thibault-martinez: Yes but the fact that the delegation are already selected gives you the reward and can prevent you from selecting other inputs that wouldn't have been required

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
```diff
@@ -0,0 +1,181 @@
...
+        let slot_commitment_id = if let Some(commitment_context) = options
+            .iter()
+            .flat_map(|opts| opts.context_inputs.iter())
+            .flat_map(|c| c.iter())
+            .find_map(|c| c.as_commitment_opt())
+        {
+            commitment_context.slot_commitment_id()
+        } else {
+            // Add a commitment context input with the latest slot commitment
+            let latest_id = self.client().get_info().await?.node_info.status.latest_commitment_id;
```
thibault-martinez: It is recommended we use issuance for this instead

thibault-martinez: And maybe it should be added in `build_transaction` with the others instead

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
thibault-martinez: I don't necessarily think it makes more sense and I prefer us strictly following the TIP but alright if you prefer

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
DaughterOfMars: fixed

thibault-martinez: danke dir

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
DaughterOfMars: you mean we should not set the value here and instead do it in `build_transaction`?

thibault-martinez: Yeah, I think we could/should centralise the context input in one place, but just put a TODO if you want us to address it later

thibault-martinez: You can ref this https://github.com/iotaledger/iota-sdk/issues/1924

sdk/src/wallet/operations/transaction/high_level/delegation/delay.rs
```diff
@@ -0,0 +1,111 @@
...
+        let delegation_output = self
+            .data()
+            .await
+            .unspent_delegation_output(&delegation_id)
+            .ok_or(crate::wallet::Error::MissingDelegation(delegation_id))?
+            .clone();
+        let protocol_parameters = self.client().get_protocol_parameters().await?;
+        // Add a commitment context input with the latest slot commitment
+        let slot_commitment_id = self.client().get_info().await?.node_info.status.latest_commitment_id;
+        let context_input = ContextInput::from(CommitmentContextInput::new(slot_commitment_id));
```
thibault-martinez: TODO ref https://github.com/iotaledger/iota-sdk/issues/1924

sdk/src/wallet/operations/transaction/input_selection.rs
thibault-martinez, line 1: I guess you now also need to handle reward if a burn has been set?

DaughterOfMars: wish granted

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -420,6 +430,10 @@ impl InputSelection {
 
         self.validate_transitions()?;
 
+        // Only keep mana rewards for selected inputs
```
thibault-martinez: But why do we have more to start with?

DaughterOfMars: We shouldn't, but the user could pass in whatever so I figured I'd check

Thoralf-M: Wouldn't it be better to return an error in this case?

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
```diff
@@ -0,0 +1,149 @@
...
+impl<S: 'static + SecretManage> Wallet<S>
+where
+    crate::wallet::Error: From<S::Error>,
+    crate::client::Error: From<S::Error>,
+{
+    /// Creates a delegation output.
+    /// ```ignore
+    /// let params = CreateDelegationParams {
+    ///     address: None,
+    ///     delegated_amount: 200,
```
Thoralf-M: Maybe 2_000_000 would be a more realistic value here, otherwise one just gets an error because it doesn't cover the storage deposit

sdk/src/wallet/operations/transaction/high_level/delegation/delay.rs
```diff
@@ -0,0 +1,96 @@
...
+            .await
+            .unspent_delegation_output(&delegation_id)
+            .ok_or(crate::wallet::Error::MissingDelegation(delegation_id))?
+            .output
+            .clone();
+        let protocol_parameters = self.client().get_protocol_parameters().await?;
+
+        let min_delegation_amount = delegation_output.minimum_amount(protocol_parameters.storage_score_parameters());
+
+        // TODO: Should we return an error if `reclaim_excess == true` and `can_reclaim == false`?
```
Thoralf-M: I think that makes sense to return an error if that doesn't work out, because if one provides `reclaim_excess == true` one probably expect that it delegates all the time
And if we notice that this error gets annoying because it happens often and we would actually be fine without, we an also remove again

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -228,14 +232,27 @@ impl InputSelection {
...
     fn filter_inputs(&mut self) {
         self.available_inputs.retain(|input| {
             // TODO what about other kinds?
-            // Filter out non basic/account/foundry/nft outputs.
+            // Filter out non basic/account/foundry/nft/delegation outputs.
             if !input.output.is_basic()
                 && !input.output.is_account()
                 && !input.output.is_foundry()
                 && !input.output.is_nft()
+                && !input.output.is_delegation()
```
thibault-martinez: This is annoying, I don't think delegation should be generally available. If they are not explicitly required/burnt, I don't think they should be potentially selected for other requirements like amount etc

sdk/src/types/block/output/account.rs
```diff
@@ -388,6 +388,12 @@ impl AccountOutput {
         self.features.block_issuer().is_some()
     }
 
+    /// Returns whether the output can claim rewards based on its current and next state in a transaction.
+    pub fn can_claim_rewards(&self, next_state: Option<&Self>) -> bool {
+        // TODO: should destroying an account add rewards?
+        self.features().staking().is_some() && next_state.map_or(false, |o| o.features().staking().is_none())
```
thibault-martinez: What if it's 2 different accounts ?

thibault-martinez: Or we just expect this to be called with the right input/output?

sdk/tests/client/input_selection/delegation_outputs.rs
```diff
@@ -0,0 +1,100 @@
...
+        },
+        slot::SlotIndex,
+    },
+};
+use pretty_assertions::assert_eq;
+
+use crate::client::{BECH32_ADDRESS_ED25519_0, SLOT_INDEX};
+
+#[test]
+fn remainder_needed_for_mana() {
```
thibault-martinez, line 28: The remainder being generated mana ?

sdk/src/client/api/block_builder/input_selection/mod.rs
DaughterOfMars: I suppose

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
DaughterOfMars: fair

sdk/src/wallet/operations/transaction/high_level/delegation/delay.rs
DaughterOfMars: okie dokie

sdk/src/client/api/block_builder/input_selection/mod.rs
DaughterOfMars: okay I can remove it

sdk/src/types/block/output/account.rs
DaughterOfMars: yeah that

sdk/tests/client/input_selection/delegation_outputs.rs
DaughterOfMars: in this case it's because of the reward mana

sdk/src/types/block/output/account.rs
thibault-martinez: Ok fair enough

sdk/tests/client/input_selection/delegation_outputs.rs
thibault-martinez: Well, 100 stored + 100 reward = 200, only potential is not explicitly mentioned. But I guess it's different PoVs. Just wanted to know what was intended but in the end it's the same thing.

sdk/src/wallet/operations/transaction/input_selection.rs
```diff
@@ -78,16 +78,46 @@ where
...
+                if wallet_data.locked_outputs.contains(output_id) {
                     return Err(crate::wallet::Error::CustomInput(format!(
-                        "provided custom input {input} is already used in another transaction",
+                        "provided custom input {output_id} is already used in another transaction",
                     )));
                 }
+                if let Some(input) = wallet_data.outputs.get(output_id) {
+                    if input
+                        .output
+                        .can_claim_rewards(outputs.iter().find(|o| input.output.chain_id() == o.chain_id()))
```
thibault-martinez: Maybe should use `or_from_output_id` in the input side

sdk/src/wallet/operations/transaction/input_selection.rs
```diff
@@ -122,12 +153,26 @@ where
...
```
thibault-martinez: same

sdk/src/wallet/operations/transaction/build_transaction.rs
```diff
@@ -48,6 +52,34 @@ where
             }
 
             inputs.push(Input::Utxo(UtxoInput::from(*input.output_id())));
+
+            if let Some(reward) = selected_transaction_data.mana_rewards.get(input.output_id()) {
+                mana_rewards += *reward;
+                context_inputs.insert(ContextInput::from(RewardContextInput::new(idx as _)?));
+                if !context_inputs.iter().any(|c| c.kind() == CommitmentContextInput::KIND) {
```
thibault-martinez: Can you fit that into L110?

thibault-martinez: Avoid loops and better to have it in a single place

sdk/src/wallet/operations/transaction/build_transaction.rs
```diff
@@ -36,8 +36,12 @@ where
 
         let mut inputs: Vec<Input> = Vec::new();
         let mut context_inputs = HashSet::new();
+        let mut mana_rewards = 0;
```
thibault-martinez, line 40: What's the point of this?

DaughterOfMars: it accumulates the total mana rewards for semantic validation (which isn't part of this PR)

sdk/src/wallet/operations/transaction/build_transaction.rs
DaughterOfMars: I'd rather have a variable like in the other PR

sdk/src/client/api/block_builder/input_selection/mod.rs
DaughterOfMars: done

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
DaughterOfMars: done

sdk/src/wallet/operations/transaction/high_level/delegation/delay.rs
DaughterOfMars: I ended up refactoring so that this error is not possible

sdk/src/wallet/operations/transaction/build_transaction.rs
DaughterOfMars: fixed it

sdk/src/wallet/operations/transaction/build_transaction.rs
DaughterOfMars: added a comment

sdk/src/wallet/operations/transaction/build_transaction.rs
```diff
@@ -48,6 +55,30 @@ where
...
+            if output.as_delegation().delegation_id().is_null() {
+                *output = DelegationOutputBuilder::from(output.as_delegation())
+                    .with_start_epoch(protocol_parameters.delegation_start_epoch(latest_slot_commitment_id))
+                    .finish_output()?;
+            } else {
+                *output = DelegationOutputBuilder::from(output.as_delegation())
+                    .with_end_epoch(protocol_parameters.delegation_end_epoch(latest_slot_commitment_id))
+                    .finish_output()?;
+            }
+            needs_commitment_context = true;
```
thibault-martinez, line 82: creation AND transition need one, yeah ?

sdk/src/wallet/operations/transaction/build_transaction.rs
```diff
@@ -48,6 +55,30 @@ where
...
+                context_inputs.insert(ContextInput::from(RewardContextInput::new(idx as _)?));
+                needs_commitment_context = true;
+            }
+        }
+
+        for output in selected_transaction_data
+            .outputs
+            .iter_mut()
+            .filter(|o| o.is_delegation())
+        {
```
thibault-martinez, line 71: Honestly I don't think this whole block should be here. I get it that you need the latest_slot_commitment_id and here is the ideal place but later when we refactor the block options and have issuance set as first request, I think we should move this to its relevant place. But we'll see that later.

sdk/src/types/block/output/account.rs
```diff
@@ -388,6 +388,12 @@ impl AccountOutput {
         self.features.block_issuer().is_some()
     }
 
+    /// Returns whether the output can claim rewards based on its current and next state in a transaction.
+    pub fn can_claim_rewards(&self, next_state: Option<&Self>) -> bool {
+        // TODO: should destroying an account add rewards?
```
thibault-martinez: Not super clear to me, I'll ask tomorrow

thibault-martinez: Actually, TIP says the following so I guess so

- All of the following conditions hold:
        - The _Staking Feature_ is removed on the output side of the transaction or the account is destroyed.
        - A _Reward Input_ must be present whose `Index` references the account input.
        - The Mana amount on the input side of the transaction is increased by at most the amount defined in
          [TIP-40 (Validator Rewards)](../TIP-0040/tip-0040.md#validator-rewards).

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
```diff
@@ -0,0 +1,149 @@
...
+            }
+            None => self.address().await.inner().clone(),
+        };
+
+        let output = DelegationOutputBuilder::new_with_amount(
+            params.delegated_amount,
+            DelegationId::null(),
+            params.validator_address,
+        )
+        .add_unlock_condition(AddressUnlockCondition::new(address.clone()))
```
thibault-martinez: clone not needed

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
```diff
@@ -0,0 +1,149 @@
...
+        .finish_output()?;
+
+        let transaction = self.prepare_transaction([output], options).await?;
+
+        Ok(PreparedCreateDelegationTransaction {
```
thibault-martinez, line 107: Why do we need to return the delegationID?

thibault-martinez: We don't do it for the other chains

sdk/src/wallet/operations/transaction/build_transaction.rs
DaughterOfMars: I believe so

sdk/src/wallet/operations/transaction/build_transaction.rs
DaughterOfMars: okie

sdk/src/types/block/output/account.rs
DaughterOfMars: oh okay

sdk/src/wallet/operations/transaction/high_level/delegation/create.rs
DaughterOfMars: just because it's annoying to get otherwise? I dunno

sdk/src/types/block/output/delegation.rs
```diff
@@ -248,22 +262,20 @@ impl DelegationOutput {
     /// Creates a new [`DelegationOutputBuilder`] with a provided amount.
     pub fn build_with_amount(
         amount: u64,
-        delegated_amount: u64,
```
thibault-martinez, line 251: Honestly this is very confusing and I don't like it. If `delegated_amount` is the one you actually set, this function and all others should be called `build_with_delegated_amount` and the param should keep its name. I don't like that we are deviating from the TIP, I don't want to delay this PR even further but I think I would like to change that back in the future.

DaughterOfMars: \>:c

thibault-martinez (dismissed):

Thoralf-M (dismissed):

thibault-martinez (approved):

Thoralf-M (approved):

PR #1872, opened 2024-01-23
Add CommitmentContextInput if required

### LINKED ISSUE #1877: Add CommitmentContextInput if required
Waiting for iota-core to be updated.

thibault-martinez:
Just made it to work but it's not pretty, will think of something better.
Fixes https://github.com/iotaledger/iota-sdk/issues/1877


sdk/src/wallet/operations/transaction/build_transaction.rs
```diff
@@ -49,7 +48,6 @@ where
             }
 
             inputs.push(Input::Utxo(UtxoInput::from(*input.output_id())));
-            inputs_for_signing.push(input.clone());
```
thibault-martinez, line 52: Was there any reason to do this?

Thoralf-M: Checked the history and no, I probably have overseen that it can be used directly

thibault-martinez: Thanks

DaughterOfMars (approved):

Thoralf-M (approved):

PR #1881, opened 2024-01-24
Node/hex: bigint fixes

### LINKED ISSUE #1839: [Bug]: mismatch between Nodejs `bigint` types and (hex) `string` types
### Issue description

In the bindings, several interfaces are typed as BigInts, but expect or return hex encoded strings. This should be improved. Examples:
- TokenScheme: https://github.com/iotaledger/iota-sdk/blob/develop/bindings/nodejs/lib/types/block/output/token-scheme.ts#L39-L42
- CreateNativeTokenParams: https://github.com/iotaledger/iota-sdk/blob/develop/bindings/nodejs/lib/types/wallet/transaction-options.ts#L70
- There might be others. I haven't performed an exhaustive search

### Version

1.1.4

### Expected behaviour

Correct typings: if hex encoded amounts are returned then they should be typed as `HexEncodedAmount`, if it is of Javascript BigInt types, then they should be typed as `bigint`. Refrain from returning numbers, as that will result in lost precision.

### Actual behaviour

Incorrect typings for bigints. They return hexadecimal strings

Yes

1. Create a new TokenScheme with circulating and total supply of `1000000000000000000000000000000n` BigInt type
2. Will result in a total supply of 1329227995784915872903807060280344576n, because it is being interpreted as a hex string

### Errors

_No response_

### Duplicate declaration

kwek20:
# Description of change

Fixes a few cases where hexstring wasnt parsed properly

**== Open question: ==**
Is there any way you can think of that makes this a breaking change?The original use still works: (It doesnt need `new INativeToken`)
```
let native: INativeToken = {
    id: "",
    amount: BigInt(1)
}
```

The alternative is changing the methods that use INativeToken to preparse like here:
https://github.com/iotaledger/iota-sdk/blob/2b44b886538b3c5dbbaf55a8ace38e8f44273bfe/bindings/nodejs/lib/wallet/account.ts#L550

fixes #1839

## Type of change

Ran examples which touch the relevant code like list-outputs and create nativetoken

### DIFF
--- a/bindings/nodejs/lib/types/models/native-token.ts
+++ b/bindings/nodejs/lib/types/models/native-token.ts
@@ -1,17 +1,20 @@
 // Copyright 2023 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+import { Transform } from 'class-transformer';
 import type { HexEncodedString } from '../utils/hex-encoding';
+import { hexToBigInt } from '../utils/hex-encoding';
 /**
  * Native token.
  */
-export interface INativeToken {
+export class INativeToken {
     /**
      * Identifier of the native token.
      */
-    id: HexEncodedString;
+    id!: HexEncodedString;
     /**
      * Amount of native tokens of the given Token ID.
      */
-    amount: bigint;
+    @Transform(value => hexToBigInt(value.value))
+    amount!: bigint;
 }
--- a/bindings/nodejs/lib/types/models/native-token.ts
+++ b/bindings/nodejs/lib/types/models/native-token.ts
@@ -15,6 +15,6 @@ export class INativeToken {
     /**
      * Amount of native tokens of the given Token ID.
      */
-    @Transform(value => hexToBigInt(value.value))
+    @Transform((value) => hexToBigInt(value.value))
     amount!: bigint;
 }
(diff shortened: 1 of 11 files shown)


kwek20:
@Tuditi Ive addressed the 2 cases you mention, and checked others. Do you think this is a breaking change or its fine?

Tuditi:
It's fine, it can be seen as a fix. Based on the typings the developer would expect a BigInt.

DaughterOfMars (dismissed):

Thoralf-M (changes requested):
Changelog entry

bindings/nodejs/lib/types/models/native-token.ts
```diff
@@ -1,17 +1,20 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
+import { Transform } from 'class-transformer';
 import type { HexEncodedString } from '../utils/hex-encoding';
+import { hexToBigInt } from '../utils/hex-encoding';
 /**
  * Native token.
  */
-export interface INativeToken {
+export class INativeToken {
```
Alex6323, line 10: Should we rename this type then in 2.0 (remove the `I` prefix)?

kwek20: yea, ill make at ticket

Thoralf-M (dismissed):

DaughterOfMars (dismissed):

Alex6323 (dismissed):

Thoralf-M (approved):

thibault-martinez (approved):

DaughterOfMars (approved):

PR #1897, opened 2024-01-26
Remove sync reisssue

### LINKED ISSUE #1867: Remove auto reissue transaction
## Description

Should we remove the auto reissue for transactions during syncing, since they require additional mana?

## Motivation

https://github.com/iotaledger/iota-sdk/pull/1756#discussion_r1461875596
> I guess we shouldn't do this automatically for now (or never), since it also requires more mana which doesn't get allotted if we just submit the same payload

Thoralf-M:
# Description of change

Remove sync reissue because it doesn't work this way anymore when there was already a block sent, since it costs additional mana, which should be allotted before reissuing a new block and also the referenced commitment quickly gets too old
Now it only reissues a tx if no block was created before (so should only happen once)

Also updated the spent metadata setting to None, to not have warn logs even though nothing changed

Fixes https://github.com/iotaledger/iota-sdk/issues/1867

## Type of change

Created a tx without issuing a block (modified this part of the code), then synced the wallet and it issued it in a block


sdk/src/wallet/operations/syncing/transactions.rs
```diff
@@ -151,66 +150,46 @@ where
...
+                let reissued_block = self
+                    .submit_signed_transaction(transaction.payload.clone(), None)
+                    .await?;
+                transaction.block_id.replace(reissued_block);
+                updated_transactions.push(transaction);
             }
         }
         drop(wallet_data);
 
-        for mut transaction in transactions_to_reissue {
```
DaughterOfMars, line 205: Any reason this was here before?

Thoralf-M: Yes, it reissued txs that were still pending. This PR removes it since this requires additional Mana for the block which should be allotted first and it also might not work at all if there is a commitment context input, because it could be too old at this point

DaughterOfMars: okie

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1900, opened 2024-01-29
Properly set Mana of output side accounts

thibault-martinez:


sdk/src/client/api/block_builder/input_selection/transition.rs
```diff
@@ -55,11 +55,19 @@ impl InputSelection {
         // Remove potential sender feature because it will not be needed anymore as it only needs to be verified once.
         let features = input.features().iter().filter(|feature| !feature.is_sender()).cloned();
 
-        let builder = AccountOutputBuilder::from(input)
+        let mut builder = AccountOutputBuilder::from(input)
             .with_account_id(account_id)
             .with_foundry_counter(u32::max(highest_foundry_serial_number, input.foundry_counter()))
             .with_features(features);
 
+        if input.is_block_issuer() {
```
thibault-martinez, line 63: Should we just do it all the time?

sdk/src/client/api/block_builder/input_selection/requirement/amount.rs
```diff
@@ -239,17 +239,17 @@ impl InputSelection {
...
-                output.chain_id().unwrap(),
-                new_amount
-            );
-
             let new_output = match output {
-                Output::Account(output) => AccountOutputBuilder::from(&*output)
-                    .with_amount(new_amount)
-                    .finish_output()?,
+                Output::Account(output) => {
+                    // Mana generated or stored by an issuer account is locked to that account.
```
Thoralf-M: Why is this relevant for amount?

thibault-martinez: Eww that was another test but I forgot to remove this

thibault-martinez: reverted

sdk/src/client/api/block_builder/input_selection/transition.rs
Thoralf-M: Can we use this Mana still for an allotment then?

thibault-martinez: Allotment to the same account you mean? That's a good question, I'll ask

Thoralf-M: Yes, I think we can but my question was more like if ISA still can do it if we set it here on the output already

sdk/src/wallet/operations/transaction/high_level/minting/create_native_token.rs
```diff
@@ -134,15 +134,20 @@ where
...
             .ok_or_else(|| crate::wallet::Error::MintingFailed("Missing account output".to_string()))?;
 
-        if let Output::Account(account_output) = &account_output.output {
-            // Create the new account output with the same feature blocks, just updated foundry_counter.
+        if let Output::Account(account_output) = &account_output_data.output {
+            // Create the new account output with the same features, just updated mana and foundry_counter.
             let new_account_output_builder = AccountOutputBuilder::from(account_output)
                 .with_account_id(account_id)
+                .with_mana(account_output_data.output.available_mana(
+                    &self.client().get_protocol_parameters().await?,
```
Alex6323: I would avoid this 2nd request and instead change line L135 to get the protocol parameters once.

thibault-martinez: There are no 2nd request as it's cached but I've changed it anyway.

sdk/src/client/api/block_builder/input_selection/transition.rs
thibault-martinez: No, if you put it in inputs it won't be available for BIC anymore. But we would only be able to allot to this specific account and we don't know that in ISA atm

Thoralf-M: But isn't that something we probably want most of the time (to allot to this account)?

Thoralf-M: If that makes sense maybe at least put a TODO or a debug log, otherwise might be hard to find later why this amount is not available

thibault-martinez: https://github.com/iotaledger/iota-sdk/issues/1918

Thoralf-M (approved):

sdk/src/client/api/block_builder/input_selection/transition.rs
thibault-martinez: Added TODO in the code as well

DaughterOfMars (approved):

PR #1906, opened 2024-01-29
Add mana rewards to semantic checks

### LINKED ISSUE #1310: Handle mana rewards in Delegation Output destruction
https://github.com/iotaledger/tips-draft/blob/tip40/tips/TIP-0040/tip-0040.md#consumed-outputs

- [ ] https://github.com/iotaledger/iota-sdk/pull/1700/files#diff-0cd9fdcd945509b7f15ba76e36be88c3db111d6c0967dd15327efe148d5a9421R194
- [ ] https://github.com/iotaledger/iota-sdk/pull/1831/files#diff-d902708cabe60524f3e8b5aa7a1d9d469085e72150ee311dffe15f53d175bb63R133

DaughterOfMars:
# Description of change

Adds mana rewards to semantic validation. They must be passed in, so a user should either call `build_transaction` using a wallet or they will have to manually get the rewards using a client, add them up, and pass them in.

Closes #1310
Dependent on #1869


Thoralf-M (changes requested):
Python and nodejs bindings now also need a new param in the function

DaughterOfMars:
Does that mean you agree with the methodology?

Thoralf-M (commented):
Didn't we say we want to have the mana rewards optional, so it only errors if a value is provided?

bindings/python/iota_sdk/utils.py
```diff
@@ -220,13 +220,14 @@ def verify_secp256k1_ecdsa_signature(
 
     @staticmethod
     def verify_transaction_semantic(
-            transaction: Transaction, inputs: List[InputSigningData], protocol_parameters: ProtocolParameters, unlocks: Optional[List[Unlock]] = None) -> str:
+            transaction: Transaction, inputs: List[InputSigningData], protocol_parameters: ProtocolParameters, unlocks: Optional[List[Unlock]] = None, mana_rewards: int = 0) -> str:
```
Thoralf-M: Should be optional with default None now
```suggestion
            transaction: Transaction, inputs: List[InputSigningData], protocol_parameters: ProtocolParameters, unlocks: Optional[List[Unlock]] = None, mana_rewards: Optional[int] = None) -> str:
```

Thoralf-M (dismissed):

sdk/src/types/block/semantic/mod.rs
```diff
@@ -281,7 +286,15 @@ impl<'a> SemanticValidationContext<'a> {
...
-            } else {
+            // If mana rewards were provided or none of the inputs can claim rewards
+            // then the mismatch should return an error. Otherwise the mismatch may
+            // still be acceptable.
+            } else if self.mana_rewards.is_some()
+                || !self
+                    .input_chains
+                    .iter()
+                    .any(|(chain_id, input)| input.can_claim_rewards(self.output_chains.get(chain_id).copied()))
+            {
```
thibault-martinez: Maybe if you keep the rewards Hashmap from `Selected` this would be a little bit more defined/verified and we could move the code to the transitions where I think it should be

DaughterOfMars: Would you expect this to still allow the transaction to pass validation if there is no reward provided for an input and the amounts don't match? Or should there be a new error to tell the user that they may need to provide rewards for the input?

thibault-martinez: I think I would error, just an extra layer of verification, what do you think ?

thibault-martinez (changes requested):
Need to update PreparedTransactionData and SignedTransactionData in the bindings

bindings/nodejs/lib/types/wallet/signed-transaction-data.ts
```diff
@@ -13,4 +14,6 @@ export class SignedTransactionData {
     /** Signed inputs data. */
     @Type(() => InputSigningData)
     inputsData!: InputSigningData;
+    /** Mana rewards by input */
+    manaRewards?: { [output_id: OutputId]: string };
```
thibault-martinez: Should this be u64? Or NumericString? I'm a bit confused with all the options

bindings/python/iota_sdk/utils.py
```diff
@@ -220,13 +220,14 @@ def verify_secp256k1_ecdsa_signature(
 
     @staticmethod
     def verify_transaction_semantic(
-            transaction: Transaction, inputs: List[InputSigningData], protocol_parameters: ProtocolParameters, unlocks: Optional[List[Unlock]] = None) -> str:
+            transaction: Transaction, inputs: List[InputSigningData], protocol_parameters: ProtocolParameters, unlocks: Optional[List[Unlock]] = None, mana_rewards: Optional[int] = None) -> str:
```
thibault-martinez: Not  Optional[int] anymore?

bindings/nodejs/lib/utils/utils.ts
```diff
@@ -454,21 +455,27 @@ export class Utils {
...
+     * @param manaRewards The total mana rewards claimed in the transaction.
      *
      * @returns The conflict reason.
      */
     static verifyTransactionSemantic(
         transaction: SignedTransactionPayload,
         inputs: InputSigningData[],
+        protocolParameters: ProtocolParameters,
         unlocks?: Unlock[],
+        manaRewards?: NumericString,
```
thibault-martinez: Not anymore

bindings/nodejs/lib/types/utils/bridge/utils.ts
```diff
@@ -222,7 +222,9 @@ export interface __VerifyTransactionSemantic__ {
     data: {
         transaction: SignedTransactionPayload;
         inputs: InputSigningData[];
+        protocolParameters: ProtocolParameters;
         unlocks?: Unlock[];
+        manaRewards?: NumericString;
```
thibault-martinez: Not anymore

bindings/nodejs/lib/types/client/prepared-transaction-data.ts
```diff
@@ -24,6 +24,10 @@ export class PreparedTransactionData {
      * Optional remainder output information
      */
     remainders?: Remainder[];
+    /**
+     * Mana rewards by input.
+     */
+    manaRewards?: { [output_id: OutputId]: string };
```
thibault-martinez: string?

bindings/python/iota_sdk/utils.py
DaughterOfMars: ah, good call

sdk/src/types/block/semantic/mod.rs
```diff
@@ -31,12 +32,16 @@ pub struct SemanticValidationContext<'a> {
     pub(crate) unlocks: Option<&'a [Unlock]>,
     pub(crate) input_amount: u64,
     pub(crate) input_mana: u64,
+    pub(crate) mana_rewards: BTreeMap<OutputId, u64>,
+    pub(crate) reward_context_inputs: HashMap<OutputId, RewardContextInput>,
+    pub(crate) commitment_context_input: Option<CommitmentContextInput>,
+    pub(crate) bic_context_input: Option<BlockIssuanceCreditContextInput>,
```
thibault-martinez, line 38: There can be more than one, but I can change that in my PR

DaughterOfMars: alright!

sdk/src/types/block/semantic/state_transition.rs
```diff
@@ -387,10 +458,18 @@ impl StateTransitionVerifier for DelegationOutput {
...
 
     fn destruction(
+        output_id: &OutputId,
         _current_state: &Self,
-        _context: &SemanticValidationContext<'_>,
+        context: &SemanticValidationContext<'_>,
     ) -> Result<(), StateTransitionError> {
-        // TODO handle mana rewards
+        // If a mana reward was provided but no reward context input exists
+        if context.mana_rewards.get(output_id).is_some() && !context.reward_context_inputs.contains_key(output_id) {
```
thibault-martinez, line 466: What if there is no mana reward ?

sdk/src/types/block/semantic/state_transition.rs
```diff
@@ -387,10 +458,18 @@ impl StateTransitionVerifier for DelegationOutput {
...
+            return Err(StateTransitionError::MissingRewardInput);
+        }
+        if !context.commitment_context_input.is_none() {
```
thibault-martinez:
```suggestion
        if context.commitment_context_input.is_none() {
```

sdk/src/types/block/semantic/state_transition.rs
DaughterOfMars: Then you're not claiming rewards I guess

thibault-martinez (approved):

sdk/src/types/block/semantic/state_transition.rs
thibault-martinez: I feel like it would better detect mistakes if that was an error but alright

DaughterOfMars: Well, is it possible to destroy a delegation and not claim rewards?

thibault-martinez: I don't think it's possible no. Just saying we could check the registered rewards actually match the destructions just in case

DaughterOfMars: Well if it's not possible then maybe we should not allow missing rewards full stop for destroyed delegations

Thoralf-M (approved):

sdk/src/types/block/semantic/state_transition.rs
thibault-martinez: I feel like that's what I'm saying

thibault-martinez: But let's merge for now, I need to go to bed

PR #1915, opened 2024-01-30
Some more alias -> account renames

Thoralf-M:


thibault-martinez (dismissed):

bindings/core/src/method/client.rs
```diff
@@ -301,7 +301,7 @@ pub enum ClientMethod {
     //////////////////////////////////////////////////////////////////////
     // Node indexer API
     //////////////////////////////////////////////////////////////////////
-    /// Fetch alias/basic/NFT/foundry output IDs
+    /// Fetch account/basic/NFT/foundry output IDs
```
DaughterOfMars: can it do anchor/delegation?

Thoralf-M: Added them

DaughterOfMars: there are other comments that might be missing them too in these changes

Thoralf-M: Updated another place, one is left, but there is a TODO since we didn't filter these types out yet

DaughterOfMars: Alrighty

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1916, opened 2024-01-30
Bring back slot_commitment_id_index test

thibault-martinez:


Thoralf-M (dismissed):

sdk/tests/types/slot.rs
```diff
@@ -1,40 +1,25 @@
...
 use packable::PackableExt;
+use pretty_assertions::assert_eq;
 
-// #[test]
-// fn slot_commitment_id_index() {
-// Test from https://github.com/iotaledger/tips-draft/blob/tip46/tips/TIP-0046/tip-0046.md#slot-commitment-id-1
+#[test]
+fn slot_commitment_id_index() {
+    // Test from https://github.com/iotaledger/tips-draft/blob/tip46/tips/TIP-0046/tip-0046.md#slot-commitment-id-1
+    let file = std::fs::read_to_string("./tests/types/fixtures/slot_commitment.json").unwrap();
```
DaughterOfMars: Add a struct to define the types and then just call https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html?

thibault-martinez: If we had more than one fixture maybe but here I'm not sure there is any point tbh

DaughterOfMars: I mean it's not a lot to add and it helps clarity

thibault-martinez: Ok, will do

thibault-martinez: done

DaughterOfMars (approved):

Thoralf-M (approved):

Alex6323 (approved):

Issue #1726, opened 2023-12-01
`Bech32::from_str` not doable for `MultiAddress`


thibault-martinez:
Add a better error and see where it could go wrong.

thibault-martinez:
Won't do

PR #1919, opened 2024-01-30
Change Bech32Address serialization

thibault-martinez:
Fixes https://github.com/iotaledger/iota-sdk/issues/1726


sdk/src/client/node_api/indexer/query_parameters.rs
```diff
@@ -67,7 +67,7 @@ pub struct OutputQueryParameters {
     /// Filters outputs based on the presence of a specific native token.
     native_token: Option<TokenId>,
     /// Returns outputs that are unlockable by the bech32 address.
-    unlockable_by_address: Option<Bech32Address>,
+    unlockable_by_address: Option<String>,
```
thibault-martinez: Should we have another type, Bech32AddressString, to at least validate it's correct Bech32?

DaughterOfMars: I think that's reasonable

sdk/src/types/block/address/bech32.rs
```diff
@@ -222,11 +228,19 @@ impl<T: core::borrow::Borrow<Bech32Address>> From<T> for Address {
...
-
 impl<T: AsRef<str> + Send> ConvertTo<Bech32Address> for T {
     fn convert(self) -> Result<Bech32Address, Error> {
         Bech32Address::try_from_str(self)
     }
 }
+
+/// A valid bech32 encoded address.
+#[derive(Clone, Debug, Eq, PartialEq, Hash, AsRef, Deref, Ord, PartialOrd)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
thibault-martinez: Should serde fail if it's an invalid bech32? Probably?

DaughterOfMars: Thought that was the point of the type

bindings/core/src/method/wallet.rs
```diff
@@ -365,7 +365,7 @@ pub enum WalletMethod {
     Send {
         #[serde(with = "iota_sdk::utils::serde::string")]
         amount: u64,
-        address: Bech32Address,
+        address: String,
```
DaughterOfMars, line 417: Missed?

sdk/src/types/block/address/bech32.rs
```diff
@@ -216,17 +223,81 @@ impl PartialEq<str> for Bech32Address {
     }
 }
 
+impl PartialEq<Bech32AddressString> for Bech32Address {
+    fn eq(&self, other: &Bech32AddressString) -> bool {
+        self.to_string().eq(&other.0)
```
Alex6323, line 228: it's probably slightly more future proof if you use `to_bech32_address_string` instead of `to_string` here.

thibault-martinez:
Won't do

Issue #1740, opened 2023-12-05
Merge TransactionOptions into a new BlockOptions?


thibault-martinez:
Need IssuerId as BlockOptions, do we need more than that? We currently pass that + TransactionOptions in most APIs now.

- [ ] Maybe add issuance to it to avoid having to call it twice ? (https://github.com/iotaledger/iota-sdk/pull/1739/files#diff-bb97bc42ed78814b7c87abe4a31292fcfa12055d3bf69e2dd7ed05666178fbb0R102)

Alex6323:
Have the alternatives already been discussed? Pretty much settled on the `BlockOptions` wrapper, or still open to just renaming `TransactionOptions` to `BlockOptions` with additional fields?

thibault-martinez:
There is only one issuance call anymore, no need to do this.

PR #1922, opened 2024-01-31
Feat: Block options

kwek20:
# Description of change

Adds a struct wrapping TransactionOptions and 2 new fields called BlockOptions

Fixes https://github.com/iotaledger/iota-sdk/issues/1740

## Type of change


sdk/src/wallet/operations/transaction/build_transaction.rs
```diff
@@ -27,12 +27,13 @@ where
     pub(crate) async fn build_transaction(
         &self,
         selected_transaction_data: Selected,
-        options: impl Into<Option<TransactionOptions>> + Send,
+        options: impl Into<Option<BlockOptions>> + Send,
```
Alex6323, line 30: bit weird API to provide a `BlockOptions` to a method called `build_transaction`, but okay. Atm we're not using its `issuer_id` field. Is this supposed to be used when we iterate the inputs to search for an account output with a block issuer feature?

PR #1923, opened 2024-01-31
Fix signing with implicit account creation and ed25519 addresses

Thoralf-M:
# Description of change

Fix signing with implicit account creation and ed25519 addresses, without this change it would add a second signature unlock with the same signature, which is not allowed.

## Type of change

Signed a tx with inputs that have both address types and the same backed Ed25519 address

### DIFF
--- a/sdk/src/client/secret/mod.rs
+++ b/sdk/src/client/secret/mod.rs
@@ -569,8 +569,8 @@ where
 
         // Convert restricted and implicit addresses to Ed25519 address, so they're the same entry in `block_indexes`.
         let required_address = match required_address {
-            Address::Restricted(restricted) => restricted.address().clone(),
             Address::ImplicitAccountCreation(implicit) => Address::Ed25519(*implicit.ed25519_address()),
+            Address::Restricted(restricted) => restricted.address().clone(),
             _ => required_address,
         };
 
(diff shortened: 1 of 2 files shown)


sdk/src/client/secret/mod.rs
```diff
@@ -567,14 +567,15 @@ where
             .required_address(slot_index, protocol_parameters.committable_age_range())?
             .ok_or(crate::client::Error::ExpirationDeadzone)?;
 
-        let required_address = if let Address::Restricted(restricted) = &required_address {
-            restricted.address()
-        } else {
-            &required_address
+        // Convert restricted and implicit addresses to Ed25519 address, so they're the same entry in `block_indexes`.
+        let required_address = match required_address {
```
thibault-martinez, line 571: Same thing in leger nano ?

sdk/src/client/secret/mod.rs
```diff
@@ -567,14 +567,15 @@ where
...
+            Address::Restricted(restricted) => restricted.address().clone(),
+            Address::ImplicitAccountCreation(implicit) => Address::Ed25519(*implicit.ed25519_address()),
```
thibault-martinez, line 572: Do you mind switching both pls? So they are in order

sdk/src/client/secret/mod.rs
Thoralf-M: Good catch

sdk/src/client/secret/mod.rs
Thoralf-M: Done

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1932, opened 2024-02-01
BlockIssuer semantic validation

### LINKED ISSUE #1853: BlockIssuerFeature transition semantic check
https://github.com/iotaledger/tips/blob/tip42/tips/TIP-0042/tip-0042.md#block-issuer-feature

thibault-martinez:
Fixes #1853


sdk/src/types/block/semantic/mod.rs
```diff
@@ -125,10 +131,36 @@ impl<'a> SemanticValidationContext<'a> {
...
+                        }
+                        let entry = self.block_issuer_mana.entry(account_id).or_default();
+                        entry.0 = entry
+                            .0
+                            .checked_add(consumed_output.available_mana(
+                                &self.protocol_parameters,
+                                output_id.transaction_id().slot_index(),
+                                self.transaction.creation_slot(),
+                            )?)
+                            .ok_or(Error::ConsumedManaOverflow)?;
```
Thoralf-M, line 157: Is this only to check the mana overflow? Isn't that also checked later again for the whole tx?

Thoralf-M: Ah entry is mutable even without mut keyword

sdk/src/types/block/semantic/mod.rs
```diff
@@ -240,19 +272,63 @@ impl<'a> SemanticValidationContext<'a> {
...
             };
 
-            if let Some(sender) = features.and_then(|f| f.sender()) {
+            if let Some(sender) = features.and_then(Features::sender) {
                 if !self.unlocked_addresses.contains(sender.address()) {
                     return Ok(Some(TransactionFailureReason::SenderFeatureNotUnlocked));
                 }
             }
 
+            // TODO: should be done only towards block issuers
```
Thoralf-M: Will you do it in this PR? Having a comment here which mentions that this is adding the `Account Out_locked`/Mana locked for an account would also be good

thibault-martinez: Will probably do it in this PR yeah, still thinking what the best way is

thibault-martinez: Done

sdk/src/types/block/semantic/mod.rs
```diff
@@ -13,32 +13,35 @@ use primitive_types::U256;
...
     protocol::ProtocolParameters,
+    slot::SlotCommitmentId,
     unlock::Unlock,
     Error,
 };
 
 ///
 pub struct SemanticValidationContext<'a> {
     pub(crate) transaction: &'a Transaction,
+    pub(crate) transaction_id: TransactionId,
```
DaughterOfMars, line 28: Can't you just get this from the transaction?

sdk/src/types/block/semantic/mod.rs
```diff
@@ -105,17 +110,18 @@ impl<'a> SemanticValidationContext<'a> {
 
     ///
     pub fn validate(mut self) -> Result<Option<TransactionFailureReason>, Error> {
-        // Validation of inputs.
-        let mut has_implicit_account_creation_address = false;
-
-        self.commitment_context_input = self.transaction.context_inputs().commitment().copied();
+        self.commitment_context_input = self
```
DaughterOfMars, line 111: Seems weird to even store this in a field, since it's so easy to retrieve from the transaction

sdk/src/types/block/semantic/mod.rs
thibault-martinez: Yes we can, but I'm storing it to avoid having to compute it in a loop

sdk/src/types/block/semantic/mod.rs
```diff
@@ -105,17 +110,18 @@ impl<'a> SemanticValidationContext<'a> {
...
+            .transaction
+            .context_inputs()
+            .commitment()
+            .map(|c| c.slot_commitment_id());
 
-        self.bic_context_input = self
+        self.bic_context_inputs = self
```
DaughterOfMars: same

sdk/src/types/block/semantic/mod.rs
thibault-martinez: It's really extra verbose if we don't :/

sdk/src/types/block/semantic/mod.rs
thibault-martinez: It really makes it much more easy to work with if we do the filter map here tbh

thibault-martinez: But maybe this one is used only in this function actually, but `commitment_context_input` I think I'd keep tbh

sdk/src/types/block/semantic/mod.rs
```diff
@@ -240,19 +272,64 @@ impl<'a> SemanticValidationContext<'a> {
...
             };
 
-            if let Some(sender) = features.and_then(|f| f.sender()) {
+            if let Some(sender) = features.and_then(Features::sender) {
                 if !self.unlocked_addresses.contains(sender.address()) {
                     return Ok(Some(TransactionFailureReason::SenderFeatureNotUnlocked));
                 }
             }
 
+            // Handling `Account Out_locked` mana.
```
DaughterOfMars: weird comment

sdk/src/types/block/semantic/mod.rs
```diff
@@ -262,7 +339,7 @@ impl<'a> SemanticValidationContext<'a> {
             self.output_mana = self.output_mana.checked_add(mana).ok_or(Error::CreatedManaOverflow)?;
 
             // Add allotted mana
-            for mana_allotment in self.transaction.allotments() {
+            for mana_allotment in (*self.transaction.allotments()).iter() {
```
DaughterOfMars: can't just be `self.transaction.allotments().iter()`?

thibault-martinez: Ah yeah sure, thanks. A remnant from before I changed the function I guess.

sdk/src/types/block/semantic/mod.rs
DaughterOfMars: But you can do it before the loop in the validate method?

sdk/src/types/block/semantic/mod.rs
thibault-martinez: Changed

sdk/src/types/block/semantic/mod.rs
thibault-martinez: Removed

sdk/src/types/block/semantic/mod.rs
DaughterOfMars: Can't you just store it in a local variable for the method?

thibault-martinez: It's used in a lot of state transitions as well

sdk/src/types/block/semantic/mod.rs
thibault-martinez: It's already needed in `SemanticValidationContext::new` for something else

sdk/src/types/block/semantic/mod.rs
DaughterOfMars: Alright, fair enough 🤷‍♀️

sdk/src/types/block/semantic/mod.rs
thibault-martinez: Changed

sdk/src/types/block/semantic/mod.rs
```diff
@@ -105,17 +108,18 @@ impl<'a> SemanticValidationContext<'a> {
...
-
-        self.commitment_context_input = self.transaction.context_inputs().commitment().copied();
+        self.commitment_context_input = self
+            .transaction
+            .context_inputs()
+            .commitment()
+            .map(|c| c.slot_commitment_id());
 
-        self.bic_context_input = self
+        let bic_context_inputs = self
```
Alex6323, line 117: not sure the binding's name is still appropriate ... `bic_accounts` maybe?

thibault-martinez: Idk, I'd like to retain the information that this actually comes from the context inputs field, otherwise it's a bit confusing and we could think it's the block issuer accounts coming from inputs or outputs.

sdk/src/types/block/semantic/mod.rs
```diff
@@ -105,17 +108,18 @@ impl<'a> SemanticValidationContext<'a> {
...
             .transaction
             .context_inputs()
             .iter()
-            .find_map(|c| c.as_block_issuance_credit_opt())
-            .copied();
+            .filter_map(|c| c.as_block_issuance_credit_opt().map(|bic| *bic.account_id()))
+            .collect::<HashSet<_>>();
```
Alex6323: I think you can use `block_issuance_credits` method on the `ContextInputs` type here.

sdk/src/types/block/semantic/mod.rs
Alex6323: hm, okay, was checking whether this name was kept on purpose

sdk/src/types/block/semantic/mod.rs
thibault-martinez: Indeed, changed, thanks

sdk/src/types/block/semantic/mod.rs
```diff
@@ -240,19 +270,63 @@ impl<'a> SemanticValidationContext<'a> {
...
                 }
             }
 
+            if let Some(unlock_conditions) = created_output.unlock_conditions() {
+                if let (Some(address), Some(timelock)) = (unlock_conditions.address(), unlock_conditions.timelock()) {
+                    if let Address::Account(account_address) = address.address() {
+                        if let Some(entry) = self.block_issuer_mana.get_mut(account_address.account_id()) {
+                            let past_bounded_slot_index = self
+                                .protocol_parameters
+                                .past_bounded_slot(self.commitment_context_input.unwrap());
```
Alex6323: do we need to document that `unwrap`?

sdk/src/types/block/semantic/mod.rs
```diff
@@ -125,10 +129,36 @@ impl<'a> SemanticValidationContext<'a> {
...
+
         for (index, (output_id, consumed_output)) in self.inputs.iter().enumerate() {
             let (amount, consumed_native_token, unlock_conditions) = match consumed_output {
                 Output::Basic(output) => (output.amount(), output.native_token(), output.unlock_conditions()),
-                Output::Account(output) => (output.amount(), None, output.unlock_conditions()),
+                Output::Account(output) => {
+                    if output.features().block_issuer().is_some() {
+                        let account_id = output.account_id_non_null(output_id);
+
+                        if self.commitment_context_input.is_none() {
```
Alex6323, line 143: just checking ... so at this point any commitment context input will do to pass validation, no constraints?

sdk/src/types/block/semantic/mod.rs
```diff
@@ -13,32 +13,34 @@ use primitive_types::U256;
...
     pub(crate) reward_context_inputs: HashMap<OutputId, RewardContextInput>,
-    pub(crate) commitment_context_input: Option<CommitmentContextInput>,
-    pub(crate) bic_context_input: Option<BlockIssuanceCreditContextInput>,
     pub(crate) input_native_tokens: BTreeMap<TokenId, U256>,
     pub(crate) input_chains: HashMap<ChainId, (&'a OutputId, &'a Output)>,
     pub(crate) output_amount: u64,
     pub(crate) output_mana: u64,
     pub(crate) output_native_tokens: BTreeMap<TokenId, U256>,
     pub(crate) output_chains: HashMap<ChainId, (OutputId, &'a Output)>,
+    pub(crate) block_issuer_mana: HashMap<AccountId, (u64, u64)>,
```
Alex6323, line 43: I think code readability could profit from replacing this u64 tuple to a struct with some named fields, but I would understand if you think it's not worth the effort.

thibault-martinez: Indeed, I don't think this is worth it

sdk/src/types/block/semantic/mod.rs
thibault-martinez: Correct

sdk/src/types/block/semantic/mod.rs
Thoralf-M: Or needs to return `TransactionFailureReason::TimelockCommitmentInputMissing`

thibault-martinez: Turns out the unwrap wasn't really justified, but I have used `BlockIssuerCommitmentInputMissing` instead

Thoralf-M (dismissed):

sdk/src/types/block/semantic/mod.rs
```diff
@@ -308,6 +385,12 @@ impl<'a> SemanticValidationContext<'a> {
             }
         }
 
+        for (_, (account_input_mana, account_output_mana)) in &self.block_issuer_mana {
```
Alex6323: could probably just iterate the values

sdk/src/types/block/semantic/mod.rs
```diff
@@ -308,6 +385,12 @@ impl<'a> SemanticValidationContext<'a> {
...
+            if self.input_mana - account_input_mana < self.output_mana - account_output_mana {
```
Alex6323, line 389: we are absolutely sure this can never panic and is strictly less?

thibault-martinez: Yes, rhs is a subset of lhs

sdk/src/types/block/semantic/mod.rs
thibault-martinez: Done

Alex6323 (approved):

Thoralf-M (approved):

PR #1935, opened 2024-02-01
Remove OutputData::is_spent and rely on OutputConsumptionMetadata only

### LINKED ISSUE #1718: Properly set or remove output metadata set spent
Guess we just don't update it here

_Originally posted by @Thoralf-M in https://github.com/iotaledger/iota-sdk/pull/1714#discussion_r1410406481_

Thoralf-M:
# Description of change

Remove OutputData::is_spent and rely on OutputConsumptionMetadata only so we don't have multiple fields to update which could get out of sync

Closes #1718 

## Type of change

### DIFF
--- a/sdk/src/wallet/operations/syncing/outputs.rs
+++ b/sdk/src/wallet/operations/syncing/outputs.rs
@@ -47,7 +47,6 @@ where
                     metadata: *output_with_meta.metadata(),
                     output: output_with_meta.output().clone(),
                     output_id_proof: output_with_meta.output_id_proof().clone(),
-                    is_spent: output_with_meta.metadata().is_spent(),
                     network_id,
                     remainder,
                 }
@@ -72,7 +71,8 @@ where
             match wallet_data.outputs.get_mut(&output_id) {
                 // set unspent
                 Some(output_data) => {
-                    output_data.is_spent = false;
+                    log::warn!("Removing spent output metadata for {output_id}, because it's still unspent");
+                    output_data.metadata.spent = None;
                     unspent_outputs.push((output_id, output_data.clone()));
                     outputs.push(OutputWithMetadata::new(
                         output_data.output.clone(),
--- a/sdk/src/wallet/types/mod.rs
+++ b/sdk/src/wallet/types/mod.rs
@@ -42,8 +42,6 @@ pub struct OutputData {
     pub output: Output,
     /// The output ID proof
     pub output_id_proof: OutputIdProof,
-    /// If an output is spent
-    pub is_spent: bool,
     /// Network ID
     pub network_id: u64,
     pub remainder: bool,
--- a/sdk/src/wallet/types/mod.rs
+++ b/sdk/src/wallet/types/mod.rs
@@ -48,6 +48,11 @@ pub struct OutputData {
 }
 
 impl OutputData {
+    /// Returns whether the [`OutputMetadata`] is spent or not.
+    pub fn is_spent(&self) -> bool {
+        self.metadata.is_spent()
+    }
+
     pub fn input_signing_data(
         &self,
         wallet_data: &WalletData,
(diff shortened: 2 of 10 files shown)


sdk/src/wallet/operations/syncing/outputs.rs
```diff
@@ -72,7 +71,8 @@ where
             match wallet_data.outputs.get_mut(&output_id) {
                 // set unspent
                 Some(output_data) => {
-                    output_data.is_spent = false;
+                    log::warn!("Removing spent output metadata for {output_id}, because it's still unspent");
+                    output_data.metadata.spent = None;
```
thibault-martinez, line 75: Can you remind me why it was spent in the first place?

Thoralf-M: Usually this shouldn't happen, but I think if it got spent and then there is a chain switch which makes it unspent again, the wallet wouldn't be able to use it

thibault-martinez: Thanks

sdk/src/wallet/types/mod.rs
```diff
@@ -42,8 +42,6 @@ pub struct OutputData {
     pub output: Output,
     /// The output ID proof
     pub output_id_proof: OutputIdProof,
-    /// If an output is spent
-    pub is_spent: bool,
```
DaughterOfMars, line 46: Maybe replace this with `pub fn is_spent(&self)`

sdk/src/wallet/update.rs
```diff
@@ -66,9 +69,20 @@ where
                     wallet_data.unspent_outputs.remove(&output_id);
                     // Update spent data fields
                     if let Some(output_data) = wallet_data.outputs.get_mut(&output_id) {
-                        // TODO https://github.com/iotaledger/iota-sdk/issues/1718
-                        // output_data.metadata.set_spent(true);
-                        output_data.is_spent = true;
+                        if !output_data.metadata.is_spent() {
+                            log::warn!(
+                                "[SYNC] Setting output as spent {} without having the OutputConsumptionMetadata",
```
thibault-martinez:
```suggestion
                                "[SYNC] Setting output {} as spent without having the OutputConsumptionMetadata",
```

sdk/src/wallet/update.rs
```diff
@@ -160,8 +174,20 @@ where
         }
 
         for output_to_unlock in &spent_output_ids {
-            if let Some(output) = wallet_data.outputs.get_mut(output_to_unlock) {
-                output.is_spent = true;
+            if let Some(output_data) = wallet_data.outputs.get_mut(output_to_unlock) {
+                if !output_data.metadata.is_spent() {
+                    log::warn!(
+                        "[SYNC] Setting output as spent {} without having the OutputConsumptionMetadata",
```
thibault-martinez:
```suggestion
                        "[SYNC] Setting output {} as spent without having the OutputConsumptionMetadata",
```

sdk/src/wallet/types/mod.rs
Thoralf-M: Good idea

sdk/src/wallet/update.rs
```diff
@@ -66,9 +69,20 @@ where
...
+                            log::warn!(
+                                "[SYNC] Setting output {} as spent without having the OutputConsumptionMetadata",
+                                output_id
+                            );
+                            // Set 0 values because we don't have the actual metadata and also couldn't get it, probably
+                            // because it got pruned.
+                            output_data.metadata.spent = Some(OutputConsumptionMetadata::new(
+                                0.into(),
+                                TransactionId::new([0u8; TransactionId::LENGTH]),
+                                None,
```
thibault-martinez, line 82: Do you think these placeholder values could lead to anything wrong?

Thoralf-M: Well if it's used to fetch more information about the tx that consumed it, it would fail, but there is no case where it can't fail if we don't get this data

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1936, opened 2024-02-01
Add BlockState::Accepted

thibault-martinez:


sdk/src/types/api/core.rs
```diff
@@ -339,31 +339,33 @@ pub struct SubmitBlockResponse {
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub enum BlockState {
-    // Stored but not confirmed.
+    /// Stored but not confirmed.
     Pending,
-    // Confirmed with the first level of knowledge.
+    /// TODO
```
DaughterOfMars: TODO

bindings/python/iota_sdk/types/block/metadata.py
```diff
@@ -33,16 +33,18 @@ class BlockState(Enum):
 
     Attributes:
         Pending: Stored but not confirmed.
+        Accepted: TODO
```
DaughterOfMars: TODO

sdk/src/types/api/core.rs
thibault-martinez: This variant is not yet in the TIP 🤷🏻

DaughterOfMars: probably should find a good description for it before merging tho

thibault-martinez: Done

bindings/python/iota_sdk/types/block/metadata.py
thibault-martinez: Done

DaughterOfMars (approved):

bindings/python/iota_sdk/types/block/metadata.py
```diff
@@ -32,10 +32,10 @@ class BlockState(Enum):
     """Describes the state of a block.
 
     Attributes:
-        Pending: Stored but not confirmed.
-        Accepted: TODO
-        Confirmed: Confirmed with the first level of knowledge.
-        Finalized: Included and can no longer be reverted.
+        Pending: Stored but not accepted/confirmed.
+        Accepted: Valid block referenced by some validators.
```
Thoralf-M, line 36: No better definition than "some"?

Thoralf-M (approved):

bindings/python/iota_sdk/types/block/metadata.py
thibault-martinez: I think it's fine if we compare it to Confirmed we understand that it's less than 2/3 but I can add it if you want

Thoralf-M: I just would myself understand what exactly this means, is a single reference from a single validator already enough to have it accepted?

thibault-martinez: I would say so, yes. I can draft and wait for a better comment 🤷🏻

Thoralf-M: No need to block for this

PR #1934, opened 2024-02-01
Rename wallet data

Alex6323:
# Description of change

1. Renames `WalletData` to `WalletLedger` and corresponding field/variable names.
2. Moves `address`, `bip_path`, and `alias` out of `WalletLedger` into `Wallet`

## Links to any relevant issues

Closes #1903

PR #1941, opened 2024-02-02
Improve `WalletBuilder`

### LINKED ISSUE #1904: Add `WalletBuilder::with_address_generation_method` or something similar
## Description

Replace `WalletBuilder::with_bip_path` and `WalletBuilder::with_address` with a unifying builder method that ensures that `Wallet::bip_path` and `Wallet::address` cannot conflict in some way.

## Motivation

We can't allow fields `address` and `bip_path` have unrelated values.

Yes/No.

Alex6323:
## Description

This PR unifies the way of address building in the `WalletBuilder`.

Update: I realized during implementation of the alternative, that it doesn't provide any real benefits, and that is was much better to just use the already existing builder methods, and handle the different possible scenarios in the `finish` of the builder. This is mainly because only then all the parts are in place to check the validity of the provided values. Those need to be checked against the possibly loaded values from the storage. Also ... for ed25519 addresses and a provided or loaded secret manager and a provided or loaded bip path we can only ensure there, that they belong together.

This PR now makes improvements to the `WalletBuilder` by making sure, that a wallet can't be built from inconsistent provided values and stored state.

Builds on top of #1934.

Closes #1904
Partially addresses #2016


thibault-martinez (changes requested):
Conflicts

sdk/src/wallet/core/builder.rs
```diff
@@ -212,16 +212,8 @@ where
...
-            self.alias = loaded_wallet_builder.as_ref().and_then(|builder| builder.alias.clone());
-        }
-
-        if self.address.is_none() {
-            self.address = loaded_wallet_builder
-                .as_ref()
-                .and_then(|builder| builder.address.clone());
-        }
+        // TODO if address and/or bip path were provided we should check whether they might
+        // conflict with stored data in the wallet db.
```
Thoralf-M: Open an issue for this?

Alex6323: https://github.com/iotaledger/iota-sdk/issues/2016

sdk/src/wallet/core/builder.rs
```diff
@@ -178,46 +175,95 @@ where
...
+            .clone()
+            .ok_or(crate::wallet::Error::MissingParameter("client_options"))?
+            .finish()
+            .await?;
+
+        match (self.address.as_ref(), self.bip_path.as_ref()) {
+            (Some(address), Some(bip_path)) => {
+                // verify that the address is derived from the provided bip path.
+                if let Some(backing_ed25519_address) = address.inner.backing_ed25519() {
+                    self.verify_ed25519_address(backing_ed25519_address, bip_path).await?;
```
Thoralf-M: This seems wrong to me here, the secret manager shouldn't be needed at this point if the address is there, otherwise the secret manager needs to be unlocked/ready every time we build a wallet with existing storage?

Alex6323: good point, we only need to check it if it hasn't been stored yet

Alex6323: Changed

sdk/src/wallet/core/builder.rs
```diff
@@ -178,46 +175,102 @@ where
...
+                            bip_path.coin_type,
+                            bip_path.account,
+                            bip_path.address_index..bip_path.address_index + 1,
+                            GenerateAddressOptions {
+                                internal: bip_path.change != 0,
+                                ledger_nano_prompt: false,
+                            },
+                        )
+                        // Panic: if it didn't return an Err, then there must be at least one address
+                        .await?[0];
```
Thoralf-M: Can we make a function that's generating the address here and that's also called in `verify_ed25519_address()`?

sdk/src/wallet/core/builder.rs
```diff
@@ -178,46 +175,102 @@ where
             self.secret_manager = secret_manager;
         }
 
-        let restored_bip_path = loaded_wallet_builder.as_ref().and_then(|builder| builder.bip_path);
+        let mut skip_verification = true;
```
thibault-martinez: I'd personally flip the bool and call it `verify_*` since you end up negating it in the condition anyway, reads better IMO

sdk/src/wallet/core/builder.rs
```diff
@@ -178,46 +175,102 @@ where
...
+        let loaded_address = loaded_wallet_builder
+            .as_ref()
+            .and_then(|builder| builder.address.clone());
+
+        // May use a previously stored address it wasn't provided
```
thibault-martinez:
```suggestion
        // May use a previously stored address if it wasn't provided
```

sdk/src/wallet/core/builder.rs
```diff
@@ -178,46 +175,102 @@ where
...
+        if let Some(address) = &self.address {
+            if let Some(loaded_address) = &loaded_address {
+                if address != loaded_address {
+                    return Err(crate::wallet::Error::WalletAddressMismatch(address.clone()));
```
thibault-martinez, line 187: I know it was already like that, but providing only one element in a mismatch error is a bit weird...

sdk/src/wallet/core/builder.rs
```diff
@@ -178,46 +175,102 @@ where
...
+                if !skip_verification {
+                    // verify that the address is derived from the provided bip path.
+                    if let Some(backing_ed25519_address) = address.inner.backing_ed25519() {
+                        self.verify_ed25519_address(backing_ed25519_address, bip_path).await?;
+                    } else {
+                        return Err(crate::wallet::Error::InvalidParameter("address/bip_path mismatch"));
+                    }
+                }
+                self.address.replace(address.clone());
+                self.bip_path = Some(*bip_path);
```
thibault-martinez: Do we really need these two lines?

sdk/src/wallet/core/builder.rs
```diff
@@ -178,46 +175,102 @@ where
...
+            }
+            (Some(address), None) => {
+                self.address.replace(address.clone());
+                self.bip_path = None;
```
thibault-martinez: same?

sdk/src/wallet/core/builder.rs
Alex6323: nope, we don't

sdk/src/wallet/core/builder.rs
Alex6323: nope, we don't

sdk/src/wallet/core/builder.rs
Alex6323: yup, wanted to create an issue for that

Alex6323: https://github.com/iotaledger/iota-sdk/issues/2045

sdk/src/wallet/core/builder.rs
```diff
@@ -178,46 +175,78 @@ where
...
+            }
+            (Some(address), None) => {}
+            (None, Some(bip_path)) => {
+                self.address.replace(Bech32Address::new(
+                    client.get_bech32_hrp().await?,
+                    self.generate_ed25519_address(bip_path).await?,
+                ));
+            }
+            (None, None) => {
+                return Err(crate::wallet::Error::MissingParameter("address, bip_path"));
```
Thoralf-M: Does this make sense? Or `|` since we need only one
```suggestion
                return Err(crate::wallet::Error::MissingParameter("address or bip_path"));
```

sdk/src/wallet/core/builder.rs
```diff
@@ -178,46 +175,78 @@ where
...
+                if verify_address {
+                    // verify that the address is derived from the provided bip path.
+                    if let Some(backing_ed25519_address) = address.inner.backing_ed25519() {
+                        self.verify_ed25519_address(backing_ed25519_address, bip_path).await?;
+                    } else {
+                        return Err(crate::wallet::Error::InvalidParameter("address/bip_path mismatch"));
+                    }
+                }
+            }
+            (Some(address), None) => {}
```
thibault-martinez:
```suggestion
            (Some(_address), None) => {}
```

thibault-martinez: warning

sdk/src/wallet/core/builder.rs
thibault-martinez: It's actually `and` here

Thoralf-M: Why?

Thoralf-M: We only need one?

thibault-martinez: Ah yeah correct indeed, my bad

Thoralf-M (approved):

sdk/src/wallet/core/builder.rs
```diff
@@ -178,46 +175,78 @@ where
...
+            }
+            (Some(_address), None) => {}
+            (None, Some(bip_path)) => {
+                self.address.replace(Bech32Address::new(
+                    client.get_bech32_hrp().await?,
+                    self.generate_ed25519_address(bip_path).await?,
+                ));
+            }
+            (None, None) => {
+                return Err(crate::wallet::Error::MissingParameter("address or bip_path"));
```
thibault-martinez, line 241: Wondering if we shouldn't create an address from a default BIP path anyway

thibault-martinez: We do in the CLI wallet...

Alex6323: which is a "user" of the sdk, so I would say there it's fine, but it feels weird to me if we do that already on the sdk level :shrug:

Alex6323: maybe create an `?` issue about this?

thibault-martinez (approved):

PR #1951, opened 2024-02-05
Add rewards to mana balance

### LINKED ISSUE #1948: Add rewards to the Mana balance?

DaughterOfMars:
# Description of change

Adds mana rewards to balance calculation.

Closes #1948

### DIFF
--- a/sdk/src/wallet/operations/balance.rs
+++ b/sdk/src/wallet/operations/balance.rs
@@ -69,6 +69,7 @@ where
                         output_id.transaction_id().slot_index(),
                         slot_index,
                     )?;
+                    // Add mana rewards
                     if let Ok(response) = self.client().get_output_mana_rewards(output_id, slot_index).await {
                         balance.mana.rewards += response.rewards;
                     }
@@ -101,6 +102,10 @@ where
                 Output::Delegation(delegation) => {
                     // Add amount
                     balance.base_coin.total += delegation.amount();
+                    // Add mana rewards
+                    if let Ok(response) = self.client().get_output_mana_rewards(output_id, slot_index).await {
+                        balance.mana.rewards += response.rewards;
+                    }
                     // Add storage deposit
                     balance.required_storage_deposit.delegation += storage_cost;
                     if !wallet_data.locked_outputs.contains(output_id) {
@@ -132,9 +137,6 @@ where
                             output_id.transaction_id().slot_index(),
                             slot_index,
                         )?;
-                        if let Ok(response) = self.client().get_output_mana_rewards(output_id, slot_index).await {
-                            balance.mana.rewards += response.rewards;
-                        }
 
                         // Add storage deposit
                         if output.is_basic() {
@@ -210,10 +212,6 @@ where
                                     output_id.transaction_id().slot_index(),
                                     slot_index,
                                 )?;
-                                if let Ok(response) = self.client().get_output_mana_rewards(output_id, slot_index).await
-                                {
-                                    balance.mana.rewards += response.rewards;
-                                }
 
                                 // Add storage deposit
                                 if output.is_basic() {
--- a/sdk/src/wallet/operations/balance.rs
+++ b/sdk/src/wallet/operations/balance.rs
@@ -1,8 +1,6 @@
 // Copyright 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::HashMap;
-
 use primitive_types::U256;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


sdk/src/wallet/types/balance.rs
```diff
@@ -85,6 +85,8 @@ pub struct ManaBalance {
     pub(crate) total: DecayedMana,
     /// Available mana.
     pub(crate) available: DecayedMana,
+    /// Mana rewards.
+    pub(crate) rewards: u64,
```
thibault-martinez, line 89: I wonder if we should split into staker/delegator rewards

DaughterOfMars: 🤷‍♀️

sdk/src/wallet/operations/balance.rs
```diff
@@ -271,6 +285,7 @@ where
                         output_data.output_id.transaction_id().slot_index(),
                         slot_index,
                     )?;
+                    mana_rewards.remove(&output_data.output_id);
```
Thoralf-M: Shouldn't the rewards stay if it's an account or the first created delegation output? Because just transition these should not remove the rewards?

DaughterOfMars: Can you explain more?

Thoralf-M: If you have a staking account and do a tx where you just transition the account, this doesn't affect the rewards at all? Or for delegation, if you transition into delayed delegation, this also changes nothing for the rewards?

DaughterOfMars: but these are just locked outputs?

Thoralf-M: Locked output means in this case that they're used as input in pending transactions

DaughterOfMars: right so we can't really know if those rewards are claimed right

Thoralf-M: If they were claimed already, then you wouldn't get them from the API?

Thoralf-M: Also even if the tx would claim them, as long as it's not confirmed, I see no reason to not still show the rewards

DaughterOfMars: isn't that true of all locked outputs though?

Thoralf-M: Yes, actually would never remove any the rewards

DaughterOfMars: If you say so 🤷‍♀️

sdk/src/wallet/types/balance.rs
Thoralf-M: Usually you do only a single thing I guess, wouldn't split unless someone requests it

Thoralf-M (dismissed):

sdk/src/wallet/types/balance.rs
thibault-martinez: Fair enough

sdk/src/wallet/operations/balance.rs
```diff
@@ -129,6 +132,9 @@ where
                             output_id.transaction_id().slot_index(),
                             slot_index,
                         )?;
+                        if let Ok(response) = self.client().get_output_mana_rewards(output_id, slot_index).await {
+                            balance.mana.rewards += response.rewards;
+                        }
```
thibault-martinez: Should we do these only if the output is delegation/account? That could be a LOT of requests for nothing otherwise

Thoralf-M: Hm actually at this point it can't be either of them, since we match on the output types in L63
Instead if should be moved in this closure `Output::Delegation(delegation) => {`

sdk/src/wallet/operations/balance.rs
```diff
@@ -204,6 +210,10 @@ where
                                     output_id.transaction_id().slot_index(),
                                     slot_index,
                                 )?;
+                                if let Ok(response) = self.client().get_output_mana_rewards(output_id, slot_index).await
+                                {
+                                    balance.mana.rewards += response.rewards;
+                                }
```
Thoralf-M: Same here, not needed since it can't be an Account or Delegation output

sdk/src/wallet/operations/balance.rs
thibault-martinez: Ah yeah, even better

DaughterOfMars: done

thibault-martinez (approved):

Thoralf-M (approved):

Issue #1876, opened 2024-01-23
Should `ProtocolParameters` impl `Default`?


Alex6323:
## Description

We should think about whether that really makes sense considering:
* values change over time through fluid protocol upgrades (what happens then? Updating the defaults?)
* there are several networks in parallel (not clear why defaulting to a particular one of those?)
* buggy as shown with https://github.com/iotaledger/iota-sdk/issues/1808

thibault-martinez:
Closed by https://github.com/iotaledger/iota-sdk/pull/1969

Issue #1549, opened 2023-11-01
Expose locked_outputs to users?


Thoralf-M:
## Description

Expose `locked_outputs` or basically a duplication of it to users, so they manually can lock some outputs they don't want to be consumed in transactions? 
Having two methods like `.lock_output(output_id)`, `.unlock_output(output_id)`  that add/remove this output id to/from the locked_outputs field of the wallet.

## Motivation

For example one has a basic output with some metadata feature that one would like to keep and not consume in other transactions, then it would be nice if one can just call `.lock_output(output_id)`, `.unlock_output(output_id)` to make it not available for transactions or later available again.

## Requirements

1. Add  `.lock_output(output_id)` or `.lock_output(output_id, "my precious metadata output".to_string())` and `.unlock_output(output_id)` which make outputs not or available to the input selection
2. Make sure that users don't mess up transactions and that https://github.com/iotaledger/iota-sdk/blob/8b55ee58ff725864e78ec8c1ddc154b95f54c9de/sdk/src/wallet/core/builder.rs#L276 isn't messing up user locked outputs

## Open questions (optional)

Currently locked_outputs is a HashSet `locked_outputs: HashSet<OutputId>` that is automatically filled/cleared based on selecting inputs/transactions, if we allow to just unlock any output id there, users could mess it up (unlock outputs that are already used as inputs in txs) which could lead to conflicting transactions. And user defined outputs would be also removed automatically here https://github.com/iotaledger/iota-sdk/blob/8b55ee58ff725864e78ec8c1ddc154b95f54c9de/sdk/src/wallet/core/builder.rs#L276 if that's not changed.
Therefore it might make sense to create a new HashSet for user defined locked outputs to keep them separated. Or maybe a `HashMap<OutputId, Reason>` would be better? Then users could even specify a reason for what they locked an output. `Reason` could be an enum like this
```Rust
pub enum Reason {
    // Used as input in a transaction.
    TransactionInput,
    // Locked by the user.
    Custom(String),
}
```
Then `.lock_output(output_id, "my precious metadata output".to_string())`

## Are you planning to do it yourself in a pull request?

No.

Thoralf-M:
As a simpler but not so convenient alternative, `forbidden_inputs` could be added to the TransactionOptions

PR #1961, opened 2024-02-06
Add automatic mana allotment

### LINKED ISSUE #1816: Automatic Mana Allotment
Manual Mana Allotment has been added in https://github.com/iotaledger/iota-sdk/pull/1762.
Need to add a way to automatically allot the right amount of mana for a specific block.
See https://github.com/iotaledger/iota-sdk/issues/1282

DaughterOfMars:
# Description of change

Adds automatic mana allotment to ISA. These changes move the context input addition to a new requirement, as well as adds the mana allotment requirement. In order to facilitate these changes, the slot index was upgraded to a slot commitment ID and input selection now requires an issuer ID. This also means that an issuer ID is now required to prepare a transaction. The reason for this is that we must be able to put the new allotments on the proper mana allotment, or else create one with the issuer ID. If we did not, we could not be certain that the resulting work score would be correct.

The reference work score must also be passed into ISA now in order to calculate the correct mana balance between the inputs and outputs.

In the process of making the changes, I found that it seems to be a simpler change to loop over the last two requirements for `Mana` and `Allotment` until no more inputs are needed.

Closes #1816


DaughterOfMars:
Tests are failing because they do not account for mana needed to allot the block. This could be fixed in two ways:
1. Fix the protocol params we use in the tests (see #1876) so that generated mana can cover the amount
2. Add stored mana to all of the inputs.

Either way, I'd appreciate it if this was reviewed now, so I'm leaving it undrafted.

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -38,16 +40,20 @@ pub struct InputSelection {
...
     addresses: HashSet<Address>,
     burn: Option<Burn>,
     remainder_address: Option<Address>,
-    protocol_parameters: ProtocolParameters,
-    slot_index: SlotIndex,
+    slot_commitment_id: SlotCommitmentId,
     requirements: Vec<Requirement>,
     automatically_transitioned: HashSet<ChainId>,
-    mana_allotments: u64,
+    issuer_id: AccountId,
```
Thoralf-M: Would make this optional for cases where someone else issues the block or one has already still a lot BIC and doesn't want more (so no automatic allotment, if None is provided)

DaughterOfMars: it is done

sdk/src/wallet/operations/transaction/input_selection.rs
```diff
@@ -36,14 +38,27 @@ where
...
+        let slot_commitment_id = self.client().get_issuance().await?.latest_commitment.id();
+        let reference_mana_cost = if let Some(issuer_id) = issuer_id {
+            Some(
+                self.client()
+                    .get_account_congestion(&issuer_id, None)
+                    .await?
+                    .reference_mana_cost,
+            )
+        } else {
+            None
```
Thoralf-M, line 54: Don't we want first account_id as issuer_id if nothing else is provided? Otherwise it would not automatically allot in all the wallet tx methods?

sdk/src/wallet/operations/transaction/prepare_transaction.rs
```diff
@@ -62,6 +66,8 @@ where
                 RemainderValueStrategy::CustomAddress(address) => Some(address.clone()),
             });
 
+        let issuer_id = issuer_id.into();
```
Thoralf-M: Related to https://github.com/iotaledger/iota-sdk/pull/1961/files#r1482572913, probably better to set the issuer_id here already and add a field in the TransactionOptions where one can specify it as well?
So if provided issuer_id, use it
If not provided, use from TransactionOptions and if there also not provided, use first account

sdk/src/wallet/operations/transaction/input_selection.rs
DaughterOfMars: That decision should be made further up and passed in here

DaughterOfMars: Actually, maybe it's fine

Thoralf-M: Yeah added another comment, I think we should do it further up where we have access to the tx options

sdk/src/wallet/operations/transaction/prepare_transaction.rs
Thoralf-M: So now you moved the isser_id in the TransactionOptions, but if the user provides nothing there, then we also allot nothing?

DaughterOfMars: yes. I can still make it default to the first account address, just not 100% sure where

Thoralf-M: Probably in prepare_transaction(), just wondering if we want to allow having None in the wallet

DaughterOfMars: I did it in select inputs instead because that's where we actually use it. And it's fine to have None in the wallet, because then you just won't allot mana and it will have to come from previously allotted mana instead.

Thoralf-M: But all the tx methods end up calling select_inputs(), so there is no way it ends up without issuer_id?

DaughterOfMars: If the wallet doesn't have one, then it won't

Thoralf-M: Ah yeah, guess that's fine for now

sdk/src/wallet/operations/transaction/options.rs
```diff
@@ -1,43 +1,65 @@
...
-    // If custom inputs are provided only they are used. If also other additional inputs should be used,
-    // `mandatory_inputs` should be used instead.
-    #[serde(default)]
-    pub custom_inputs: Option<Vec<OutputId>>,
-    #[serde(default)]
-    pub mandatory_inputs: Option<Vec<OutputId>>,
+    /// Transaction context inputs to include.
+    pub context_inputs: Vec<ContextInput>,
+    /// Inputs that must be used for the transaction.
+    pub mandatory_inputs: BTreeSet<OutputId>,
```
Thoralf-M: So what do we think about adding `forbidden_inputs` as well?
custom_inputs originally also made it possible to prevent ISA from selecting some inputs that might should stay unspent for whatever reason
If we add `forbidden_inputs` here, this would basically solve https://github.com/iotaledger/iota-sdk/issues/1549 (the proposed way there might be nicer to use, but having this field can also do it)

DaughterOfMars: We could! Technically it's still possible the same way as before, by adding mandatory inputs and then disabling additional input selection, but your suggestion would be more flexible for sure. Alternatively, we could provide a filter closure argument and just call `retain` on the available inputs.

Thoralf-M: > and then disabling additional input selection

How can it be disabled?

Thoralf-M: Closure sounds annoying with bindings

DaughterOfMars: you can siable it with the new `allow_additional_input_selection` flag. And yes, that would not work well with bindings, you are right haha

Thoralf-M: Ah thanks, missed the addition of  `allow_additional_input_selection`

Thoralf-M: No need to add forbidden_inputs then in this PR, could still be done as part of the other issue

sdk/src/client/api/block_builder/input_selection/requirement/allotment.rs
```diff
@@ -0,0 +1,157 @@
...
+                self.requirements.push(Requirement::Allotment);
+                return Ok(additional_inputs);
+            }
+        }
+
+        Ok(Vec::new())
+    }
+
+    pub(crate) fn transaction_unlocks(&self) -> Result<Unlocks, Error> {
+        // let transaction_signing_hash = prepared_transaction_data.transaction.signing_hash();
```
Thoralf-M:
```suggestion
```

sdk/src/client/api/block_builder/input_selection/requirement/allotment.rs
```diff
@@ -0,0 +1,157 @@
...
+        let mut blocks = Vec::new();
+        let mut block_indexes = HashMap::<Address, usize>::new();
+
+        // Assuming inputs_data is ordered by address type
```
Thoralf-M: This ordering still needs to happen before L39 where the inputs are set for this tx?

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -38,16 +41,27 @@ pub struct InputSelection {
...
+    payload: Option<TaggedDataPayload>,
+    allow_additional_input_selection: bool,
+    protocol_parameters: ProtocolParameters,
+}
+
+/// Account and RMC for automatic mana allotment
+#[derive(Copy, Clone, Debug)]
+pub struct AutoManaAllotment {
+    issuer_id: AccountId,
+    reference_mana_cost: u64,
```
Thoralf-M, line 72: Derive Getters?

Thoralf-M (changes requested):
Something creates a difference of `76292` Mana (required-found) even though we should have enough and also the found amount seems to be miscalculated
The more potential Mana we get the higher the required amount gets
```
Wallet "rekt": implicit-account-transition 0x3ca99e80a0a80429cf718804a16a3902a0f1c6af35d2bb80f9317aaf38bf078df31400000000
wallet error: `insufficient mana: found 176470, required 252762`
2024-02-12 10:56:40 (UTC) wallet::wallet_cli                         ERROR wallet error: `insufficient mana: found 176470, required 252762`
Wallet "rekt": implicit-account-transition 0x3ca99e80a0a80429cf718804a16a3902a0f1c6af35d2bb80f9317aaf38bf078df31400000000
wallet error: `insufficient mana: found 184099, required 260391`
2024-02-12 10:56:49 (UTC) wallet::wallet_cli                         ERROR wallet error: `insufficient mana: found 184099, required 260391`
Wallet "rekt": implicit-account-transition 0x3ca99e80a0a80429cf718804a16a3902a0f1c6af35d2bb80f9317aaf38bf078df31400000000
wallet error: `insufficient mana: found 199357, required 275649`
2024-02-12 10:57:10 (UTC) wallet::wallet_cli                         ERROR wallet error: `insufficient mana: found 199357, required 275649`
Wallet "rekt": implicit-account-transition 0x3ca99e80a0a80429cf718804a16a3902a0f1c6af35d2bb80f9317aaf38bf078df31400000000
wallet error: `insufficient mana: found 222245, required 298536`
2024-02-12 10:57:40 (UTC) wallet::wallet_cli                         ERROR wallet error: `insufficient mana: found 222245, required 298536`
Wallet "rekt": implicit-account-transition 0x3ca99e80a0a80429cf718804a16a3902a0f1c6af35d2bb80f9317aaf38bf078df31400000000
wallet error: `insufficient mana: found 306166, required 382457`
2024-02-12 10:59:32 (UTC) wallet::wallet_cli                         ERROR wallet error: `insufficient mana: found 306166, required 382457`
Wallet "rekt": sync
Synced.
Balance {
    base_coin: BaseCoinBalance {
        total: 2000000000,
        available: 2000000000,
        voting_power: 0,
    },
    mana: ManaBalance {
        total: DecayedMana {
            stored: 2000,
            potential: 762914,
        },
        available: DecayedMana {
            stored: 2000,
            potential: 762914,
        },
        rewards: 0,
    },
    required_storage_deposit: RequiredStorageDeposit {
        basic: 63400,
        account: 0,
        foundry: 0,
        nft: 0,
        delegation: 0,
    },
    native_tokens: {},
    accounts: {},
    foundries: {},
    nfts: {},
    delegations: {},
    potentially_locked_outputs: {},
}
Wallet "rekt": outputs
Outputs:
0    0x39804c8645f9a36165f0fa887cb644b2bf9959266d48f0fb938a989866b11ef2f31400000000 ImplicitAccount Unspent
1    0x3ca99e80a0a80429cf718804a16a3902a0f1c6af35d2bb80f9317aaf38bf078df31400000000 ImplicitAccount Unspent
```

sdk/src/client/api/block_builder/input_selection/mod.rs
DaughterOfMars: why?

Thoralf-M: Just thought because it's a public struct, but maybe better to make it just pub(crate)?

DaughterOfMars: okay, that makes sense

sdk/src/wallet/operations/transaction/input_selection.rs
```diff
@@ -38,6 +38,7 @@ where
         #[cfg(feature = "participation")]
         let voting_output = self.get_voting_output().await?;
         let protocol_parameters = self.client().get_protocol_parameters().await?;
+        let creation_slot_index = self.client().get_slot_index().await?;
```
Thoralf-M: Shouldn't this rather get the slot index from the issuance commitment? Otherwise there could maybe a mismatch?
So `let creation_slot_index = slot_commitment_id.slot_index();`

sdk/src/client/api/block_builder/input_selection/requirement/allotment.rs
```diff
@@ -0,0 +1,162 @@
...
+            if !additional_inputs.is_empty() {
+                self.requirements.push(Requirement::Allotment);
+                return Ok(additional_inputs);
+            }
+        }
+
+        Ok(Vec::new())
+    }
+
+    pub(crate) fn transaction_unlocks(&self) -> Result<Unlocks, Error> {
```
Thoralf-M: How about naming this `dummy_transaction_unlocks` or something with placeholder or similar?

sdk/src/wallet/operations/transaction/input_selection.rs
DaughterOfMars: That was the issue causing the mana burn. The commitment ID is the latest ID but the slot index is more recent.

sdk/src/client/api/block_builder/input_selection/requirement/allotment.rs
DaughterOfMars: changed it to `null_transaction_unlocks`

sdk/src/wallet/operations/transaction/input_selection.rs
```diff
@@ -72,173 +82,76 @@ where
...
 
-        if let Some(burn) = burn {
+        if let Some(burn) = &options.burn {
             for delegation_id in burn.delegations() {
                 if let Some(output) = wallet_data.unspent_delegation_output(delegation_id) {
                     mana_rewards.insert(
                         output.output_id,
                         self.client()
-                            .get_output_mana_rewards(&output.output_id, slot_index)
+                            .get_output_mana_rewards(&output.output_id, creation_slot_index)
```
Thoralf-M: Is `creation_slot_index` correct here? Shouldn't it be the slot index of the commitment input that will be referenced?

DaughterOfMars: maybe! I have no idea tbh

Thoralf-M: To me it sounds so
```
If the output is in Delegating State let Delegation End be Future Bounded Epoch Index - 1, where:
Future Bounded Slot Index is given by Commitment Index + Min Committable Age where Commitment Index is the slot index of the commitment input. If no Commitment Input is present, the transaction is invalid.
Note that the presence of a Commitment Input is already required for any transaction containing a Delegation Output on the input or output side, except when it is destroyed in the Delayed Claiming State.
```
https://github.com/iotaledger/tips/blob/tip40/tips/TIP-0040/tip-0040.md#calculations-5

sdk/src/wallet/operations/transaction/high_level/minting/create_native_token.rs
```diff
@@ -105,11 +105,6 @@ where
             // Create the new account output with the same features, just updated mana and foundry_counter.
             let new_account_output_builder = AccountOutputBuilder::from(account_output)
                 .with_account_id(account_id)
-                .with_mana(account_output_data.output.available_mana(
-                    &protocol_parameters,
-                    account_output_data.output_id.transaction_id().slot_index(),
-                    self.client().get_slot_index().await?,
-                )?)
```
thibault-martinez, line 112: Does that mean we are doing this in ISA?

sdk/src/wallet/operations/transaction/input_selection.rs
```diff
@@ -72,173 +82,76 @@ where
...
             available_outputs_signing_data,
+            options.context_inputs,
             outputs,
             Some(wallet_data.address.clone().into_inner()),
-            slot_index,
+            creation_slot_index,
+            slot_commitment_id,
             protocol_parameters.clone(),
         )
+        .with_required_inputs(options.mandatory_inputs)
```
thibault-martinez: Can we rename `mandatory_inputs` in the options to `required_inputs` for consistency pls? I do prefer required over mandatory.

thibault-martinez:
> the slot index was upgraded to a slot commitment ID

Is that true? I see we take both?

sdk/src/wallet/operations/transaction/high_level/staking/extend.rs
```diff
@@ -83,8 +83,8 @@ where
                 past_bounded_epoch,
                 end_epoch,
             ));
-            options.custom_inputs = Some(vec![account_output_data.output_id]);
-            options.context_inputs = Some(vec![ContextInput::from(RewardContextInput::new(0)?)]);
+            options.mandatory_inputs = [account_output_data.output_id].into();
```
thibault-martinez: Should we not set the flag to false then?

sdk/src/wallet/operations/transaction/high_level/staking/end.rs
```diff
@@ -65,9 +65,11 @@ where
             .with_features(features)
             .finish_output()?;
 
-        let mut options = TransactionOptions::default();
-        options.custom_inputs = Some(vec![account_output_data.output_id]);
-        options.context_inputs = Some(vec![ContextInput::from(RewardContextInput::new(0)?)]);
+        let options = TransactionOptions {
+            mandatory_inputs: [account_output_data.output_id].into(),
```
thibault-martinez: Same?

sdk/src/wallet/operations/transaction/input_selection.rs
DaughterOfMars: I'm not sure that's what it's saying

sdk/src/wallet/operations/transaction/high_level/minting/create_native_token.rs
DaughterOfMars: Yes, we kind of have to because only ISA knows for sure what the creation index is

sdk/src/wallet/operations/transaction/input_selection.rs
DaughterOfMars: sure

DaughterOfMars:
> > the slot index was upgraded to a slot commitment ID
> 
> Is that true? I see we take both?

not anymore, but it was when I wrote it 😆 still not 100% sure if this is correct though

sdk/src/wallet/operations/transaction/high_level/staking/extend.rs
DaughterOfMars: No, because it's fine for ISA to select more inputs

sdk/src/wallet/operations/transaction/high_level/staking/end.rs
DaughterOfMars: same

sdk/src/wallet/operations/transaction/high_level/staking/extend.rs
thibault-martinez: I thought `custom_inputs` meant no additional inputs

DaughterOfMars: It did at one point. Now the flag controls that instead

bindings/nodejs/lib/types/wallet/transaction-options.ts
```diff
@@ -13,12 +20,8 @@ export interface TransactionOptions {
...
-    /**
-     * Custom inputs that should be used for the transaction.
-     * If custom inputs are provided, only those are used.
-     * If also other additional inputs should be used, `mandatoryInputs` should be used instead.
-     */
-    customInputs?: OutputId[];
+    /** Transaction context inputs to include. */
+    contextInputs?: ContextInput[];
     /** Inputs that must be used for the transaction. */
     mandatoryInputs?: OutputId[];
```
thibault-martinez: reame

sdk/src/wallet/operations/participation/voting.rs
```diff
@@ -111,7 +111,7 @@ where
             Some(TransactionOptions {
                 // Only use previous voting output as input.
                 custom_inputs: Some(vec![voting_output.output_id]),
```
thibault-martinez, line 113: I guess it's not compiled but you may want to remove this, resolve if not

bindings/nodejs/lib/types/wallet/transaction-options.ts
DaughterOfMars: damn casing got me

sdk/src/wallet/operations/participation/voting.rs
DaughterOfMars: not really, can just be done if these are uncommented

sdk/src/wallet/operations/transaction/high_level/staking/extend.rs
```diff
@@ -83,8 +83,8 @@ where
                 past_bounded_epoch,
                 end_epoch,
             ));
-            options.custom_inputs = Some(vec![account_output_data.output_id]);
-            options.context_inputs = Some(vec![ContextInput::from(RewardContextInput::new(0)?)]);
+            options.required_inputs = [account_output_data.output_id].into();
+            options.context_inputs = vec![ContextInput::from(RewardContextInput::new(0)?)];
```
thibault-martinez: Without the flag set to `false`, I'm not 100% sure that hardcoded 0 is valid 🤔

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -61,57 +76,80 @@ pub struct Selected {
...
-        match required_address {
-            Address::Account(account_address) => Ok(Some(Requirement::Account(*account_address.account_id()))),
-            Address::Nft(nft_address) => Ok(Some(Requirement::Nft(*nft_address.nft_id()))),
-            _ => Ok(None),
-        }
-    }
+    /// Creates a new [`InputSelection`].
+    pub fn new(
+        available_inputs: impl IntoIterator<Item = InputSigningData>,
+        context_inputs: impl IntoIterator<Item = ContextInput>,
```
thibault-martinez: I personally wouldn't have it as mandatory parameter but have a `with_context_inputs`. For the vast majority of scenario this would be empty.

DaughterOfMars: fair point!

sdk/src/wallet/operations/transaction/high_level/staking/extend.rs
DaughterOfMars: Actually, I don't think we need to pass these in at all since this can be handled by ISA now

sdk/src/wallet/operations/transaction/high_level/staking/end.rs
```diff
@@ -65,9 +65,11 @@ where
             .with_features(features)
             .finish_output()?;
 
-        let mut options = TransactionOptions::default();
-        options.custom_inputs = Some(vec![account_output_data.output_id]);
-        options.context_inputs = Some(vec![ContextInput::from(RewardContextInput::new(0)?)]);
+        let options = TransactionOptions {
+            required_inputs: [account_output_data.output_id].into(),
+            context_inputs: vec![ContextInput::from(RewardContextInput::new(0)?)],
```
thibault-martinez: Here too then?

DaughterOfMars: ah dang it

sdk/src/wallet/operations/transaction/build_transaction.rs
```diff
@@ -35,100 +31,28 @@ where
...
-                    .with_end_epoch(protocol_parameters.delegation_end_epoch(latest_slot_commitment_id))
-                    .finish_output()?;
-            }
-            needs_commitment_context = true;
         }
 
         // Build transaction
 
-        // TODO: Add an appropriate mana allotment here for this account
         let mut builder = Transaction::builder(protocol_parameters.network_id())
```
thibault-martinez: I am just seeing now that this isn't using `with_creation_slot` which means the default will use system time that isn't verified against the node. We should preferably set it to `self.client().get_slot_index()`

DaughterOfMars: but that's what gets called?

thibault-martinez: Hmm doesn't look like it

DaughterOfMars: oh, weird, I thought I changed that or something

sdk/src/wallet/operations/transaction/options.rs
```diff
@@ -1,43 +1,65 @@
...
+        output::{AccountId, OutputId},
         payload::{signed_transaction::TransactionCapabilities, tagged_data::TaggedDataPayload},
     },
 };
 
 /// Options for transactions
-#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
+#[serde(default)]
```
thibault-martinez, line 21: Does this actually propagate to the fields? If TransactionOptions is present but one of the fields is absent in serialisation that would achieve the same?

bindings/nodejs/lib/types/wallet/transaction-options.ts
```diff
@@ -13,20 +20,24 @@ export interface TransactionOptions {
...
     /** Optional note, that is only stored locally. */
     note?: string;
     /** Whether to allow sending a micro amount. */
     allowMicroAmount?: boolean;
+    /** Whether to allow the selection of additional inputs for this transaction. */
+    allowAdditionalInputSelection?: boolean;
+    /** Transaction capabilities. */
+    capabilities?: HexEncodedString;
+    /** Mana allotments for the transaction. */
+    manaAllotments?: ManaAllotment[];
```
thibault-martinez: Not sure this will serialise the same as `BTreeMap<AccountId, u64>`

bindings/python/iota_sdk/types/transaction_options.py
```diff
@@ -68,26 +71,38 @@ class TransactionOptions:
...
+                 context_inputs: Optional[List[ContextInput]] = None,
+                 required_inputs: Optional[List[OutputId]] = None,
                  burn: Optional[Burn] = None,
                  note: Optional[str] = None,
-                 allow_micro_amount: Optional[bool] = None):
+                 allow_micro_amount: Optional[bool] = None,
+                 allow_additional_input_selection: Optional[bool] = None,
+                 capabilities: Optional[HexStr] = None,
+                 mana_allotments: Optional[ManaAllotment] = None,
+                 issuer_id: Optional[AccountAddress] = None):
```
thibault-martinez: Not an Address

bindings/python/iota_sdk/types/transaction_options.py
```diff
@@ -68,26 +71,38 @@ class TransactionOptions:
...
-                 mandatory_inputs: Optional[List[OutputId]] = None,
+                 context_inputs: Optional[List[ContextInput]] = None,
+                 required_inputs: Optional[List[OutputId]] = None,
                  burn: Optional[Burn] = None,
                  note: Optional[str] = None,
-                 allow_micro_amount: Optional[bool] = None):
+                 allow_micro_amount: Optional[bool] = None,
+                 allow_additional_input_selection: Optional[bool] = None,
+                 capabilities: Optional[HexStr] = None,
+                 mana_allotments: Optional[ManaAllotment] = None,
```
thibault-martinez: Only one ManaAllotment? Plus it also wouldn't serialise as `BTreeMap<AccountId, u64>`?

sdk/tests/client/input_selection/account_outputs.rs
```diff
@@ -1644,3 +1683,191 @@ fn remainder_address_in_state_controller() {
...
+    assert!(selected.transaction.outputs().contains(&outputs[0]));
+    assert_eq!(selected.transaction.allotments().len(), 1);
+    let mana_cost = 230_000;
+    assert_eq!(
+        selected.transaction.allotments()[0],
+        ManaAllotment::new(account_id_1, mana_cost).unwrap()
+    );
+    assert_eq!(
+        selected.transaction.outputs()[1].as_account().mana(),
+        mana_input_amount - mana_cost
```
thibault-martinez, line 1741: How come it is the exact difference? What about generated/decayed mana?

sdk/tests/client/input_selection/account_outputs.rs
```diff
@@ -1644,3 +1683,191 @@ fn remainder_address_in_state_controller() {
...
+    .unwrap();
+
+    assert!(unsorted_eq(&selected.inputs_data, &inputs));
+    assert_eq!(selected.transaction.outputs().len(), 2);
+    assert!(selected.transaction.outputs().contains(&outputs[0]));
+
+    assert_eq!(selected.transaction.allotments().len(), 1);
+    assert_eq!(
+        selected.transaction.allotments()[0],
+        ManaAllotment::new(account_id_1, txn_required_mana_allotment + additional_allotment).unwrap()
```
thibault-martinez: if both manual and auto allotment are provided for the same account, should we not cap it to the auto value if it's more than the manual value?

sdk/src/client/api/block_builder/input_selection/requirement/context_inputs.rs
```diff
@@ -0,0 +1,97 @@
...
+                }
+            }
+            // Transitioning an implicit account requires a BlockIssuanceCreditContextInput.
+            if let Output::Basic(implicit_account) = &input.output {
+                if implicit_account.is_implicit_account() {
+                    log::debug!("Adding block issuance context input for transitioned implicit account output");
+                    self.context_inputs
+                        .insert(BlockIssuanceCreditContextInput::from(AccountId::from(input.output_id())).into());
+                }
+            }
```
Alex6323: else if let ?

sdk/src/client/api/block_builder/input_selection/requirement/context_inputs.rs
```diff
@@ -0,0 +1,97 @@
...
+                if account.features().block_issuer().is_some() {
+                    log::debug!("Adding block issuance context input for transitioned account output");
+                    self.context_inputs.insert(
+                        BlockIssuanceCreditContextInput::from(account.account_id_non_null(input.output_id())).into(),
+                    );
+                }
+            }
+            // Transitioning an implicit account requires a BlockIssuanceCreditContextInput.
+            if let Output::Basic(implicit_account) = &input.output {
+                if implicit_account.is_implicit_account() {
```
Alex6323: probably would keep calling it `basic` as long as we don't know whether it's actually an implicit account, but up to you.

thibault-martinez: makes sense to me

sdk/src/wallet/operations/transaction/prepare_transaction.rs
```diff
@@ -37,56 +32,13 @@ where
...
-                options.as_ref().and_then(|options| options.burn.as_ref()),
-                options.as_ref().and_then(|options| options.mana_allotments.clone()),
-            )
-            .await?;
-
-        let prepared_transaction_data = match self.build_transaction(selected_transaction_data.clone(), options).await {
-            Ok(res) => res,
-            Err(err) => {
-                // unlock outputs so they are available for a new transaction
-                self.unlock_inputs(&selected_transaction_data.inputs).await?;
```
thibault-martinez, line 86: Has this been moved somewhere else?

sdk/src/wallet/operations/transaction/prepare_transaction.rs
```diff
@@ -37,56 +32,13 @@ where
...
-                return Err(err);
-            }
-        };
+        let prepared_transaction_data = self.select_inputs(outputs, options.clone()).await?;
```
thibault-martinez: Do we need to clone?

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -38,80 +44,101 @@ pub struct InputSelection {
...
-    pub inputs: Vec<InputSigningData>,
-    /// Provided and created outputs.
-    pub outputs: Vec<Output>,
-    /// Remainder outputs information.
-    pub remainders: Vec<RemainderData>,
-    /// Mana rewards by input.
-    pub mana_rewards: HashMap<OutputId, u64>,
+/// Account and RMC for automatic mana allotment
+#[derive(Copy, Clone, Debug)]
+pub(crate) struct AutoManaAllotment {
```
Alex6323: just to make sure... we still want to refer to this functionality as "auto mana allotment"? I thought that term wasn't really liked and something like "minimum mana allotment" was the preferred term?

thibault-martinez: I'm not super opinionated but since the node team doesn't really like it because it's something different for them, we probably shouldn't create any ambiguity.

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
```diff
@@ -16,30 +16,41 @@ impl InputSelection {
...
+            }
 
             Ok(inputs)
         }
     }
 
     pub(crate) fn mana_sums(&self) -> Result<(u64, u64), Error> {
-        let required_mana = self.outputs.iter().map(|o| o.mana()).sum::<u64>() + self.mana_allotments;
+        let required_mana =
+            self.outputs.iter().map(|o| o.mana()).sum::<u64>() + self.mana_allotments.values().sum::<u64>();
```
thibault-martinez: Just checking, we don't need to care about auto mana allotment yet at this point?

sdk/src/client/api/block_builder/input_selection/requirement/context_inputs.rs
```diff
@@ -0,0 +1,97 @@
...
+use crate::{
+    client::secret::types::InputSigningData,
+    types::block::{
+        context_input::{BlockIssuanceCreditContextInput, CommitmentContextInput, RewardContextInput},
+        output::{AccountId, DelegationOutputBuilder, Output},
+    },
+};
+
+impl InputSelection {
+    pub(crate) fn fulfill_context_inputs_requirement(&mut self) -> Result<Vec<InputSigningData>, Error> {
```
thibault-martinez, line 14: Can this be called more than once?

sdk/tests/client/input_selection/account_outputs.rs
DaughterOfMars: That's how it was before, but I think if you provide an amount manually you probably don't want that amount reduced

sdk/src/client/api/block_builder/input_selection/requirement/context_inputs.rs
DaughterOfMars: sure

sdk/src/client/api/block_builder/input_selection/requirement/context_inputs.rs
DaughterOfMars: 🤷‍♀️

sdk/src/wallet/operations/transaction/prepare_transaction.rs
DaughterOfMars: no, it's not needed anymore because we wait to lock until after building the prepared txn succeeds

sdk/src/wallet/operations/transaction/prepare_transaction.rs
DaughterOfMars: probably not

sdk/src/client/api/block_builder/input_selection/mod.rs
DaughterOfMars: okie dokie

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
DaughterOfMars: this is run once before allotment and then if an additional allotment is needed

sdk/src/client/api/block_builder/input_selection/requirement/context_inputs.rs
DaughterOfMars: shouldn't

bindings/nodejs/lib/types/wallet/transaction-options.ts
DaughterOfMars: ah, correct

bindings/python/iota_sdk/types/transaction_options.py
DaughterOfMars: yeah but there's no `AccountId` type in python yet

thibault-martinez: `HexStr` then, otherwise it serialises with a `type`

sdk/src/wallet/operations/transaction/options.rs
DaughterOfMars: that's what the documentation says

sdk/tests/client/input_selection/account_outputs.rs
DaughterOfMars: Currently the protocol params to not generate any mana

bindings/nodejs/lib/types/wallet/transaction-options.ts
```diff
@@ -1,7 +1,14 @@
 // Copyright 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-import { AccountAddress, AccountId, Bech32Address, OutputId } from '../block';
+import {
+    AccountAddress,
+    AccountId,
+    Bech32Address,
+    ContextInput,
+    ManaAllotment,
```
Thoralf-M: 'ManaAllotment' is defined but never used

sdk/src/wallet/operations/transaction/input_selection.rs
Thoralf-M: I'm also not sure, opened an issue https://github.com/iotaledger/iota-sdk/issues/1995

sdk/tests/client/input_selection/account_outputs.rs
thibault-martinez: Ah I see, thanks

sdk/tests/client/input_selection/account_outputs.rs
```diff
@@ -1644,3 +1683,254 @@ fn remainder_address_in_state_controller() {
         }
     });
 }
+
+#[test]
+fn automatic_allot_account_mana() {
```
thibault-martinez: may want to rename automatic in the function names, or just ignore me

bindings/nodejs/lib/types/wallet/transaction-options.ts
```diff
@@ -13,20 +19,24 @@ export interface TransactionOptions {
...
     /** Optional note, that is only stored locally. */
     note?: string;
     /** Whether to allow sending a micro amount. */
     allowMicroAmount?: boolean;
+    /** Whether to allow the selection of additional inputs for this transaction. */
+    allowAdditionalInputSelection?: boolean;
+    /** Transaction capabilities. */
+    capabilities?: HexEncodedString;
+    /** Mana allotments for the transaction. */
+    manaAllotments?: { [account_id: AccountId]: u64 };
```
thibault-martinez, line 37: Would that `account_id` actually be named in the serialisation ?

DaughterOfMars: no, I don't think so

sdk/src/client/api/block_builder/input_selection/transition.rs
```diff
@@ -55,21 +55,12 @@ impl InputSelection {
         // Remove potential sender feature because it will not be needed anymore as it only needs to be verified once.
         let features = input.features().iter().filter(|feature| !feature.is_sender()).cloned();
 
-        let mut builder = AccountOutputBuilder::from(input)
+        let output = AccountOutputBuilder::from(input)
+            .with_mana(0)
```
Thoralf-M: I think we should do this only if it doesn't has an issuer feature, otherwise it's probably annoying to track the account bound mana
And also want to set it to 0 for NFT outputs?

thibault-martinez: block issuer feature *

Thoralf-M: Yes that's what's I meant :D

sdk/src/client/api/block_builder/input_selection/requirement/context_inputs.rs
DaughterOfMars: now it is

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
thibault-martinez: Ah I didn't get at first that the allotment requirement would change this field, makes sense now

sdk/src/wallet/operations/transaction/input_selection.rs
Thoralf-M: Philipp answered on the issue, we need the commitment slot index here

sdk/src/client/api/block_builder/input_selection/requirement/ed25519.rs
```diff
@@ -9,7 +9,7 @@ impl InputSelection {
     fn selected_unlocks_ed25519_address(&self, input: &InputSigningData, address: &Address) -> bool {
         let required_address = input
             .output
-            .required_address(self.slot_index, self.protocol_parameters.committable_age_range())
+            .required_address(self.creation_slot, self.protocol_parameters.committable_age_range())
```
Thoralf-M: This needs the commitment slot index instead of the tx creation slot

sdk/src/client/api/block_builder/input_selection/requirement/ed25519.rs
```diff
@@ -22,7 +22,7 @@ impl InputSelection {
     fn available_has_ed25519_address(&self, input: &InputSigningData, address: &Address) -> bool {
         let required_address = input
             .output
-            .required_address(self.slot_index, self.protocol_parameters.committable_age_range())
+            .required_address(self.creation_slot, self.protocol_parameters.committable_age_range())
```
Thoralf-M: This needs the commitment slot index instead of the tx creation slot

sdk/src/client/api/block_builder/input_selection/requirement/allotment.rs
```diff
@@ -0,0 +1,215 @@
...
+    pub(crate) fn null_transaction_unlocks(&self) -> Result<Unlocks, Error> {
+        let mut blocks = Vec::new();
+        let mut block_indexes = HashMap::<Address, usize>::new();
+
+        // Assuming inputs_data is ordered by address type
+        for (current_block_index, input) in self.selected_inputs.iter().enumerate() {
+            // Get the address that is required to unlock the input
+            let required_address = input
+                .output
+                .required_address(self.creation_slot, self.protocol_parameters.committable_age_range())?
```
Thoralf-M: This needs the commitment slot index instead of the tx creation slot

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -244,6 +366,56 @@ impl InputSelection {
...
+        transaction_capabilities: impl Into<TransactionCapabilities>,
+    ) -> Self {
+        self.transaction_capabilities = transaction_capabilities.into();
+        self
+    }
+
+    fn required_account_nft_addresses(&self, input: &InputSigningData) -> Result<Option<Requirement>, Error> {
+        let required_address = input
+            .output
+            .required_address(self.creation_slot, self.protocol_parameters.committable_age_range())?
```
Thoralf-M: This needs the commitment slot index instead of the tx creation slot

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -267,14 +439,14 @@ impl InputSelection {
...
-            if unlock_conditions.is_timelocked(self.slot_index, self.protocol_parameters.min_committable_age()) {
+            if unlock_conditions.is_timelocked(self.creation_slot, self.protocol_parameters.min_committable_age()) {
                 return false;
             }
 
             let required_address = input
                 .output
                 // Account transition is irrelevant here as we keep accounts anyway.
-                .required_address(self.slot_index, self.protocol_parameters.committable_age_range())
+                .required_address(self.creation_slot, self.protocol_parameters.committable_age_range())
```
Thoralf-M: This needs the commitment slot index instead of the tx creation slot

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -267,14 +439,14 @@ impl InputSelection {
             // PANIC: safe to unwrap as non basic/account/foundry/nft outputs are already filtered out.
             let unlock_conditions = input.output.unlock_conditions().unwrap();
 
-            if unlock_conditions.is_timelocked(self.slot_index, self.protocol_parameters.min_committable_age()) {
+            if unlock_conditions.is_timelocked(self.creation_slot, self.protocol_parameters.min_committable_age()) {
```
Thoralf-M: This needs the commitment slot index instead of the tx creation slot

sdk/src/client/api/block_builder/input_selection/remainder.rs
```diff
@@ -12,22 +15,23 @@ use crate::{
...
     // Gets the remainder address from configuration of finds one from the inputs.
-    fn get_remainder_address(&self) -> Result<Option<(Address, Option<Bip44>)>, Error> {
+    pub(crate) fn get_remainder_address(&self) -> Result<Option<(Address, Option<Bip44>)>, Error> {
         if let Some(remainder_address) = &self.remainder_address {
             // Search in inputs for the Bip44 chain for the remainder address, so the ledger can regenerate it
             for input in self.available_inputs.iter().chain(self.selected_inputs.iter()) {
                 let required_address = input
                     .output
-                    .required_address(self.slot_index, self.protocol_parameters.committable_age_range())?
+                    .required_address(self.creation_slot, self.protocol_parameters.committable_age_range())?
```
Thoralf-M: This needs the commitment slot index instead of the tx creation slot

sdk/src/client/api/block_builder/input_selection/remainder.rs
```diff
@@ -40,18 +44,18 @@ impl InputSelection {
         for input in &self.selected_inputs {
             let required_address = input
                 .output
-                .required_address(self.slot_index, self.protocol_parameters.committable_age_range())?
+                .required_address(self.creation_slot, self.protocol_parameters.committable_age_range())?
```
Thoralf-M: This needs the commitment slot index instead of the tx creation slot

sdk/src/client/api/block_builder/input_selection/remainder.rs
```diff
@@ -116,46 +120,53 @@ impl InputSelection {
...
-                        .with_mana(output.mana() + mana_diff)
-                        .finish_output()?,
-                    _ => panic!("only account, nft can be automatically created and can hold mana"),
+                .iter_mut()
+                .filter_map(|o| sort_order.get(&o.kind()).map(|order| (*order, o)))
+                .collect::<BTreeMap<_, _>>();
+            // Find the first value that matches the remainder address
+            if let Some(output) = ordered_outputs.into_values().find(|o| {
+                matches!(o.required_address(
+                    self.creation_slot,
```
Thoralf-M: This needs the commitment slot index instead of the tx creation slot

sdk/src/client/api/block_builder/input_selection/remainder.rs
```diff
@@ -167,33 +178,54 @@ impl InputSelection {
...
+
+        // Mana can potentially be added to an appropriate existing output instead of a new remainder output
+        let mana_remainder = selected_mana > required_mana
+            && remainder_address.map_or(true, |remainder_address| {
+                self.outputs
+                    .iter()
+                    .filter(|o| o.is_basic() || o.is_account() || o.is_anchor() || o.is_nft())
+                    .find(|o| {
+                        matches!(o.required_address(
+                        self.creation_slot,
```
Thoralf-M: This needs the commitment slot index instead of the tx creation slot

sdk/src/client/api/block_builder/input_selection/requirement/sender.rs
```diff
@@ -50,7 +50,7 @@ impl InputSelection {
                     for input in self.selected_inputs.iter() {
                         let required_address = input
                             .output
-                            .required_address(self.slot_index, self.protocol_parameters.committable_age_range())?
+                            .required_address(self.creation_slot, self.protocol_parameters.committable_age_range())?
```
Thoralf-M: This needs the commitment slot index instead of the tx creation slot

Thoralf-M: Probably good to also rename the param in `required_address()` to `commitment_slot_index`

DaughterOfMars: done

sdk/src/wallet/operations/transaction/input_selection.rs
thibault-martinez: can this be resolved?

DaughterOfMars: comment was hidden stupid github

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -38,80 +47,116 @@ pub struct InputSelection {
     required_inputs: HashSet<OutputId>,
     forbidden_inputs: HashSet<OutputId>,
     selected_inputs: Vec<InputSigningData>,
-    outputs: Vec<Output>,
+    context_inputs: HashSet<ContextInput>,
+    provided_outputs: Vec<Output>,
+    added_outputs: Vec<Output>,
```
thibault-martinez, line 52: Are `automatically_transitioned` also added to this ?

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -265,12 +263,19 @@ impl InputSelection {
             }
         }
 
+        let outputs = self
+            .added_outputs
+            .into_iter()
+            .chain(self.remainders.storage_deposit_returns)
+            .chain(self.remainders.data.iter().map(|r| r.output.clone()))
+            .collect::<Vec<_>>();
```
thibault-martinez, line 270: Just checking, this doesn't include provided, is that correct?

sdk/src/client/api/block_builder/input_selection/mod.rs
DaughterOfMars: yes

sdk/src/client/api/block_builder/input_selection/mod.rs
DaughterOfMars: goof

sdk/src/client/api/block_builder/input_selection/mod.rs
thibault-martinez: Do we still need `automatically_transitioned`?

sdk/src/client/api/block_builder/input_selection/requirement/mod.rs
```diff
@@ -82,7 +82,7 @@ impl InputSelection {
     pub(crate) fn outputs_requirements(&mut self) {
         let inputs = self.available_inputs.iter().chain(self.selected_inputs.iter());
 
-        for output in &self.outputs {
+        for output in self.provided_outputs.iter().chain(&self.added_outputs) {
```
thibault-martinez, line 81: `non_remainder_outputs`?

sdk/src/client/api/block_builder/input_selection/mod.rs
DaughterOfMars: possibly not, but I tried to remove it and ran into some weirdness so I don't want to touch it

sdk/src/client/api/block_builder/input_selection/requirement/mod.rs
DaughterOfMars: can't do it here because of borrowing

thibault-martinez: I see

sdk/src/client/api/block_builder/input_selection/mod.rs
thibault-martinez: A topic for later

sdk/src/client/api/block_builder/input_selection/requirement/amount.rs
```diff
@@ -247,13 +247,17 @@ impl InputSelection {
...
         // Only consider automatically transitioned outputs.
-        let outputs = self.outputs.iter_mut().filter(|output| {
-            output
-                .chain_id()
-                .as_ref()
-                .map(|chain_id| self.automatically_transitioned.contains(chain_id))
-                .unwrap_or(false)
-        });
+        let outputs = self
+            .provided_outputs
```
thibault-martinez: We could remove if we want to loop a bit less? automatically transitioned can never be part of provided?

sdk/src/client/api/block_builder/input_selection/requirement/amount.rs
```diff
@@ -50,7 +50,7 @@ impl InputSelection {
             }
         }
 
-        for output in self.all_outputs() {
+        for output in self.non_remainder_outputs() {
```
thibault-martinez, line 53: why?

DaughterOfMars: because this runs in remainder and it doesn't need to look at them, and it breaks if it looks at them in the wrong place

sdk/src/client/api/block_builder/input_selection/requirement/amount.rs
DaughterOfMars: good point!

thibault-martinez (dismissed):

Thoralf-M (dismissed):
More changes in new PR(s)

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -249,7 +249,7 @@ impl InputSelection {
                 .0;
             let added_mana = self.remainders.added_mana;
             if let Some(output) = self.get_output_for_added_mana(&remainder_address) {
-                log::debug!("Adding {added_mana} excess input mana to output with address {remainder_address}");
+                println!("Adding {added_mana} excess input mana to output with address {remainder_address}");
```
Thoralf-M: :trollface:

thibault-martinez (approved):

Thoralf-M (approved):

PR #501, opened 2023-05-30
Fix burn interface, undo burn changes in old binding

Thoralf-M:
# Description of change

Fix burn interface, undo burn changes in old binding since we don't have the new burn method there on the Rust side

## Type of change

- Bug fix (a non-breaking change which fixes an issue)

## How the change has been tested

Running examples

## Change checklist

- [ ] I have followed the contribution guidelines for this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

PR #1962, opened 2024-02-07
Add classes for IDs with slot index

### LINKED ISSUE #1947: bindings: add IDs slot index getter

Thoralf-M:
# Description of change

Add classes for IDs with slot index

#1947 

## Type of change

Added tests, running wallet examples


bindings/nodejs/lib/types/block/id.ts
```diff
@@ -2,6 +2,41 @@
...
+    }
+    inspect() {
+        return this.id;
+    }
+    toString() {
+        return this.id;
+    }
+    toJSON() {
+        return this.id;
+    }
```
Thoralf-M: Is there some way to have a class without named field? Haven't found a way

VmMad: Did you mean to do something like this?
```typescript
class IdWithSlotIndex extends String {
    slotIndex(): SlotIndex {
        const numberString = super.toString().slice(66, 74);
        const chunks = [];
        for (
            let i = 0, charsLength = numberString.length;
            i < charsLength;
            i += 2
        ) {
            chunks.push(numberString.substring(i, i + 2));
        }
        const separated = chunks.map((n) => parseInt(n, 16));
        const buf = Uint8Array.from(separated).buffer;
        const view = new DataView(buf);
        return view.getUint32(0, true);
    }
}
```

if you do this you lose access to `.slotIndex()` if you do something like `new IdWithSlotIndex("0x..").toString()`

msarcev: Maybe extend the String class and just add a method (tried this in devtools, seams to work)
```
class IdWithSlotIndex extends String {
    slotId() {
        this.length > 8 ? this.slice(-8) : this
    }
}
```
This way when you instantiate a "string" with `new IdWithSlotIndex("someBlockId")` you get the capabilities of the added function

Thoralf-M: Yes, thanks!

Thoralf-M: @msarcev but don't you see this as problem if it's available for all strings? Because most strings don't have a slot index encoded at the end

msarcev: I don’t get the question 🤔 
It won’t be “available for all strings”, only strings that are instantiated with `new IdWithSlotIndex(…` , no ?

bindings/nodejs/lib/types/block/id.ts
```diff
@@ -21,13 +43,18 @@ export type NftId = HexEncodedString;
 /**
  * A Block ID represented as hex-encoded string.
  */
-export type BlockId = HexEncodedString;
+export class BlockId extends IdWithSlotIndex {}
```
msarcev, line 46: Here we could get away with `BlockId` still being a type
`type BlockId = IdWithSlotIndex` if that's better 🤔 
You won't be able to instantiate it as a class, but you should still be able to do:
```
const blockId: BlockId = new IdWithSlotIndex("12345678")
```
Again, not sure if it's better, just cleaner maybe.

bindings/nodejs/lib/types/block/id.ts
msarcev: Ah, maybe you mean that you can put any string in there, even ones that are not "blockId/transactionId" 🤔 
That was also the case with your first solution though.
@VmMad suggested that we could add checks (maybe in the constructor) that assert if you passed something like a blockId (maybe by length). This would also mean handling errors if construction of that thing fails.
Wdyt ?

msarcev: We could do something like:
```
export class IdWithSlotIndex extends String {
    private _isValid: boolean = true;

    constructor(id: string) {
        super(id)

        if (id.length < 64) {
            this._isValid = false;
        }
    }

    slotIndex(): SlotIndex {
        // ...
    }

    get isValid() {
        return this._isValid
    }
}
```
Then you can use it like:
const blockId = new IdWithSlotIndex("123123123")
```
if (blockId.isValid) {
    // use it
}
```

msarcev: @Thoralf-M ☝🏻

bindings/nodejs/lib/types/block/id.ts
VmMad: Or maybe just rename `IdWithSlotIndex` to `BlockId`.

```tsx
  const slot = BlockId("0x...").slotIndex();
```

So it is also usable as type 

```typescript
  const blockId: BlockId = BlockId("0x...");
```

Thoralf-M: But then you can't call `.slotIndex()` on a blockid?

bindings/nodejs/lib/types/block/id.ts
Thoralf-M: > It won’t be “available for all strings”, only strings that are instantiated with new IdWithSlotIndex(… , no ?

Yes, sorry misread something

Sounds weird to me to have an `isValid()` method
But do I get it right that you prefer to use `new IdWithSlotIndex("0x...").slotIndex()` over `new BlockId("0x...").slotIndex()` if `BlockId` stays a type for a string instead of a class?

msarcev: Maybe combine everything and instead of even having this `IdWithSlotIndex` class, make all "string" types that have a slot index encoded be a subclass of `String` with the added method explicitly.
```
class BlockId extends String {
    slotIndex() {
        // call to shared util
    }
}

class TransactionId extends String {
    slotIndex() {
        // call to shared util
    }
}
```
This way we can construct with `new BlockId("123123")` fine

bindings/nodejs/lib/types/block/id.ts
msarcev: @Thoralf-M In which case you mean?
For the example i suggested up here you can do.
```
const blockId: BlockId = new IdWithSlotIndex("12345678")
blockId.slotIndex()
```
Cause the `BlockId` type is just an alias for `IdWithSlotIndex`

Thoralf-M: Now I understand, thanks
So do you prefer this, or https://github.com/iotaledger/iota-sdk/pull/1962#discussion_r1481491384? :smile:

msarcev: At this point https://github.com/iotaledger/iota-sdk/pull/1962#discussion_r1481491384 is maybe the cleanest i think (no indirection by avoiding the extra type and it's clear that each "id thing" is a string class with extra util methods). @VmMad wdyt ?

VmMad: @msarcev You know, at this point I think the solution @Thoralf-M did looks good to me. Because in case another method is added to the class `IdWithSlotIndex`, like `isValid()`, all the other classes that extend from it,  have it available, like `OutputId` or `SlotCommitmentId` which already extend `IdWithSlotIndex`. This way creating and referencing the util function multiple times in different classes.

TLDR; to me like it is done right now is fine, as they will be types correctly without needing an extra type
```tsx
export class BlockId extends IdWithSlotIndex {}
export class TransactionId extends IdWithSlotIndex {}
// ...
```

msarcev: Agreed, we can go with this imo.

thibault-martinez (commented):
- [x] iota-sdk/bindings/nodejs/lib/types/utils/bridge/utils.ts
- [x] iota-sdk/bindings/nodejs/lib/types/wallet/participation.ts
- [x] iota-sdk/bindings/nodejs/lib/types/wallet/signed-transaction-data.ts
- [x] iota-sdk/bindings/nodejs/lib/types/wallet/wallet.ts
- [x] iota-sdk/bindings/nodejs/lib/utils/utils.ts
- [x] iota-sdk/bindings/nodejs/lib/types/models/api/block-id-response.ts
- [x] iota-sdk/bindings/nodejs/lib/types/models/api/tips-response.ts
- [x] iota-sdk/bindings/nodejs/lib/types/models/block-metadata.ts
- [x] iota-sdk/bindings/nodejs/lib/types/client/prepared-transaction-data.ts
- [x] iota-sdk/bindings/nodejs/lib/types/models/api/plugins/indexer/output-response.ts

Thoralf-M:
@thibault-martinez classes aren't allowed in map things `An index signature parameter type must be 'string', 'number', 'symbol', or a template literal type.ts(1268)`
Also wasn't sure about the other places because they're interfaces, but apparently that's fine, so changed these

bindings/nodejs/examples/client/05-get-address-balance.ts
```diff
@@ -45,7 +51,9 @@ async function run() {
         });
 
         // Get outputs by their IDs
-        const addressOutputs = await client.getOutputs(outputIdsResponse.items);
+        const addressOutputs = await client.getOutputs(
+            outputIdsResponse.items.map((id) => new OutputId(id)),
```
thibault-martinez, line 55: No map needed anymore?

bindings/nodejs/examples/how_tos/client/get-outputs.ts
```diff
@@ -33,7 +33,9 @@ async function run() {
         console.log('First output of query:');
         console.log('ID: ', outputIdsResponse.items[0]);
 
-        const outputs = await client.getOutputs(outputIdsResponse.items);
+        const outputs = await client.getOutputs(
+            outputIdsResponse.items.map((id) => new OutputId(id)),
```
thibault-martinez, line 37: No map needed anymore?

bindings/nodejs/examples/client/05-get-address-balance.ts
Thoralf-M: Still needed because outputIdsResponse is an interface

thibault-martinez: Hmm ok, looks like a double wrap to me but if you say so

thibault-martinez (dismissed):

bindings/nodejs/examples/client/05-get-address-balance.ts
Thoralf-M: It's because the class is in an interface and then it's not actually this class object, this was also the reason why I initially didn't set this type there

Alex6323 (dismissed):

bindings/nodejs/lib/types/block/id.ts
```diff
@@ -2,6 +2,28 @@
...
 
 import { HexEncodedString } from '../utils';
+import { SlotIndex } from './slot';
+
+/**
+ * Base class for IDs with a hex encoded slot index at the end.
+ */
+export class IdWithSlotIndex extends String {
+    slotIndex(): SlotIndex {
+        const numberString = super.slice(66, 74);
```
DaughterOfMars: use `length - 8` instead?

bindings/nodejs/lib/types/utils/bridge/utils.ts
```diff
@@ -224,7 +224,7 @@ export interface __VerifyTransactionSemantic__ {
         inputs: InputSigningData[];
         protocolParameters: ProtocolParameters;
         unlocks?: Unlock[];
-        manaRewards?: { [outputId: OutputId]: NumericString };
+        manaRewards?: { [outputId: HexEncodedString]: NumericString };
```
DaughterOfMars, line 227: we definitely can't use OutputId here? I know it can't be used in the dict but is there another way?

bindings/nodejs/lib/types/block/id.ts
Thoralf-M: No, there is the output index afterwards for output ids

bindings/nodejs/lib/types/utils/bridge/utils.ts
Thoralf-M: No idea what other way there could be

bindings/nodejs/lib/types/block/id.ts
DaughterOfMars: sure but that should be called on the inner transaction ID not the output ID

Thoralf-M: Why? Then you would first need to get a transaction ID from the output ID

bindings/nodejs/lib/types/utils/bridge/utils.ts
DaughterOfMars: there is a `Map` type in js, could probably use that instead: 
```js
manaRewards?: Map<OutputId, NumericString>;
```

Thoralf-M: I think there were some issues with the Map type, for example the serialization by default ends up in an empty object
And for things with a BTreeMap we also undid it here https://github.com/iotaledger/iota-sdk/pull/501

DaughterOfMars: :(

bindings/nodejs/lib/types/block/id.ts
DaughterOfMars: yeah, that's how we do it in rust

Thoralf-M: Changed

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1963, opened 2024-02-07
Merge develop into 2.0

thibault-martinez:


bindings/nodejs/lib/types/block/output/output.ts
```diff
@@ -139,6 +139,10 @@ abstract class CommonOutput extends Output {
         }
         return undefined;
     }
+
+    getUnlockConditions(): UnlockCondition[] {
+        return this.unlockConditions;
+    }
```
Thoralf-M, line 145: Should be deleted since the field is readable

sdk/tests/client/input_selection/native_tokens.rs
```diff
@@ -1487,6 +1487,73 @@ fn multiple_remainders() {
     });
 }
 
+// #[test]
+// fn higher_nts_count_but_below_max_native_tokens() {
```
Thoralf-M, line 1491: I guess here we should now have >=65 inputs with different native tokens and a low amount and then other input(s) with a higher amount which can be used alone to fulfill the output amount requirement

Thoralf-M (approved):

sdk/tests/client/input_selection/native_tokens.rs
thibault-martinez: Yeah I thought it was annoying in the merge PR so I just commented it out for now 😄

PR #1964, opened 2024-02-07
Python: `tox -e all`

Alex6323:
# Description of change

I am suggesting to add an `all` enviroment to our `tox.ini` to be pretty certain that the Python CI will pass.

Would do a separate PR for `develop` if needed.


bindings/python/tox.ini
```diff
@@ -21,7 +21,7 @@ commands =
 description = Run format checker {basepython}
 deps = autopep8
 commands =
-    autopep8 --exclude .venv  --diff --recursive --aggressive --max-line-length 120 --exit-code .
+    autopep8 --exclude env  --diff --recursive --aggressive --max-line-length 120 --exit-code .
```
Thoralf-M, line 24: Nice, I already wondered why this command runs so slow :joy:

Thoralf-M (approved):

bindings/python/tox.ini
```diff
@@ -48,3 +48,17 @@ deps =
 commands =
     pip install .
     pylint --disable=missing-function-docstring,missing-class-docstring --rcfile=.pylintrc tests/**/*.py 
+
+[testenv:all]
```
thibault-martinez, line 52: Just checking, is there no way to call the other commands instead of duplicating?

Alex6323: I had the same thought. So that's what I tried at first. It's possible but not directly supported. You would need to call `tox -e {...}` multiple times which is more costly than this. The tipping point for me was that I wanted the formatter to modify in place, and not just show the diff.

thibault-martinez: Alright, thanks

thibault-martinez (approved):

PR #1969, opened 2024-02-07
Feat/remove protocol params default

### LINKED ISSUE #1876: Should `ProtocolParameters` impl `Default`?

DaughterOfMars:
# Description of change

Removes the default impls for protocol params and the inner structures, and replaces them with the `iota_mainnet_v3_protocol_parameters` function for testing. This was heavily dependent on packable changes so those were made in https://github.com/iotaledger/common-rs/pull/76. Additionally, the `NetworkInfo` struct has been made optional on the client because the values cannot be defaulted.

Closes #1876


sdk/src/client/core.rs
```diff
@@ -107,24 +108,45 @@ impl ClientInner {
...
-                if current_time < last_sync {
-                    return Ok(self.network_info.read().await.clone());
-                }
+            self.fetch_network_info().await
+        }
+
+        #[cfg(not(target_family = "wasm"))]
+        Ok(if let Some(info) = &*self.network_info.read().await {
+            info.clone()
+        } else {
```
Thoralf-M: This should have the timestamp logic from fetch_network_info, otherwise we will never update the network info from the node

sdk/src/client/node_manager/syncing.rs
```diff
@@ -110,14 +111,21 @@ impl ClientInner {
...
                     .clone();
+                if let Some(network_info) = network_info.as_mut() {
+                    network_info.tangle_time = info.status.relative_accepted_tangle_time;
+                    network_info.protocol_parameters = protocol_parameters;
+                } else {
+                    network_info.replace(NetworkInfo {
+                        protocol_parameters,
+                        tangle_time: info.status.relative_accepted_tangle_time,
+                    });
+                }
```
Thoralf-M: Does anything speak against just always replacing it?

sdk/src/client/core.rs
DaughterOfMars: That's what it did before, and relied on sync to update it

sdk/src/client/node_manager/syncing.rs
DaughterOfMars: no, I don't know why I did this lol

sdk/src/client/core.rs
Thoralf-M: Ah nvm I thought this is the syncing code

sdk/src/types/block/payload/signed_transaction/transaction.rs
```diff
@@ -328,16 +328,14 @@ impl WorkScore for Transaction {
     }
 }
 
-fn verify_network_id<const VERIFY: bool>(network_id: &u64, visitor: &ProtocolParameters) -> Result<(), Error> {
-    if VERIFY {
```
Thoralf-M, line 332: Just checking, visitor doesn't need to be made optional and then a check if it's Some()?

DaughterOfMars: No, because these are only called if it is `Some`

Thoralf-M (dismissed):

bindings/core/Cargo.toml
```diff
@@ -11,6 +11,7 @@ publish = false
 [dependencies]
 iota-sdk = { path = "../../sdk", default-features = false, features = [
     "wallet",
+    "samples",
```
thibault-martinez: Is this really a good name? Not very descriptive of what it contains tbh

sdk/src/client/node_manager/syncing.rs
```diff
@@ -3,6 +3,7 @@
 
 #[cfg(not(target_family = "wasm"))]
 use {
+    crate::client::NetworkInfo,
     crate::types::block::PROTOCOL_VERSION,
```
thibault-martinez: merge crate imports

sdk/src/types/block/context_input/mod.rs
```diff
@@ -87,7 +87,7 @@ pub(crate) type ContextInputCount =
 #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Deref, Packable)]
 #[packable(unpack_error = Error, with = |e| e.unwrap_item_err_or_else(|p| Error::InvalidContextInputCount(p.into())))]
 pub struct ContextInputs(
-    #[packable(verify_with = verify_context_inputs_packable)] BoxedSlicePrefix<ContextInput, ContextInputCount>,
+    #[packable(verify_with = verify_context_inputs)] BoxedSlicePrefix<ContextInput, ContextInputCount>,
```
thibault-martinez, line 90: Just making sure, since `ContextInputs` doesn't have a packable `unpack_visitor` attribute, it'll still be forwarded a `()` and then calling the verify if the parent type does have a visitor, right?

bindings/core/Cargo.toml
DaughterOfMars: suggestions?

sdk/src/types/block/context_input/mod.rs
DaughterOfMars: Basically, yes

sdk/src/client/builder.rs
```diff
@@ -275,11 +292,10 @@ impl ClientBuilder {
 
 /// Struct containing network related information
 // TODO do we really want a default?
```
Thoralf-M: Can be removed now?
```suggestion
```

Thoralf-M (dismissed):

sdk/tests/client/signing/account.rs
```diff
@@ -202,6 +206,7 @@ async fn account_reference_unlocks() -> Result<()> {
         )
         .with_outputs(outputs)
         .with_creation_slot(slot_index + 1)
+        .with_capabilities([TransactionCapabilityFlag::BurnMana])
```
thibault-martinez, line 209: Why are we adding that?

Thoralf-M: It was added before automatic mana allotment so it doesn't fail because of mana remainder, but now we can remove it

thibault-martinez: I tried removing them and tests are failing

thibault-martinez: But since it's signing tests, maybe it's not that bad to keep them since it's not really the point of the test

Thoralf-M: Yeah one would need to adjust them if it should pass without it

sdk/src/client/core.rs
```diff
@@ -46,7 +46,7 @@ impl core::ops::Deref for Client {
 pub struct ClientInner {
     /// Node manager
     pub(crate) node_manager: RwLock<NodeManager>,
-    pub(crate) network_info: RwLock<NetworkInfo>,
+    pub(crate) network_info: RwLock<Option<NetworkInfo>>,
```
thibault-martinez: Should we not request the info from the node if no ProtocolParameters were provided? Sounds a bit annoying to have an Option here. Not being able to have infos from both the user and the node seems unlikely and should error?

DaughterOfMars: I don't disagree, I'll try it

sdk/tests/client/input_selection/account_outputs.rs
```diff
@@ -1809,84 +1809,86 @@ fn min_allot_account_mana_additional() {
...
-    let account_id_2 = AccountId::from_str(ACCOUNT_ID_2).unwrap();
-
-    let additional_allotment = 1000;
-    let txn_required_mana_allotment = 271_000;
-    // The account does not have enough to cover the requirement
-    let account_mana = txn_required_mana_allotment - 100;
-    // But there is additional available mana elsewhere
-    let additional_available_mana = additional_allotment + 111;
+// #[test]
+// fn min_allot_account_mana_cannot_select_additional() {
```
thibault-martinez, line 1821: Why?

DaughterOfMars: Reliant on #2036

thibault-martinez: Ok, let's not forgot to uncomment/fix it there then, put a comment or resolve this

Thoralf-M (changes requested):
Failing tests

sdk/src/client/core.rs
```diff
@@ -31,6 +31,7 @@ use crate::{
 #[derive(Clone)]
 pub struct Client {
     pub(crate) inner: Arc<ClientInner>,
+    pub(crate) network_info: Arc<RwLock<NetworkInfo>>,
```
thibault-martinez, line 34: Can you ELI5 why it's been moved to the Client pls?

sdk/src/client/node_manager/syncing.rs
```diff
@@ -69,7 +66,7 @@ impl ClientInner {
 
         log::debug!("sync_nodes");
         let mut healthy_nodes = HashSet::new();
-        let mut network_nodes: HashMap<String, Vec<(ProtocolParameters, Node, Option<u64>)>> = HashMap::new();
+        let mut network_nodes: HashMap<String, Vec<(_, Node, Option<u64>)>> = HashMap::new();
```
thibault-martinez, line 69: Just curious, why?

sdk/src/wallet/core/builder.rs
```diff
@@ -247,6 +247,30 @@ where
...
+            .await?;
+
+        // May create a default Ed25519 wallet address if there's a secret manager.
+        let wallet_address = if self.address.is_none() && self.secret_manager.is_some() {
+            let bech32_hrp = client.network_info.read().await.protocol_parameters.bech32_hrp();
+            let address = self.create_default_wallet_address(bech32_hrp).await?;
+            address
+        } else {
+            return Err(crate::wallet::Error::MissingParameter("address"));
+        };
```
thibault-martinez: This all looks like the result of a failed conflict resolution?

sdk/src/client/core.rs
DaughterOfMars: The ClientInner is needed to call network methods, like to fetch the info. But I can't create the ClientInner without the info.

sdk/src/client/node_manager/syncing.rs
DaughterOfMars: Just didn't feel like importing it with the feature gate

sdk/src/wallet/core/builder.rs
DaughterOfMars: I re-added it because it didn't make sense to me that we can't generate an address now

thibault-martinez: But `self.address` can't be None at this point

Thoralf-M: Why not? If you provide the bip path it should generate it in L237?

DaughterOfMars: oh, missed that haha

sdk/src/wallet/core/builder.rs
```diff
@@ -247,6 +247,21 @@ where
...
 
+        // May use a previously stored wallet address if it wasn't provided
+        if self.address.is_none() {
+            self.address = loaded_wallet_builder
+                .as_ref()
+                .and_then(|builder| builder.address.clone());
+        }
+
+        // Create the node client.
+        let client = self
```
thibault-martinez: Already done L215

sdk/src/wallet/core/builder.rs
```diff
@@ -247,6 +247,21 @@ where
             self.alias = loaded_wallet_builder.as_ref().and_then(|builder| builder.alias.clone());
         }
 
+        // May use a previously stored wallet address if it wasn't provided
+        if self.address.is_none() {
```
thibault-martinez: Already done L184

sdk/src/client/node_manager/syncing.rs
thibault-martinez: Fair enough

thibault-martinez (dismissed):

bindings/core/src/method/utils.rs
```diff
@@ -120,13 +154,19 @@ pub enum UtilsMethod {
...
+        output_id: OutputId,
+    },
     /// Computes the slot commitment id from a slot commitment.
     #[serde(rename_all = "camelCase")]
-    ComputeSlotCommitmentId { slot_commitment: SlotCommitment },
+    ComputeSlotCommitmentId {
+        slot_commitment: SlotCommitment,
+    },
     /// Returns the hex representation of the serialized output bytes.
     #[serde(rename_all = "camelCase")]
```
Thoralf-M: Not needed, but also not really a problem

Thoralf-M (dismissed):

bindings/core/src/method/utils.rs
Thoralf-M: The `#[serde(rename_all = "camelCase")]`

thibault-martinez (approved):

Thoralf-M (dismissed):

Alex6323 (dismissed):

thibault-martinez (changes requested):
Python tests failing

thibault-martinez (commented):
Still failing :trollface:

thibault-martinez (approved):

Thoralf-M (approved):

PR #1978, opened 2024-02-09
Remove Multi and Restricted addresses

thibault-martinez:


thibault-martinez:
I will just redo this when/if it actually happens. Solving conflicts is simply not worth the effort.

PR #1979, opened 2024-02-09
Replace reissue_transaction_until_included() with wait_for_transaction_acceptance()

### LINKED ISSUE #1930: Change reissue_transaction_until_included() to await_block_inclusion()?
Change reissue_transaction_until_included() to something like await_block_inclusion()? Reissuing doesn't work the same way anymore as before (if one block was already successfully sent) and requires more mana which needs to be alloted first, therefore it's probably better to just wait for inclusion of the block and if the tx doesn't get confirmed create a new one.

Thoralf-M:
# Description of change

Replace reissue_transaction_until_included() with wait_for_transaction_acceptance() because we can't simply reissue blocks anymore, since they cost Mana. So now it only issues the block once when there was no block issued yet (should happen only very rarely because of some node/network issue) and then just checks the transaction acceptance.
Also changed the interval to milliseconds and set the default to 500, so it can return faster.

Closes [#1930](https://github.com/iotaledger/iota-sdk/issues/1930)

## Type of change

Running this example

```Rust
use iota_sdk::{
    types::block::output::{
        unlock_condition::{AddressUnlockCondition, ExpirationUnlockCondition},
        NftOutputBuilder, UnlockCondition,
    },
    wallet::Result,
    Wallet,
};

#[tokio::main]
        .name("wallet.log")
        .level_filter(log::LevelFilter::Debug);

    let config = fern_logger::LoggerConfig::build()
        .with_output(logger_output_config)
        .finish();

    fern_logger::logger_init(config).unwrap();
    // This example uses secrets in environment variables for simplicity which should not be done in production.
    dotenvy::dotenv().ok();

        .with_storage_path(&"./cli/stardust-cli-wallet-db")
        .finish()
        .await?;

    // May want to ensure the wallet is synced before sending a transaction.
    let balance = wallet.sync(None).await?;

    // Get the first nft
    if let Some(nft_id) = balance.nfts().first() {
        // Set the stronghold password
        wallet.set_stronghold_password("test".to_string()).await?;

        let nft = NftOutputBuilder::new_with_amount(1_000_000, *nft_id)
            .with_unlock_conditions([
                UnlockCondition::Address(AddressUnlockCondition::new(wallet.address().await)),
                UnlockCondition::Expiration(ExpirationUnlockCondition::new(
                    wallet.address().await,
                    wallet.client().get_slot_index().await? + 5000,
                )?),
            ])
            .finish_output()?;

        let transaction = wallet.send_outputs(vec![nft], None).await?;
        println!("Transaction sent: {}", transaction.transaction_id);

        println!("Tx: {:?}", transaction);

        let block_id = wallet
            .wait_for_transaction_acceptance(&transaction.transaction_id, Some(1), None)
            .await?;

        println!("Block included: /block/{}", block_id);
    } else {
        println!("No available NFTs");
    }

    Ok(())
}
```


bindings/nodejs/lib/types/error.ts
```diff
@@ -44,7 +44,7 @@ export type ClientErrorName =
     | 'secretManagerMismatch'
     | 'healthyNodePoolEmpty'
     | 'taggedData'
-    | 'tangleInclusion'
+    | 'TransactionAcceptance'
```
DaughterOfMars: casing?

sdk/src/client/error.rs
```diff
@@ -113,9 +113,9 @@ pub enum Error {
     /// Error when building tagged_data blocks
     #[error("error when building tagged_data block: {0}")]
     TaggedData(String),
-    /// The block cannot be included into the Tangle
-    #[error("block ID `{0}` couldn't get included into the Tangle")]
-    TangleInclusion(String),
+    /// The transaction could not get accepted
+    #[error("transaction ID `{0}` couldn't get accepted")]
```
DaughterOfMars:
```suggestion
    /// The transaction could not be accepted
    #[error("transaction ID `{0}` couldn't be accepted")]
```

sdk/src/wallet/operations/reissue.rs
```diff
@@ -16,48 +16,69 @@ use crate::{
...
 
-        if let Some(transaction) = transaction {
-            if transaction.inclusion_state == InclusionState::Confirmed {
-                return transaction.block_id.ok_or(Error::MissingParameter("block id"));
-            }
+        if transaction.inclusion_state == InclusionState::Accepted
+            || transaction.inclusion_state == InclusionState::Confirmed
+            || transaction.inclusion_state == InclusionState::Finalized
+        {
+            return transaction.block_id.ok_or(Error::MissingParameter("block id"));
```
DaughterOfMars: is it possible to get the block ID if it's missing somehow?

sdk/src/wallet/operations/reissue.rs
```diff
@@ -16,48 +16,69 @@ use crate::{
...
+            None => {
+                let wallet_data = self.data().await;
+                let issuer_id = transaction
+                    .payload
+                    .transaction()
+                    .allotments()
+                    .first()
+                    .map(|a| *a.account_id())
+                    .or_else(|| wallet_data.first_account_id())
+                    .ok_or(Error::AccountNotFound)?;
```
DaughterOfMars: This feels very iffy to me. Maybe we should pass the id in instead?

Thoralf-M: But if it's a different one than allotted it could burn the mana from the wrong account

bindings/nodejs/lib/types/error.ts
Thoralf-M:
```suggestion
    | 'transactionAcceptance'
```

sdk/src/wallet/operations/reissue.rs
Thoralf-M: It should never be missing because we set it when we set the inclusion_state to confirmed (the others we currently never set, but we will also have the block id then)
So we could probably unwrap, but error seems better in case we mess something up at some point

Independent of the above, we could try to get it, but then it's also not guaranteed since it might be pruned

sdk/src/wallet/operations/reissue.rs
DaughterOfMars: but there's no guarantee that this is right either :c

Thoralf-M: You said we have some issue with the reissuing part in syncing (because of the secret manager or something), don't we have the same issue here?
Because then I maybe just remove this part here?

bindings/nodejs/examples/how_tos/advanced_transactions/advanced_transaction.ts
```diff
@@ -63,7 +63,7 @@ async function run() {
         console.log(`Transaction sent: ${transaction.transactionId}`);
 
         console.log('Waiting until included in block...');
```
thibault-martinez: :trollface:

bindings/nodejs/examples/how_tos/wallet/consolidate-outputs.ts
```diff
@@ -68,7 +68,7 @@ async function run() {
         console.log('Transaction sent: %s', transaction.transactionId);
 
         // Wait for the consolidation transaction to get confirmed
-        const blockId = wallet.reissueTransactionUntilIncluded(
+        const blockId = wallet.awaitTransactionAcceptance(
```
thibault-martinez: May want to change L76?

bindings/python/examples/how_tos/wallet/consolidate_outputs.py
```diff
@@ -50,7 +50,7 @@
 print('Transaction sent: ', transaction.transaction_id)
 
 # Wait for the consolidation transaction to get confirmed
-block_id = wallet.reissue_transaction_until_included(
+block_id = wallet.await_transaction_acceptance(
```
thibault-martinez: Change L57?

bindings/python/iota_sdk/wallet/wallet.py
```diff
@@ -717,10 +717,10 @@ def prepare_transaction(
...
-    def reissue_transaction_until_included(
+    def await_transaction_acceptance(
             self, transaction_id: TransactionId, interval=None, max_attempts=None) -> BlockId:
-        """Reissues a transaction sent from the wallet for a provided transaction id until it's
-        included (referenced by a milestone). Returns the included block id.
+        """Checks the transaction state for a provided transaction id until it's accepted. Interval in milliseconds. Returns the block id that
+        contains this transaction.
         """
         return BlockId(self._call_method(
             'reissueTransactionUntilIncluded', {
```
thibault-martinez: :trollface:

sdk/src/wallet/operations/reissue.rs
Alex6323: sounds like if `block id` isn't there then it's a bug in the sdk? so panicking is probably what we want then.

bindings/nodejs/examples/how_tos/advanced_transactions/advanced_transaction.ts
Thoralf-M: Not really wrong though `'accepted': the transaction is included.`:trollface:

sdk/src/wallet/operations/reissue.rs
```diff
@@ -16,48 +16,69 @@ use crate::{
...
+            .ok_or_else(|| Error::TransactionNotFound(*transaction_id))?;
 
-        if let Some(transaction) = transaction {
-            if transaction.inclusion_state == InclusionState::Confirmed {
-                return transaction.block_id.ok_or(Error::MissingParameter("block id"));
-            }
+        if transaction.inclusion_state == InclusionState::Accepted
+            || transaction.inclusion_state == InclusionState::Confirmed
+            || transaction.inclusion_state == InclusionState::Finalized
+        {
```
Alex6323: would it make sense to add an `is_accepted` method (and maybe others), that returns true if `self == Accepted | Confirmed | Finalized` (pseudocode) since certain states imply others?

Thoralf-M: Currently I think it's not worth it, so far this is the only place where it's like that and in other places you probably want to differentiate more

sdk/src/wallet/operations/reissue.rs
thibault-martinez: I am not super opinionated but I'm fine with panicking if the others prefer that

sdk/src/wallet/operations/reissue.rs
```diff
@@ -66,76 +87,43 @@ where
...
-                            ClientError::TangleInclusion(first_block_id.to_string())
-                        } else {
-                            e
-                        }
-                    })?;
-                    return Ok(included_block.id(&protocol_parameters));
-                }
+                })?;
+                let protocol_parameters = self.client().get_protocol_parameters().await?;
+                return Ok(included_block.id(&protocol_parameters));
```
Alex6323: maybe we can put this in an anonymous function, for example:
```rust
let handle_failed = || { ... }
```
and call this in L109 instead of this `failed` boolean. Bit annoying with the `return` but if it's not too ugly then might be worth it.

sdk/src/wallet/operations/reissue.rs
```diff
@@ -16,48 +16,69 @@ use crate::{
...
 where
     Error: From<S::Error>,
     crate::client::Error: From<S::Error>,
 {
-    /// Reissues a transaction sent from the account for a provided transaction id until it's
-    /// included (referenced by a milestone). Returns the included block id.
-    pub async fn reissue_transaction_until_included(
+    /// Checks the transaction state for a provided transaction id until it's accepted. Interval in milliseconds.
+    /// Returns the block id that contains this transaction.
+    pub async fn await_transaction_acceptance(
```
Alex6323: bit nitpicky but maybe the `await_` prefix is not so good since it's one of Rust's keywords and `await_transaction_acceptance.await?` reads weird. Therefore I suggest something like:
```rust
async fn block_on_transaction_acceptance(...)
```

sdk/src/wallet/operations/reissue.rs
Thoralf-M: I'm not sure what that's supposed to achieve, can you maybe do a full code suggestion?

sdk/src/wallet/operations/reissue.rs
Thoralf-M: Removed it

sdk/src/wallet/operations/reissue.rs
Thoralf-M: Changed

sdk/src/wallet/operations/reissue.rs
thibault-martinez: I don't know about that tbh... that function doesn't really block, that's what await means. Whether it's blocking or not depends on the caller.

Thoralf-M: I don't really care, can also undo, maybe let's wait for an opinion from @DaughterOfMars

Alex6323: `await` means that it cannot make immediate progress, hence is blocked by some other resource to become available. It could be blocked by some required client request to be answered or some lock to become available. Typically if there are other tasks, then the runtime switches to another, which means that the overall application isn't blocked, but I get your point. I called it like that because in most examples in order to continue with it we need to wait for the transaction to get accepted, hence the example is blocked from further continuing until that happened. I still think you're probably right, and it should be named differently. Maybe there's no good alternative :shrug:

thibault-martinez: > I called it like that because in most examples in order to continue with it we need to wait for the transaction to get accepted, hence the example is blocked from further continuing until that happened.

That's exactly what I'm saying, this is a caller behaviour. If `block_on` is in the name, I understand that it doesn't yield control to the scheduler, but it absolutely does here. But maybe it's just me, idk.

Alex6323: yup, I agree that `block_on` can't be used anymore in another context. So that would be confusing. Maybe `wait_on`, but that's almost not any better than the `await_` prefix? So ... :shrug:

sdk/src/wallet/operations/reissue.rs
Alex6323: okay

sdk/src/wallet/operations/reissue.rs
Alex6323: hmm, had a 2nd look and what I suggested is not even necessary. Can you not just move that code executed if `failed == true` into the `TransactionState::Failed` match branch itself?

sdk/src/wallet/operations/reissue.rs
Alex6323: Please revert :trollface:

sdk/src/wallet/operations/reissue.rs
Thoralf-M: Makes sense, done

sdk/src/wallet/operations/reissue.rs
DaughterOfMars: well that issue is probably not an issue here, but ok

DaughterOfMars:
I don't really like `BlockOnTransactionAcceptance`, what about `WaitForTransactionAcceptance`?

thibault-martinez:
> I don't really like `BlockOnTransactionAcceptance`, what about `WaitForTransactionAcceptance`?

@DaughterOfMars it has been discussed here https://github.com/iotaledger/iota-sdk/pull/1979#discussion_r1485938851 if you want to comment

sdk/src/wallet/operations/reissue.rs
DaughterOfMars: I do think `WaitFor` is better than `Await` in this case.

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
```diff
@@ -0,0 +1,97 @@
...
+            .data()
+            .await
+            .transactions
+            .get(transaction_id)
+            .cloned()
+            .ok_or_else(|| Error::TransactionNotFound(*transaction_id))?;
+
+        if transaction.inclusion_state == InclusionState::Accepted
+            || transaction.inclusion_state == InclusionState::Confirmed
+            || transaction.inclusion_state == InclusionState::Finalized
```
DaughterOfMars:
```suggestion
        if matches!(transaction.inclusion_state, InclusionState::Accepted | InclusionState::Confirmed | InclusionState::Finalized)
```

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            return Ok(transaction
+                .block_id
+                // Safe to unwrap, we always set the block id together with any of these inclusion states.
+                .expect("missing block id in TransactionWithMetadata"));
+        }
+
+        if transaction.inclusion_state == InclusionState::Conflicting
+            || transaction.inclusion_state == InclusionState::UnknownPruned
```
DaughterOfMars:
```suggestion
        if matches!(transaction.inclusion_state, InclusionState::Conflicting | InclusionState::UnknownPruned)
```

DaughterOfMars (dismissed):

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
```diff
@@ -0,0 +1,98 @@
...
+    types::{
+        api::core::TransactionState,
+        block::{payload::signed_transaction::TransactionId, BlockId},
+    },
+    wallet::{types::InclusionState, Error, Wallet},
+};
+
+// Time in milliseconds
+const DEFAULT_AWAIT_TX_ACCEPTANCE_INTERVAL: u64 = 500;
+const DEFAULT_AWAIT_TX_ACCEPTANCE_MAX_AMOUNT: u64 = 80;
```
thibault-martinez: Do you want to adapt these?

sdk/src/wallet/operations/mod.rs
```diff
@@ -18,9 +18,9 @@ pub(crate) mod output_consolidation;
 /// The module for participation
 #[cfg(feature = "participation")]
 pub(crate) mod participation;
-/// The module for reissuing blocks or transactions
-pub(crate) mod reissue;
 /// The module for synchronization of the wallet
 pub(crate) mod syncing;
 /// The module for transactions
 pub(crate) mod transaction;
+/// The module for awaiting on transaction acceptance
```
thibault-martinez: Adapt?

Thoralf-M: Done, not sure if that's good though

DaughterOfMars (dismissed):

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
```diff
@@ -0,0 +1,98 @@
...
+                transaction_id, transaction.inclusion_state
+            ))
+            .into());
+        }
+
+        let block_id = transaction
+            .block_id
+            .ok_or_else(|| ClientError::TransactionAcceptance(transaction_id.to_string()))?;
+
+        let duration = std::time::Duration::from_millis(interval.unwrap_or(DEFAULT_WAIT_FOR_TX_ACCEPTANCE_INTERVAL));
```
thibault-martinez: Nit but `Duration::from_millis` is const so you could have that as constant instead of the value

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
```diff
@@ -0,0 +1,98 @@
...
+        for _ in 0..max_attempts.unwrap_or(DEFAULT_WAIT_FOR_TX_ACCEPTANCE_MAX_AMOUNT) {
+            #[cfg(target_family = "wasm")]
+            gloo_timers::future::TimeoutFuture::new(duration.as_millis() as u32).await;
+
+            #[cfg(not(target_family = "wasm"))]
+            tokio::time::sleep(duration).await;
```
thibault-martinez: Should we put the sleep after the checks? Just in case the first API call is enough

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
```diff
@@ -0,0 +1,98 @@
...
+
+            let block_metadata = self.client().get_block_metadata(&block_id).await?;
+            if let Some(transaction_state) = block_metadata.transaction_metadata.map(|m| m.transaction_state) {
+                match transaction_state {
+                    TransactionState::Accepted | TransactionState::Confirmed | TransactionState::Finalized => {
+                        return Ok(block_id);
+                    }
+                    TransactionState::Failed => {
+                        // Check if the transaction got reissued in another block and confirmed there
+                        let included_block = self.client().get_included_block(transaction_id).await.map_err(|e| {
```
thibault-martinez: Call `get_included_block_metadata` to avoid having to compute the block ID ?

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
Thoralf-M: But the interval is provided as param, so you think it should be `interval.map(std::time::Duration::from_millis).unwrap_or(DEFAULT_WAIT_FOR_TX_ACCEPTANCE_INTERVAL)`?

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
Thoralf-M: Good call

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
thibault-martinez: Somehow my brain didn't parse `interval`, just ignore me

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
Thoralf-M: Probably isn't, but moved it, doesn't matter much

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
Thoralf-M: Did it this way now

thibault-martinez (dismissed):

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
```diff
@@ -10,8 +12,7 @@ use crate::{
     wallet::{types::InclusionState, Error, Wallet},
 };
 
-// Time in milliseconds
-const DEFAULT_WAIT_FOR_TX_ACCEPTANCE_INTERVAL: u64 = 500;
+const DEFAULT_WAIT_FOR_TX_ACCEPTANCE_INTERVAL: Duration = Duration::from_millis(500);
 const DEFAULT_WAIT_FOR_TX_ACCEPTANCE_MAX_AMOUNT: u64 = 80;
```
DaughterOfMars: `DEFAULT_WAIT_FOR_TX_ACCEPTANCE_MAX_ATTEMPTS`?

sdk/src/wallet/operations/wait_for_tx_acceptance.rs
```diff
@@ -62,14 +63,10 @@ where
...
-            #[cfg(target_family = "wasm")]
-            gloo_timers::future::TimeoutFuture::new(duration.as_millis() as u32).await;
-
-            #[cfg(not(target_family = "wasm"))]
-            tokio::time::sleep(duration).await;
-
+        let duration = interval
+            .map(std::time::Duration::from_millis)
+            .unwrap_or(DEFAULT_WAIT_FOR_TX_ACCEPTANCE_INTERVAL);
+        for _ in 0..max_attempts.unwrap_or(DEFAULT_WAIT_FOR_TX_ACCEPTANCE_MAX_AMOUNT) {
```
DaughterOfMars: not 100% sure but I think this might do the option check every iteration for no reason

DaughterOfMars: actually, no way it does that

thibault-martinez (approved):

DaughterOfMars (approved):

PR #1980, opened 2024-02-09
Change the way permanodes are handled

### LINKED ISSUE #1620: Keep permanode support as it is?
Do we want permanode support in the client like it is now or should we change/remove it? Currently it's a bit messy, one can add a permanode with `with_permanode()` and in `get_request()` have a `prefer_permanode` bool parameter we hardcoded for each endpoint, which will then use a permanode first if available and otherwise a normal node.
Nodeinfo doesn't work because the response is different, posting a block and getting tips will never be possible, but for users it's not clear what is supported or not.
Permanodes also have other API endpoints one can call with `call_plugin_route()`, but currently only if one would provide a permanode as normal node, because there `prefer_permanode` is set to false

Thoralf-M:
# Description of change

Change the way permanodes are handled by adding a permanode field to the Node struct and removing extra permanode options in the client options. By using a combination of primary nodes and nodes one is able to control in what order nodes/permanodes will be queried

Fixes https://github.com/iotaledger/iota-sdk/issues/1620

## Type of change

```TS
import { Client, initLogger } from '@iota/sdk';

async function run() {
    initLogger();

    const client = await Client.create({
        primaryNodes: [{ url: "https://some-chronicle-node", permanode: true }],
    });

    try {
        const theBlock = await client.getBlock("0xa59f5ed1a51e12e36cd2ca84c57503819c6bc45b8e85ebdec389b98e7865665f93240000");
        console.log(theBlock);
    } catch (error) {
        console.error('Error: ', error);
    }
}

run().then(() => process.exit());
```
prints the block and logs 
```
2024-02-15 09:10:39 (UTC) iota_sdk::client::node_manager::syncing    [94mDEBUG[0m sync_nodes
2024-02-15 09:10:39 (UTC) iota_sdk::client::node_manager::http_client [94mDEBUG[0m GET: 132 ms for 200 OK https://some-chronicle-node/api/core/v3/info
2024-02-15 09:10:39 (UTC) iota_sdk_bindings_core::method_handler::call_method [94mDEBUG[0m Client method: GetBlock { block_id: BlockId { id: "0xa59f5ed1a51e12e36cd2ca84c57503819c6bc45b8e85ebdec389b98e7865665f93240000", slot_index: SlotIndex(9363) } }
2024-02-15 09:10:40 (UTC) iota_sdk::client::node_manager::http_client [94mDEBUG[0m GET: 210 ms for 200 OK https://some-chronicle-node/api/core/v3/blocks/0xa59f5ed1a51e12e36cd2ca84c57503819c6bc45b8e85ebdec389b98e7865665f93240000
2024-02-15 09:10:40 (UTC) iota_sdk_bindings_core::method_handler::call_method [94mDEBUG[0m Client response: Block(BlockDto { inner: UnsignedBlockDto { header: BlockHeaderDto { protocol_version: 3, network_id: 3650798313638353144, issuing_time: 1707987767280827038, slot_commitment_id: SlotCommitmentId { id: "0xcf8588654d9758cef4b01531964fed09674832c0c3281c147d7f5e4e31a9f1ed89240000", slot_index: SlotIndex(9353) }, latest_finalized_slot: SlotIndex(9353), issuer_id: AccountId(0x486e8dca1297e572f2a02ca0fba07fc3f1c6ce0037ff2fb48a06ef90a16955c6) }, body: Validation(ValidationBlockBodyDto { kind: 1, strong_parents: {BlockId { id: "0x6525c0af76d97e79fe4489f1d93a2596fff64a27eb800d964f23167d2174d71393240000", slot_index: SlotIndex(9363) }}, weak_parents: {}, shallow_like_parents: {}, highest_supported_version: 3, protocol_parameters_hash: ProtocolParametersHash(0xfe2abe6c1360ed5079f1a0f5ad8830238bd1b63face69f2945b00fd0a0720875) }) }, signature: Ed25519Signature { public_key: 0x64f38ad410566cb3a5b9f5555e9bb1a6e5cd3d316f9df622c9169d3f8a51e6c2, signature: 0xbe9da2d9adab6fb9b30fbecbb628ad1a9fc8f95f42b963bbae04232b2004d33a3fed968db77cb279278a23d8756babb86b66f8f6cbb6ec65e6164e23f0b6e005 } })
```

```py
from iota_sdk import Client, Node, BlockId

client = Client(
    primary_nodes=[Node("https://some-chronicle-node", permanode=True)])

block = client.get_block(
    BlockId('0x64436506c09999442fb51ffe9e7e5c2f7cd397845f7903d05f12d1e136702ffb90000000'))
print(block)
```

```Rust
use std::str::FromStr;

use iota_sdk::{
    client::{node_manager::node::Node, Client, Result},
    types::block::BlockId,
};
use url::Url;

#[tokio::main]
    let client = Client::builder()
        .with_primary_node(Node {
            url: Url::parse("https://some-chronicle-node")?,
            auth: None,
            disabled: false,
            permanode: true,
        })?
        .with_ignore_node_health()
        .finish()
        .await?;

    let block = client
        .get_block(&BlockId::from_str(
            "0x64436506c09999442fb51ffe9e7e5c2f7cd397845f7903d05f12d1e136702ffb90000000",
        )?)
        .await?;

    println!("{block:?}");

    Ok(())
}
```


sdk/src/client/node_manager/builder.rs
```diff
@@ -87,45 +77,36 @@ impl NodeManagerBuilder {
...
-    pub(crate) fn with_primary_node(mut self, url: &str, auth: Option<NodeAuth>) -> Result<Self> {
-        let mut url = validate_url(Url::parse(url)?)?;
-        if let Some(auth) = &auth {
-            if let Some((name, password)) = &auth.basic_auth_name_pwd {
-                url.set_username(name)
-                    .map_err(|_| crate::client::Error::UrlAuth("username"))?;
-                url.set_password(Some(password))
-                    .map_err(|_| crate::client::Error::UrlAuth("password"))?;
-            }
+    pub(crate) fn with_primary_nodes(mut self, urls: &[&str]) -> Result<Self> {
```
DaughterOfMars, line 88: where auth go

DaughterOfMars (dismissed):
nice!

sdk/src/client/node_manager/builder.rs
Thoralf-M: I have decided to have it only in `with_primary_node()` in the `Node` struct, otherwise the array here gets ugly

DaughterOfMars: okie dokie

bindings/nodejs/lib/types/client/network.ts
```diff
@@ -49,6 +49,8 @@ export interface INode {
     auth?: IAuth;
     /** Whether the node is disabled or not. */
     disabled?: boolean;
+    /** Whether the node is a permanode or not. */
+    permanode?: boolean;
```
thibault-martinez, line 53: Does it not need to be `Option<>` in Rust to be equivalent to `?` ? Or does `#[serde(default)]` achieve the same?

bindings/python/iota_sdk/client/client.py
```diff
@@ -33,8 +33,7 @@ class Client(NodeCoreAPI, NodeIndexerAPI, HighLevelAPI, ClientUtils):
     def __init__(
         self,
         nodes: Optional[Union[str, List[str]]] = None,
-        primary_node: Optional[str] = None,
-        permanode: Optional[str] = None,
+        primary_nodes: Optional[Union[str, List[str]]] = None,
```
thibault-martinez: Can you reverse nodes and primary_nodes please?

sdk/src/client/node_manager/syncing.rs
```diff
@@ -66,62 +68,104 @@ impl ClientInner {
...
-                    .protocol_parameters_by_version(PROTOCOL_VERSION)
-                    .expect("missing v3 protocol parameters")
-                    .parameters
-                    .clone();
+                network_info.protocol_parameters = parameters.clone();
             }
 
-            for (info, node_url) in nodes {
-                healthy_nodes.insert(node_url.clone(), info.clone());
+            for (_info, node_url, _tangle_time) in nodes {
```
Alex6323: Not a big deal, but I think we could do something like this:
```rust
healthy_nodes.extend(nodes.iter().map(|n| n.node_url).cloned())
```

bindings/nodejs/lib/types/client/network.ts
Thoralf-M: For JS -> Rust it makes no difference, other way around permanode will never be undefined, but that's fine

sdk/src/client/node_api/core/routes.rs
```diff
@@ -376,19 +377,49 @@ impl ClientInner {
...
 
-        self.get_request(path, None, false, false).await
+        self.get_request(path, None, false).await
     }
 }
 
 impl Client {
+    /// GET /api/core/v3/info endpoint
+    pub(crate) async fn get_permanode_info(node: &Node) -> Result<PermanodeInfoResponse> {
+        let mut node = node.clone();
```
Alex6323: bit weird to expect a `&Node` when we actually need a clone. Can we not just expect the same arguments and create a `Node` struct as in `get_node_info`?

Thoralf-M: We need a clone in any case because of the url, but I changed it so this function accepts `Node`

sdk/src/client/node_manager/builder.rs
```diff
@@ -87,45 +77,36 @@ impl NodeManagerBuilder {
...
-            disabled: false,
-        }));
         Ok(self)
     }
 
-    pub(crate) fn with_permanode(mut self, url: &str, auth: impl Into<Option<NodeAuth>>) -> Result<Self> {
-        let mut url = validate_url(Url::parse(url)?)?;
-        let auth = auth.into();
-        if let Some(auth) = &auth {
+    pub(crate) fn with_primary_node(mut self, mut node: Node) -> Result<Self> {
```
Alex6323: Ignore if you want, but if you can spare the time you could do a bit of cleanup in this PR and order the methods a bit: I'ld suggest something like `with_node`, `with_nodes`, `with_primary_node`, `with_primary_nodes`. The rest of the pack is probably relatively arbitrary.

Thoralf-M: Reordered them, but a bit different

sdk/src/client/node_manager/syncing.rs
```diff
@@ -66,63 +68,103 @@ impl ClientInner {
...
-        for (network_id, node) in &network_nodes {
-            if node.len() > most_nodes.1 {
-                most_nodes.0 = network_id;
-                most_nodes.1 = node.len();
+        // Get the network_name that most nodes have in common
+        let mut most_nodes = ("network_name", 0);
+        for (network_name, nodes) in &network_nodes {
+            if nodes.len() > most_nodes.1 {
+                most_nodes.0 = network_name;
+                most_nodes.1 = nodes.len();
```
Alex6323: this can be simplified into a one-liner by using `max_by_key`

Alex6323 (dismissed):

sdk/src/client/node_manager/syncing.rs
Thoralf-M: Good call, can actually simplify even more with the part below

Alex6323: yeah, looks good

Alex6323 (approved):

sdk/src/client/node_api/core/routes.rs
```diff
@@ -57,10 +57,11 @@ impl ClientInner {
         url.set_path(PATH);
         let status = crate::client::node_manager::http_client::HttpClient::new(DEFAULT_USER_AGENT.to_string())
             .get(
-                Node {
+                &Node {
                     url,
                     auth: None,
                     disabled: false,
+                    permanode: false,
```
thibault-martinez, line 64: permanode doesn't have a health route?

Thoralf-M: Doesn't matter here, we just need to set something

thibault-martinez: Alright

DaughterOfMars (approved):

thibault-martinez (approved):

Issue #1918, opened 2024-01-30
ISA: allow account bound mana to be used as allotment


thibault-martinez:
If that makes sense maybe at least put a TODO or a debug log, otherwise might be hard to find later why this amount is not available

_Originally posted by @Thoralf-M in https://github.com/iotaledger/iota-sdk/pull/1900#discussion_r1471744917_

Thoralf-M:
Done in https://github.com/iotaledger/iota-sdk/pull/1961

PR #1986, opened 2024-02-13
Allow using account mana for allotments, fix semantic validation

Thoralf-M:
# Description of change

Allow using account Mana for allotments, fix semantic validation (only add allotments Mana to output Mana once)

#1918 

## Type of change

Sending a transaction with only accounts as inputs

### DIFF
--- a/sdk/src/client/api/block_builder/input_selection/requirement/allotment.rs
+++ b/sdk/src/client/api/block_builder/input_selection/requirement/allotment.rs
@@ -13,7 +13,7 @@ use crate::{
         address::Address,
         input::{Input, UtxoInput},
         mana::ManaAllotment,
-        output::{AccountOutputBuilder, Output},
+        output::{AccountOutputBuilder, ChainId, NftOutputBuilder, Output},
         payload::{signed_transaction::Transaction, SignedTransactionPayload},
         signature::Ed25519Signature,
         unlock::{AccountUnlock, NftUnlock, ReferenceUnlock, SignatureUnlock, Unlock, Unlocks},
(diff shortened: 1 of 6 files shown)


sdk/src/types/block/semantic/mod.rs
```diff
@@ -221,6 +221,14 @@ impl<'a> SemanticValidationContext<'a> {
             }
         }
 
+        // Add allotted mana
+        for mana_allotment in self.transaction.allotments() {
```
thibault-martinez, line 225: Super nit picking here but can you put it right after the outputs loop instead of before pls?

Thoralf-M: Why? In the tx the allotments are before outputs (I'm still fine with changing it, also initially thought about moving it below, only did not because of the tx structure)

thibault-martinez: No worries, just works better in my head for some reason, ignore me.

sdk/src/client/api/block_builder/input_selection/requirement/allotment.rs
```diff
@@ -68,27 +67,96 @@ impl InputSelection {
...
-            {
-                log::debug!(
-                    "Reducing account mana of {} by {required_allotment_mana} for allotment",
-                    output.as_account().account_id()
-                );
-                let new_mana = output.mana() - required_allotment_mana;
-                *output = match output {
-                    Output::Account(a) => AccountOutputBuilder::from(&*a).with_mana(new_mana).finish_output()?,
+        if self.mana_allotments[&issuer_id].0 < required_allotment_mana {
+            let (selected_mana, required_mana) = self.mana_sums_without_allotments()?;
```
DaughterOfMars, line 71: I think you'd be better off using the regular method and subtracting the specific allotment for this issuer

Thoralf-M: Yes, that's one reason why I wrote `still needs cleanup`

Thoralf-M:
Commits were partially picked for https://github.com/iotaledger/iota-sdk/pull/1961

PR #1997, opened 2024-02-14
TransactionFailureReason updates

### LINKED ISSUE #1954: Remaining TransactionFailureReason questions

thibault-martinez:
Fixes https://github.com/iotaledger/iota-sdk/issues/1954


DaughterOfMars (dismissed):

bindings/python/iota_sdk/types/transaction_metadata.py
```diff
@@ -25,96 +25,98 @@ class TransactionFailureReason(Enum):
...
-            13: "Under- or overflow in Mana calculations.",
-            14: "Inputs and outputs do not contain the same amount of Mana.",
-            15: "Mana decay creation slot/epoch index exceeds target slot/epoch index.",
-            16: "Native token amount must be greater than zero.",
-            17: "Native token sum exceeds max value of a uint256.",
-            18: "Native token sums are unbalanced.",
-            19: "Multi address length and multi unlock length do not match.",
-            20: "Multi address unlock threshold not reached.",
-            21: "Multi unlocks can't be nested.",
+            1: "Transaction is conflicting.",
```
Thoralf-M: In Rust this is `transaction was conflicting and was rejected.`

thibault-martinez: Good catch, changed

Thoralf-M (approved):

DaughterOfMars (approved):

PR #2008, opened 2024-02-16
Update TransactionFailureReason

thibault-martinez:
https://github.com/iotaledger/iota.go/pull/692


sdk/src/types/block/semantic/unlock.rs
```diff
@@ -13,58 +13,63 @@ impl SemanticValidationContext<'_> {
...
 
-                if signature
-                    .is_valid(self.transaction_signing_hash.as_ref(), ed25519_address)
-                    .is_err()
-                {
-                    return Err(TransactionFailureReason::UnlockSignatureInvalid);
-                }
+                    if signature
+                        .is_valid(self.transaction_signing_hash.as_ref(), ed25519_address)
+                        .is_err()
```
DaughterOfMars: hmm, weird. Does `is_valid` also return a bool we should check?

thibault-martinez: `pub fn is_valid(&self, message: &[u8], address: &Ed25519Address) -> Result<(), Error>`

thibault-martinez: Can fail for multiple reasons

DaughterOfMars: probably shouldn't be called `is_valid` then, unless this isn't in our lib?

sdk/src/types/block/semantic/unlock.rs
```diff
@@ -13,58 +13,63 @@ impl SemanticValidationContext<'_> {
...
-            }
-            (Address::Account(account_address), Unlock::Account(unlock)) => {
-                // PANIC: indexing is fine as it is already syntactically verified that indexes reference below.
-                if let (output_id, Output::Account(account_output)) = self.inputs[unlock.index() as usize] {
-                    if &account_output.account_id_non_null(output_id) != account_address.account_id() {
-                        // TODO https://github.com/iotaledger/iota-sdk/issues/1954
-                        return Err(TransactionFailureReason::SemanticValidationFailed);
-                    }
+                Unlock::Reference(_) => {
+                    // TODO actually check that it was unlocked by the same signature.
```
Thoralf-M, line 34: I know it was there before already, but what does this mean?

sdk/src/types/block/semantic/unlock.rs
thibault-martinez: What would would you call it?

sdk/src/types/block/semantic/unlock.rs
```diff
@@ -13,58 +13,63 @@ impl SemanticValidationContext<'_> {
...
-                    }
-                    if !self.unlocked_addresses.contains(address) {
-                        // TODO https://github.com/iotaledger/iota-sdk/issues/1954
-                        return Err(TransactionFailureReason::SemanticValidationFailed);
+            (Address::Nft(nft_address), unlock) => {
+                if let Unlock::Nft(unlock) = unlock {
+                    // PANIC: indexing is fine as it is already syntactically verified that indexes reference below.
+                    if let (output_id, Output::Nft(nft_output)) = self.inputs[unlock.index() as usize] {
+                        if &nft_output.nft_id_non_null(output_id) != nft_address.nft_id() {
+                            return Err(TransactionFailureReason::ChainAddressUnlockInvalid);
```
Thoralf-M, line 63: This is really the correct error here? Same for accounts

sdk/src/types/block/semantic/unlock.rs
DaughterOfMars: `validate` probably

sdk/src/types/block/semantic/unlock.rs
Thoralf-M: Ah I guess yes

thibault-martinez: I would say so, yes

sdk/src/types/block/semantic/unlock.rs
thibault-martinez: Ok I can change it

thibault-martinez: Done

DaughterOfMars (approved):

Thoralf-M (approved):

sdk/src/types/block/semantic/unlock.rs
Thoralf-M: The referenced signature could be different, so we need to get it by index and compare the hashed public key with the pubkeyhash of the address

Issue #1883, opened 2024-01-25
Add TIP API response tests to Python and Nodejs


thibault-martinez:
And re-enable https://github.com/iotaledger/iota-sdk/pull/1687/files#diff-e55a410885ce608d553cac1a8ae845c88684ced36475b8747771c262c049378dR34

Thoralf-M:
Done in Python, for nodejs we test the protocol parameters here https://github.com/iotaledger/iota-sdk/blob/9ad22f7556a5510a014577b83a145794a530ff37/bindings/nodejs/tests/client/infoMethods.spec.ts#L79 and for the other types we would need to rewrite it with something like https://zod.dev/?id=objects as the interfaces aren't available at runtime and that's not worth it just for this test

PR #2009, opened 2024-02-16
Python: add API responses test

Thoralf-M:
# Description of change

Python: add API responses test

#1883 

## Type of change

Running the test


Alex6323 (approved):

bindings/python/tests/test_api_responses.py
```diff
@@ -0,0 +1,91 @@
...
+    #                   "get-transaction-metadata-by-id-response-example.json")
+    # GET /api/core/v3/commitments/{commitmentId}
+    test_api_response(SlotCommitment, "get-commitment-response-example.json")
+    # GET /api/core/v3/commitments/{commitmentId}/utxo-changes
+    # TODO: enable when https://github.com/iotaledger/iota-sdk/issues/2020 is fixed
+    # test_api_response(UtxoChangesResponse,
+    #                   "get-utxo-changes-response-example.json")
+    # GET /api/core/v3/commitments/{commitmentId}/utxo-changes/full
+    # test_api_response(UtxoChangesFullResponse,
+    #   "get-utxo-changes-full-response-example.json")
```
thibault-martinez, line 91: There is so much we can't do atm?

thibault-martinez: Ah, didn't see there was a comment for each disabled test :(

bindings/python/iota_sdk/types/node_info.py
```diff
@@ -279,6 +279,7 @@ class ProtocolParameters:
     bech32_hrp: str
     storage_score_parameters: StorageScoreParameters
     work_score_parameters: WorkScoreParameters
+    mana_parameters: ManaParameters
```
thibault-martinez, line 282: Not like it matters but does order has an influence in Python?

bindings/python/tests/test_api_responses.py
Thoralf-M: Yeah there are a few things that prevent a lot to pass

bindings/python/iota_sdk/types/node_info.py
Thoralf-M: When you create the class that parameters are ordered like this and also to json will use it, apart from that optional alvalues must be last, but that's about it

This here I only noticed because I initially tried to compare the json without sorting (which then didn't work because of the optional fields)

bindings/python/tests/test_api_responses.py
```diff
@@ -0,0 +1,91 @@
...
+from typing import Generic, TypeVar
+from json import load, loads, dumps
+from iota_sdk import RoutesResponse, CongestionResponse, ManaRewardsResponse, ValidatorResponse, ValidatorsResponse, CommitteeResponse, IssuanceBlockHeaderResponse, Block, OutputResponse, SlotCommitment
+
+
+base_path = '../../sdk/tests/types/api/fixtures/'
+T = TypeVar("T")
+
+
+def test_api_responses():
```
thibault-martinez, line 13: Just curious, we don't have `SubmitBlockResponse` in Python?

bindings/python/tests/test_api_responses.py
```diff
@@ -0,0 +1,91 @@
...
+    def test_api_response(cls_type: Generic[T], path: str):
```
thibault-martinez, line 14: I think that's the first time I see some kind of generic in Python

thibault-martinez (approved):

bindings/python/tests/test_api_responses.py
thibault-martinez: I guess bindings core just return the ID

Thoralf-M: Correct

bindings/python/tests/test_api_responses.py
Thoralf-M: Probably because you don't really need them if you ignore the type checks :trollface:

PR #2011, opened 2024-02-16
Validate if address has been unlocked by same signature

### LINKED ISSUE #2010: Validate if from reference unlock referenced signature is the correct one
https://github.com/iotaledger/iota-sdk/pull/2008#discussion_r1492469439

thibault-martinez:
Fixes #2010

### DIFF
--- a/sdk/src/types/block/semantic/unlock.rs
+++ b/sdk/src/types/block/semantic/unlock.rs
@@ -1,6 +1,8 @@
 // Copyright 2023 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use crypto::hashes::{blake2b::Blake2b256, Digest};
+
 use crate::types::block::{
     address::Address,
     output::{Output, OutputId},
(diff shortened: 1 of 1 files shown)


sdk/src/types/block/semantic/unlock.rs
```diff
@@ -30,11 +32,18 @@ impl SemanticValidationContext<'_> {
 
                     self.unlocked_addresses.insert(address.clone());
                 }
-                Unlock::Reference(_) => {
-                    // TODO actually check that it was unlocked by the same signature.
+                Unlock::Reference(unlock) => {
                     if !self.unlocked_addresses.contains(address) {
                         return Err(TransactionFailureReason::DirectUnlockableAddressUnlockInvalid);
                     }
```
Thoralf-M, line 38: This we can remove then?

thibault-martinez: No? We are not actually validating the signature here

sdk/src/types/block/semantic/unlock.rs
```diff
@@ -30,11 +32,18 @@ impl SemanticValidationContext<'_> {
...
+
+                    let Signature::Ed25519(signature) = self.unlocks.unwrap()[unlock.index() as usize]
```
Thoralf-M, line 41: Is this safe to index at this point?

thibault-martinez: Yes, but I'll put a comment

sdk/src/types/block/semantic/unlock.rs
Thoralf-M: True, but if there is a signature unlock with an invalid signature it would also fail when validating it, but fine

sdk/src/types/block/semantic/unlock.rs
thibault-martinez: Done

Thoralf-M (approved):

DaughterOfMars (approved):

PR #2026, opened 2024-02-19
Enable Nova Protocol on Ledger Nano App

shufps:
Enabled Nova Protocol for Ledger Nano

### DIFF
--- a/sdk/src/client/secret/ledger_nano.rs
+++ b/sdk/src/client/secret/ledger_nano.rs
@@ -16,8 +16,8 @@ use crypto::{
     },
 };
 use iota_ledger_nano::{
-    api::errors::APIError, get_app_config, get_buffer_size, get_ledger, get_opened_app, LedgerBIP32Index,
-    Packable as LedgerNanoPackable, TransportTypes,
+    api::constants::Protocols, api::errors::APIError, get_app_config, get_buffer_size, get_ledger, get_opened_app,
+    LedgerBIP32Index, Packable as LedgerNanoPackable, TransportTypes,
 };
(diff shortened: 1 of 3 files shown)


sdk/src/client/secret/ledger_nano.rs
```diff
@@ -159,7 +160,7 @@ impl SecretManage for LedgerSecretManager {
         let lock = self.mutex.lock().await;
 
         // get ledger
-        let ledger = get_ledger(coin_type, bip32_account, self.is_simulator).map_err(Error::from)?;
+        let ledger = get_ledger(coin_type, Protocols::Nova, bip32_account, self.is_simulator).map_err(Error::from)?;
```
thibault-martinez: Nothing that should be blocking now but if ever you want to change in a future version: enum names are usually singular in Rust

thibault-martinez: And I would personally put that parameter first

thibault-martinez (dismissed):

Thoralf-M (dismissed):

sdk/src/client/secret/ledger_nano.rs
shufps: Ok, changed :blush:

shufps:
ah something happend ... sorry have to revert my last commit, it cargo fmted everything :scream:

Thoralf-M (approved):

thibault-martinez (approved):

sdk/src/client/secret/ledger_nano.rs
thibault-martinez: 🙏🏻

PR #2035, opened 2024-02-20
Fix bindings types

### LINKED ISSUE #1975: Fix remainderValueStrategy CustomAddress type in bindings
Fix remainderValueStrategy CustomAddress type in bindings, it's just an address now and not Bip44Address

Thoralf-M:
# Description of change

Fix RemainderValueStrategy in bindings, fix balance in Python

 Fixes https://github.com/iotaledger/iota-sdk/issues/1975 

## Type of change

Running examples
```
transaction = wallet.send_with_params(params, TransactionOptions(
    remainder_value_strategy=RemainderValueStrategy.ReuseAddress))

transaction = wallet.send_with_params(params, TransactionOptions(remainder_value_strategy=RemainderValueStrategyCustomAddress(
    Utils.parse_bech32_address("rms1qpszqzadsym6wpppd6z037dvlejmjuke7s24hm95s9fg9vpua7vluaw60xu"))))


const response = await wallet.send(amount, address, { remainderValueStrategy: { strategy: 'ReuseAddress' } as ReuseAddress });
const response = await wallet.send(amount, address, { remainderValueStrategy: { strategy: 'CustomAddress', value: Utils.parseBech32Address('rms1qrrv7flg6lz5cssvzv2lsdt8c673khad060l4quev6q09tkm9mgtupgf0h0') } });

```

### DIFF
--- a/bindings/python/iota_sdk/types/transaction_options.py
+++ b/bindings/python/iota_sdk/types/transaction_options.py
@@ -3,7 +3,8 @@
 
 from enum import Enum
 from typing import Optional, List, Union
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+from iota_sdk.types.address import Address
 from iota_sdk.types.burn import Burn
 from iota_sdk.types.common import HexStr, json
 from iota_sdk.types.context_input import ContextInput
@@ -17,30 +18,10 @@ class RemainderValueStrategyCustomAddress:
     """Remainder value strategy for custom addresses.
 
     Attributes:
-        address: An address to move the remainder value to.
-        key_index: The address key index.
-        internal: Determines if an address is a public or an internal (change) address.
-        used: Indicates whether an address has been used already.
+        value: An address to move the remainder value to.
     """
-
-    address: str
-    key_index: int
-    internal: bool
-    used: bool
-
-    def to_dict(self) -> dict:
-        """Custom dict conversion.
-        """
-
-        return {
-            'strategy': 'CustomAddress',
-            'value': {
-                'address': self.address,
-                'keyIndex': self.key_index,
-                'internal': self.internal,
-                'used': self.used
-            }
-        }
+    strategy: str = field(default_factory=lambda: 'CustomAddress', init=False)
+    value: Address
 
 
 class RemainderValueStrategy(Enum):
@@ -57,7 +38,6 @@ class RemainderValueStrategy(Enum):
 
         return {
             'strategy': self.name,
-            'value': self.value[0]
         }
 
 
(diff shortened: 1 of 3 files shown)


bindings/python/iota_sdk/types/transaction_options.py
```diff
@@ -3,7 +3,8 @@
 
 from enum import Enum
 from typing import Optional, List, Union
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+from iota_sdk.types.address import Address
```
Alex6323, line 7: closing the gaps again I see :trollface:

bindings/python/iota_sdk/types/transaction_options.py
```diff
@@ -17,30 +18,10 @@ class RemainderValueStrategyCustomAddress:
     """Remainder value strategy for custom addresses.
 
     Attributes:
-        address: An address to move the remainder value to.
-        key_index: The address key index.
-        internal: Determines if an address is a public or an internal (change) address.
-        used: Indicates whether an address has been used already.
+        value: An address to move the remainder value to.
```
Alex6323, line 21: `strategy` missing?

bindings/python/iota_sdk/types/transaction_options.py
Thoralf-M: There was no gap?
Also if you ask me it's stupid to do if the formatter doesn't do it

bindings/python/iota_sdk/types/transaction_options.py
Thoralf-M: Not sure if it should be part of the comment since users don't set it

Thoralf-M: Can add it if you think it should still be there

bindings/python/iota_sdk/types/transaction_options.py
Alex6323: well it's also stupid to add two empty lines (and the formatter does THAT), but you're right ... there was only a gap line in the display on GH, not in the code.

bindings/python/iota_sdk/types/transaction_options.py
Alex6323: I don't care ... but I noticed that we do it sometimes.

Alex6323 (approved):

bindings/python/iota_sdk/types/transaction_options.py
```diff
@@ -57,7 +38,6 @@ def to_dict(self) -> dict:
 
         return {
             'strategy': self.name,
-            'value': self.value[0]
```
thibault-martinez, line 60: Why is value not part of the serialisation?

thibault-martinez: Oh the diff mislead me, this is RemainderValueStrategy

thibault-martinez: Why is it done like that btw? Why an enum just for ReuseAddress?

bindings/python/iota_sdk/types/transaction_options.py
```diff
@@ -79,28 +59,15 @@ class TransactionOptions:
...
-        self.required_inputs = required_inputs
-        self.burn = burn
-        self.note = note
-        self.allow_micro_amount = allow_micro_amount
-        self.allow_additional_input_selection = allow_additional_input_selection
-        self.capabilities = capabilities
-        self.mana_allotments = mana_allotments
-        self.issuer_id = issuer_id
+    remainder_value_strategy: Optional[Union[RemainderValueStrategy,
+                                             RemainderValueStrategyCustomAddress]] = None
```
thibault-martinez, line 63: Should we define this Union as an alias like we did for other enums?

bindings/python/iota_sdk/types/transaction_options.py
Thoralf-M: Yes the two empty lines are really stupid

bindings/python/iota_sdk/types/transaction_options.py
Thoralf-M: We had a second option to generate a new remainder address and didn't you also say that we might add a new remainder option in the future?
Currently I could think of a remainder option that will use the account address instead of the ed25519 one

bindings/python/iota_sdk/types/transaction_options.py
Thoralf-M: I think it's not worth it atm since this is the only place where it's used

bindings/python/iota_sdk/types/transaction_options.py
thibault-martinez: I'm confused, that's not really what I'm asking. We have `RemainderValueStrategy` which is an enum that has one `ReuseAddress` variant and then we have `RemainderValueStrategyCustomAddress` which is not part of the enum?

Thoralf-M: Ah it's not part of the enum because you can't have class things as enum variants

thibault-martinez: bullshit

thibault-martinez (approved):

PR #2036, opened 2024-02-20
Allotment touchups

### LINKED ISSUE #2006: ISA should not mutate user provided outputs
Currently done here https://github.com/iotaledger/iota-sdk/pull/1961/files#diff-2af52cf4dc5f8c01cbbf7ad942241735421cba6c689beeb1baf2c0204c672a26R138-R139
Needed for account transitions where no other inputs with enough mana are available for the allotment

DaughterOfMars:
# Description of change

In order to provide more fine-grained control over ISA, I have added a new flag to the transaction options to control whether ISA can use account mana for allotments. It will only remove mana for the minimum allotment and only from an account output with the `account_id == issuer_id`. This extends to the CLI, so that for staking operations users can now set this flag.

I also reworked the way that the mana requirement selects inputs so that it can also add the output mana to the equation. This means that the calculation will be the same regardless of whether the inputs are passed in or added later.

### DIFF
--- a/cli/src/wallet_cli/mod.rs
+++ b/cli/src/wallet_cli/mod.rs
@@ -80,6 +80,7 @@ pub enum WalletCommand {
         /// The staking period (in epochs). Will default to the staking unbonding period.
         staking_period: Option<u32>,
         /// Whether to allot mana from the account output.
+        #[arg(long, default_value_t = true)]
         allot_from_account: bool,
     },
     /// Burn an amount of native token.
@@ -167,6 +168,7 @@ pub enum WalletCommand {
         /// The Account ID of the staking account.
         account_id: AccountId,
         /// Whether to allot mana from the account output.
+        #[arg(long, default_value_t = true)]
         allot_from_account: bool,
     },
     /// Exit the CLI wallet.
@@ -178,6 +180,7 @@ pub enum WalletCommand {
         /// The number of additional epochs to add to the staking period.
         additional_epochs: u32,
         /// Whether to allot mana from the account output.
+        #[arg(long, default_value_t = true)]
         allot_from_account: bool,
     },
     /// Request funds from the faucet.
--- a/cli/src/wallet_cli/mod.rs
+++ b/cli/src/wallet_cli/mod.rs
@@ -79,6 +79,8 @@ pub enum WalletCommand {
         fixed_cost: u64,
         /// The staking period (in epochs). Will default to the staking unbonding period.
         staking_period: Option<u32>,
+        /// Whether to allot mana from the account output.
+        allot_from_account: bool,
     },
     /// Burn an amount of native token.
     BurnNativeToken {
@@ -164,6 +166,8 @@ pub enum WalletCommand {
     EndStaking {
         /// The Account ID of the staking account.
         account_id: AccountId,
+        /// Whether to allot mana from the account output.
+        allot_from_account: bool,
     },
     /// Exit the CLI wallet.
     Exit,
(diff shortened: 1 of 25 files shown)


Thoralf-M (changes requested):
in_allot_account_mana_requirement_twice' panicked at sdk\tests\client\input_selection\account_outputs.rs:1937:6:
called `Result::unwrap()` on an `Err` value: InsufficientMana { found: 1100, required: 1880 }

cli/src/wallet_cli/mod.rs
```diff
@@ -79,6 +79,8 @@ pub enum WalletCommand {
         fixed_cost: u64,
         /// The staking period (in epochs). Will default to the staking unbonding period.
         staking_period: Option<u32>,
+        /// Whether to allot mana from the account output.
+        allot_from_account: bool,
```
Thoralf-M, line 84: This is required? Sounds annoying, would default to true so one doesn't has to provide it every time
Same for the other commands

sdk/src/types/block/output/account.rs
```diff
@@ -395,6 +397,40 @@ impl AccountOutput {
...
+        let generation_amount = self.amount().saturating_sub(min_deposit);
+        let stored_mana = protocol_parameters.mana_with_decay(self.mana(), creation_index, target_index)?;
+        let potential_mana =
+            protocol_parameters.generate_mana_with_decay(generation_amount, creation_index, target_index)?;
+
+        Ok(DecayedMana {
+            stored: stored_mana,
+            potential: potential_mana,
+        })
+    }
```
Thoralf-M, line 432: I'm wondering if we should create a macro for these two. Does anyone else have an opinion on this?

Thoralf-M: Btw this change alone would be good for an own small PR

cli/src/wallet_cli/mod.rs
DaughterOfMars: good point. Should I default to true or false though 🤔

Thoralf-M: I would default to true, what's your opinion on this @thibault-martinez since you opened the issue :trollface:

sdk/src/types/block/output/account.rs
DaughterOfMars: 🤷‍♀️  I don't really feel like it's worth it

Alex6323: agree with Chloé, don't think it's worth it, but you may want to create an investigate-issue for this if you feel we should have another look at it.

Thoralf-M: Not that important :trollface:

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
```diff
@@ -233,43 +236,41 @@ impl InputSelection {
...
     pub(crate) fn mana_sums(&self, include_remainders: bool) -> Result<(u64, u64), Error> {
-        let required_mana = if include_remainders {
-            self.all_outputs().map(|o| o.mana()).sum::<u64>() + self.remainders.added_mana
-        } else {
-            self.non_remainder_outputs().map(|o| o.mana()).sum::<u64>()
-        } + self.mana_allotments.values().sum::<u64>();
+        let mut required_mana =
+            self.non_remainder_outputs().map(|o| o.mana()).sum::<u64>() + self.mana_allotments.values().sum::<u64>();
+        if include_remainders {
+            required_mana += self.remainder_outputs().map(|o| o.mana()).sum::<u64>() + self.remainders.added_mana;
```
Thoralf-M, line 273: `self.remainders.added_mana` is the mana amount we added to existing outputs so we don't need an own remainder output, right? Would be good to have a comment here, hard to understand otherwise

DaughterOfMars: yes, that is correct

Thoralf-M (approved):

Alex6323 (approved):

Issue #1990, opened 2024-02-13
Add send-mana to bindings


thibault-martinez:

thibault-martinez:
Closed by https://github.com/iotaledger/iota-sdk/pull/2068

PR #2038, opened 2024-02-20
Add send_mana and prepare_send_mana

### LINKED ISSUE #2028: Add send-mana to rust core and CLI
With option to set a storage deposit return unlock condition to get the amount back

thibault-martinez:
Fixes https://github.com/iotaledger/iota-sdk/issues/2028

### DIFF
--- a/cli/src/wallet_cli/mod.rs
+++ b/cli/src/wallet_cli/mod.rs
@@ -24,8 +24,8 @@ use iota_sdk::{
     utils::ConvertTo,
     wallet::{
         types::OutputData, BeginStakingParams, ConsolidationParams, CreateDelegationParams, CreateNativeTokenParams,
-        Error as WalletError, MintNftParams, OutputsToClaim, SendNativeTokenParams, SendNftParams, SendParams,
-        SyncOptions, TransactionOptions, Wallet,
+        Error as WalletError, MintNftParams, OutputsToClaim, SendManaParams, SendNativeTokenParams, SendNftParams,
+        SendParams, SyncOptions, TransactionOptions, Wallet,
     },
     U256,
 };
@@ -1046,7 +1046,8 @@ pub async fn send_mana_command(
     address: impl ConvertTo<Bech32Address>,
     mana: u64,
 ) -> Result<(), Error> {
-    let transaction = wallet.send_mana(mana, address, None).await?;
+    let params = SendManaParams::new(mana, address.convert()?);
+    let transaction = wallet.send_mana(params, None).await?;
 
     println_log_info!(
         "Transaction sent:\n{:?}\n{:?}",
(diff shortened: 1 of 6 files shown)


sdk/src/wallet/operations/transaction/prepare_output.rs
```diff
@@ -113,7 +113,7 @@ where
         let min_amount_basic_output =
             BasicOutput::minimum_amount(&Address::from(Ed25519Address::null()), storage_score_params);
 
-        let min_required_storage_deposit = first_output.minimum_amount(storage_score_params);
+        let min_required_storage_deposit = first_output.amount();
```
thibault-martinez, line 116: Equivalent, right?

Thoralf-M: Yes, if the calculation there is not different inside :trollface:

sdk/src/wallet/operations/transaction/high_level/send_mana.rs
```diff
@@ -0,0 +1,95 @@
...
+        log::debug!("[TRANSACTION] prepare_send_mana");
+        let return_strategy = params.return_strategy.unwrap_or_default();
+        let storage_score_params = self.client().get_storage_score_parameters().await?;
+
+        let mut output_builder = BasicOutputBuilder::new_with_minimum_amount(storage_score_params)
+            .with_mana(params.mana)
+            .add_unlock_condition(AddressUnlockCondition::new(params.address));
+
+        match return_strategy {
+            ReturnStrategy::Return => {
```
Thoralf-M, line 77: Clippy not complaining to use `if return_strategy == ReturnStrategy::Return{}` instead?

thibault-martinez: No, I'm using a match so it fails when/if we add a new strategy, wouldn't fail with the if. I guess Clippy understands that.

sdk/src/wallet/operations/transaction/prepare_output.rs
thibault-martinez: Don't think it is

Thoralf-M (dismissed):

cli/src/wallet_cli/mod.rs
```diff
@@ -272,6 +272,16 @@ pub enum WalletCommand {
         #[arg(long, default_value_t = false)]
         allow_micro_amount: bool,
     },
+    /// Send mana.
+    SendMana {
+        /// Address to send mana to, e.g. rms1qztwng6cty8cfm42nzvq099ev7udhrnk0rw8jt8vttf9kpqnxhpsx869vr3.
```
DaughterOfMars:
```suggestion
        /// Recipient address, e.g. rms1qztwng6cty8cfm42nzvq099ev7udhrnk0rw8jt8vttf9kpqnxhpsx869vr3.
```

DaughterOfMars (commented):
Bindings?

thibault-martinez:
> Bindings?

https://github.com/iotaledger/iota-sdk/issues/1990

DaughterOfMars (approved):

Thoralf-M (approved):

PR #2048, opened 2024-02-22
Python: rework OutputId so it works in deserialization

### LINKED ISSUE #2020: Fix Python OutputId deserialization
Change Python OutputId deserialization so it works for OutputMetadata, OutputWithMetadata, currently fails with `AttributeError: 'str' object has no attribute 'keys'`

Thoralf-M:
# Description of change

Python: rework OutputId so it works in deserialization

Fixes https://github.com/iotaledger/iota-sdk/issues/2020

## Type of change

Running tests

### DIFF
--- a/bindings/python/iota_sdk/types/output_id.py
+++ b/bindings/python/iota_sdk/types/output_id.py
@@ -2,24 +2,39 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from dataclasses import dataclass
-
-from iota_sdk.types.common import HexStr, json
+from iota_sdk.types.common import json
 from iota_sdk.types.output import Output
 from iota_sdk.types.transaction_id import TransactionId
 
 
-class OutputId(dict):
+class OutputId(str):
     """Represents an output ID.
 
     Attributes:
         output_id: The unique id of an output.
-        transaction_id: The transaction id associated with the output.
-        output_index: The index of the output within a transaction.
-
     """
 
-    def __init__(self, transaction_id: TransactionId, output_index: int):
+    def __new__(cls, output_id: str):
         """Initialize OutputId
+        """
+        if len(output_id) != 78:
+            raise ValueError(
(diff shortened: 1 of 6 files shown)


DaughterOfMars (dismissed):
This is fine I guess. What was the reason it wouldn't work as a class with two fields?

Thoralf-M:
> This is fine I guess. What was the reason it wouldn't work as a class with two fields?

Because it needs to be a dict for multiple fields and then it fails to deserialize because the field in the data is just a string `AttributeError: 'str' object has no attribute 'keys'`
This could probably also solved in some other way with a custom deserialization method, but didn't found how this can be done and to me this now also seems to be cleaner overall

Alex6323 (dismissed):

bindings/python/iota_sdk/types/output_id.py
```diff
@@ -2,64 +2,66 @@
...
-        # Validate that it has only valid hex characters
-        int(transaction_id[2:], 16)
-        output_index_hex = (output_index).to_bytes(2, "little").hex()
-        self.output_id = transaction_id + output_index_hex
-        self.transaction_id = transaction_id
-        self.output_index = output_index
+                'output_id length must be 78 characters with 0x prefix')
+        if not output_id.startswith('0x'):
+            raise ValueError('output_id must start with 0x')
+        instance = super().__new__(cls, output_id)
```
thibault-martinez, line 27: What is `super` here? `str`?

Thoralf-M: Yes

thibault-martinez: So apart from length and 0x, we don't really check that it's hexadecimal ?

Thoralf-M: No, added

thibault-martinez (approved):

Alex6323 (approved):

PR #2046, opened 2024-02-21
The big error refactor

DaughterOfMars:
# Description of change

This PR breaks up the gigantic `block::Error` type into many different errors. This necessitates some changes to error handling, which I think are annoying but acceptable. For users of the lib, they get the benefits of more useful errors while also being able to generally ignore casting them up using libs like `eyre` (see the CLI).

You may be asking "Why, oh god why? This PR is 228 files, what is wrong with you?" and to that I have no good answer. I am sorry.

PR #2058, opened 2024-02-23
Use TransactionFailureReason as semantic validation error

### LINKED ISSUE #1565: Refactor semantic errors

### LINKED ISSUE #2002: Add TxFailureReason ErrInvalidChainAddressUnlock and ErrInvalidDirectUnlockableAddressUnlock
- [x] Would it be possible to have an amount overflow error like the mana one? Could happen with private tangle configurations.
- [x] What error should be used on invalid unlock method? Like a Signature for an AccountAddress?
- [x] What error should be used if an NFT/Account/Anchor unlock point to a different but same type NFT/Account/Anchor?

thibault-martinez:


DaughterOfMars:
FYI this is also done by #2046

bindings/core/src/method_handler/utils.rs
```diff
@@ -129,7 +129,11 @@ pub(crate) fn call_utils_method_internal(method: UtilsMethod) -> Result<Response
                 protocol_parameters,
             );
 
-            Response::TransactionFailureReason(context.validate()?)
+            if let Err(e) = context.validate() {
+                Response::TransactionFailureReason(Some(e))
+            } else {
+                Response::TransactionFailureReason(None)
+            }
```
Thoralf-M: Don't we now want to return an error in the bindings as well?

thibault-martinez: So removing ` Response::TransactionFailureReason` and letting `validate` error?

Thoralf-M: Yes, but not sure if that's really better, that's why I made it a question

thibault-martinez: Now that it's an actual error, maybe it should behave like one and not return an option yeah

thibault-martinez: I think it's slightly better actually

thibault-martinez: Done

sdk/src/client/secret/mod.rs
```diff
@@ -671,9 +671,9 @@ where
...
     validate_signed_transaction_payload_length(&tx_payload)?;
 
-    let conflict = verify_semantic(&inputs_data, &tx_payload, mana_rewards, protocol_parameters.clone())?;
+    let conflict = verify_semantic(&inputs_data, &tx_payload, mana_rewards, protocol_parameters.clone());
 
-    if let Some(conflict) = conflict {
+    if let Err(conflict) = conflict {
         log::debug!("[sign_transaction] conflict: {conflict:?} for {:#?}", tx_payload);
         return Err(Error::TransactionSemantic(conflict));
     }
```
Thoralf-M: You can do 
```Rust
    verify_semantic(&inputs_data, &tx_payload, mana_rewards, protocol_parameters.clone())
        .inspect_err(|e| {
            log::debug!("[sign_transaction] conflict: {e:?} for {tx_payload:#?}");
        })
        .map_err(Error::TransactionSemantic)?;
```
copied from https://github.com/iotaledger/iota-sdk/pull/2046 :trollface:

sdk/tests/client/signing/account.rs
```diff
@@ -110,9 +110,9 @@ async fn sign_account_state_transition() -> Result<()> {
         &tx_payload,
         prepared_transaction_data.mana_rewards,
         protocol_parameters,
-    )?;
+    );
```
Thoralf-M: Just do this and remove the if let Err() below?
```suggestion
    )?;
```

Thoralf-M: Same in many other tests or with unwrap

thibault-martinez: I guess it was done this way to include the `tx_payload` but I don't think it's really necessary in tests so yeah I can remove

thibault-martinez: Done

Thoralf-M: Yeah usually shouldn't be needed there and if it panics we can add it just to investigate

sdk/src/client/secret/mod.rs
thibault-martinez: stolen

Thoralf-M (approved):

sdk/src/types/block/semantic/mod.rs
```diff
@@ -145,36 +144,41 @@ impl<'a> SemanticValidationContext<'a> {
...
-                            .ok_or(Error::ConsumedManaOverflow)?;
+                            .checked_add(
+                                consumed_output
+                                    .available_mana(
+                                        &self.protocol_parameters,
+                                        output_id.transaction_id().slot_index(),
+                                        self.transaction.creation_slot(),
+                                    )
+                                    // Unwrap is fine as we already checked both slot indices against each others.
+                                    .unwrap(),
```
DaughterOfMars, line 163:
```suggestion
                                    .unwrap(), // Unwrap is fine as we already checked both slot indices against each others.
```

DaughterOfMars (approved):
We cool with the few instances of `SemanticValidationFailed`?

thibault-martinez:
> We cool with the few instances of `SemanticValidationFailed`?

Yeah it's just failsafe but it should actually never happen

sdk/src/types/block/semantic/mod.rs
thibault-martinez: I'll keep it as is if you don't mind, we usually put them on the line before

PR #2067, opened 2024-02-26
Python: change `BlockState` and `TransactionState` enums to hold `str` variants

### LINKED ISSUE #2019: Python: update BlockState and TransactionState enums
The numbers don't work for deserialization from the responses, so we need a str, but do we still need a way to convert the numbers to the variants?
`Pending = 'pending'` instead of `Pending = 0` seems to work fine for the API response test

Alex6323:
# Description of change

Changes `BlockState` and `TransactionState` enum variants from `int`egers to `str`ings.

Closed #2019 

## Open questions

Should we still add methods to convert from/to `int`s?

### DIFF
--- a/bindings/python/iota_sdk/types/block/metadata.py
+++ b/bindings/python/iota_sdk/types/block/metadata.py
@@ -1,33 +1,7 @@
 # Copyright 2023 IOTA Stiftung
 # SPDX-License-Identifier: Apache-2.0
 
-from __future__ import annotations
 from enum import Enum, IntEnum
-from dataclasses import dataclass
-from typing import Optional
-
-from iota_sdk.types.common import json
-from iota_sdk.types.block.id import BlockId
-from iota_sdk.types.transaction_metadata import TransactionFailureReason, TransactionState
-
-
-@json
-@dataclass
-class BlockMetadata:
-    """Represents the metadata of a block.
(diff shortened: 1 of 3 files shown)


Thoralf-M (changes requested):
Should be possible to enable the two tests with this issue in https://github.com/iotaledger/iota-sdk/blob/2.0/bindings/python/tests/test_api_responses.py

DaughterOfMars (approved):

bindings/python/iota_sdk/types/block/metadata.py
```diff
@@ -1,36 +1,10 @@
...
-from typing import Optional
 
-from iota_sdk.types.common import json
-from iota_sdk.types.block.id import BlockId
-from iota_sdk.types.transaction_metadata import TransactionFailureReason, TransactionState
 
-
-@json
-@dataclass
-class BlockMetadata:
```
thibault-martinez, line 16: Why?

Thoralf-M (approved):

bindings/python/iota_sdk/types/block/metadata.py
Thoralf-M: There is BlockMetadataResponse instead now

thibault-martinez (approved):

PR #2068, opened 2024-02-26
Bindings: send mana

### LINKED ISSUE #1990: Add send-mana to bindings

Alex6323:
# Description of change

Adds `sendMana`  wallet method to the bindings.

Closes #1990

### DIFF
--- a/bindings/core/src/method/wallet.rs
+++ b/bindings/core/src/method/wallet.rs
@@ -315,6 +315,13 @@ pub enum WalletMethod {
         #[serde(default)]
         options: Option<TransactionOptions>,
     },
+    /// Prepare to send mana.
+    /// Expected response: [`PreparedTransaction`](crate::Response::PreparedTransaction)
+    PrepareSendMana {
+        params: SendManaParams,
+        #[serde(default)]
+        options: Option<TransactionOptions>,
+    },
     /// Prepare to send native tokens.
     /// Expected response: [`PreparedTransaction`](crate::Response::PreparedTransaction)
     PrepareSendNativeTokens {
@@ -441,7 +448,7 @@ pub enum WalletMethod {
     /// Send mana.
     /// Expected response: [`SentTransaction`](crate::Response::SentTransaction)
     SendMana {
-        params: Box<SendManaParams>,
+        params: SendManaParams,
         #[serde(default)]
         options: Option<TransactionOptions>,
     },
--- a/bindings/core/src/method/wallet.rs
+++ b/bindings/core/src/method/wallet.rs
@@ -445,13 +445,6 @@ pub enum WalletMethod {
         #[serde(default)]
         options: Option<TransactionOptions>,
     },
-    /// Send mana.
-    /// Expected response: [`SentTransaction`](crate::Response::SentTransaction)
-    SendMana {
-        params: SendManaParams,
-        #[serde(default)]
-        options: Option<TransactionOptions>,
-    },
     /// Set the alias of the wallet.
     /// Expected response: [`Ok`](crate::Response::Ok)
     SetAlias { alias: String },
--- a/bindings/core/src/method/wallet.rs
+++ b/bindings/core/src/method/wallet.rs
@@ -28,8 +28,8 @@ use iota_sdk::{
     },
     wallet::{
         BeginStakingParams, ClientOptions, ConsolidationParams, CreateAccountParams, CreateDelegationParams,
-        CreateNativeTokenParams, FilterOptions, MintNftParams, OutputParams, OutputsToClaim, SendNativeTokenParams,
-        SendNftParams, SendParams, SyncOptions, TransactionOptions,
+        CreateNativeTokenParams, FilterOptions, MintNftParams, OutputParams, OutputsToClaim, SendManaParams,
+        SendNativeTokenParams, SendNftParams, SendParams, SyncOptions, TransactionOptions,
     },
     U256,
 };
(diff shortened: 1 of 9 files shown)


bindings/core/src/method/wallet.rs
```diff
@@ -438,6 +438,13 @@ pub enum WalletMethod {
         #[serde(default)]
         options: Option<TransactionOptions>,
     },
+    /// Send mana.
+    /// Expected response: [`SentTransaction`](crate::Response::SentTransaction)
+    SendMana {
+        params: Box<SendManaParams>,
```
thibault-martinez: Clippy asked for it?

thibault-martinez (changes requested):
Also needs a binding for `prepare_send_mana`

Thoralf-M (commented):
Also add the prepare_ methods?

Alex6323:
> Also add the prepare_ methods?

:rocket:

bindings/core/src/method/wallet.rs
Alex6323: Not really ... I just assumed it would complain, and we do it already for some other variant. Also ... even if it's not complaining atm, it could in the future, should we ever add more params.

thibault-martinez: If it actually does not complain, I would remove it since it has a runtime cost. And this is not breaking change so we can always change later if it starts complaining at some point.

Alex6323: It didn't complain so I removed it.

bindings/nodejs/lib/wallet/wallet.ts
```diff
@@ -1705,6 +1706,59 @@ export class Wallet {
...
+        ) as Response<TransactionWithMetadata>;
+        return plainToInstance(TransactionWithMetadata, parsed.payload);
+    }
+
+    /**
+     * Prepare to send mana.
+     *
+     * @param params Amount, Address, and Return Strategy.
+     * @param transactionOptions Additional transaction options.
+     * @returns The sent transaction.
```
Thoralf-M:
```suggestion
     * @returns The prepared transaction.
```

bindings/python/iota_sdk/wallet/wallet.py
```diff
@@ -805,6 +805,29 @@ def send_outputs(
             }
         ))
 
+    def send_mana(
+            self, params: SendManaParams, options: Optional[TransactionOptions] = None) -> TransactionWithMetadata:
+        """Send mana.
+        """
+        return TransactionWithMetadata.from_dict(self._call_method(
```
thibault-martinez: should call `prepare_send_mana` like the other send bindings

Thoralf-M: Not 100% sure about this since it requires two calls to the Rust side, but if we want it everywhere we also don't need the direct methods like `SendMana` in the bindings core

thibault-martinez: Ah but I'm just realising now that you shouldn't have added the send bindings to the core, we don't do that for the other ones?

thibault-martinez: Ah we commented at the same time. But that's what we're doing everywhere already

Thoralf-M: Not everywhere, we also have Send, SendWithParams, SendOutputs

Thoralf-M: But yeah we should probably do it everywhere to have less "duplication"

thibault-martinez: Yeah but the 3 you mentioned are the base ones, not the high level ones

Thoralf-M: They're also in the high level mod :trollface:
`SendOutputs` is the only not high level one I would say, but even high level or not, why have it different for them?

Alex6323: what's the plan?

bindings/core/src/method/wallet.rs
```diff
@@ -438,6 +445,13 @@ pub enum WalletMethod {
         #[serde(default)]
         options: Option<TransactionOptions>,
     },
+    /// Send mana.
+    /// Expected response: [`SentTransaction`](crate::Response::SentTransaction)
+    SendMana {
```
DaughterOfMars: We don't actually have any other send methods because the bindings just calls `.send()` on them or whatever

bindings/python/iota_sdk/wallet/wallet.py
Thoralf-M: Removing `SendMana` from the bindings core is the plan https://github.com/iotaledger/iota-sdk/pull/2068#discussion_r1504854601 :trollface:
For the others I opened an issue https://github.com/iotaledger/iota-sdk/issues/2081

Alex6323:
Lgtm. Thanks!

Thoralf-M (approved):

DaughterOfMars (approved):

bindings/nodejs/lib/types/wallet/bridge/index.ts
```diff
@@ -2,7 +2,6 @@ import type {
     __AnnounceCandidacyMethod__,
     __PrepareBurnMethod__,
     __PrepareClaimOutputsMethod__,
-    __ClaimOutputsMethod__,
```
thibault-martinez, line 5: And the actual core binding would be removed later?

Thoralf-M: Added it here https://github.com/iotaledger/iota-sdk/issues/2081, but the nodejs part wasn't used anyways

thibault-martinez (approved):

PR #2070, opened 2024-02-26
Remove Bip44Address, AddressWithUnspentOutputs::{key_index, internal}

### LINKED ISSUE #2025: Remove `AddressWithUnspentOutputs::{key_index, internal}`

Thoralf-M:
# Description of change

Remove Bip44Address, AddressWithUnspentOutputs::{key_index, internal} as they are no longer needed
Also made `AddressWithUnspentOutputs` private since we don't need it exposed

Closes https://github.com/iotaledger/iota-sdk/issues/2025

## Type of change

`cargo test`

### DIFF
--- a/sdk/src/types/block/protocol/samples.rs
+++ b/sdk/src/types/block/protocol/samples.rs
@@ -242,7 +242,7 @@ impl ProtocolParameters {
         self.rewards_parameters.final_target_rewards_rate = (self.token_supply()
             * self.rewards_parameters().reward_to_generation_ratio() as u64
             * self.mana_parameters().generation_rate() as u64)
-            >> self.mana_parameters().generation_rate_exponent() - self.slots_per_epoch_exponent();
+            >> (self.mana_parameters().generation_rate_exponent() - self.slots_per_epoch_exponent());
         let bootstrapping_duration_years =
             self.rewards_parameters().bootstrapping_duration() as f64 * self.epochs_per_year().exp();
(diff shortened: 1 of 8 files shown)


Alex6323 (approved):

sdk/src/types/block/protocol/samples.rs
```diff
@@ -242,7 +242,7 @@ impl ProtocolParameters {
         self.rewards_parameters.final_target_rewards_rate = (self.token_supply()
             * self.rewards_parameters().reward_to_generation_ratio() as u64
             * self.mana_parameters().generation_rate() as u64)
-            >> self.mana_parameters().generation_rate_exponent() - self.slots_per_epoch_exponent();
+            >> (self.mana_parameters().generation_rate_exponent() - self.slots_per_epoch_exponent());
```
thibault-martinez, line 245: We're 100% sure about that? Because it actually does change the computation

Thoralf-M: It doesn't change https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cd273bf1d04773378b475cd7c2d47d6e

Thoralf-M: https://doc.rust-lang.org/reference/expressions.html#expression-precedence `-` has precedence over `>>`

thibault-martinez: Yeah I just read this page, interesting, I would have expected the opposite, thanks

thibault-martinez (approved):

PR #2075, opened 2024-02-26
Automatically set transaction capability flags in ISA

### LINKED ISSUE #2005: Don't ignore TransactionCapabilityFlag in InputSelection
Currently one can provide `.with_transaction_capabilities([TransactionCapabilityFlag::BurnMana])`, but nothing is done with this, same with the others

DaughterOfMars:
# Description of change

This PR removes the ability to pass in transaction capability flags and instead sets them based on the `Burn`. To support `TransactionCapabilityFlag::BurnMana`, a boolean field was added to this struct. If there is excess mana in the "initial mana balance" then it will be burned. *Nothing else will be burned*, and remainders will still be created for mana from selected inputs. This way there are no surprise mana burns, everything that the user intends to burn must be specifically provided.

Closes #2005


Thoralf-M:
> This way there are no surprise mana burns, everything that the user intends to burn must be specifically provided.

Wouldn't it be better to let the user provide the exact Mana amount to be burned instead of `excess mana in the "initial mana balance"`?

bindings/nodejs/lib/types/client/burn.ts
```diff
@@ -6,6 +6,10 @@ import { AccountId, FoundryId, NftId, TokenId } from '../block/id';
 
 /** A DTO for [`Burn`] */
 export interface Burn {
+    /** Burn initial excess mana */
```
Thoralf-M: Please elaborate in the doc comments what `initial excess mana` is

Thoralf-M (dismissed):

Thoralf-M (dismissed):

sdk/tests/client/input_selection/account_outputs.rs
```diff
@@ -1269,6 +1274,10 @@ fn account_burn_should_validate_account_sender() {
     .select()
     .unwrap();
 
+    assert_eq!(
+        selected.transaction.capabilities(),
+        &TransactionCapabilities::from([TransactionCapabilityFlag::DestroyAccountOutputs])
+    );
```
thibault-martinez, line 1280: There is another Burn in `account_burn_should_validate_account_address`, do you want to add a check for it?

sdk/src/client/api/block_builder/input_selection/requirement/native_tokens.rs
```diff
@@ -60,6 +63,10 @@ impl InputSelection {
         output_native_tokens.merge(melted_native_tokens)?;
 
         if let Some(burn) = self.burn.as_ref() {
+            if !burn.native_tokens.is_empty() {
+                self.transaction_capabilities
+                    .add_capability(TransactionCapabilityFlag::BurnNativeTokens);
+            }
             output_native_tokens.merge(NativeTokensBuilder::from(burn.native_tokens.clone()))?;
```
thibault-martinez, line 68: Could even put this line in the if, less stuff to construct for nothing

sdk/src/client/api/block_builder/input_selection/remainder.rs
```diff
@@ -237,10 +237,15 @@ impl InputSelection {
...
         // Mana can potentially be added to an appropriate existing output instead of a new remainder output
-        let mana_remainder = selected_mana > required_mana
+        let mut mana_remainder = selected_mana > required_mana
             && remainder_address.map_or(true, |remainder_address| {
                 !self.output_for_added_mana_exists(&remainder_address)
             });
+        // If we are burning mana, we may not need a mana remainder
+        if self.burn.as_ref().map_or(false, |b| b.mana()) {
+            let initial_excess = self.initial_mana_excess()?;
+            mana_remainder &= selected_mana > required_mana + initial_excess;
```
thibault-martinez, line 242: &= is bitwise but then it also works for booleans?

DaughterOfMars: booleans are bits...kiiiind of

but yeah it works the same

sdk/src/client/api/block_builder/input_selection/requirement/native_tokens.rs
```diff
@@ -59,7 +62,9 @@ impl InputSelection {
         input_native_tokens.merge(minted_native_tokens)?;
         output_native_tokens.merge(melted_native_tokens)?;
 
-        if let Some(burn) = self.burn.as_ref() {
+        if let Some(burn) = self.burn.as_ref().filter(|burn| !burn.native_tokens.is_empty()) {
```
thibault-martinez, line 65: Clippy doesn't complain about !is_empty instead of is_some?

DaughterOfMars: it's not an option

Thoralf-M (dismissed):

sdk/src/client/api/block_builder/input_selection/requirement/native_tokens.rs
thibault-martinez: I'm dumb

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -160,10 +160,7 @@ impl InputSelection {
         self.available_inputs
             .retain(|input| !self.forbidden_inputs.contains(input.output_id()));
 
-        // This is to avoid a borrow of self since there is a mutable borrow in the loop already.
-        let required_inputs = std::mem::take(&mut self.required_inputs);
-
-        for required_input in required_inputs {
+        for required_input in self.required_inputs.clone() {
```
thibault-martinez, line 163: Why not keep the `take` since we're not using `required_inputs` anymore?

sdk/src/client/api/block_builder/input_selection/mod.rs
```diff
@@ -270,6 +267,19 @@ impl InputSelection {
             }
         }
 
+        // If we're burning generated mana, set the capability flag.
+        if self.burn.as_ref().map_or(false, |b| b.generated_mana()) {
+            // Get the mana sums with generated mana to see whether there's a difference.
+            if !self
+                .transaction_capabilities
+                .has_capability(TransactionCapabilityFlag::BurnMana)
+                && input_mana < self.total_selected_mana(true)?
```
thibault-martinez, line 276: I'm a bit confused by this. Why is this describing a burn if there is no output mana involved in the comparison?

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
```diff
@@ -268,20 +283,32 @@ impl InputSelection {
...
-            selected_mana += self.total_mana(input)?;
+            selected_mana += self.total_mana(input, include_generated)?;
         }
-        Ok((selected_mana, required_mana))
+
+        Ok(selected_mana)
     }
 
-    fn total_mana(&self, input: &InputSigningData) -> Result<u64, Error> {
+    fn total_mana(&self, input: &InputSigningData, include_generated: impl Into<Option<bool>>) -> Result<u64, Error> {
```
thibault-martinez: Since this fn is private I'd take a bool directly and do the unwrap/default in `total_selected_mana` instead

thibault-martinez: Otherwise we do this in a loop for nothing

DaughterOfMars: fair point

sdk/src/client/api/block_builder/input_selection/mod.rs
DaughterOfMars: We are using it now, in `initial_mana_excess`

sdk/src/client/api/block_builder/input_selection/mod.rs
DaughterOfMars: because we calculated `input_mana` without generated mana, so if there's a difference that means the output mana was also calculated without the generated mana, thus we have burned mana

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
DaughterOfMars: actually, the problem is then we have to add the check for the burn at every call point

sdk/src/client/api/block_builder/input_selection/mod.rs
thibault-martinez: Ok so for my understanding, the compiler wouldn't say anything if we take and then reuse later?

sdk/src/client/api/block_builder/input_selection/mod.rs
thibault-martinez: > because we calculated input_mana without generated mana

Are you sure about that?
`input_mana` is calculated from `self.total_selected_mana(None)` and None defaults to `include_generated = true`?

sdk/src/client/api/block_builder/input_selection/mod.rs
DaughterOfMars: no because take leaves an empty vec

sdk/src/client/api/block_builder/input_selection/mod.rs
DaughterOfMars: yes, so if you are trying to burn mana, then my above statement is true, otherwise the amounts will always be the same and the flag will not be set

sdk/src/client/api/block_builder/input_selection/requirement/mana.rs
```diff
@@ -259,6 +260,22 @@ impl InputSelection {
         Ok(added_inputs)
     }
 
+    pub(crate) fn initial_mana_excess(&self) -> Result<u64, Error> {
```
Thoralf-M, line 263: This doesn't need to consider mana rewards?

Thoralf-M: Ah that's added in total_mana()

DaughterOfMars: It does! Check total_mana

Thoralf-M (approved):

thibault-martinez (approved):

PR #2086, opened 2024-02-28
Remove remaining mentions of milestones

### LINKED ISSUE #1041: Remove remaining mentions of milestones

thibault-martinez:


sdk/examples/client/block/01_block_confirmation_time.rs
```diff
@@ -60,19 +60,5 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    // if let Some(ms_index) = metadata.referenced_by_milestone_index {
-    //     let ms = client.get_milestone_by_index(ms_index).await?;
-    //     println!(
-    //         "Block {block_id} got confirmed by milestone {ms_index} at timestamp {}.",
-    //         ms.essence().timestamp()
-    //     );
-    // } else {
-    //     println!("Block {block_id} is not confirmed.")
-    // }
-
```
thibault-martinez, line 76: Not sure there is actually an equivalent to this tbh

thibault-martinez (commented):
FYI there is more mentions of milestones in participation code but I have purposely ignored that for now.

sdk/examples/client/block/01_block_confirmation_time.rs
Thoralf-M: Maybe at L53 it should compute the difference in seconds from sending until it got confirmed? Or we remove this example altogether

cli/src/wallet_cli/mod.rs
```diff
@@ -1170,10 +1180,12 @@ pub async fn transactions_command(wallet: &Wallet, show_details: bool) -> Result
...
             } else {
-                let transaction_time = to_utc_date_time(tx.timestamp)?;
-                let formatted_time = transaction_time.format("%Y-%m-%d %H:%M:%S UTC").to_string();
-
-                println_log_info!("{:<5}{}\t{}", i, tx.transaction_id, formatted_time);
+                println_log_info!(
+                    "{:<5}{}\t{}",
+                    i,
+                    tx.transaction_id,
+                    tx.payload.transaction().creation_slot()
```
Thoralf-M: What do you think about also converting the slot to the utc_date_time? Seeing if it was sent yesterday at 9 PM is probably more interesting than slot 1424 (would still keep the slot though)

sdk/src/wallet/mod.rs
```diff
@@ -87,10 +88,10 @@ use crate::{
...
 #[serde(rename_all = "camelCase")]
 pub struct FilterOptions {
-    /// Filter all outputs where the booked milestone index is below the specified timestamp
-    pub lower_bound_booked_timestamp: Option<u32>,
-    /// Filter all outputs where the booked milestone index is above the specified timestamp
-    pub upper_bound_booked_timestamp: Option<u32>,
+    /// Filter all outputs where the included slot is below the specified slot.
+    pub included_below_slot: Option<SlotIndex>,
+    /// Filter all outputs where the included slot is above the specified slot.
+    pub included_above_slot: Option<SlotIndex>,
```
Thoralf-M: Maybe it's just me, but to with these names it's not clear if it would return outputs that match the condition or output that don't match it

Thoralf-M: Maybe it's enough if the comments say what exactly happens
Currently seems to do the opposite of the Rust iter filter() method which does this
```
Creates an iterator which uses a closure to determine if an element should be yielded.

Given an element the closure must return true or false. The returned iterator will yield only the elements for which the closure returns true.
```

cli/src/wallet_cli/mod.rs
DaughterOfMars: Agreed

thibault-martinez: On it

sdk/examples/client/block/01_block_confirmation_time.rs
thibault-martinez: Added the duration

sdk/src/wallet/mod.rs
thibault-martinez: Actually I think I messed it up and did the opposite

thibault-martinez: I reversed them, do you still want me to update a comment or something?

Thoralf-M: Hm I think you reversed the "meaning" compared to what it was before now
Could still make sense though, but yes please update the comment so it's clear if an output that matches the condition will be returned or not

Thoralf-M (approved):

Alex6323 (approved):

PR #2092, opened 2024-02-28
Allow ignoring missing mana rewards

### LINKED ISSUE #2034: Should mana_rewards be mandatory in semantic validation?
To allow validation of everything apart from the mana if the rewards are not known
Might can do it in a smart way and only ignore if the tx could actually have staking or delegation rewards

DaughterOfMars:
# Description of change

If no mana rewards map is provided to semantic validation, the rewards checks are now ignored.

Closes #2034


thibault-martinez (changes requested):
Failing CI

sdk/src/types/block/semantic/mod.rs
```diff
@@ -404,7 +404,9 @@ impl<'a> SemanticValidationContext<'a> {
 
         if self.input_mana != self.output_mana {
             if self.input_mana > self.output_mana {
-                if !self.transaction.has_capability(TransactionCapabilityFlag::BurnMana) {
+                if (self.mana_rewards.is_some() || self.reward_context_inputs.is_empty())
+                    && !self.transaction.has_capability(TransactionCapabilityFlag::BurnMana)
+                {
                     return Err(TransactionFailureReason::CapabilitiesManaBurningNotAllowed);
                 }
             } else {
```
Thoralf-M: Don't we also have to not return the error in the else if None was provided?

DaughterOfMars: no because it's possible you aren't burning mana there

sdk/src/utils/serde.rs
```diff
@@ -277,3 +277,32 @@ pub mod mana_rewards {
...
+    use serde::{Deserialize, Deserializer};
+
+    use crate::types::block::output::OutputId;
+
+    pub fn serialize<S: serde::Serializer>(
+        mana_rewards: &Option<BTreeMap<OutputId, u64>>,
+        s: S,
+    ) -> Result<S::Ok, S::Error> {
+        match mana_rewards {
+            Some(bytes) => super::mana_rewards::serialize(bytes, s),
```
thibault-martinez: bytes?

sdk/src/types/block/semantic/mod.rs
```diff
@@ -404,7 +404,9 @@ impl<'a> SemanticValidationContext<'a> {
 
         if self.input_mana != self.output_mana {
             if self.input_mana > self.output_mana {
-                if !self.transaction.has_capability(TransactionCapabilityFlag::BurnMana) {
+                if (self.mana_rewards.is_some() || self.reward_context_inputs.is_empty())
```
thibault-martinez, line 407: why is_empty?

thibault-martinez: why is_empty?

sdk/src/utils/serde.rs
DaughterOfMars: err ahh err ahh

sdk/src/types/block/semantic/mod.rs
DaughterOfMars: it's the inverse of this
```
if self.mana_rewards.is_none() && !self.reward_context_inputs.is_empty() {
    Ok(())
} else if !self.transaction.has_capability(TransactionCapabilityFlag::BurnMana) {
    return Err(TransactionFailureReason::CapabilitiesManaBurningNotAllowed)
}
```

sdk/src/types/block/semantic/mod.rs
Thoralf-M: But if the tx claims rewards which aren't provided here, there would also be a mismatch?

DaughterOfMars: yes, that's the else there

Thoralf-M: Yeah and my point is, if `self.mana_rewards.is_none()` it shouldn't error then?

Thoralf-M: Having a test with a transaction that claims mana would probably be the best to see if it really works

DaughterOfMars: The logic is this but reversed
```
if self.mana_rewards.is_none() && !self.reward_context_inputs.is_empty() {
    Ok(())
} else if !self.transaction.has_capability(TransactionCapabilityFlag::BurnMana) {
    return Err(TransactionFailureReason::CapabilitiesManaBurningNotAllowed)
}
```
The intention was to only skip the validation here if you actually have tried to claim rewards. The reward context inputs is a convenient way to check. So yes there is a situation where you have passed `None` mana rewards and still get this error, which I think is fine?

Thoralf-M: I'm talking about this case `self.input_mana < self.output_mana`, where it now would return an error in L413?

Thoralf-M: input_mana being smaller because the mana rewards aren't known in the validation because none was provided, but was added to the outputs

Thoralf-M: Providing rewards and none both seem to fail
```
diff --git a/sdk/tests/client/input_selection/delegation_outputs.rs b/sdk/tests/client/input_selection/delegation_outputs.rs
index ad48472b2..bdf4e70a6 100644
--- a/sdk/tests/client/input_selection/delegation_outputs.rs
+++ b/sdk/tests/client/input_selection/delegation_outputs.rs
@@ -105,4 +105,21 @@ fn remainder_needed_for_mana() {
                 .sum::<u64>(),
         selected.transaction.outputs().iter().map(|o| o.mana()).sum::<u64>()
     );
+
+    let inputs = inputs
+        .iter()
+        .map(|input| (input.output_id(), &input.output))
+        .collect::<Vec<(&OutputId, &iota_sdk::types::block::output::Output)>>();
+
+    iota_sdk::types::block::semantic::SemanticValidationContext::new(
+        &selected.transaction,
+        &inputs,
+        None,
+        // Providing the rewards and providing None both fails
+        // None,
+        Some(std::collections::BTreeMap::from([(delegation_output_id, mana_rewards)])),
+        protocol_parameters.clone(),
+    )
+    .validate()
+    .unwrap();
 }

```

DaughterOfMars: Ah, I got input and output mana mixed up 😵‍💫

Thoralf-M: The None case works now, but `Some(std::collections::BTreeMap::from([(delegation_output_id, mana_rewards)])),` still fails

sdk/src/types/block/semantic/mod.rs
```diff
@@ -239,7 +239,7 @@ impl<'a> SemanticValidationContext<'a> {
                 )
                 .ok_or(TransactionFailureReason::ManaOverflow)?;
 
-            if let Some(mana_rewards) = self.mana_rewards.get(*output_id) {
+            if let Some(mana_rewards) = self.mana_rewards.as_ref().and_then(|r| r.get(*output_id)) {
                 self.input_mana
```
Thoralf-M:
```suggestion
                self.input_mana = self.input_mana
```

Thoralf-M (dismissed):

Thoralf-M (approved):

thibault-martinez (approved):

PR #2100, opened 2024-02-29
feat: Wallet events improvements

### LINKED ISSUE #2054: Emit block signing event with signing input
So the signing_input can be displayed in Firefly for ledger nano, similar to `TransactionProgressEvent::PreparedTransactionSigningHash`

We should probably also change the numbers of the TransactionProgress events so they match the order we emit them (SigningTransaction before PreparedTransactionSigningHash/PreparedTransaction) https://github.com/iotaledger/iota-sdk/blob/23623631126768f082def54531f7aaf9077b56be/sdk/src/wallet/operations/transaction/sign_transaction.rs#L39

marc2332:
# Description of change

- Reorder wallet events (emit `SigningTransaction` before `PreparedTransactionSigningHash`, emit `Broadcasting` after signing the block)
- Add a new `PreparedBlockSigningInput` event
- Updated Nodejs and python bindings

Fixes https://github.com/iotaledger/iota-sdk/issues/2054

## Type of change

- Enhancement

Firefly and SDK tests

## Change checklist

- [x] I have added tests that prove my fix is effective or that my feature works
    - [x] Update the tests accordingly
    - [x] Avoid relying on `TransactionSigningHash` for the block signing hash


sdk/src/types/block/core/block.rs
```diff
@@ -57,6 +64,11 @@ impl UnsignedBlock {
         [self.header.hash(), self.body.hash()].concat()
     }
 
+    /// Return the Blake2b hash of the block's signing input.
+    pub fn signing_hash(&self) -> BlockSigningHash {
+        BlockSigningHash::new(Blake2b256::digest(self.signing_input()).into())
```
Thoralf-M: This shouldn't be hashed?

marc2332: I have thought about emitting the block event with the raw signing input, perhaps that would be better?

thibault-martinez: Contrary to TransactionSigningHash, the protocol doesn't care about BlockSigningHash so I'd say maybe we should emit the input yes

Thoralf-M: Ledger nano should also show the raw signing input and you should be able to compare it with the data from the event, otherwise there is no point in having it at all

marc2332: :100:

bindings/nodejs/lib/types/wallet/event.ts
```diff
@@ -121,12 +121,14 @@ enum TransactionProgressType {
...
     PreparedTransaction = 2,
-    /** Prepared transaction signing hash hex encoded, required for blindsigning with a Ledger Nano. */
-    PreparedTransactionSigningHash = 3,
     /** Signing the transaction. */
-    SigningTransaction = 4,
+    SigningTransaction = 3,
+    /** Prepared transaction signing hash hex encoded, required for blindsigning with a Ledger Nano. */
+    PreparedTransactionSigningHash = 4,
+    /** Prepared block signing input, required for blindsigning with a Ledger Nano. */
+    PrepareBlockSigningInput = 5,
```
DaughterOfMars: Missing the `d` in prepared?

sdk/src/wallet/operations/transaction/submit_transaction.rs
```diff
@@ -22,11 +22,14 @@ where
     ) -> crate::wallet::Result<BlockId> {
         log::debug!("[TRANSACTION] submit_signed_transaction");
 
+        let block_id = self
+            .submit_basic_block(Some(Payload::from(payload)), issuer_id, true)
+            .await?;
+
         #[cfg(feature = "events")]
         self.emit(WalletEvent::TransactionProgress(TransactionProgressEvent::Broadcasting))
             .await;
```
Thoralf-M: Still not sure why the order should be changed `Broadcasting` is not after it was broadcasted and the function returns right afterwards

marc2332: Then `Broadcasting` doesn't make sense. Because it was being emitted even before the block has been signed, which might take a while since you could be signing with ledger, perhaps I should rename it to `Broadcasted` and leave it with the new order ? What do you think?

Thoralf-M: Or move it after signing? Signing was added afterwards and this wasn't moved then

marc2332: Sure, that would do too

marc2332: Done @Thoralf-M

Thoralf-M (changes requested):
Please also update python https://github.com/iotaledger/iota-sdk/blob/2.0/bindings/python/iota_sdk/types/event.py

marc2332:
CI is failing but should be fixed by https://github.com/iotaledger/iota-sdk/pull/2131

bindings/python/examples/exchange/4_listen_events.py
```diff
@@ -35,8 +35,8 @@ def callback(event):
     received_event = True
 
 
-# Only interested in new outputs here.
-wallet.listen(callback, [WalletEventType.NewOutput])
+# Only interested when inputs are being selected.
+wallet.listen(callback, [WalletEventType.SelectingInputs])
```
Thoralf-M: Why did you change this? Input selection will not run in this example and makes no sense for it anyways :sweat_smile:

marc2332: Right, I went too fast replacing, let me see

marc2332: Yeah, this is wrong.

Thoralf-M: Ah sorry I linked the wrong event file, the other is added in another PR and there is https://github.com/iotaledger/iota-sdk/issues/2121 so no point in you doing any changes now

marc2332: Okay, just reverted those changes which definitely did not make any sense, but, now I fail to see what exactly is there to change in the python bindings? `WalletEventType` looks just fine

Thoralf-M: Yeah nothing to change in this PR, the TransactionProgress part is just completely missing, but will be added in another PR

marc2332: yeah, saw that too, but since it's the same for the other variants I supposed it was normal

sdk/src/wallet/operations/block.rs
```diff
@@ -39,13 +41,23 @@ where
             }
         }
 
+        #[cfg(feature = "events")]
+        self.emit(WalletEvent::TransactionProgress(
+            TransactionProgressEvent::PreparedBlockSigningInput(unsigned_block.signing_input()),
```
Thoralf-M: Don't want to prefix_hex::encode this? So it's also a String like `PreparedTransactionSigningHash`

marc2332: oh yes, I like the sound of that

bindings/nodejs/lib/types/wallet/event.ts
```diff
@@ -223,11 +234,17 @@ class PreparedTransactionSigningHashProgress extends TransactionProgress {
...
 
 /**
- * A 'signing transaction' progress.
+ * A 'prepared block input' progress.
  */
-class SigningTransactionProgress extends TransactionProgress {
-    constructor() {
-        super(TransactionProgressType.SigningTransaction);
+class PreparedBlockSigningInputProgress extends TransactionProgress {
+    blockSigningInput: string;
```
Thoralf-M:
```suggestion
    blockSigningInput: HexEncodedString;
```

bindings/nodejs/lib/types/wallet/event.ts
```diff
@@ -223,11 +234,17 @@ class PreparedTransactionSigningHashProgress extends TransactionProgress {
...
+
+    /**
+     * @param signingHash The signing hash of the block.
+     */
+    constructor(signingInput: string) {
```
Thoralf-M:
```suggestion
    constructor(signingInput: HexEncodedString) {
```

sdk/src/wallet/operations/transaction/submit_transaction.rs
```diff
@@ -22,11 +20,10 @@ where
...
-        #[cfg(feature = "events")]
-        self.emit(WalletEvent::TransactionProgress(TransactionProgressEvent::Broadcasting))
-            .await;
+        let block_id = self
+            .submit_basic_block(Some(Payload::from(payload)), issuer_id, true)
+            .await?;
 
-        self.submit_basic_block(Some(Payload::from(payload)), issuer_id, true)
-            .await
+        Ok(block_id)
```
Thoralf-M: Can now just be `self.submit_basic_block(Some(Payload::from(payload)), issuer_id, true).await` again

Thoralf-M (approved):

thibault-martinez (approved):
Thanks

PR #2101, opened 2024-02-29
Validator route pagination

### LINKED ISSUE #1118: Update validator route with pagination
## Description

Regarding https://github.com/iotaledger/iota-sdk/pull/1080
and https://github.com/iotaledger/iota-sdk/issues/1053 

The pagination wasnt specified yet and should be properly added when the tip updates this

Alex6323:
# Description of change

There wasn't much to do except for fixing a few issues as the functionality was basically there already. However, I added an example to fetch the validators in a paginated way.

#1118


bindings/nodejs/examples/client/get-validators.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright 2021-2023 IOTA Stiftung
```
Thoralf-M:
```suggestion
// Copyright 2024 IOTA Stiftung
```

sdk/examples/client/node_api_core/get_validators.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright 2022 IOTA Stiftung
```
Thoralf-M:
```suggestion
// Copyright 2024 IOTA Stiftung
```

sdk/examples/client/node_api_core/get_validators.rs
```diff
@@ -0,0 +1,43 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! This example returns the validators of the node by querying its `/validators` endpoint.
+//! The result is paginated with a page size of 1. You can provide a LIST_INDEX to request
+//! a particular validator.
```
Thoralf-M: What's this LIST_INDEX? Are the validators sorted by stake or what is this good for?

sdk/src/client/node_api/core/routes.rs
```diff
@@ -147,8 +147,8 @@ impl Client {
     ) -> Result<ValidatorsResponse> {
         const PATH: &str = "api/core/v3/validators";
         let query = query_tuples_to_query_string([
-            page_size.into().map(|i| ("pageSize", i.to_string())),
-            cursor.into().map(|i| ("cursor", i)),
+            page_size.into().map(|page_size| ("pageSize", page_size.to_string())),
+            cursor.into().map(|cursor| ("cursor", cursor)),
```
Thoralf-M: Don't you think we should automatically request the next pages and merge them together like we did for the outputs https://github.com/iotaledger/iota-sdk/blob/d9eea41a7b86ad1b4c3bd39e8f9ac5a40b70d5e8/sdk/src/client/node_api/indexer/mod.rs#L34?

sdk/examples/client/node_api_core/get_validators.rs
Alex6323: this is part of the cursor

Thoralf-M: Sure, but what does it represent?

Thoralf-M: Just read the TIP `Returns information of all stakers (registered validators) and if they are active, ordered by their holding stake.` so if you provide 5 here, it will return the staker with the 5th largest stake, right?

Alex6323: yes

sdk/src/client/node_api/core/routes.rs
Alex6323: seems like something we can provide as an extra, but that's not always what the user wants ... I mean the full list?

sdk/examples/client/node_api_core/get_validators.rs
Alex6323: I'll update the docs for this example with that text so it's clear to the user.

sdk/src/client/node_api/core/routes.rs
Thoralf-M: Actually I think if the page size can be 1000 we never need it, so not worth it

sdk/examples/client/node_api_core/get_validators.rs
DaughterOfMars: should that not be obfuscated so users don't have to build cursors like that?

Alex6323: You're right. Actually the only reason why I did this is to circumvent the bug in the node (it doesn't accept its own cursor last time I tried :trollface: ) , and then for some reason liked the example a little too much. I'll change`em.

sdk/examples/client/node_api_core/get_validators.rs
```diff
@@ -0,0 +1,41 @@
+// Copyright 2024 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! This example returns the validators of the node by querying its `/validators` endpoint.
+//! The result is paginated with a page size of 1. You can provide a LIST_INDEX to request
```
Thoralf-M: Outdated

Alex6323: Fixed

sdk/examples/client/node_api_core/get_validators.rs
Alex6323: Changed.

Thoralf-M (dismissed):

Thoralf-M (dismissed):

sdk/src/client/node_api/core/routes.rs
```diff
@@ -138,8 +138,8 @@ impl Client {
     ) -> Result<ValidatorsResponse, ClientError> {
         const PATH: &str = "api/core/v3/validators";
         let query = query_tuples_to_query_string([
-            page_size.into().map(|i| ("pageSize", i.to_string())),
-            cursor.into().map(|i| ("cursor", i)),
+            page_size.into().map(|page_size| ("pageSize", page_size.to_string())),
+            cursor.into().map(|cursor| ("cursor", cursor)),
```
thibault-martinez: Don't do this if you're not doing it in the whole file

thibault-martinez: I would personally not do this change as we use single letter closures a lot, but if you really prefer, do it in the whole file, no need to be inconsistent

Alex6323: > Don't do this if you're not doing it in the whole file

But doing that for the whole file would bloat the PR with unrelated changes depending on the number of unrelated occurences. 

> I would personally not do this change as we use single letter closures a lot, but if you really prefer, do it in the whole file, no need to be inconsistent

But that is not why I made the change: while `i` commonly represents `index` like in other places of this file, it does not usually represent `size` or a `cursor`. I don't mind single letter closures at all, but SUGGESTED this little change. I am fine with using `n` and `c` instead, or reverting completelty. But what am I supposed to do? Keep it a draft as long as I make suggetions??

Also ... your comment sounded a bit salty for no good reason. Do you want people to make suggestions, or not? Sometimes they necessarily won't be to your liking, but I have no problem with reverting the change.

thibault-martinez: My comment was not salty at all and I have nothing against your change, I did say you can do it. You are free to suggest whatever you want but the review is here to elaborate on these suggestions. I definitely think we should either apply this change to the whole file (5 more lines is not going to bloat the PR), or not do it. Pick what you prefer.

Thoralf-M (approved):

thibault-martinez (approved):

PR #2102, opened 2024-02-29
Make ISA a transaction builder

### LINKED ISSUE #1991: Turn ISA into a fully fledged TransactionBuilder?
- [ ] Merge `prepare_transaction` and `select_inputs`?
- [x] Remove `try_select_inputs` comments
- [ ] Remove `validate_transitions`?
- [ ] `filter_inputs` still needed?

### LINKED ISSUE #2123: Fix mana allotments

DaughterOfMars:
# Description of change

This is mostly just a rename right now. If there are any suggestions on additional changes, let me know.

I also added a version of `build_transaction` to the client. It should fetch available inputs using a list of provided addresses, and otherwise just does everything that isn't specific wallet functionality.

## Related Issues

Closes #1991


sdk/src/client/api/block_builder/options.rs
```diff
@@ -28,14 +28,14 @@ pub struct TransactionOptions {
...
-    /// Specifies what needs to be burned during input selection.
+    /// Specifies what needs to be burned during iransaction builder.
     pub burn: Option<Burn>,
     /// A string attached to the transaction.
     pub note: Option<String>,
     /// Whether to allow sending a micro amount.
     pub allow_micro_amount: bool,
     /// Whether to allow the selection of additional inputs for this transaction.
-    pub allow_additional_input_selection: bool,
+    pub allow_additional_transaction_builder: bool,
```
DaughterOfMars: 😆 whoopsie

sdk/src/client/api/block_builder/transaction_builder/mod.rs
```diff
@@ -427,7 +538,7 @@ impl InputSelection {
     }
 
     /// Disables selecting additional inputs.
-    pub fn disable_additional_input_selection(mut self) -> Self {
+    pub fn disable_additional_transaction_builder(mut self) -> Self {
```
thibault-martinez: :trollface:

sdk/tests/client/transaction_builder/account_outputs.rs
```diff
@@ -5,7 +5,7 @@ use std::str::FromStr;
 
 use iota_sdk::{
     client::{
-        api::input_selection::{Burn, Error, InputSelection, Requirement},
+        api::transaction_builder::{Burn, Error, Requirement, TransactionBuilder},
```
thibault-martinez: I know it was already like that but do you think it's actually fine to have such generic symbol names like `Requirement` exported? Maybe `TransactionRequirement` would be a bit more explicit.

sdk/src/client/api/block_builder/transaction_builder/mod.rs
```diff
@@ -40,12 +46,133 @@ use crate::{
...
+
+        if !options.allow_additional_input_selection {
+            transaction_builder = transaction_builder.disable_additional_transaction_builder();
+        }
+
+        if let Some(capabilities) = options.capabilities {
+            transaction_builder = transaction_builder.with_transaction_capabilities(capabilities)
+        }
+
+        let prepared_transaction_data = transaction_builder.select()?;
```
thibault-martinez: Rename to build/finish?

sdk/src/client/api/block_builder/transaction_builder/mod.rs
DaughterOfMars: dangit!!!!

sdk/tests/client/transaction_builder/account_outputs.rs
DaughterOfMars: I do think it's generally fine, since anything can be renamed at the local level

sdk/src/client/api/block_builder/transaction_builder/mod.rs
DaughterOfMars: good call

thibault-martinez: Absolutely my fault and I'm happy to do the change for you but we should actually go with `finish` IMO. In the rest of the codebase, build usually gives us a builder and finish gives us the final type.

DaughterOfMars: alright

sdk/src/types/block/semantic/mod.rs
```diff
@@ -334,9 +334,11 @@ impl<'a> SemanticValidationContext<'a> {
                 Output::Delegation(output) => (output.amount(), 0, None, None),
             };
 
-            if let Some(sender) = features.and_then(Features::sender) {
-                if !self.unlocked_addresses.contains(sender.address()) {
-                    return Err(TransactionFailureReason::SenderFeatureNotUnlocked);
+            if self.unlocks.is_some() {
```
thibault-martinez, line 337: Also add it for the issuer feature checks?

sdk/src/wallet/events/types.rs
```diff
@@ -192,8 +192,8 @@ pub struct TransactionInclusionEvent {
 #[derive(Clone, Debug, Eq, PartialEq)]
 #[non_exhaustive]
 pub enum TransactionProgressEvent {
-    /// Performing input selection.
-    SelectingInputs,
+    /// Building a transaction.
+    BuildingTransaction,
```
thibault-martinez, line 196: Also need to change this in the bindings

sdk/src/client/api/block_builder/options.rs
```diff
@@ -24,11 +23,9 @@ pub struct TransactionOptions {
     pub remainder_value_strategy: RemainderValueStrategy,
     /// An optional tagged data payload.
     pub tagged_data_payload: Option<TaggedDataPayload>,
-    /// Transaction context inputs to include.
-    pub context_inputs: Vec<ContextInput>,
     /// Inputs that must be used for the transaction.
     pub required_inputs: BTreeSet<OutputId>,
-    /// Specifies what needs to be burned during input selection.
+    /// Specifies what needs to be burned in the transaction.
```
thibault-martinez, line 28: This comment needs to be changed in bindings as well

sdk/src/wallet/operations/transaction/high_level/minting/create_native_token.rs
```diff
@@ -120,28 +120,26 @@ where
...
 
-                    if let Some(foundry_metadata) = params.foundry_metadata {
-                        foundry_builder = foundry_builder.add_immutable_feature(foundry_metadata);
-                    }
+                if let Some(foundry_metadata) = params.foundry_metadata {
+                    foundry_builder = foundry_builder.add_immutable_feature(foundry_metadata);
+                }
 
-                    foundry_builder.finish_output()?
-                }, // Native Tokens will be added automatically in the remainder output in try_select_inputs()
```
Thoralf-M, line 143: Keep the comments (modified)? There is another one in mint_native_token.rs

sdk/src/client/api/block_builder/transaction.rs
```diff
@@ -27,63 +28,85 @@ const SINGLE_UNLOCK_LENGTH: usize = 1 + 1 + Ed25519Signature::PUBLIC_KEY_LENGTH
...
-) -> crate::client::error::Result<()> {
-    let signed_transaction_payload_bytes = signed_transaction_payload.pack_to_vec();
-    if signed_transaction_payload_bytes.len() > MAX_TX_LENGTH_FOR_BLOCK_WITH_8_PARENTS {
-        return Err(Error::InvalidSignedTransactionPayloadLength {
-            length: signed_transaction_payload_bytes.len(),
-            max_length: MAX_TX_LENGTH_FOR_BLOCK_WITH_8_PARENTS,
-        });
+impl SignedTransactionPayload {
+    /// Verifies that the signed transaction payload doesn't exceed the block size limit with 8 parents.
+    pub fn validate_length(&self) -> crate::client::error::Result<()> {
```
Thoralf-M: Shouldn't we move this in the transaction builder finish()? Since we also do the semantic validation there

DaughterOfMars: We don't sign there?

thibault-martinez: I've created https://github.com/iotaledger/iota-sdk/issues/2119 btw, I'm not even sure we need these methods at all...

sdk/src/wallet/operations/transaction/high_level/minting/create_native_token.rs
DaughterOfMars: I don't really see the purpose of them tbh

sdk/src/client/api/block_builder/transaction.rs
Thoralf-M: Ah yeah but there is the other one on Transaction, no need to change anything in this PR then :+1:

bindings/python/iota_sdk/types/event.py
```diff
@@ -19,3 +19,22 @@ class WalletEventType(IntEnum):
     SpentOutput = 2
     TransactionInclusion = 3
     TransactionProgress = 4
```
thibault-martinez, line 21: I know it was already like that but surely that won't work with the nested `TransactionProgressEvent`?

sdk/src/wallet/operations/transaction/high_level/minting/create_native_token.rs
Thoralf-M: If you think that it's clear without it what's going on :shrug:

bindings/python/iota_sdk/types/event.py
DaughterOfMars: not sure 🤷‍♀️ make an issue?

thibault-martinez: Sure

thibault-martinez: https://github.com/iotaledger/iota-sdk/issues/2121

sdk/tests/client/transaction_builder/delegation_outputs.rs
```diff
@@ -97,18 +98,7 @@ fn remainder_needed_for_mana() {
         &inputs,
         None,
         None,
-        protocol_parameters.clone(),
-    )
-    .validate()
-    .unwrap();
-
-    // validating with rewards
```
thibault-martinez, line 105: Can't we make it pass instead?

DaughterOfMars: It does pass, I removed it because this is already checked by the original call now

thibault-martinez: Ah ok I see, thanks

sdk/src/client/api/block_builder/transaction_builder/mod.rs
```diff
@@ -40,14 +43,133 @@ use crate::{
...
+    /// Builds a transaction using the given inputs, outputs, addresses, and options.
+    pub async fn build_transaction(
+        &self,
+        addresses: impl IntoIterator<Item = Address>,
+        outputs: impl IntoIterator<Item = Output>,
+        options: TransactionOptions,
+    ) -> crate::client::Result<PreparedTransactionData> {
+        let outputs = outputs.into_iter().collect::<Vec<_>>();
+        let addresses = addresses.into_iter().collect::<Vec<_>>();
+        // Voting output needs to be requested before to prevent a deadlock
```
Thoralf-M: Wrong comment

sdk/src/client/api/block_builder/transaction_builder/mod.rs
```diff
@@ -40,14 +43,133 @@ use crate::{
...
+            Some(self.get_account_congestion(&issuer_id, None).await?.reference_mana_cost)
+        } else {
+            None
+        };
+        let remainder_address = match options.remainder_value_strategy {
+            RemainderValueStrategy::ReuseAddress => None,
+            RemainderValueStrategy::CustomAddress(address) => Some(address),
+        };
+
+        let hrp = self.get_bech32_hrp().await?;
```
Thoralf-M: Can also just use `protocol_parameters.bech32_hrp()`

sdk/src/client/api/block_builder/transaction_builder/mod.rs
```diff
@@ -40,14 +43,133 @@ use crate::{
     },
 };
 
-/// Working state for the input selection algorithm.
+impl Client {
+    /// Builds a transaction using the given inputs, outputs, addresses, and options.
+    pub async fn build_transaction(
```
Thoralf-M, line 49: Wait, why is this on the client?

sdk/src/client/api/block_builder/transaction_builder/mod.rs
DaughterOfMars: thanks

sdk/src/client/api/block_builder/transaction_builder/mod.rs
DaughterOfMars: Because it can be :)

Thoralf-M: Tbh not sure if we want this here, it's a bit limited and might be annoying to maintain
Would rather put this in an example

DaughterOfMars: how is it limited?

sdk/src/client/api/block_builder/transaction_builder/requirement/context_inputs.rs
```diff
@@ -1,26 +1,28 @@
...
+        for input in &self.selected_inputs {
             match &input.output {
                 // Transitioning an issuer account requires a BlockIssuanceCreditContextInput.
                 Output::Account(account) => {
                     if account.features().block_issuer().is_some() {
                         log::debug!("Adding block issuance context input for transitioned account output");
-                        self.context_inputs.insert(
+                        self.bic_context_inputs.insert(
                             BlockIssuanceCreditContextInput::from(account.account_id_non_null(input.output_id()))
                                 .into(),
```
thibault-martinez: I guess this and other .into() are not needed anymore

DaughterOfMars: tru

sdk/src/client/api/block_builder/transaction_builder/requirement/context_inputs.rs
```diff
@@ -49,7 +51,7 @@ impl InputSelection {
 
             if self.mana_rewards.get(input.output_id()).is_some() {
                 log::debug!("Adding reward and commitment context input for output claiming mana rewards");
-                self.context_inputs.insert(RewardContextInput::new(idx as _)?.into());
+                self.reward_context_inputs.insert(*input.output_id());
```
thibault-martinez, line 55: So it's only later that we actually convert to `RewardContextInput`?

sdk/src/client/api/block_builder/transaction_builder/requirement/context_inputs.rs
```diff
@@ -81,25 +83,16 @@ impl InputSelection {
...
-                .context_inputs
-                .iter()
-                .any(|c| c.kind() == CommitmentContextInput::KIND)
-        {
+        if needs_commitment_context && self.commitment_context_input.is_none() {
             // TODO https://github.com/iotaledger/iota-sdk/issues/1740
-            self.context_inputs
-                .insert(CommitmentContextInput::new(self.latest_slot_commitment_id).into());
+            self.commitment_context_input
+                .replace(CommitmentContextInput::new(self.latest_slot_commitment_id).into());
```
thibault-martinez: Doesn't really make any difference but `replace` while we're under an `is_none` condition feels a bit weird

sdk/src/client/api/block_builder/transaction_builder/mod.rs
Thoralf-M: Because just the addresses are provided there is `chain: None,`, so after building the user has to know which input needs to be signed with what

sdk/src/client/api/block_builder/transaction_builder/mod.rs
```diff
@@ -40,14 +43,132 @@ use crate::{
...
+            RemainderValueStrategy::ReuseAddress => None,
+            RemainderValueStrategy::CustomAddress(address) => Some(address),
+        };
+
+        let hrp = protocol_parameters.bech32_hrp();
+
+        let mut available_input_ids = HashSet::new();
+        for address in &addresses {
+            let mut cursor = None;
+            loop {
```
Thoralf-M: This loop and cursor stuff isn't needed, that's already handled inside of `output_ids()`

sdk/src/wallet/operations/transaction/build_transaction.rs
```diff
@@ -144,32 +147,30 @@ where
             protocol_parameters.clone(),
         )
         .with_required_inputs(options.required_inputs)
-        .with_forbidden_inputs(forbidden_inputs)
```
thibault-martinez, line 143: I don't really have a string opinion, do you think this is better UX-wise? It is easier to specify what is not available or to remove it from what it available?

sdk/src/client/api/block_builder/transaction_builder/requirement/context_inputs.rs
DaughterOfMars: Yes, because the order of the inputs isn't set yet so picking an index now is a bad idea

sdk/src/client/api/block_builder/transaction_builder/requirement/context_inputs.rs
DaughterOfMars: just less verbose

sdk/src/client/api/block_builder/transaction_builder/mod.rs
DaughterOfMars: oh nice!

sdk/src/client/api/block_builder/transaction_builder/requirement/context_inputs.rs
thibault-martinez: One letter less verbose :trollface: ?

sdk/src/client/api/block_builder/transaction_builder/mod.rs
thibault-martinez: So this is new, right? This hasn't been moved from somewhere else?

sdk/src/client/api/block_builder/transaction_builder/mod.rs
```diff
@@ -156,16 +278,7 @@ impl InputSelection {
...
         ]);
 
-        // Removes forbidden inputs from available inputs.
-        self.available_inputs
-            .retain(|input| !self.forbidden_inputs.contains(input.output_id()));
-
         for required_input in self.required_inputs.clone() {
-            // Checks that required input is not forbidden.
-            if self.forbidden_inputs.contains(&required_input) {
-                return Err(Error::RequiredInputIsForbidden(required_input));
```
thibault-martinez, line 166: Remove the variant then?

sdk/src/wallet/operations/transaction/build_transaction.rs
DaughterOfMars: Yeah IMO it's much easier to just not provide the inputs you don't want to have available

sdk/src/client/api/block_builder/transaction_builder/mod.rs
Thoralf-M: Yes it's new

DaughterOfMars: is there another way we could provide that, you think?

Thoralf-M: Together with the address is probably the best, but even then I'm not sure about this. If users want to do something with foundries , they would also need to get the outputs themselves first to get account outputs, so they can get their addresses and provide it
And there might be more things

DaughterOfMars: sure, but that's not completely unreasonable for a stateless transaction builder. Would there need to be a chain per address?

Thoralf-M: Depends on how flexible you want it to be :shrug:

DaughterOfMars: I mean the wallet only uses the one chain so

thibault-martinez: Kinda agree with Thoralf on this, not sure it's super useful and in general I would add things only when necessary/asked. I don't have a strong opinion though.

DaughterOfMars: Really though if we provide the chain similarly to how the wallet has one then the functionality will be identical right?

Thoralf-M: What's the point then in having it, if it's doing the same thing? :trollface:
The way the inputs(outputs) are retrieved and filtered is a bit different, apart from that it's doing the same I guess

Thoralf-M (approved):

thibault-martinez (approved):

PR #2108, opened 2024-03-01
Clean up wallet syncing code

### LINKED ISSUE #1905: Cleanup wallet syncing code
## Description

The current syncing logic is still kinda based on the previous multiple accounts wallet. With only a single account now we should be able to simplify this logic.

## Motivation

Some cleanup might help making this code a bit easier to maintain.

Yes/No.

Alex6323:
# Description of change

An attempt at cleaning up some wallet syncing code.

#1905


bindings/core/src/response.rs
```diff
@@ -300,13 +300,15 @@ pub enum Response {
     /// Response for:
     /// - [`ClaimableOutputs`](crate::method::WalletMethod::ClaimableOutputs)
     OutputIds(Vec<OutputId>),
+    // TODO: update bindings in a separate PR
     /// Response for:
     /// - [`GetOutput`](crate::method::WalletMethod::GetOutput)
-    OutputData(Option<Box<OutputData>>),
+    OutputData(Option<Box<OutputWithExtendedMetadata>>),
+    // TODO: update bindings in a separate PR
```
Thoralf-M: With the rename or what?

Alex6323: yup

sdk/src/wallet/types/address.rs
```diff
@@ -1,18 +1,34 @@
...
+    types::block::{address::Bech32Address, output::OutputId},
+    wallet::OutputWithExtendedMetadata,
+};
 
-/// An account address with unspent output_ids for unspent outputs.
-#[derive(Debug, Getters, Setters, Clone, Serialize, Deserialize, Eq, PartialEq)]
+#[derive(Debug, Getters, Setters, Clone, Serialize, Deserialize, Eq, PartialEq, derive_more::Deref)]
+#[serde(rename_all = "camelCase")]
+#[getset(get = "pub")]
+pub(crate) struct AddressWithUnspentOutputIds {
```
Thoralf-M, line 15: Why add this and not just use AddressWithUnspentOutputs where the unspent_outputs are empty as long as there are none? There are some places where you do conversions between the two types that don't seem necessary if it would be just a single type

bindings/core/src/response.rs
Thoralf-M: Would put the TODO on the OutputData classes in the bindings then with `rename`, here it's not really clear since the types itself are still the same

sdk/src/wallet/types/address.rs
Alex6323: They were truly separate types at first with one holding the ids and the other holding the outputs, but I had to change it. I would still keep the two types, but `AddressWithUnspentOutputs` can just hold an `AddressWithUnspentOutputIds` as a field, so we don't duplicte.

Thoralf-M: If you think that's not more annoying because of the nesting

Alex6323: it's actually making things less annoying as I don't need to destructure in a few cases

bindings/core/src/response.rs
Alex6323: The Python and the NodeJs `OutputData` has more fields {address, chain} than the Rust version used to. Did we forget to keep those types in sync?

sdk/src/wallet/operations/syncing/addresses/output_ids/account_foundry.rs
```diff
@@ -77,13 +77,9 @@ impl<S: 'static + SecretManage> Wallet<S> {
...
 
-        let mut output_ids = HashSet::new();
         let results: Vec<Result<OutputIdsResponse, WalletError>> = futures::future::try_join_all(tasks).await?;
-
-        for res in results {
-            let foundry_output_ids = res?;
-            output_ids.extend(foundry_output_ids.items);
-        }
+        let responses: Vec<OutputIdsResponse> = results.into_iter().collect::<Result<Vec<_>, _>>()?;
+        let output_ids: HashSet<OutputId> = HashSet::from_iter(responses.into_iter().flat_map(|res| res.items));
```
Thoralf-M: I know there was a HashSet also before, but this shouldn't be necessary as one Foundry can only be controlled be one account

sdk/src/wallet/operations/syncing/addresses/output_ids/mod.rs
```diff
@@ -230,20 +229,21 @@ impl<S: 'static + SecretManage> Wallet<S> {
...
         options: &SyncOptions,
-    ) -> Result<(Vec<AddressWithUnspentOutputs>, Vec<OutputId>), WalletError> {
+    ) -> Result<(Vec<AddressWithUnspentOutputIds>, Vec<SpentOutput>), WalletError> {
         log::debug!("[SYNC] start get_output_ids_for_addresses");
         let address_output_ids_start_time = Instant::now();
 
-        let mut addresses_with_outputs = Vec::new();
+        let mut addresses_with_unspent_outputs = Vec::new();
         // spent outputs or account/nft/foundries that don't get synced anymore, because of other sync options
+        // TODO: can we rename this to `spent_or_ignored_outputs`? (they are ignored because ^)
```
Thoralf-M: Fine with me if you find that easier to understand

sdk/src/wallet/types/mod.rs
```diff
@@ -32,23 +32,30 @@ use crate::{
     wallet::WalletError,
 };
 
-/// An output with metadata
+// TODO: This type is heavily used in the syncing logic so I found it appropriate to change its name in this PR,
```
Thoralf-M: What's the TODO here? :trollface:

sdk/src/wallet/operations/syncing/addresses/output_ids/mod.rs
Alex6323: not necessarily easier to understand, but more concise than `not anymore synced`

sdk/src/wallet/types/mod.rs
Alex6323: removing my explanation of the rename once you had a chance to voice your opinion on the change :smile:

sdk/src/wallet/operations/syncing/addresses/output_ids/account_foundry.rs
Alex6323: if you're absolutely sure I can remove the HashSet

Thoralf-M: Node needs to return wrong information for it  to be be wrong, but that would be no different to the other functions where we get output ids

sdk/src/wallet/operations/syncing/addresses/output_ids/mod.rs
Thoralf-M: But renaming OutputData to OutputWithExtendedMetadata :trollface:

bindings/core/src/response.rs
Thoralf-M: Yes

sdk/src/wallet/operations/syncing/addresses/output_ids/mod.rs
```diff
@@ -276,23 +276,24 @@ impl<S: 'static + SecretManage> Wallet<S> {
...
-                if !output_ids.is_empty() {
+                if !new_unspent_output_ids.is_empty() {
                     // outputs we had before, but now not anymore, got spent or are account/nft/foundries that don't
                     // get synced anymore because of other sync options
-                    for output_id in address.output_ids {
-                        if !output_ids.contains(&output_id) {
-                            spent_or_not_anymore_synced_outputs.push(output_id);
+                    for output_id in address.unspent_output_ids {
+                        if !new_unspent_output_ids.contains(&output_id) {
+                            spent_or_not_anymore_synced_outputs.push(output_id.into());
```
Thoralf-M: Check clippy, this `.into()` isn't necessary

sdk/src/wallet/types/address.rs
```diff
@@ -4,15 +4,27 @@
...
-    /// The address.
-    pub(crate) address: Bech32Address,
-    /// Output ids
-    pub(crate) output_ids: Vec<OutputId>,
+    #[serde(flatten)]
+    pub(crate) address_with_unspent_output_ids: AddressWithUnspentOutputIds,
+    pub(crate) unspent_outputs: Vec<OutputWithExtendedMetadata>,
 }
+
+pub(crate) type SpentOutput = OutputId;
```
Thoralf-M: Would prefer this
```suggestion
pub(crate) type SpentOutputId = OutputId;
```

sdk/src/wallet/operations/syncing/addresses/output_ids/mod.rs
Alex6323: remnant of me trying to make it a new-type instead of a type alias ... thanks

sdk/src/wallet/operations/syncing/addresses/output_ids/mod.rs
Alex6323: There are a few options I considered:
* `OutputData` -> i.e. keeping it ... could mean anything and nothing
* `OutputWithExtMetadata` -> kinda looks ugly, although valid option, too
* `OutputWithExtData` or `OutputWithData` -> kinda looks ugly and this `Data` thing again that can mean ...
* `OutputWithMetadataExt` -> Ext suffix usually used for traits
* `OutputWithExtendedMetadata` -> long, but most accurate and unambiguous

sdk/src/wallet/operations/syncing/addresses/output_ids/mod.rs
Alex6323: Removed

bindings/core/src/response.rs
Alex6323: Updated the bindings

sdk/src/wallet/types/mod.rs
Alex6323: removed

sdk/src/wallet/operations/syncing/addresses/output_ids/account_foundry.rs
Alex6323: removed

sdk/src/wallet/operations/syncing/addresses/output_ids/mod.rs
Alex6323: Already done.

sdk/src/wallet/operations/syncing/mod.rs
```diff
@@ -18,13 +18,18 @@ use crate::{
...
+            Balance, OutputWithExtendedMetadata,
+        },
         Wallet, WalletError,
     },
 };
 
 impl<S: 'static + SecretManage> Wallet<S> {
-    /// Set the fallback SyncOptions for account syncing.
+    // TODO: can we rename this to `set_fallback_sync_options`? It's no biggie, but if I read `default` in Rust
+    // if feels like it would use `SyncOptions::default()`, but that's not what we mean here, right?
```
Thoralf-M: Yes it's not the Rust default, but it's also exposed to the bindings and might be weird there?

Alex6323: not sure, but I am fine with not doing the change

Thoralf-M: With the definition from here https://stackoverflow.com/a/76104060 it sounds like default would be the correct one :trollface:

Alex6323: I agree with the reasoning in that answer, but it's not considering the Rust aspect. `fn default_sync_options(&self)` (method below) is confusing, but ... we could explicitly mention in the doc comment, that it doesn't return `SyncOptions::default()` necessarily.

Alex6323: removed

bindings/python/iota_sdk/types/output_with_extended_metadata.py
```diff
@@ -12,20 +12,20 @@
...
+        output_id: The corresponding output ID.
         network_id: The network ID the output belongs to.
         remainder: Whether the output represents a remainder amount.
     """
-    output_id: OutputId
-    metadata: OutputMetadata
     output: Output
+    metadata: OutputMetadata
     output_id_proof: OutputIdProof
+    output_id: OutputId
```
DaughterOfMars: isn't the ID already in the metadata?

DaughterOfMars (commented):
tbh I don't like the name. Maybe `ExtendedOutputMetadata` would be better? I would even be happier just leaving it `OutputData`...

Alex6323:
> tbh I don't like the name. Maybe `ExtendedOutputMetadata` would be better? I would even be happier just leaving it `OutputData`...

yeah, believe me when I say I am not fond of this name either, but I remember us all not really liking `...Data` types (looking at you `WalletData`), because that's just a lazy name giving you no clue at all about what it contains, i.e. everything is data. 

Anyways in this case we can make an exception, I guess :man_shrugging:

bindings/python/iota_sdk/types/output_with_extended_metadata.py
Alex6323: yes it is, and I mentioned this weirdness already, but should probable be dealt with separately as this would require some discussion. AFAIR this is done on purpose, because the metadata can't be trusted 100%.

thibault-martinez:
> > tbh I don't like the name. Maybe `ExtendedOutputMetadata` would be better? I would even be happier just leaving it `OutputData`...
> 
> yeah, believe me when I say I am not fond of this name either, but I remember us all not really liking `...Data` types (looking at you `WalletData`), because that's just a lazy name giving you no clue at all about what it contains, i.e. everything is data.
> 
> Anyways in this case we can make an exception, I guess 🤷‍♂️

But we said we'll revert the rename anyway?

bindings/nodejs/lib/types/wallet/output.ts
```diff
@@ -15,28 +14,23 @@ export enum OutputsToClaim {
...
 
-/** An output with metadata */
-export class OutputData {
-    /** The identifier of an Output */
-    outputId!: OutputId;
-    /** The metadata of the output */
-    metadata!: OutputMetadataResponse;
-    /** The actual Output */
+/** An output with extended metadata */
+export class OutputWithExtendedMetadata {
```
Thoralf-M: Still renamed here

bindings/python/iota_sdk/types/output_with_extended_metadata.py
```diff
@@ -12,20 +12,20 @@
 
 @json
 @dataclass
-class OutputData:
-    """Output data.
+class OutputWithExtendedMetadata:
```
Thoralf-M: And here

bindings/nodejs/lib/types/wallet/output.ts
Alex6323: Removed

bindings/python/iota_sdk/types/output_with_extended_metadata.py
Alex6323: Fixed

cli/src/wallet_cli/mod.rs
```diff
@@ -1389,38 +1389,38 @@ async fn print_wallet_address(wallet: &Wallet) -> Result<(), Error> {
     let mut delegations = Vec::new();
     let mut anchors = Vec::new();
 
-    for output_data in wallet.ledger().await.unspent_outputs().values() {
-        let output_id = output_data.output_id;
+    for output_with_ext_metadata in wallet.ledger().await.unspent_outputs().values() {
```
Thoralf-M: You still want to call it `output_with_ext_metadata` even though the struct isn't called like that? Many renames because of that

Alex6323: probably not

Thoralf-M (approved):

DaughterOfMars (approved):

PR #1928, opened 2024-02-01
Align Core API types and routes

Alex6323:
# Description of change

Updates/Aligns types and Core API routes with [TIP-48](https://editor.swagger.io/?url=https://raw.githubusercontent.com/iotaledger/tips/tip48/tips/TIP-0048/openapi3-core.yaml)

## Links to any relevant issues

Closes #1921 and #1977

PR #2112, opened 2024-03-01
NodeJs: add missing utxo routes

Alex6323:
* NodeJs binding: Adds `GetOutputMetadata` and `GetOutputWithMetadata`;

Also contains other fixes and doc alignments that were missed in #1928:
* `postBlockRaw` now receives the bytes from binding side
* `GetRewards` was renamed to `GetOutputManaRewards`
* The NodeJs binding still lacked some `Raw` methods;
* Some actual optional values weren't properly typed;
* The NodeJs binding docs weren't fully aligned yet;
* `transaction-failure-reason.ts` wasn't exported;
* Removes the `getNetworkInfo` method (bc there is `getProtocolParameters`)


bindings/nodejs/lib/client/client.ts
```diff
@@ -378,42 +402,60 @@ export class Client {
...
+        const response = await this.methodHandler.callMethod({
+            name: 'getOutputWithMetadata',
+            data: {
+                outputId,
+            },
+        });
+
+        const parsed = JSON.parse(
+            response,
+        ) as Response<OutputWithMetadataResponse>;
```
thibault-martinez, line 451: When do we need that stuff exactly?

Alex6323: good question ... we might not ... I'll check

bindings/nodejs/lib/client/client.ts
```diff
@@ -604,6 +761,19 @@ export class Client {
         return plainToInstance(UTXOInput, parsed.payload);
     }
 
+    // Other routes.
+
+    /**
+     * Get the network related information such as network_id.
```
Thoralf-M: This just returns protocol parameters now, let's remove it instead, since we have `GetProtocolParameters`?

bindings/python/iota_sdk/client/_node_core_api.py
```diff
@@ -84,18 +84,19 @@ def get_info(self, url: str, auth=None) -> InfoResponse:
 
     # Accounts routes.
 
-    def get_account_congestion(self, account_id: HexStr) -> CongestionResponse:
+    def get_account_congestion(self, account_id: HexStr, work_score: Optional[int]) -> CongestionResponse:
```
Thoralf-M: We should have a default None value, same for the other methods with optional params below
```suggestion
    def get_account_congestion(self, account_id: HexStr, work_score: Optional[int] = None) -> CongestionResponse:
```

bindings/python/iota_sdk/client/_node_core_api.py
```diff
@@ -162,15 +163,15 @@ def post_block(self, block: Block) -> BlockId:
             'block': block
         })
 
-    def post_block_raw(self, block: Block) -> BlockId:
+    def post_block_raw(self, block_bytes: List[int]) -> BlockId:
```
Thoralf-M: There is a bytes type we can use
```suggestion
    def post_block_raw(self, block_bytes: bytes) -> BlockId:
```

bindings/nodejs/lib/client/client.ts
Alex6323: so, I checked the response and it doesn't seem to make any difference, but someone other than me did this for classes, and probably for a good reason to make sure they are properly instantiated.

bindings/python/iota_sdk/client/_node_core_api.py
Alex6323: yup

bindings/nodejs/lib/client/client.ts
Alex6323: makes sense

Alex6323: Removed

bindings/python/iota_sdk/client/_node_core_api.py
```diff
@@ -84,18 +84,19 @@ def get_info(self, url: str, auth=None) -> InfoResponse:
...
-            'accountId': account_id
+            'accountId': account_id,
+            'workScore': work_score
         }))
 
     # Rewards routes.
 
     def get_output_mana_rewards(
-            self, output_id: OutputId, slot_index: SlotIndex) -> ManaRewardsResponse:
+            self, output_id: OutputId, slot_index: Optional[SlotIndex]) -> ManaRewardsResponse:
```
Thoralf-M:
```suggestion
            self, output_id: OutputId, slot_index: Optional[SlotIndex] = None) -> ManaRewardsResponse:
```

Thoralf-M: Still more places in other methods below

sdk/src/client/node_api/core/routes.rs
```diff
@@ -124,22 +126,10 @@ impl Client {
...
-
-        self.get_request(PATH, query.as_deref(), false).await
-    }
-
     // Validators routes.
 
-    /// Returns information of all registered validators and if they are active.
-    /// GET JSON to /api/core/v3/validators
+    /// Returns information of all stakers (registered validators) and if they are active, ordered by their holding
+    /// stake. GET /api/core/v3/validators
```
Thoralf-M:
```suggestion
    /// stake. 
    /// GET /api/core/v3/validators
```

thibault-martinez (approved):

Thoralf-M (approved):

PR #2116, opened 2024-03-01
Fix balance locking

### LINKED ISSUE #2113: Fix balance in wasm
https://github.com/iotaledger/iota-sdk/pull/2077 made it panic in wasm

```
iota_sdk_wasm.js:846 panicked at /iota-sdk/sdk/src/wallet/operations/balance.rs:36:9:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

DaughterOfMars:
# Description of change

Hopefully this fixes the issue with balance.


sdk/src/wallet/operations/balance.rs
```diff
@@ -22,340 +22,345 @@ where
...
+            tokio::task::spawn_blocking(move || tokio::runtime::Handle::current().block_on(wallet.balance_inner()))
+                .await?
+        }
+        #[cfg(target_family = "wasm")]
+        {
+            self.balance_inner().await
+        }
+    }
+
+    async fn balance_inner(&self) -> Result<Balance> {
```
Thoralf-M: This all isn't needed as it changes nothing? Not sure how we ended up like this, but looks like it works perfectly fine without any weird task spawning/blocking now
```suggestion
```

DaughterOfMars: It works of course but the entire purpose of blocking here was to avoid deadlocking because we're doing some very unsafe operations

Thoralf-M: But is this actually still the case? To me it looks like we don't have anything after `let wallet_ledger` that could result in a deadlock

DaughterOfMars: the call to the client can

Thoralf-M: It just returns a reference, no lock involved?

DaughterOfMars: I meant the call using the client to the API 😆

Thoralf-M (dismissed):

Thoralf-M (dismissed):

sdk/src/wallet/operations/balance.rs
```diff
@@ -24,334 +26,328 @@ impl<S: 'static + SecretManage> Wallet<S> {
...
+                Output::Account(account) => {
+                    // Add amount
+                    balance.base_coin.total += account.amount();
+                    balance.mana.total += output.decayed_mana(
+                        &protocol_parameters,
+                        output_id.transaction_id().slot_index(),
+                        slot_index,
+                    )?;
+                    // Add mana rewards
+                    reward_outputs.insert(*output_id);
```
thibault-martinez: Should we check if it's staking? Accounts are going to be used a lot but very few of them will be staking, less API calls?

DaughterOfMars: sure, why not

DaughterOfMars: done

Thoralf-M (approved):

thibault-martinez (approved):

PR #2120, opened 2024-03-04
Handle failed transactions/blocks

### LINKED ISSUE #2111: Handle potential problems from submitting blocks
Handle potential problems from submitting the signed transaction in a block
https://github.com/iotaledger/iota-sdk/blob/37d52aae84960f087e0687d741e9e2e820ce2e7b/sdk/src/wallet/operations/transaction/mod.rs#L133-L146
In stardust we just reissued the transaction payload in new blocks, but with Mana and commitmentcontextinputs this isn't working fine anymore.
A block could potentially be submitted, but an error returned. A block could be submitted, but failed.
We should probably try to submit the block multiple times depending on the error we get and if it fails make sure to also set the transaction state correct and unlock its inputs

Thoralf-M:
# Description of change

Handle issues from posting the block, if we can't get it posted 3 times, we will return None for the block id and during syncing when we also can't get the block, we will set the tx state as conflicting and unlock its inputs so they're available for new transactions again.

https://github.com/iotaledger/iota-sdk/issues/2111

## Type of change

cli wallet and

```Rust
#[tokio::test]
async fn send_amount_in_loop() -> Result<(), Box<dyn std::error::Error>> {
        .name("wallet.log")
        .level_filter(log::LevelFilter::Debug);
    let config = fern_logger::LoggerConfig::build()
        .with_output(logger_output_config)
        .finish();
    fern_logger::logger_init(config).unwrap();

    let storage_path = "test-storage/send_amount";
    setup(storage_path)?;

    println!("Creating wallet with block issuer account...");
    let wallet_0 = make_wallet(storage_path, None, None).await?;
    request_funds(&wallet_0).await?;

    for i in 0..1000 {
        match wallet_0
            .send_with_params([SendParams::new(1_000_000, wallet_0.address().await)?], None)
            .await
        {
            Ok(tx) => {
                println!("Sent tx {i}");
                if let Some(block_id) = tx.block_id {
                    println!(" in block {block_id}");
                    wallet_0
                        .wait_for_transaction_acceptance(&tx.transaction_id, None, None)
                        .await?;
                }
            }
            Err(err) => {
                println!("{err:?}");
            }
        }
        wallet_0.sync(None).await.unwrap();
    }

    tear_down(storage_path)
}
```
with a modified operations/block.rs so we get the case where no block was sent
```Rust
        let block_id = if crate::types::block::rand::bool::rand_bool() {
            self.client().post_block(&block).await?
        } else {
            // Just so it returns an error for testing, not relevant
            return Err(WalletError::InvalidVotingPower);
        };
 ```

### CODE REVIEW

File: sdk/src/wallet/operations/block.rs [hunk 1]
```diff
@@ -29,17 +31,18 @@ where
             Some(id) => id,
             None => {
                 let current_slot = self.client().get_slot_index().await?;
+                let network_id = self.client().get_network_id().await?;
```
DaughterOfMars: this is part of protocol params, so can we just move the call below up and reuse it?

Thoralf-M: Sure

sdk/src/wallet/operations/block.rs
```diff
@@ -67,9 +71,20 @@ where
...
+            if let Ok(block_id) = self.client().post_block(&block).await {
+                log::debug!("submitted block {}", block_id);
+                return Ok(block_id);
+            }
+            tokio::time::sleep(std::time::Duration::from_secs(attempt)).await;
+        }
+
+        log::debug!("submitting block {block:?}");
+
         let block_id = self.client().post_block(&block).await?;
```
DaughterOfMars: it's strange for this code to be duplicated just for the return value. Is there no error for running out of attempts?

DaughterOfMars (dismissed):

sdk/src/wallet/operations/block.rs
Thoralf-M: Currently we don't have one, ofc we could add one, but I don't think it's really worth it

DaughterOfMars: any way to remove the duplicated code then?

Thoralf-M: Yes, but not sure if that's really better now

sdk/src/wallet/core/mod.rs
```diff
@@ -299,9 +299,12 @@ impl WalletLedger {
     }
 
     // Returns the first possible unexpired block issuer Account id, which can be an implicit account.
-    pub fn first_block_issuer_account_id(&self, current_slot: SlotIndex) -> Option<AccountId> {
+    pub fn first_block_issuer_account_id(&self, current_slot: SlotIndex, network_id: u64) -> Option<AccountId> {
```
thibault-martinez, line 302: Seems annoying :(

Thoralf-M: Do you have a better idea?

Thoralf-M: Current slot is also annoying, but it's only on the WalletLedger, from the Wallet you get it without providing any params, so I don't think it's that bad

Alex6323 (dismissed):

sdk/src/wallet/core/mod.rs
thibault-martinez: Do we even need the check? I guess there are plenty of places where we could do this and don't

Thoralf-M: Don't see where else this would be an issue, here it was an issue for me because I changed the node and then it selected the account from another network (in the input selection we filter this out, but for the issuer we didn't)

thibault-martinez: Ok

thibault-martinez (approved):

DaughterOfMars (approved):

sdk/src/wallet/operations/block.rs
DaughterOfMars: I think it is, because it returns the error

PR #2135, opened 2024-03-05
Replace hexPublicKeyToBech32Address by publicKeyHash

### LINKED ISSUE #2080: Change HexPublicKeyToBech32Address to PublicKeyHash?
Conversion to bech32 address can be done with `AddressToBech32` and to get the public key hash one would need to convert it from bech32 back which is more annoying I think

In the bindings core we could replace it with `Blake2b256Hash` which could also be called in `ComputeAccountId, ComputeNftId` so we have less code there

Thoralf-M:
# Description of change

Replacey hexPublicKeyToBech32Address by publicKeyHash
Use Blake2b256Hash instead of ComputeAccountId/ComputeNftId/PublicKeyHash in bindings core so we have less to maintain there

Closes https://github.com/iotaledger/iota-sdk/issues/2080

## Type of change

Tests


bindings/nodejs/examples/how_tos/sign_and_verify_ed25519/sign-ed25519.ts
```diff
@@ -65,8 +65,8 @@ async function run() {
             `Public key: ${ed25519Signature.publicKey}\nSignature: ${ed25519Signature.signature}`,
         );
 
-        const bech32Address = Utils.hexPublicKeyToBech32Address(
-            ed25519Signature.publicKey,
+        const bech32Address = Utils.AddressToBech32(
```
Thoralf-M:
```suggestion
        const bech32Address = Utils.addressToBech32(
```

bindings/nodejs/lib/types/client/bridge/client.ts
```diff
@@ -357,8 +357,8 @@ export interface __NftIdToBech32Method__ {
     };
 }
 
-export interface __HexPublicKeyToBech32AddressMethod__ {
-    name: 'hexPublicKeyToBech32Address';
+export interface __PublicKeyHashMethod__ {
```
thibault-martinez: Do we even need a bridge if there is no binding for it?

Thoralf-M: No, forgot to remove

thibault-martinez (approved):

DaughterOfMars (approved):

PR #2152, opened 2024-03-06
Better transitions

### LINKED ISSUE #2176: Transaction builder seems to be able to select too many inputs
And can then panic
```
thread 'tokio-runtime-worker' panicked at sdk/src/client/api/block_builder/transaction_builder/mod.rs:589:68:
called `Result::unwrap()` on an `Err` value: RewardIndex(InvalidBoundedU16(129))
```

DaughterOfMars:
# Description of change

Add some transitions that the transaction builder can do automatically. Attempt to put excess chain amount and mana on the chained output. Completely disallow editing provided outputs.


sdk/src/client/api/block_builder/transaction_builder/transition.rs
```diff
@@ -172,7 +287,80 @@ impl TransactionBuilder {
...
             Output::Nft(nft_input) => self.transition_nft_input(nft_input, input.output_id()),
+            Output::Basic(basic_output) => self.transition_implicit_account_input(basic_output, input.output_id()),
             _ => Ok(None),
         }
     }
 }
+
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum AccountChange {
```
Thoralf-M, line 287: What about other changes like adding a metadata feature? Would we need an enum variant for all of that or can we still do that different?

DaughterOfMars: I think that's definitely possible, I just added only the stuff we currently use

Thoralf-M: What do you mean with it's possible? Would we need to add them to the enum variant or can you transition an output and do the changes without using this enum?

DaughterOfMars: You can do it manually still by adding the output, but that output cannot be changed to add mana or anything so sometimes you need an additional remainder. We could make it possible to do it entirely in the builder using this `Transitions` type

Thoralf-M: Hm and what if you want to do multiple changes to the output and still get mana remainder added to it? Like start staking and adding a metadata feature in the same tx

DaughterOfMars: Currently that would not be possible. Do you think it's worth adding this kind of thing?

Thoralf-M: Not sure, but looks like this enum would end up being another account output builder which doesn't really seem to be a good solution to me

DaughterOfMars: Well, we can always change it if there's a need in the future 🤷‍♀️

sdk/src/client/api/block_builder/transaction_builder/mod.rs
```diff
@@ -377,25 +380,47 @@ impl TransactionBuilder {
...
+
+        let mut added_amount_mana = HashMap::<Option<ChainId>, (u64, u64)>::new();
+        for (chain_id, added_amount) in self.remainders.added_amount.drain() {
+            added_amount_mana.entry(chain_id).or_default().0 = added_amount;
+        }
+        for (chain_id, added_mana) in self.remainders.added_mana.drain() {
+            added_amount_mana.entry(chain_id).or_default().1 = added_mana;
+        }
+
+        for (chain_id, (added_amount, added_mana)) in added_amount_mana {
```
Thoralf-M, line 394: Why is this split up between potential many outputs and not just added to a single one?

DaughterOfMars: The idea is to keep mana on the same chain ID when possible

sdk/src/client/api/block_builder/transaction_builder/error.rs
```diff
@@ -25,9 +26,16 @@ use crate::types::block::{
 pub enum TransactionBuilderError {
     #[error("additional inputs required for {0:?}, but additional input selection is disabled")]
     AdditionalInputsRequired(Requirement),
+    #[error("account {0} is already staking")]
+    AlreadyStaking(AccountId),
     /// Can't burn and transition an output at the same time.
     #[error("can't burn and transition an output at the same time, chain ID: {0}")]
     BurnAndTransition(ChainId),
+    #[error("account id {account_id} cannot end staking until {end_epoch}")]
```
Alex6323:
```suggestion
    #[error("account {account_id} cannot end staking until {end_epoch}")]
```

sdk/src/client/api/block_builder/transaction_builder/remainder.rs
```diff
@@ -109,13 +108,80 @@ impl TransactionBuilder {
             .get_remainder_address()?
             .ok_or(TransactionBuilderError::MissingInputWithEd25519Address)?;
 
+        let remainder_builder =
+            BasicOutputBuilder::new_with_minimum_amount(self.protocol_parameters.storage_score_parameters())
+                .add_unlock_condition(AddressUnlockCondition::new(Address::from(Ed25519Address::from(
+                    [0; 32],
+                ))));
```
Alex6323: maybe use `Ed25519Address::null()` instead?

DaughterOfMars: good call

Alex6323 (dismissed):

sdk/src/client/api/block_builder/transaction_builder/requirement/amount.rs
```diff
@@ -47,40 +47,39 @@ impl TransactionBuilder {
...
+                found: input_amount,
+                required: output_amount,
+            });
+        }
+
+        let mut priority_map = PriorityMap::<AmountPriority>::generate(&mut self.available_inputs);
+        loop {
+            let Some(input) = priority_map.next(output_amount - input_amount, self.latest_slot_commitment_id) else {
+                break;
+            };
```
Alex6323: `while let` ?

DaughterOfMars: done, thanks

Alex6323 (dismissed):

sdk/tests/client/transaction_builder/basic_outputs.rs
```diff
@@ -2575,12 +2575,13 @@ fn automatic_allotment_provided_in_and_output() {
 
     let selected = TransactionBuilder::new(
         inputs.clone(),
-        outputs.clone(),
+        None,
```
Thoralf-M: This should stay `outputs.clone(),` the test is called `automatic_allotment_provided_in_and_output`

Thoralf-M: Then it fails
```
thread 'client::transaction_builder::basic_outputs::automatic_allotment_provided_in_and_output' panicked at sdk/tests/client/transaction_builder/basic_outputs.rs:2573:6:
called `Result::unwrap()` on an `Err` value: InsufficientAmount { found: 1000000, required: 1014100 }
```
:cry:

DaughterOfMars: I changed it because you cannot change the provided outputs, remember

DaughterOfMars: But I can add another output to cover the remainder I guess

DaughterOfMars: Actually, ignore that, there may be something wrong, I'll investigate

DaughterOfMars: The mana calculation needed to factor in the auto allotment in order for the amount requirement to know whether it needed a remainder for mana. It's fixed now.

Alex6323 (dismissed):

Thoralf-M (changes requested):
Conflict

Thoralf-M (approved):

Alex6323 (approved):

PR #2153, opened 2024-03-06
Clean up requirements

DaughterOfMars:
# Description of change

Requirements currently return vectors of inputs, but that's inefficient and half the time we just return a new empty vec, so I removed it.


sdk/src/client/api/block_builder/transaction_builder/requirement/account.rs
```diff
@@ -42,7 +36,13 @@ impl TransactionBuilder {
                 "{account_id:?} requirement already fulfilled by {:?}",
                 input.output_id()
             );
-            return Ok(Vec::new());
+            return Ok(());
+        }
+
+        if !self.allow_additional_input_selection {
```
thibault-martinez, line 42: Can we not move this to `select_input`?
Even if it means passing it the current Requirement

DaughterOfMars: We can't, because we call `select_input` on required inputs too 😄

thibault-martinez: image

DaughterOfMars: Honestly tho it's fine because the check is not a lot of duplicated code and it allows us to easily add the requirement too 🤷‍♀️

thibault-martinez (approved):

Thoralf-M (approved):

PR #2159, opened 2024-03-07
Predict how many slots it will take to generate mana

### LINKED ISSUE #2098: Util function to get slots to generate a certain amount of mana
If one has outputs and wants to know how long one has to wait until enough Mana is generated so one can issue a block

DaughterOfMars:
# Description of change

I found a decent way to do this calculation which basically goes like this:
1. Calculate the mana generated per slot without decay
2. Calculate the number of slots it would take to reach the requirement with that generation rate (this is the earliest possible slot)
    - If the number of slots is more than u32::MAX then we either don't have enough amount or the required mana is crazy high
3. Check how much mana we actually would have generated in that number of slots
    - If we netted no mana, we know we don't have enough generation amount to get there
4. Check if we generated enough
    - If not, reduce the required mana by the amount that was generated in those slots and repeat from 1

The nice thing about this method is that approaches the solution within a bounded number of loops (though I don't know the exact bound, it's probably 7 🤷‍♀️ ).

Closes #2098


sdk/src/types/block/output/account.rs
```diff
@@ -440,6 +439,12 @@ impl AccountOutput {
             potential: potential_mana,
         })
     }
+
+    /// Returns the generation amount of the output.
+    pub fn generation_amount(&self, protocol_parameters: &ProtocolParameters) -> u64 {
```
Thoralf-M: Perhaps add something mentioning mana? Just reading this one pprobably has no idea what this is good for

DaughterOfMars: good call

sdk/src/types/block/mana/parameters.rs
```diff
@@ -196,6 +196,37 @@ impl ProtocolParameters {
...
+    ) -> Result<u32, ManaError> {
+        let mut num_slots = 0;
+        let mana_generated_per_epoch = self
+            .mana_parameters()
+            .generate_mana(generation_amount, self.slots_per_epoch());
+        let mut required_mana_remaining = required_mana;
+        loop {
+            // Get the minimum number of slots required to achieve the needed mana (i.e. not including decay)
+            num_slots +=
+                u32::try_from(1 + (required_mana_remaining * self.slots_per_epoch() as u64) / mana_generated_per_epoch)
```
Thoralf-M: Probably missing an extra check for mana_generated_per_epoch in case its 0

thread 'client::transaction_builder::outputs::insufficient_mana2' panicked at sdk/src/types/block/mana/parameters.rs:214:35:
attempt to divide by zero

```
#[test]
fn insufficient_mana2() {
    let protocol_parameters = iota_mainnet_protocol_parameters();

    let min_amount = BasicOutputBuilder::new_with_minimum_amount(protocol_parameters.storage_score_parameters())
        .with_mana(0)
        .add_unlock_condition(AddressUnlockCondition::new(
            Address::try_from_bech32(BECH32_ADDRESS_ED25519_0).unwrap(),
        ))
        .finish_output()
        .unwrap()
        .amount();

    let inputs = build_inputs(
        [
            (
                Basic {
                    amount: min_amount,
                    mana: 0,
                    address: Address::try_from_bech32(BECH32_ADDRESS_ED25519_0).unwrap(),
                    native_token: None,
                    sender: None,
                    sdruc: None,
                    timelock: None,
                    expiration: None,
                },
                None,
            ),
            (
                Basic {
                    amount: min_amount,
                    mana: 0,
                    address: Address::try_from_bech32(BECH32_ADDRESS_ED25519_0).unwrap(),
                    native_token: None,
                    sender: None,
                    sdruc: None,
                    timelock: None,
                    expiration: None,
                },
                None,
            ),
        ],
        Some(SLOT_INDEX),
    );

    let outputs = build_outputs([Basic {
        amount: 28200,
        mana: 10000,
        address: Address::try_from_bech32(BECH32_ADDRESS_ED25519_0).unwrap(),
        native_token: None,
        sender: None,
        sdruc: None,
        timelock: None,
        expiration: None,
    }]);

    let selected = TransactionBuilder::new(
        inputs.clone(),
        outputs.clone(),
        [Address::try_from_bech32(BECH32_ADDRESS_ED25519_0).unwrap()],
        SLOT_INDEX,
        SLOT_COMMITMENT_ID,
        protocol_parameters.clone(),
    )
    .finish()
    .unwrap_err();

    let TransactionBuilderError::InsufficientMana {
        found,
        required,
        slots_remaining,
    } = selected
    else {
        panic!("expected insufficient mana error")
    };
    assert_eq!(found, 0);
    assert_eq!(required, 10000);
}
```

DaughterOfMars: Ah, good point, thank you

Thoralf-M: If generation_amount is very low and doesn't generate enough mana it still fails because `mana_generated_per_epoch` will still be 0, so probably needs the check for `mana_generated_per_epoch`(for generation_amount it's not needed?)
```Rust
        if mana_generated_per_epoch == 0 {
            return Err(ManaError::InsufficientGenerationAmount);
        }
```

DaughterOfMars: Hmm, fair point

sdk/src/types/block/mana/parameters.rs
```diff
@@ -196,6 +196,43 @@ impl ProtocolParameters {
...
+        }
+        let mut required_mana_remaining = required_mana;
+        loop {
+            // Get the minimum number of slots required to achieve the needed mana (i.e. not including decay)
+            num_slots +=
+                u32::try_from(1 + (required_mana_remaining * self.slots_per_epoch() as u64) / mana_generated_per_epoch)
+                    .map_err(|_| ManaError::InsufficientGenerationAmount)?;
+            // Get the actual values after than many slots
+            let decayed_mana = stored_mana - self.mana_with_decay(stored_mana, 0, num_slots)?;
+            let generated_mana = self.generate_mana_with_decay(generation_amount, 0, num_slots)?;
```
Thoralf-M: Just noticed there is ManaParameter::annual_decay_factor_percentage shouldn't this have an effect on the decay depending on the slot range? Currently it doesn't seem to make any difference, but do we need an issue to handle this?

Thoralf-M (approved):

sdk/src/types/block/mana/parameters.rs
DaughterOfMars: I don't think so? But maybe I don't understand it

Thoralf-M: Checked again now, it's handled inside of generate_mana_with_decay() and actually makes a difference, so this needs to be changed

Thoralf-M (changes requested):
Providing a large mana amount like `1_000_000_000` in the output in `insufficient_mana()` panics
```Rust
thread 'client::transaction_builder::outputs::insufficient_mana' panicked at sdk/src/types/block/mana/parameters.rs:222:13:
attempt to add with overflow
```

sdk/src/types/block/mana/parameters.rs
Thoralf-M: So for example changing `let generated_mana = self.generate_mana_with_decay(generation_amount, 0, num_slots)?;` to `let generated_mana = self.generate_mana_with_decay(generation_amount, 100_000, 100_000 + num_slots)?;` 
changes the required slots from 66734 to 66746 because the decay at higher slot indexes increases (in the insufficient_mana() test the difference is too small to notice, but if you increase the mana on the outputside to 1000000 it gets noticeable)

sdk/tests/client/transaction_builder/outputs.rs
```diff
@@ -591,3 +591,97 @@ fn transition_no_more_than_needed_for_nft_amount() {
...
+    let TransactionBuilderError::InsufficientMana {
+        found,
+        required,
+        slots_remaining,
+    } = selected
+    else {
+        panic!("expected insufficient mana error")
+    };
+    assert_eq!(found, 0);
+    assert_eq!(required, 10000);
```
Thoralf-M, line 661: Shouldn't this value be higher because of the allotment?

sdk/tests/client/transaction_builder/outputs.rs
```diff
@@ -591,3 +591,97 @@ fn transition_no_more_than_needed_for_nft_amount() {
...
+#[test]
+fn insufficient_mana() {
+    let protocol_parameters = iota_mainnet_protocol_parameters();
+
+    let inputs = build_inputs(
+        [
+            (
+                Basic {
+                    amount: 1_000_000,
+                    mana: 0,
```
Thoralf-M, line 604: This amount doesn't seem to make any difference for the required slots, increasing this to 9000 still requires 665 slots, the same we get now if it's 0 (I added an `assert_eq!(slots_remaining, 665);` locally)

sdk/src/types/block/mana/parameters.rs
DaughterOfMars: Okay, that's fixable

sdk/tests/client/transaction_builder/outputs.rs
DaughterOfMars: I didn't specify any allotment?

sdk/tests/client/transaction_builder/outputs.rs
DaughterOfMars: You mean the stored mana?

Thoralf-M: Yes

sdk/tests/client/transaction_builder/outputs.rs
Thoralf-M: Thought it would automatically add one but default is None

sdk/tests/client/transaction_builder/outputs.rs
DaughterOfMars: That's definitely confusing because it is for sure factored in to the calculation

DaughterOfMars: Oh, I see the problem

DaughterOfMars: fixed

sdk/tests/client/transaction_builder/outputs.rs
```diff
@@ -591,3 +591,97 @@ fn transition_no_more_than_needed_for_nft_amount() {
...
+                    expiration: None,
+                },
+                None,
+            ),
+        ],
+        Some(SLOT_INDEX),
+    );
+    let outputs = build_outputs([Basic {
+        amount: 2_000_000,
+        mana: 10000,
```
Thoralf-M, line 632: Still overflows if mana is set to 1_000_000_000
```
thread 'client::transaction_builder::outputs::insufficient_mana' panicked at sdk/src/types/block/mana/parameters.rs:229:13:
attempt to add with overflow
```

sdk/src/types/block/mana/parameters.rs
```diff
@@ -196,6 +196,59 @@ impl ProtocolParameters {
...
+        loop {
+            // Get the minimum number of slots required to achieve the needed mana (i.e. not including decay)
+            let additional_slots = u32::try_from(
+                (required_mana_remaining as u128 * self.slots_per_epoch() as u128)
+                    / mana_generated_per_epoch.max(1) as u128,
+            )
+            .map_err(|_| ManaError::InsufficientGenerationAmount)?;
+
+            num_slots += additional_slots.max(1);
+            // Get the actual values after than many slots
```
Thoralf-M:
```suggestion
            // Get the actual values after that many slots
```

sdk/src/types/block/mana/parameters.rs
```diff
@@ -196,6 +196,59 @@ impl ProtocolParameters {
...
+            let decayed_mana =
+                stored_mana - self.mana_with_decay(stored_mana, current_slot, current_slot + num_slots)?;
+            let generated_mana =
+                self.generate_mana_with_decay(generation_amount, current_slot, current_slot + num_slots)?;
```
Thoralf-M, line 235: I think after this, current_slot should be increased by num_slots, so if running another round in the loop it starts to generate from where it ended and not at the same index again

DaughterOfMars: no, the mana is always generated from the original slot to ensure the final check is always correct

Thoralf-M: But that's wrong? Which final check?

Thoralf-M: Ah got it, `num_slots` is updated in the loop, all good

Thoralf-M (approved):
Good job!

Alex6323 (approved):

PR #2161, opened 2024-03-08
NodeJs: sweet improvements :strawberry: 

### LINKED ISSUE #2016: `WalletBuilder`: Handle conflicts between provided and restored values
we need to handle scenarios in which a user provides some value to the wallet builder which could also have been restored, and the provided value conflicts with the stored one.

- [x] wallet address (#1941)
- [x] bip path (#1941)
- [x] alias (#2161)

Alex6323:
# Description of change

This PR:
* ~adds a new `WalletError` variant: `AliasMismatch { provided: String, expected: String }`;~
* ~returns that error in case a provided alias conflicts with the one stored in the wallet db;~

UPDATE: after a call we decided to remove this alias mismatch detection from this PR,
because we don't think there's a real problem with the builder overwriting the alias
stored in the db since this is non-critical data. I kept the other changes unrelated to
the wallet alias, which are:
* NodeJs: consistency changes and improve `wallet.spec.ts` tests

Fixes #2016 

Integration tests.


Thoralf-M (dismissed):

DaughterOfMars:
Shouldn't we just overwrite it? I mean it's just an alias.

Alex6323:
> Shouldn't we just overwrite it? I mean it's just an alias.

could maybe be used to trick someone to enter the stronghold password ... I would be conservative here and not do that. I'ld rather have some explicit command that says `change_alias` or something (similar to `change_stronghold_password`).

Thoralf-M (approved):

DaughterOfMars (approved):
Maybe change the name of the PR now lol

PR #2164, opened 2024-03-11
CLI: adds `FAUCET_URL` environment variable

### LINKED ISSUE #2166: CLI: add `FAUCET_URL` environment variable
## Description

We have the `NODE_URL`  environment variable, that allows convenient configuration to which node the CLI wallet connects to, but we don't have that for the faucet, which usually also changes when you change the node (for example from private tangle to alphanet).

Alex6323:
# Description of change

Adds `FAUCET_URL` environment variable.

Fixed #2166

### DIFF
--- a/cli/src/wallet_cli/mod.rs
+++ b/cli/src/wallet_cli/mod.rs
@@ -179,8 +179,7 @@ pub enum WalletCommand {
     },
     /// Request funds from the faucet.
     Faucet {
-        /// Address the faucet sends the funds to, defaults to the wallet address.
-        #[arg(short, long)]
+        /// Address the faucet sends the funds to. If not provided, the command defaults to the wallet address.
         address: Option<Bech32Address>,
         /// URL of the faucet.
         #[arg(short, long, value_name = "URL", env = "FAUCET_URL", default_value = DEFAULT_FAUCET_URL)]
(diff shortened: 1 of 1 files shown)


cli/src/wallet_cli/mod.rs
```diff
@@ -177,10 +179,11 @@ pub enum WalletCommand {
     },
     /// Request funds from the faucet.
     Faucet {
-        /// Address the faucet sends the funds to, defaults to the wallet address.
+        /// Address the faucet sends the funds to. If not provided, the command defaults to the wallet address.
         address: Option<Bech32Address>,
-        /// URL of the faucet, default to <http://localhost:8088/api/enqueue>.
-        url: Option<String>,
+        /// URL of the faucet.
+        #[arg(short, long, value_name = "URL", env = "FAUCET_URL", default_value = DEFAULT_FAUCET_URL)]
```
Thoralf-M, line 185: Is the default visible in the cli? If not I would put it in the comment back so one sees how the format is (with `/api/enqueue`)

Alex6323: yes, it's visible

Thoralf-M (approved):

thibault-martinez (approved):

PR #2167, opened 2024-03-11
Cli: override default init params with env vars on first run

Alex6323:
# Description of change

Environment variables weren't overriding `InitParameters::default()` values when running `./wallet` for the first time.

That means that the behavior for `./wallet` and `./wallet init` were different, because only the latter was overriding the defaults with the environment variables.


cli/src/cli.rs
```diff
@@ -337,7 +337,15 @@ pub async fn new_wallet(cli: Cli) -> Result<Option<Wallet>, Error> {
             Some(wallet)
         } else {
             // init new wallet with default init parameters
-            let init_params = InitParameters::default();
+            let mut init_params = InitParameters::default();
+
+            if let Some(stronghold_snapshot_path) = std::env::var("STRONGHOLD_SNAPSHOT_PATH").ok() {
+                init_params.stronghold_snapshot_path = stronghold_snapshot_path;
+            }
+            if let Some(node_url) = std::env::var("NODE_URL").ok() {
```
Thoralf-M: Use if let Ok() without `.ok()`?

Alex6323: I don't mind changing if you insist, but I did it like this for readability as it reads like: "If there's (some) stronghold snapshot path" defined in the `.env` file, then use it. `if let Ok(...)` reads worse in this case imo because it's not so clear we're doing some fallible operation here. It's more like a "GET" which usually returns an `Option<T>`.

Thoralf-M: To me it just looks more confusing with the useless `.ok()`

DaughterOfMars: I'd agree to change it

Alex6323: Changed

Thoralf-M: Thanks

Thoralf-M (approved):

DaughterOfMars (approved):

PR #2169, opened 2024-03-11
NodeJs: reorg api response type files

### LINKED ISSUE #2078: NodeJs binding: improve dir/file structure for API types
## Description

Currently we have a lot of files that contain only one type. We should reduce those and logically group the types they contain, e.g. by route group.

## Motivation

* reduces the amount of files 
* it's easier to find stuff

Alex6323:
# Description of change

Brings the file structure in line with the V3 Core API route groups.

Fixes #2078


Thoralf-M (dismissed):

bindings/nodejs/lib/client/client.ts
```diff
@@ -67,14 +67,14 @@ import {
     OutputWithMetadataResponse,
     NetworkMetricsResponse,
 } from '../types/models/api';
-import { RoutesResponse } from '../types/models/api/routes-response';
+import { RoutesResponse } from '../types/models/api/node/routes-response';
```
DaughterOfMars: still `-response`?

DaughterOfMars: Overall confused about the places with/without this suffix

Alex6323: I think I agree with you. I also don't like all those folders now btw. So I'll suggest something different: one file per route group, which means: `accounts.ts`, `blocks.ts`, and so on ... without any suffixes.

Alex6323: I removed all the `-response` suffixes from the files and made the overall structure flat, so all response types live in their respective route group as specified by TIP-48.

bindings/nodejs/lib/types/models/api/response.ts
```diff
@@ -1,21 +1,22 @@
...
-         * The code for the error response.
-         */
-        code: string;
-        /**
-         * A more descriptive version of the error.
-         */
-        message: string;
-    };
-}
+// TODO: can this be removed? Nothing seems to require it :thinking:
```
Thoralf-M: What about all the `const parsed = JSON.parse(response) as Response<BasicOutput>;`?

Thoralf-M: Ah that's something else

Alex6323: that's using the generic `Response<T> type`, not this one.

Thoralf-M: Would delete this file :+1:

Alex6323: Removed

Thoralf-M (approved):

DaughterOfMars (approved):

PR #2177, opened 2024-03-13
Improve wait_for_transaction_acceptance

Thoralf-M:
# Description of change

Don't error with NotFound for the first attempts in wait_for_transaction_acceptance as the node may not have fully processed the tx at this point

## Type of change

Sending transactions with many outputs and then calling `wait_for_transaction_acceptance()` on them which failed before

```
2024-03-13 08:37:57 (UTC) iota_sdk::client::api::wait_for_tx_acceptance DEBUG [wait_for_transaction_acceptance]
2024-03-13 08:37:57 (UTC) iota_sdk::client::node_manager::http_client DEBUG GET: 36 ms for 404 Not Found http://localhost:8050/api/core/v3/transactions/0x8a502496881e05fa5ed60a86c605e8c93d21fdaeb07220dd41fbf884d9ec16d5e6a50000/metadata
2024-03-13 08:37:58 (UTC) iota_sdk::client::node_manager::http_client DEBUG GET: 35 ms for 200 OK http://localhost:8050/api/core/v3/transactions/0x8a502496881e05fa5ed60a86c605e8c93d21fdaeb07220dd41fbf884d9ec16d5e6a50000/metadata
```


Alex6323 (dismissed):

sdk/src/client/api/wait_for_tx_acceptance.rs
```diff
@@ -25,17 +25,31 @@ impl Client {
...
-        for _ in 0..max_attempts.unwrap_or(DEFAULT_WAIT_FOR_TX_ACCEPTANCE_MAX_ATTEMPTS) {
-            let transaction_metadata = self.get_transaction_metadata(transaction_id).await?;
-
-            match transaction_metadata.transaction_state {
-                TransactionState::Accepted | TransactionState::Committed | TransactionState::Finalized => {
-                    return Ok(());
+        for attempt in 0..max_attempts.unwrap_or(DEFAULT_WAIT_FOR_TX_ACCEPTANCE_MAX_ATTEMPTS) {
+            // Ignore not found for the first few attempts, as the node may not have fully processed the transaction
+            // yet.
+            let transaction_metadata = if attempt < 3 {
```
DaughterOfMars: This seems...bad. Is there no way for us to do this better?

Thoralf-M: Don't know, happy about suggestions

DaughterOfMars: Well, the node should always be able to find the transaction, right? It may just not get accepted, in which case it will have a failure reason. So why not just ignore failures where the transaction is not found? Obviously that means this will hang forever if you call it on a transaction you never created, but realistically is that an issue?

Thoralf-M: > Well, the node should always be able to find the transaction, right?

Not if it didn't get the transaction yet (because it's not the node you submitted the block to and the nodes didn't gossip it to this node yet) or haven't fully processed it which is exactly what happened here as you can see in the debug logs of the PR description. First request it didn't return it, ~half a second later it did (the transaction had > 100 outputs)

DaughterOfMars: Yeah but I mean it will eventually find it, always, assuming you sent it

Thoralf-M: But that doesn't help if we return an error because it wasn't found the first attempt

Thoralf-M: Nvm just understood you comment now :joy: Yes I guess we can also just always ignore not found errors

Thoralf-M: Wouldn't even hang forever as the attempts are limited

Thoralf-M: Updated

DaughterOfMars (approved):

Alex6323 (approved):

PR #2184, opened 2024-03-15
Replace first_account_id by first_block_issuer_account_id

Thoralf-M:
# Description of change

Replace first_account_id by first_block_issuer_account_id as we are only interested in those, with normal accounts it can return an error as the returned account can't be used to issue blocks
Derive Eq because of clippy

## Type of change

Running a test that created an account without block issuer feature and using the allot-mana cli command

### DIFF
--- a/cli/src/wallet_cli/mod.rs
+++ b/cli/src/wallet_cli/mod.rs
@@ -443,11 +443,12 @@ pub async fn address_command(wallet: &Wallet) -> Result<(), Error> {
 
 // `allot-mana` command
 pub async fn allot_mana_command(wallet: &Wallet, mana: u64, account_id: Option<AccountId>) -> Result<(), Error> {
-    let account_id = {
-        let wallet_ledger = wallet.ledger().await;
-        account_id
-            .or_else(|| wallet_ledger.first_account_id())
-            .ok_or(WalletError::AccountNotFound)?
+    let account_id = match account_id {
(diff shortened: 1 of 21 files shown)


cli/src/wallet_cli/mod.rs
```diff
@@ -443,11 +443,12 @@ pub async fn address_command(wallet: &Wallet) -> Result<(), Error> {
...
-            .ok_or(WalletError::AccountNotFound)?
+    let account_id = match account_id {
+        Some(account_id) => account_id,
+        None => wallet
+            .first_block_issuer_account_id()
+            .await?
+            .ok_or(WalletError::AccountNotFound)?,
     };
 
     let transaction = wallet.allot_mana([ManaAllotment::new(account_id, mana)?], None).await?;
```
DaughterOfMars, line 454: Hmm, of course for this command we probably want to distinguish between `issuer_id` and `account_id` we're allotting to.

Thoralf-M: Why? Allotment for the issuer account is automatically added as well
https://github.com/iotaledger/iota-sdk/blob/d2aea56a0a7e562b09de8295782d9e72e8c98975/sdk/src/client/api/block_builder/transaction_builder/mod.rs#L164-L165

DaughterOfMars (approved):

Alex6323 (approved):

PR #2186, opened 2024-03-15
Add custom debug impl for `Capabilities<Flag>`

### LINKED ISSUE #2047: CLI: Hidden field exposed in display of transaction details
## Description

```sh
Wallet "core": tx 0xfc84973449a282c60831fd20ceac016ba8495bb5f18fa8232a4f8ffafed18d8d53010000
```
For the transaction capabilities this shows:
``` 
capabilities: Capabilities {
    bytes: [],
    _flag: PhantomData<iota_sdk::types::block::payload::signed_transaction::transaction::TransactionCapabilityFlag>,
},
```
It's not the end of the world, but not really nice to reveal that to the user.

Alex6323:

Fixes #2047

### DIFF
--- a/sdk/src/types/block/capabilities.rs
+++ b/sdk/src/types/block/capabilities.rs
@@ -62,7 +62,7 @@ impl<Flag> std::fmt::Debug for Capabilities<Flag> {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         f.debug_struct("Capabilities")
             .field("bytes", &self.bytes)
-            .field("type", &std::any::type_name::<Flag>())
+            .field("type", &std::any::type_name::<Flag>().rsplit_once("::").unwrap().1)
             .finish()
     }
 }
--- a/sdk/src/types/block/capabilities.rs
+++ b/sdk/src/types/block/capabilities.rs
@@ -58,11 +58,11 @@ impl<Flag> Capabilities<Flag> {
     }
 }
 
-impl<Flag> std::fmt::Debug for Capabilities<Flag> {
+impl<Flag> core::fmt::Debug for Capabilities<Flag> {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         f.debug_struct("Capabilities")
             .field("bytes", &self.bytes)
-            .field("type", &std::any::type_name::<Flag>().rsplit_once("::").unwrap().1)
+            .field("flag", &std::any::type_name::<Flag>().rsplit_once("::").unwrap().1)
             .finish()
     }
 }
--- a/sdk/src/types/block/capabilities.rs
+++ b/sdk/src/types/block/capabilities.rs
@@ -32,7 +32,7 @@ impl From<Infallible> for CapabilityError {
 }
 
 /// A list of bitflags that represent capabilities.
-#[derive(Debug, Deref)]
+#[derive(Deref)]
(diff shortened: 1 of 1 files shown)


sdk/src/types/block/capabilities.rs
```diff
@@ -58,6 +58,15 @@ impl<Flag> Capabilities<Flag> {
     }
 }
 
+impl<Flag> core::fmt::Debug for Capabilities<Flag> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("Capabilities")
+            .field("bytes", &self.bytes)
+            .field("flag", &std::any::type_name::<Flag>().rsplit_once("::").unwrap().1)
```
DaughterOfMars: sad `no_std` noises

DaughterOfMars: just add a `Debug` bound for the flag and call that

sdk/src/types/block/capabilities.rs
```diff
@@ -58,6 +58,15 @@ impl<Flag> Capabilities<Flag> {
     }
 }
 
+impl<Flag> core::fmt::Debug for Capabilities<Flag> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("Capabilities")
+            .field("bytes", &self.bytes)
+            .field("flag", &core::any::type_name::<Flag>().rsplit_once("::").unwrap().1)
```
Thoralf-M: For what is the flag in the output good? Looks pretty useless to me
```
            capabilities: Capabilities {
                bytes: [],
                flag: "TransactionCapabilityFlag",
            },
```

Alex6323: It's supposed to show the same information as before, just without the unnecessary fluff, that reveals Rust details (like the fully qualified path to the type, and the hidden PhantomData)

Thoralf-M: > just without the unnecessary fluff, that reveals Rust details 

But the whole flag field is that, not just exactly the PhantomData (the whole field exists only for the PhantomData?)?

sdk/src/types/block/capabilities.rs
DaughterOfMars: ^^^

sdk/src/types/block/capabilities.rs
Alex6323: yeah, I initially removed it completely, but I think @thibault-martinez wanted to preserve the distinction between `TransactionCapabilityFlag` and `AddressCapabilityFlag`.

sdk/src/types/block/capabilities.rs
Alex6323: you're right .. thanks

sdk/src/types/block/capabilities.rs
Alex6323: My take on this is that we don't need to print it, because it's clear from the surrounding context, that it's the transaction capabilities, not the address ones, but :man_shrugging:

Alex6323: Actually, forget what I said ... I changed it so that it now prints a list of the enabled capabilities (the bytes themselves aren't helpful in the debug, so I removed them from the output).

Thoralf-M (approved):

DaughterOfMars (approved):

PR #2187, opened 2024-03-15
Python: fix event types

### LINKED ISSUE #2105: Python bindings should handle scientific notation of floats
As present in the MetricsResponse of the API

### LINKED ISSUE #2121: Is Python `TransactionProgress` correct?

Alex6323:

Fixes #2121


Thoralf-M:
Isn't https://github.com/iotaledger/iota-sdk/issues/2105 not needed at all anymore?

bindings/python/iota_sdk/types/event.py
```diff
@@ -28,13 +28,15 @@ class TransactionProgressEvent(IntEnum):
         BuildingTransaction (0): Building a transaction.
```
Thoralf-M, line 28: Can't comment there but in the class above there is `TransactionProgress = 4`
The issue was about where having it separated as class like it is now is the correct way, did you check this?
So for example when you get a transaction progress event here https://github.com/iotaledger/iota-sdk/blob/2.0/bindings/python/examples/exchange/4_listen_events.py#L27 can you use this TransactionProgressEvent in a useful way?

Thoralf-M: Just tried it with this (ignore that it doesn't handle other event types :trollface:)
```Python
import json
import os

from dotenv import load_dotenv
from iota_sdk import Wallet, WalletOptions, WalletEventType, SendParams, TransactionProgressEvent

load_dotenv()

for env_var in ['WALLET_DB_PATH', 'FAUCET_URL']:
    if env_var not in os.environ:
        raise Exception(".env WALLET_DB_PATH is undefined, see .env.example")

wallet = Wallet(WalletOptions(storage_path=os.environ.get('WALLET_DB_PATH')))


def callback(event):
    """Callback function for the event listener"""
    event = json.loads(event)
    print('Event:', event)
    if event['progress']['type'] == TransactionProgressEvent.BuildingTransaction:
        print("BuildingTransaction")


wallet.listen(callback, [WalletEventType.TransactionProgress])

wallet.prepare_send([SendParams(wallet.address(), 1000000)])
```
and it prints
```
Event: {'type': 4, 'progress': {'type': 0}}
BuildingTransaction
```

Thoralf-M (approved):

DaughterOfMars (approved):

PR #2188, opened 2024-03-15
Fix DecayedMana Getters

Thoralf-M:
Was not actually derived without `#[getset(get = "pub")]`

### DIFF
--- a/sdk/src/types/block/output/mod.rs
+++ b/sdk/src/types/block/output/mod.rs
@@ -28,7 +28,7 @@ pub mod unlock_condition;
 use core::ops::RangeInclusive;
 
 use derive_more::From;
-use getset::Getters;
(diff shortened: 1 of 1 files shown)


sdk/src/types/block/output/mod.rs
```diff
@@ -408,6 +408,7 @@ pub trait MinimumOutputAmount: StorageScore {
 /// Decayed stored and potential Mana of an output.
 #[derive(Debug, Default, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Getters, derive_more::AddAssign)]
```
DaughterOfMars: `CopyGetters` instead?

DaughterOfMars (approved):

thibault-martinez (approved):

PR #2189, opened 2024-03-15
Remove more secret manager methods from wallet

### LINKED ISSUE #1765: Remove all secret manager methods from the wallet?
## Description

Like store_mnemonic(), get_ledger_nano_status(), change_stronghold_password()...

## Motivation

Can be called from the secret manager instead and not fail if another secret manager is used

No.

Alex6323:
# Description of change

Removes some secret manager methods { Stronghold, LedgerNano } from the wallet that are just forwarding calls. Instead you're now supposed to call those methods on the secret manager directly.

Fixes #1765


Alex6323:
We decided against this change.

PR #2191, opened 2024-03-18
Update private tangle tests so they work with 2.0

Thoralf-M:
# Description of change

Update private tangle tests so they work with 2.0 and fix bugs related to them

## Type of change

Running tests


sdk/tests/client/node_api/mod.rs
```diff
@@ -28,25 +30,64 @@ use crate::client::common::{setup_client_with_node_health_ignored, FAUCET_URL};
 // THIS SEED SERVES FOR TESTING PURPOSES! DON'T USE THIS SEED IN PRODUCTION!
 const DEFAULT_DEVELOPMENT_SEED: &str = "0x256a818b2aac458941f7274985a410e57fb750f3a3a67969ece5bd9ae7eef5b2";
 
-// Sends a tagged data block to the node to test against it.
-async fn setup_tagged_data_block(secret_manager: &SecretManager) -> BlockId {
+// Sends a tagged data block to the node to test against it.*
```
DaughterOfMars: star?

sdk/tests/client/node_api/mod.rs
```diff
@@ -28,25 +30,64 @@ use crate::client::common::{setup_client_with_node_health_ignored, FAUCET_URL};
...
+            break;
+        }
+        if i == 29 {
+            panic!("Faucet no longer wants to hand over coins");
+        }
+        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
+    }
+    // Wait until account is read to issue blocks
+    for _ in 0..60 {
+        if let Ok(_r) = client.get_account_congestion(&account_id, None).await {
```
DaughterOfMars: `.is_ok()`?

sdk/tests/wallet/claim_outputs.rs
```diff
@@ -130,20 +133,41 @@ async fn claim_1_of_2_basic_outputs() -> Result<(), Box<dyn std::error::Error>>
...
+        )
+        .await?;
+    wallet_1
+        .wait_for_transaction_acceptance(&tx.transaction_id, None, None)
+        .await?;
 
     let storage_score_params = wallet_0.client().get_storage_score_parameters().await?;
-    // TODO more fitting value
-    let expiration_slot = wallet_0.client().get_slot_index().await? + 86400;
+    let slot_duration_in_seconds = wallet_0
```
DaughterOfMars: Could add a utility fn to protocol params to get a number of slots from a duration

sdk/tests/wallet/common/mod.rs
```diff
@@ -79,20 +83,87 @@ pub(crate) async fn make_ledger_nano_wallet(
...
+        }
+    }
+
+    let wallet_ledger = wallet.ledger().await;
+    let implicit_account = wallet_ledger
+        .implicit_accounts()
+        .next()
+        .expect("No implicit account found")
+        .clone();
+    drop(wallet_ledger);
```
DaughterOfMars:
```suggestion
    let mut attempts = 0;
    let implicit_account = loop {
        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
        wallet
            .sync(Some(SyncOptions {
                sync_implicit_accounts: true,
                ..Default::default()
            }))
            .await?;
        if let Some(account) = wallet.ledger().await.implicit_accounts().next() {
            break account;
        }
        attempts += 1;
        if attempts == 30 {
            panic!("Faucet no longer wants to hand over coins");
        }
    }
```

sdk/tests/wallet/common/mod.rs
```diff
@@ -79,20 +83,87 @@ pub(crate) async fn make_ledger_nano_wallet(
...
+        .get_account_congestion(&AccountId::from(&implicit_account.output_id), None)
+        .await
+    {
+        tries += 1;
         tokio::time::sleep(std::time::Duration::from_secs(2)).await;
-        let balance = wallet.sync(None).await?;
-        if balance.base_coin().available() > 0 {
-            return Ok(());
+        if tries > 100 {
+            panic!("Can't get_account_congestion for implicit account");
```
DaughterOfMars:
```suggestion
            panic!("Can't get account congestion for implicit account");
```

sdk/tests/wallet/common/mod.rs
```diff
@@ -79,20 +83,87 @@ pub(crate) async fn make_ledger_nano_wallet(
...
+    #[allow(unused_variables)]
+    if let Err(WalletError::Client(ClientError::TransactionBuilder(TransactionBuilderError::InsufficientMana {
+        found,
+        required,
+        slots_remaining,
+    }))) = wallet
+        .prepare_send(vec![SendParams::new(1_000_000, wallet.address().await)?; 10], None)
+        .await
+    {
+        tokio::time::sleep(std::time::Duration::from_secs(
```
DaughterOfMars: Another place where a util fn would help

sdk/tests/wallet/output_preparation.rs
```diff
@@ -356,7 +356,7 @@ async fn output_preparation() -> Result<(), Box<dyn std::error::Error>> {
     assert!(conditions.is_timelocked(0, 0));
     assert_eq!(
         conditions.is_expired(2, CommittableAgeRange { min: 0, max: 0 }),
-        Some(false)
+        Some(true)
```
DaughterOfMars, line 359: 🤔

sdk/tests/client/node_api/mod.rs
Thoralf-M: Pressed the wrong key when uncommenting I guess, removed

sdk/tests/wallet/output_preparation.rs
Thoralf-M: `expiration_slot_index: Some(SlotIndex::from(1)),` 2 is after, probably got updated while being commented and then messed up

sdk/tests/wallet/claim_outputs.rs
Thoralf-M: Would be called `slots_in_duration()`?

sdk/tests/wallet/transactions.rs
```diff
@@ -1,42 +1,43 @@
...
-//     tear_down(storage_path)
-// }
+    tear_down(storage_path_0)?;
+    tear_down(storage_path_1)
+}
 
 // #[ignore]
 // #[tokio::test]
-// async fn send_amount_127_outputs() -> Result<(), WalletError> {
+// async fn send_amount_127_outputs() -> Result<(), Box<dyn std::error::Error>> {
```
thibault-martinez, line 40: Why are the remaining not uncommented?

Thoralf-M: Just didn't want to have so many changes at once

sdk/tests/wallet/output_preparation.rs
DaughterOfMars: okie dokie

sdk/tests/wallet/claim_outputs.rs
DaughterOfMars: Maybe `fn duration_of_slots(num_slots: u32) -> Duration` and `fn slots_in_duration(duration: Duration) -> u32`

sdk/tests/wallet/common/mod.rs
Thoralf-M: How exactly does it help? You aren't starting at exactly the start of a slot, so if you wait for the full duration of a slot it can be delayed quite a few seconds

sdk/tests/wallet/claim_outputs.rs
Thoralf-M: Added only slots_in_duration() as I don't really see a point for the duration at the moment

sdk/tests/wallet/common/mod.rs
DaughterOfMars: Huh? The function would just do exactly what this does, multiply the slots by the slots per second and turn that into a duration.

sdk/tests/wallet/transactions.rs
thibault-martinez: Ok

sdk/tests/wallet/common/mod.rs
Thoralf-M: Yes sorry, I only looked for `tokio::time::sleep(std::time::Duration::from_secs(` which is also used above with 2 seconds
Changed

sdk/tests/wallet/claim_outputs.rs
Thoralf-M: Added duration_of_slots() now as well

sdk/src/types/block/protocol/mod.rs
```diff
@@ -173,6 +175,18 @@ impl ProtocolParameters {
         }
     }
 
+    #[cfg(feature = "std")]
+    /// Calculates the number of slots in a duration.
```
thibault-martinez, line 178: comment before attribute

sdk/src/wallet/operations/output_consolidation.rs
```diff
@@ -116,7 +116,6 @@ where
                     .map(|bech32| bech32.into_inner())
                     .unwrap_or_else(|| wallet_address.into_inner()),
             ),
-            allow_additional_input_selection: false,
```
DaughterOfMars, line 119: I dunno about this...I know the test was failing but this was put here on purpose. I feel like we should not consolidate more outputs than we requested

Thoralf-M: We provided the inputs in Stardust as no other inputs were required, but now we need Mana and these outputs will often not have enough Mana for the consolidation transaction
We could just return an error, but I think it makes more sense to transition a chain output to provide the Mana if possible

DaughterOfMars (approved):

sdk/src/wallet/operations/output_consolidation.rs
DaughterOfMars: alright 🤷‍♀️

thibault-martinez (approved):

PR #2210, opened 2024-03-28
Adjust wait_for_transaction_acceptance for slow indexer

Thoralf-M:
# Description of change

Adjust wait_for_transaction_acceptance for slow indexer

## Type of change

`cargo ci-tangle-test`

### DIFF
--- a/sdk/src/client/api/wait_for_tx_acceptance.rs
+++ b/sdk/src/client/api/wait_for_tx_acceptance.rs
@@ -4,8 +4,11 @@
 use std::time::Duration;
 
 use crate::{
-    client::{Client, ClientError},
-    types::{api::core::TransactionState, block::payload::signed_transaction::TransactionId},
+    client::{node_api::indexer::query_parameters::OutputQueryParameters, Client, ClientError},
+    types::{
+        api::core::TransactionState,
+        block::{address::ToBech32Ext, output::OutputId, payload::signed_transaction::TransactionId},
+    },
 };
 
 pub(crate) const DEFAULT_WAIT_FOR_TX_ACCEPTANCE_INTERVAL: Duration = Duration::from_millis(500);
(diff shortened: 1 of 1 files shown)


sdk/src/client/api/wait_for_tx_acceptance.rs
```diff
@@ -30,6 +33,41 @@ impl Client {
...
                     match transaction_metadata.transaction_state {
                         TransactionState::Accepted | TransactionState::Committed | TransactionState::Finalized => {
+                            let slot_index = self.get_slot_index().await?;
+                            let protocol_parameters = self.get_protocol_parameters().await?;
+                            if let Ok(output) = self.get_output(&OutputId::new(*transaction_id, 0)).await {
+                                if let Some(required_address) = output
+                                    .output
+                                    .required_address(slot_index, protocol_parameters.committable_age_range())?
+                                {
+                                    // Even thought the output was created already, the indexer might take some time
```
DaughterOfMars: u thought

DaughterOfMars (commented):
I need an explanation pls

sdk/src/client/api/wait_for_tx_acceptance.rs
Thoralf-M:
```suggestion
                                    // Even though the output was created already, the indexer might take some time
```

Thoralf-M: Thanks :D

Thoralf-M:
@DaughterOfMars in https://github.com/iotaledger/iota-sdk/pull/2209 the private tangle tests fail because sometimes the indexer isn't updated fast enough (output exists, but is not returned by the indexer, because it wasn't processed yet)
This means when we call `wait_for_transaction_acceptance()` and afterwards right away sync the account, then we might don't get the new outputs
That's why I added the extra logic to wait until the output id is also returned from the indexer, so when we sync it will be there. I don't like it, but at the moment I don't see any other way to handle it better. If you can think of something better, please let me know

thibault-martinez (approved):

DaughterOfMars:
Shouldn't we wait for that separately somehow?

Thoralf-M:
> Shouldn't we wait for that separately somehow?

I don't think so, wait_for_transaction_acceptance is not very useful without this
However, we could rename it to `wait_for_transaction_outputs()` or similar, if you think that's better

DaughterOfMars:
It's just that the indexer has nothing to do with the transaction acceptance, and we don't actually use any of the info we get from it here. It feels like this is not the proper place to have this logic.

Thoralf-M:
`wait_for_transaction_acceptance()` exists because we want to wait for the outputs to be available, which was true until recently
We still want to wait until the outputs are available, otherwise this function is useless as you can see in the failing tests

DaughterOfMars (approved):

PR #2219, opened 2024-04-15
Add issuer_id to Prepared and Signed TransactionData

Thoralf-M:
# Description of change

Add issuer_id to Prepared and Signed TransactionData to fix implicit account transitions with already existing account with the prepare method.

## Type of change

Added test

### DIFF
--- a/sdk/src/wallet/operations/transaction/mod.rs
+++ b/sdk/src/wallet/operations/transaction/mod.rs
@@ -90,10 +90,9 @@ where
         let block_id = match self
             .submit_basic_block(
                 Some(Payload::from(signed_transaction_data.payload.clone())),
-                options
-                    .as_ref()
-                    .and_then(|options| options.issuer_id)
-                    .or(signed_transaction_data.issuer_id),
+                signed_transaction_data
+                    .issuer_id
+                    .or_else(|| options.as_ref().and_then(|options| options.issuer_id)),
                 true,
             )
             .await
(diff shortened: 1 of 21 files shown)


sdk/src/wallet/operations/transaction/send_outputs.rs
```diff
@@ -71,6 +71,7 @@ where
     ) -> Result<PreparedTransactionData, WalletError> {
         log::debug!("[TRANSACTION] prepare_send_outputs");
         let options = options.into().unwrap_or_default();
+        log::debug!("prepare_send_outputs {options:#?}");
```
thibault-martinez, line 74: Just checking you want this to stay

Thoralf-M: Yes, I think it could help us in the future

sdk/src/wallet/operations/transaction/mod.rs
```diff
@@ -90,7 +90,10 @@ where
         let block_id = match self
             .submit_basic_block(
                 Some(Payload::from(signed_transaction_data.payload.clone())),
-                options.as_ref().and_then(|options| options.issuer_id),
+                options
+                    .as_ref()
+                    .and_then(|options| options.issuer_id)
```
thibault-martinez: Do you think option one should be preferred?

Thoralf-M: Wasn't sure about this, but probably would be a bit better to prefer the transaction data one, since this was used when the tx with the allotments was created, will change

Thoralf-M: Changed

thibault-martinez (dismissed):

thibault-martinez (approved):

Alex6323 (approved):

PR #2234, opened 2024-04-22
Fix restore_from_stronghold_snapshot() with same source and target path

Thoralf-M:
# Description of change

Fix restore_from_stronghold_snapshot() for the case when source and target path are the same. Without this change std::fs::copy() truncates the backup file, resulting in `Error: Stronghold(Client(ClientDataNotPresent))` 

Port 8084 seems to be used by Mono https://github.com/actions/runner-images/issues/2820, so replaced this port for the dashboard

## Type of change

Added test

### DIFF
--- a/.github/actions/private-tangle/setup/action.yml
+++ b/.github/actions/private-tangle/setup/action.yml
@@ -27,6 +27,11 @@ runs:
         go-version-file: "iota-core/go.mod"
         cache: false
 
+    - name: Unblock port 8084
+      shell: bash
+      run: |
+        sudo kill -9 `sudo lsof -t -i:8084`
+
     - name: Setup private tangle
       shell: bash
       # setup-go sets the PATH for the correct version, but sudo uses a different PATH by default
--- a/.github/actions/private-tangle/setup/action.yml
+++ b/.github/actions/private-tangle/setup/action.yml
@@ -27,10 +27,10 @@ runs:
         go-version-file: "iota-core/go.mod"
         cache: false
 
-    - name: Unblock port 8084
+    - name: Replace port 8084 by 8087 as it's already used by Mono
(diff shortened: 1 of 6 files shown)


.github/actions/private-tangle/setup/action.yml
```diff
@@ -27,6 +27,11 @@ runs:
         go-version-file: "iota-core/go.mod"
         cache: false
 
+    - name: Unblock port 8084
+      shell: bash
+      run: |
+        sudo kill -9 `sudo lsof -t -i:8084`
```
thibault-martinez: sus

sdk/src/wallet/core/operations/stronghold_backup/mod.rs
```diff
@@ -134,7 +134,9 @@ impl Wallet {
                 .map_err(|_| WalletError::Backup("invalid secret_manager"))?;
 
             // Copy Stronghold file so the seed is available in the new location
-            fs::copy(backup_path, new_snapshot_path)?;
+            if backup_path != new_snapshot_path {
```
DaughterOfMars: I don't think this will always work, since the equality just checks the components are the same, and they could be the same path but one is relative and one is absolute, for instance.

Thoralf-M: Good catch

.github/actions/private-tangle/setup/action.yml
Thoralf-M: Did something else now

thibault-martinez (approved):

DaughterOfMars (approved):
I guess that's fine. Wish there was a way to do this via the private tangle script or something.

PR #2237, opened 2024-04-26
Add tx options param to prepare_claim_outputs()

Thoralf-M:
# Description of change

Add tx options param to prepare_claim_outputs() so one can set the address to which the outputs will be claimed

## Type of change

### DIFF
--- a/bindings/core/src/method/wallet.rs
+++ b/bindings/core/src/method/wallet.rs
@@ -226,7 +226,11 @@ pub enum WalletMethod {
     /// Claim outputs.
     /// Expected response: [`PreparedTransaction`](crate::Response::PreparedTransaction)
     #[serde(rename_all = "camelCase")]
-    PrepareClaimOutputs { output_ids_to_claim: Vec<OutputId> },
+    PrepareClaimOutputs {
+        output_ids_to_claim: Vec<OutputId>,
+        #[serde(default)]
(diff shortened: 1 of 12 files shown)


bindings/core/src/method/wallet.rs
```diff
@@ -226,7 +226,11 @@ pub enum WalletMethod {
     /// Claim outputs.
     /// Expected response: [`PreparedTransaction`](crate::Response::PreparedTransaction)
     #[serde(rename_all = "camelCase")]
-    PrepareClaimOutputs { output_ids_to_claim: Vec<OutputId> },
+    PrepareClaimOutputs {
+        output_ids_to_claim: Vec<OutputId>,
+        #[serde(default)]
```
DaughterOfMars, line 231: skip serializing too?

Thoralf-M: Not needed, we never serialize this from Rust

DaughterOfMars: fair enough

DaughterOfMars (dismissed):

Alex6323:
looks good to me, but we could prevent some cloning with a slight refactor if we wanted to at some point.

Alex6323 (approved):

DaughterOfMars (approved):

PR #2240, opened 2024-04-29
Don't set with_minimum_amount for accounts when they have a staking feature

Thoralf-M:
# Description of change

Don't set with_minimum_amount for accounts when they have a staking feature, as this would result in an invalid transition

## Type of change

Sending txs with the cli wallet


thibault-martinez (approved):

DaughterOfMars (commented):
So they don't need the storage deposit minimum?

Thoralf-M:
We can't reduce the amount when we have the staking feature, without it, we can do the minimum

DaughterOfMars (approved):

PR #2261, opened 2025-03-26
Fix Type Annotation after `deranged` Update

wulfraem:
# Description of change

## Actual change intended here

The `deranged` crate, that can be included in the project, e.g. via `stronghold` feature, has added implementations for `usize: PartialOrd<_>` in its latest updates. This can lead to conflicts with implementations from `core`, e.g. in code related to the `client` feature, producing build errors like these:

```raw
error[E0283]: type annotations needed
   --> sdk/src/wallet/account/operations/output_claiming.rs:365:59
    |
365 | ...                   > NativeTokens::COUNT_MAX.into()
    |                       -                         ^^^^
    |                       |
    |                       type must be known at this point
    |
    = note: multiple `impl`s satisfying `usize: PartialOrd<_>` found in the following crates: `core`, `deranged`:
            - impl PartialOrd for usize;
            - impl<MIN, MAX> PartialOrd<deranged::RangedUsize<MIN, MAX>> for usize
              where the constant `MIN` has type `usize`, the constant `MAX` has type `usize`;
help: try using a fully qualified path to specify the expected types
    |
365 |                                 > <u8 as Into<T>>::into(NativeTokens::COUNT_MAX)
```

This can be reproduced by creating a new project, that uses `iota-sdk` with features `client` and `stronghold` enabled:

```bash
cargo init build-test
cd build-test
cargo add iota-sdk --features=client,stronghold
cargo build
```

This PR updates the affected calls to use fully qualified paths in the first commit.

### Linting rule changes

- `result_large_err = "allow"` due to errors related to the large error objects
- `clippy::empty_docs` as there were a lot of errors (~100) due to the empty doc blocks, that had been inserted
- disable `clippy::needless_lifetimes` in `bindings/core` due to linting errors from `derivative::Derivative`
  This could be done differently by wrapping the 16 cases where the error was thrown into an inner module with this rule disabled, and then `pub use`ing the structs back into the parent scope. Another possible solution would be to replace the `derivative` crate, but decided to suggest disabling the rule for now and postponing the decision how to refactor this to a later point in time. 

### The two `mut`s

The commits afterward fix findings from the CI. Tried to split the changes into separate commits for the respective steps to not get mixed up in one large commit.

Also, a bit unsure about the updates:

- [sdk/src/wallet/account/builder.rs](https://github.com/iotaledger/iota-sdk/pull/2261/commits/d099cc639f678981a99f5891dc3a381a515da94b#diff-bba01de6544278af5abe9d0bcfd73b702fd901b82b428884b2dc701a675dca30L61)
- [sdk/src/wallet/account/update.rs](https://github.com/iotaledger/iota-sdk/pull/2261/commits/d099cc639f678981a99f5891dc3a381a515da94b#diff-9a884de5d500884f91ea4eb4de48dabbc5c0c6bd017348f913db2e0490b229e2L282)

`clippy` complained about `&self` being mutable even it isn't required here, so removed the `mut`. Theoretically, this might allow someone modifying the related instances while the two update `async` functions are running, which might lead to an unexpected result. So, not sure if we should remove the `mut` (as currently done in this PR) or just ignore the warning in these places.

### `cargo audit`

Bumped two dependencies where fix versions, where available, not sure about the remaining 5, maybe we have to add them to the ignore list for now, as refactoring the usage of them might require more work.

### Python bindings checks

Not really an idea, why this fails with `ImportError: No module named iota_sdk.client.client.Client` to be honest. :/

## Type of change

Removed `deranged` and `time` dependencies from root `Cargo.lock` and ran `cargo build` to force updating to the newest version. Confirmed, that sdk still builds and tests run successfully locally.

Also did the same update to the `Cargo.lock` in `identity.rs` and confirmed, that got the same errors with the current state, and that I could build the project again after switching `iota-sdk` dependency to this PR's branch and applying a similar fix as done for this PR to the `identity.rs` code.

## Change checklist


wulfraem:
Closing the PR as with [`deranged@0.4.1`](https://crates.io/crates/deranged/0.4.1) being yanked in the meantime, the issue does not persist anymore. The steps to reproduce the error, described above, can now be performed without any problem. ^^

PR #2262, opened 2025-04-22
fix: clippy warnings

Thoralf-M:
Fixed clippy warnings
Checked with `cargo clippy`
Bindings/fmt will be done in other PRs

### DIFF
--- a/sdk/src/client/mod.rs
+++ b/sdk/src/client/mod.rs
@@ -27,11 +27,8 @@
 #[cfg(feature = "mqtt")]
 macro_rules! lazy_static {
     ($init:expr => $type:ty) => {{
-        static mut VALUE: Option<$type> = None;
-        static INIT: std::sync::Once = std::sync::Once::new();
-
-        INIT.call_once(|| unsafe { VALUE = Some($init) });
-        unsafe { VALUE.as_ref() }.expect("failed to get lazy static value")
+        static VALUE: std::sync::LazyLock<$type> = std::sync::LazyLock::new(|| $init);
+        &VALUE
     }};
 }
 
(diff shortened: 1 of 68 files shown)


DaughterOfMars (dismissed):

sdk/src/client/mod.rs
```diff
@@ -27,11 +27,8 @@
 #[cfg(feature = "mqtt")]
 macro_rules! lazy_static {
     ($init:expr => $type:ty) => {{
-        static mut VALUE: Option<$type> = None;
-        static INIT: std::sync::Once = std::sync::Once::new();
-
-        INIT.call_once(|| unsafe { VALUE = Some($init) });
-        unsafe { VALUE.as_ref() }.expect("failed to get lazy static value")
+        static VALUE: std::sync::LazyLock<$type> = std::sync::LazyLock::new(|| $init);
+        &VALUE
```
thibault-martinez, line 31: Did clippy suggest this?

thibault-martinez (dismissed):

sdk/src/client/mod.rs
Thoralf-M: No, this was actually a compilation error

thibault-martinez: How is that possible?

Thoralf-M: Nvm it was a clippy error and not a warning, didn't know that `cargo check` will show these

thibault-martinez (dismissed):

sdk/src/client/mod.rs
thibault-martinez: Alright, thanks

